PR #14, opened 2021-05-27
Rebusurance shield import

ZeLonewolf:
This PR demonstrates the follow technology stack elements:
1) Loading of an external SVG icon set and incorporating them into the style's sprite sheet without replicating the source repository
2) Aligning POI markers with text overlays (will be needed for nearly all highway shields)
3) Inline editing of SVG icons via script
4) Highway shields that scale based on the size of the text displayed

This also incidentally provides a workable rendering of attractive interstate highway shields.  This design will likely be revised when the style design is completed, and likely the current icon sizes will be too large when considering other elements on the map.  However, this initial PR flushes out important aspects of the technology stack that will be useful to future contributors.

This shield import brings in all of the rebusurance state shields, however, I have only tested the Interstate and US Highway shield.  Once the style design is complete, each shield will have to be hand-curated to ensure an attractive appearance on the map.


style/layers.js
```diff
@@ -29,5 +29,7 @@ var americanaLayers = [
 
   layerMotorwayLabel,
 
-  layerHighwayShieldInterstate,
+  layerHighwayShieldInterstate12,
+  layerHighwayShieldInterstate3,
+  layerHighwayShieldInterstate4,
```
akadouri, line 34: Do these numbers refer to the number of characters in the highway shield?

akadouri: Okay yeah I see it in the other file, its based on `["==", "ref_length", X]`

style/scripts/import_rebusurance.sh
```diff
@@ -0,0 +1,20 @@
...
+find "build/rebusurance-v1.0.0/image2d" -name "*.svg" -type f -print0 | while read -d $'\0' f; do
+  newfile=`echo "$f" | sed -e 's/ /_/g; s/U.S./us/g'`
+  mv -n "$f" "$newfile";
+done
+
+for i in $( ls build/rebusurance-v1.0.0/image2d/ | grep [A-Z] );
+do
+  sed -i 's/<text.*\/text>//g' "build/rebusurance-v1.0.0/image2d/$i"
+  sed -i 's/transform=""/transform="scale(0.35,0.35)"/g' "build/rebusurance-v1.0.0/image2d/$i"
+#  sed -i 's/stroke-width="2"/stroke-width="2.5"/g' "build/rebusurance-v1.0.0/image2d/$i"
```
akadouri, line 12: nit: commented out line

akadouri (approved):
Code looks good to me, still able to keep things running. The shield rendering seems a little off though... (edit: reread the PR description, think this was intentional now but will leave up just incase).
Screen Shot 2021-05-29 at 5 01 17 PM

style/layers.js
ZeLonewolf: Correct!  You can do things like resize the shields or change the fonts when there's more characters on the shield.  It's also a great candidate for looking to see if we can do a better job at avoiding copy/paste :-/

ZeLonewolf:
> Code looks good to me, still able to keep things running. The shield rendering seems a little off though... (edit: reread the PR description, think this was intentional now but will leave up just incase).
> Screen Shot 2021-05-29 at 5 01 17 PM

Oh wow... that is NOT normal

ZeLonewolf:
Oddly, I'm not seeing the same strange shield render behavior:
image

PR #16, opened 2021-05-31
Fixes for SVG shield manipulation code

### LINKED ISSUE #15: Intermittent resizing errors on SVG imported graphics
In some cases, SVG resizing has not worked completely, resulting in partially-resized images, such as:
image

Image resizing is done using a `sed` command to set the transform attribute, however, this seems to be not working the same way on all platforms.  Instead, this should be converted to an xmlstarlet command to more reliably edit the XML.

ZeLonewolf:
Fixes #15 

This fixes a number of issues with the SVG import.
1. We're now using xmlstarlet in all cases and eliminated sed hacking, which is unreliable and breaks under different CR/LF situations.  Since xmlstarlet is XML-aware, we'll avoid that problem.
2. Set the icon sizes for each shield so they don't take up so much space in the sprite sheet.  This also fixes the issue where the icons were artificially enlarged and thus reducing the frequency of shield placement by mapLibre.
3. Eliminated separate 2,3,4 digit shield styles and switched to dynamically-resizing icons.
4. Slightly reduced the size of the rendered icons.

image

### DIFF
--- a/style/layers.js
+++ b/style/layers.js
@@ -27,9 +27,8 @@ var americanaLayers = [
   layerBridgeOneway,
   layerBridgeOnewayLink,
 
+  layerMotorwayExit,
   layerMotorwayLabel,
 
-  layerHighwayShieldInterstate12,
-  layerHighwayShieldInterstate3,
-  layerHighwayShieldInterstate4,
+  layerHighwayShieldInterstate,
 ];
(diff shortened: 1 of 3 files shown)


style/layers.js
```diff
@@ -27,9 +27,8 @@ var americanaLayers = [
   layerBridgeOneway,
   layerBridgeOnewayLink,
 
+  layerMotorwayExit,
```
akadouri, line 30: Missing `layerMotorwayExit` file.

PR #17, opened 2021-06-04
Proof of concept: MUTCD symbolic signs for points of interest

impiaaa:
Driving down the freeway, I saw a sign describing what amenities could be found at the next exit. I thought they looked kinda like PoI icons, and was inspired.

image

This PR is only "general information" ("I") categories for now, though the destination symbols are actually under "destination and distance, general service" ("D9"). The SVG files were downloaded from Wikimedia Commons (which got them from the original PDF), then tweaked by hand to fit into 32x32 pixels, and then optimized with [svgo](/svg/svgo). The style's zoom levels and filter settings were copied from [gravitystorm/openstreetmap-carto](/gravitystorm/openstreetmap-carto). Unfortunately, organizing icons into folders necessitated updating spritezero-cli; my branch with that feature is [here](/impiaaa/spritezero-cli/tree/folders).


style/icons/MUTCD/I/12.svg
```diff
@@ -0,0 +1,5 @@
...
+    <circle style="fill:#fff;fill-opacity:1;stroke:none;stroke-width:0;stroke-linecap:round;stroke-miterlimit:4;stroke-dasharray:none" cx="60.676369" cy="40.922352" r="1.41111"/>
+</svg>
```
ZeLonewolf, line 5: Where are you pulling in the MUTCD files from?  Would it be feasible to pull them in with some kind of bulk import or are they scattered through wikimedia commons or what's the situation?

impiaaa: > The SVG files were downloaded from Wikimedia Commons (which got them from the original PDF), then tweaked by hand to fit into 32x32 pixels, and then optimized with svgo.

I don't think a bulk import is feasible if you want them to be pixel-aligned like I made them. If not, maybe you could import them from Commons, but a few of those still need cleanup to be made consistent. There might also be a way to automatically get them from the PDF but I haven't looked into it.

ZeLonewolf: Ahh I should probably read your comments first 😂.  There's probably no getting around making icons pixel aligned.  Be aware that raster zooms (which osm-carto uses) and vector zooms are not equivalent - it is off by exactly one level.

style/CONTRIBUTE.md
```diff
@@ -28,7 +28,7 @@ Pre-requisites:
   - **python3**, to run a local web server (`apt install python3`)
   - **nvm** (https://heynode.com/tutorial/install-nodejs-locally-nvm)
   - **xmlstarlet**, to preprocess imported SVGs for map displays (`apt install xmlstarlet`)
-  - **spritezero**, to combine icons into a stylesheet (`npm install -g @beyondtracks/spritezero-cli`)
+  - **spritezero**, to combine icons into a stylesheet (`npm install @beyondtracks/spritezero-cli`)
```
ZeLonewolf, line 32: Made these changes plus Makefile changes to master, recommend merging changes from main.

impiaaa:
Rebased, and also started [another branch](/impiaaa/openstreetmap-americana/tree/overpass) that uses the [Overpass](http://overpassfont.org) font for all map text.

kennykb:
I'd favor using someone else's icon set for PoI's, so that they all look as if they're going to the same party.

If we want to look like National Park Service maps, we could use National Park Service icons - https://github.com/nationalparkservice/symbol-library/  (These are a lot less clunky for a Web map than the paper-map ones at https://www.nps.gov/carto/app/#!/maps/symbols, and at some point I want to switch my map style over to them.)

kennykb:
Overpass is actually a bit blocky for a map, since it's designed for signage.  I'd actually like to see something along the lines of Rawlinson as a serif font (conventional for labeling certain natural features, particularly watercourses) and Hind as sans-serif (the nearest match I could find to the non-free Frutiger that NPS uses).  I really, really like the typography of the NPS maps.

PR #26, opened 2021-07-09
Parse bridge layers separately

ZeLonewolf:
This PR fixes an issue identified by @zekefarwell in which overlapping bridges render as if they're intersecting, ignoring the `layer` key.  Since the layers are defined in javascript, we can define individual layers and programmatically create a copy of each bridge layer for each layer.

We use 5 layers as the upper limit for this processing, based on the [High Five Interchange](https://en.wikipedia.org/wiki/High_Five_Interchange) in Dallas, TX, however, we may need to investigate whether there are more complex intersections that require a higher upper bound.

This assumes that above-ground layers are numbered starting at 1 and increment by 1 and therefore this is a reasonable limit.

Results:
image

### DIFF
--- a/style/index.html
+++ b/style/index.html
@@ -32,6 +32,8 @@
     <script type="text/javascript" src="constants/color.js"></script>
     <script type="text/javascript" src="constants/line.js"></script>
 
+    <script type="text/javascript" src="js/util.js"></script>
+
     <script type="text/javascript" src="layer/background.js"></script>
     <script type="text/javascript" src="layer/water.js"></script>
     <script type="text/javascript" src="layer/road_casing.js"></script>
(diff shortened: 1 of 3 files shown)


zekefarwell:
Oh nice!  Glad you found a way to handle this. 

The screenshot looks good, but I'm not quite getting that on my machine.  I think it may just be that the file `js/util.js` is missing?

image


On a minor stylistic note, the multiple calls of `americanaLayers.push(layerName)` could be simplified to fewer calls taking multiple values if you want:

``` js
americanaLayers.push(
  layerMotorwayLabel,
  layerHighwayShieldInterstate,
  layerBridgeMotorway,
  ...
);
```

ZeLonewolf:
Thanks for catching those!  Code updated.

zekefarwell (approved):
Looking good on my end now!

PR #28, opened 2021-07-10
Add NPM dependency tracking and replace python dev server with browser-sync

zekefarwell:
This PR brings `package.json` and `package-lock.json` in as tracked files so we can keep NPM dependencies in sync (just run `npm install`).  Also replaces the python http dev server with the NPM package browser-sync so we can have a file watcher and live reloading.

### DIFF
--- a/style/.gitignore
+++ b/style/.gitignore
@@ -1,5 +1,4 @@
 node_modules
 sprites
 build
-package*.json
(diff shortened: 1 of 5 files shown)


ZeLonewolf (approved):
Looks good to me & works here.  Thanks!

PR #29, opened 2021-07-10
Change prettier config to same-commit

ZeLonewolf:
Hopefully this fixes the issue we're having with prettier completion not being detected by merge checks.  This will add a prettier action to the same commit in a PR, rather than create a subsequent one.


ZeLonewolf:
This didn't fix the problem.

PR #31, opened 2021-07-11
Install on MacOS info

Mashin6:
Info and tweaks for MacOS install

### DIFF
--- a/style/CONTRIBUTE.md
+++ b/style/CONTRIBUTE.md
@@ -31,6 +31,15 @@ Pre-requisites:
     - **[spritezero][2]**, to combine icons into a sprite sheet
     - **[prettier][3]**, to keep our code style consistent
 
+- On MacOS (MacPorts):
+  - **npm** (`sudo port install npm6`)
+  - **python3** (`sudo port install python38`)
+  - **xmlstarlet** (`sudo port install xmlstarlet`)
+  - **nvm** (https://heynode.com/tutorial/install-nodejs-locally-nvm)
+  - **prettify** (`npm install --save-dev --save-exact prettier`)
+  - **spritezero** (`npm -g config set user $USER` and `sudo npm install -g @beyondtracks/spritezero-cli`)
+  - **browser-sync** (`npm install browser-sync`)
+
 [1]: https://browsersync.io/
 [2]: https://github.com/mapbox/spritezero
 [3]: https://prettier.io/
--- a/style/CONTRIBUTE.md
+++ b/style/CONTRIBUTE.md
@@ -23,6 +23,7 @@ Pre-requisites:
 
 - This guide is written for Ubuntu Linux. Users of Windows or other Linux distributions are encouraged to add OS-specific development guidelines.
 - On Ubuntu, the following development tools are required:
+
   - **npm** (`apt install npm`)
   - **nvm** (https://heynode.com/tutorial/install-nodejs-locally-nvm)
   - **xmlstarlet**, to preprocess imported SVGs for map displays (`apt install xmlstarlet`)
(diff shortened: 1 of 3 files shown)


style/Makefile
```diff
@@ -2,7 +2,7 @@ clean:
 	rm -rf sprites build icons/us_* rebusurance.zip
 
 rebusurance.zip:
-	wget -q -O rebusurance.zip https://github.com/1ec5/rebusurance/releases/download/v1.0.0/rebusurance-v1.0.0.zip 
+	curl -s -L https://github.com/1ec5/rebusurance/releases/download/v1.0.0/rebusurance-v1.0.0.zip --output rebusurance.zip
```
ZeLonewolf, line 6: This change is fine, but I expect to eventually replace this bulk import with customized-edited shield blanks based on experimentation in another branch.  There's too many changes that need to be made to the artwork to make it look right in 20x20 icons.

zekefarwell: I just [installed `wget` on my Mac with homebrew](https://formulae.brew.sh/formula/wget), but `curl` works fine too.

style/CONTRIBUTE.md
```diff
@@ -31,6 +31,15 @@ Pre-requisites:
...
     - **[prettier][3]**, to keep our code style consistent
 
+- On MacOS (MacPorts):
+  - **npm** (`sudo port install npm6`)
+  - **python3** (`sudo port install python38`)
+  - **xmlstarlet** (`sudo port install xmlstarlet`)
+  - **nvm** (https://heynode.com/tutorial/install-nodejs-locally-nvm)
+  - **prettify** (`npm install --save-dev --save-exact prettier`)
+  - **spritezero** (`npm -g config set user $USER` and `sudo npm install -g @beyondtracks/spritezero-cli`)
+  - **browser-sync** (`npm install browser-sync`)
```
zekefarwell, line 41: I think rather than MacPorts specific instructions it would be better to just list the requirements.  I use Hombrew to install some of these packages and for NodeJS/NPM I just download and run the installer from [nodejs.org](https://nodejs.org).  Shouldn't matter how packages are installed though.

Changes were also introduced in PR #28 to simplify the install process.  We now have the NPM dependencies tracked in `style/package.json` so they don't need to be installed individually.  Just run `npm install` in the style directory to install them all in one step.  

Python is no longer needed, and NVM is not a hard requirement, just one method of installing NodeJS if you need to juggle multiple versions.  NPM 7 or newer is needed and should be included with an install of NodeJS 14 or newer.

style/Makefile
Mashin6: For some reason my wget can't work with https :( and I thought curl is probably part of any linux distribution anyways...

style/CONTRIBUTE.md
Mashin6: I guess one option could be to just show both ways to install e.g. (`sudo port install npm6` | `brew install node`) OR have a lists of packages names under MacPorts and a second list for Homebrew. 

I saw the `package.json` file, but to install `spritezero` I had to switch default npm user to my user for it to install. I thought that might be useful for if someone runs into the same issue. I can simplify to `npm -g config set user $USER` + `sudo npm install`

I only now read your slack message. Makes sense to remove the python install.

zekefarwell: Hmm.  plain `npm install` in the /style directory works fine for me on a fresh clone of the project.  It installs `spritezero`, `browser-sync`, and `prettier` as local project packages in the `/style/node_modules` directory as expected.

Not sure why the user switch would be necessary but I'm guessing permissions issues related to installing modules globally (`npm -g`).  It's generally recommended to avoid using `sudo` with `npm` as it can cause various problems. Some reasons why here: https://medium.com/@ExplosionPills/dont-use-sudo-with-npm-still-66e609f5f92

Mashin6: Might be something related to how MacPorts installs npm, but on my computer default user for npm was `nobody` and it didn't have access to required folders giving me `EACCES user "nobody" does not have permission to access ...`

zekefarwell (approved):
This all looks fine to me now.  

Ideally the instructions should specify `npm7` rather than the outdated `npm6` but the Ubuntu instructions need to be changed to reflect this as well.  I can spend some time later this week on improving the setup instructions in general to make things clearer for new contributors.

PR #35, opened 2021-07-12
Administrative boundaries rendering

Mashin6:
Might still need some tweaking, Europe looks very busy on z3.

image
image
image
image


style/layers.js
```diff
@@ -41,6 +41,12 @@ americanaLayers.push(
 
   layerHighwayShieldInterstate,
 
+  layerBoundaryCountyBg,
+  layerBoundaryCounty,
+  layerBoundaryStateBg,
+  layerBoundaryState,
+  layerBoundaryCountryBg,
+  layerBoundaryCountry,
   layerPlaceCity,
```
ZeLonewolf, line 50: I think we should put the boundary layer between water and roads.  One of the screen shots has a boundary going through a highway shield which doesn't look right.  Note that the layers are applied in the order listed.

ZeLonewolf:
The casing seems a little harsh over water:
image

Perhaps we can explore changing the casing (background) to be more subtle.  Below example with alpha=0.5:
image

Perhaps even a subtle purple or blue may work as well, as that is common styling in American maps.

Also, if we're adding support for rendering `admin_level` 2 through 6 in this PR, we should include the odd-numbered ones.  For example, level 3 is missing for the UK, which would give you Scotland, England, and Wales boundaries:
image

Also, level 5 is used in Canada:
https://wiki.openstreetmap.org/wiki/Canada_admin_level

style/layers.js
Mashin6: Love this suggestion. The roads pop out more when the border does not go over them.

Mashin6:
I like the 0.5 alpha. Other option is also to remove it completely. Here are some examples from other places. b) is USA
image

It's weird. admin_level 3 is there `["in", "admin_level", 3, 4]` 
Would 5 be in the same layer with 6? I'll try to play so it doesn't look too much over plotted.
Or is there a way how to do that in Canada would render only 5 and not 6, but in US only 6 and not 5?

ZeLonewolf:
Globally, there are pretty much examples of all levels 2-11 being in actual usage.  This page has a comprehensive list:
https://wiki.openstreetmap.org/wiki/Tag:boundary%3Dadministrative#admin_level.3D.2A_Country_specific_values

weech:
Brian requested some samples from American maps in the Slack discussion, so here's a few. I have a 2010 AAA state-level map that renders national and state boundaries extremely similarly, with the only difference being that the national border uses two dashes instead of one. That map renders county lines as blue dashed lines. I also have a 2008 Rand McNally city-level map that renders county boundaries similarly to the AAA state boundaries, except with the casing over water features. It renders municipal boundaries as a somewhat transparent light blue line.

Mashin6:
Interesting.. the casing is also much wider than what we have.
Here is USGS.. counties have yellow casing.
@ZeLonewolf The main maintainer has to decide, which way to go :)  (we might need to revisit when more elements will be on the map)
image

zekefarwell:
Not showing maritime boundaries seems to be quite common.  I think we should do the same if possible.  Boundaries over lakes and rivers however do seem to be typical so it isn't as simple as just hiding boundaries over water.

Screen Shot 2021-07-13 at 1 27 59 PM

This website is a good reference for some different map style examples: https://www.maps-of-the-usa.com

ZeLonewolf:
Thanks for those examples, @weech -- Check out the boundary action on that lake!  We can achieve an effect like this by placing the boundary casing _below_ the water layer, but the boundary fill _above_ it.  This would also solve issues where the over-water boundaries don't look great with the casing over water, but do look good with casing over land:
image

ZeLonewolf:
With regard to:
> The main maintainer has to decide, which way to go

While I am happy to "break ties" if that's what's needed to keep forward momentum, I plan to seek wide feedback on design choices (especially important. fundamental elements such as boundaries) so that it reflects community opinion on what a visually-pleasing "American-style map" should look like.  Thanks @Mashin6 for starting the conversation :)

I do think the simple line borders at the lower zooms look good the way you have them.  If you want to separate those changes into a separate PR I think we're good to approve merge them more quickly, otherwise we can do it as one big PR once we resolve the styling and over-water concerns in this discussion.

ZeLonewolf:
Note that I am open to changing the color of water (to a different shade of blue, of course) as part of the solution space for attractive boundaries.

Mashin6:
@zekefarwell: awesome site. I love it. Removing maritime borders completely would need PR at OMP.

@ZeLonewolf: I think it makes sense to have them all together in one PR, we are not in a rush. I think we just need to have it roughly done (what borders to show when, line type, basic colors), and detail will be polished depending on the content of other objects that we be there.

I tried adding a bit purple and opacity, increase width and hide under water.
image
image
image
image

ZeLonewolf:
Hiding the casing underneath the water layer was definitely the right choice.  This is starting to look like a real map!

style/layer/boundary.js
```diff
@@ -0,0 +1,161 @@
+var layerBoundaryCountyBg = {
+  id: "boundary_county-bg",
+  type: "line",
+  paint: {
+    "line-color": "rgba(250, 250, 250, 1)",
```
ZeLonewolf, line 5: Since you're using the same color in multiple places, consider extracting the colors out to the color constants file:
https://github.com/ZeLonewolf/openstreetmap-americana/blob/049c0d142af77b425d5d1d54da7ecab180f3585a/style/constants/color.js#L1-L5

That way you can just change the colors in one place as you play with the style.

For example, in the road layer:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/049c0d142af77b425d5d1d54da7ecab180f3585a/style/layer/road.js#L5

Mashin6:
Just or this example I used brighter red for highways `colorMotorway = "rgba(253, 2, 79, 1)";` and made the borders a bit darker `colorBorderCasing = "rgba(136, 85, 141, 0.5)";`

image

here is the previous `var colorBorderCasing = "rgba(219, 162, 230, 0.5)";`
image

ZeLonewolf:
I noticed that the alpha=0.5 causes weird styling when the border loops upon itself, as it does on the NH/Canada boundary, so I think it's best to keep the boundary casing at alpha=1.

I came up with this, which lightens the casing (from the 1.0 alpha version of the color above) and with a slightly brighter red for the highways:

image

Values are:
`colorMotorway = "hsla(354, 82%, 45%, 1)";`
`colorBorderCasing = "rgba(224, 207, 232, 1)";`

Do we need casing on the county boundaries, or is it better to keep them without the casing?  I'm thinking that the casing doesn't add very much if it's not actually wider than the boundary lines.

Mashin6:
Cool, I like that. (I also noticed that the transparency causes troubles)

I feel ambivalent about the county casing. On one hand there is consistency. When you look at the lines you immediately know that those represent administrative boundaries. But on other hand do we need to highlight them so much? We could 1) keep the casing only for very high zooms, 2)  remove it completely, 3) revisit it later once we have all the other objects rendered.

ZeLonewolf:
Here's a quick snap I took from a AAA map:
boundaries

They all go down to just a line with no casing (in blue, of all colors) for county boundaries.  I'm not sure blue is the right color for this, but they definitely drop the casing on lesser boundaries.

zekefarwell:
Keep in mind that these highway maps we are taking inspiration from are all small scale.  This means emulating their styling is most appropriate at the lower zoom levels.  Deviating from their style as the zoom increases is also appropriate, and there we should look to larger scale maps for inspiration.  So for county boundaries it may work well to start with no casing at the zoom where they are introduced, but then add casing at a higher zoom.

ZeLonewolf:
> So for county boundaries it may work well to start with no casing at the zoom where they are introduced, but then add casing at a higher zoom

Good call on that.

I think also, if we are going to have boundary casing that it should be no smaller than 3x the width of the boundary line (i.e. an amount equal to the line thickness itself on either side), otherwise I think we want the line plain with no casing.

Mashin6:
True.. if you think about it we are creating 20 different maps with world-wide coverage, all at the same time.

So what would you two suggest to be the zoom level when the county boundaries appear and what zoom their casing becomes visible?

The casing is currently 2x boundary line.. I can definitely go more thicker

ZeLonewolf:
Just eyeballing it, I'm thinking we wouldn't want to start county boundary casing until somewhere around zoom 10-11 and (when we add them) perhaps we add city/town boundaries at the same zoom where we add county casing?

ZeLonewolf:
New county/town boundaries look great!

I did notice at zoom 6 we have solid national boundaries and dashed state boundaries.  For consistency, I think that the dashed boundaries and the dashed+cased boundaries should both start at zoom 7, and I recommend the state boundaries move to a thin solid line at zoom 6, the same way it looks at zoom 5.  Current zoom 6:
image

I'm noticing an oddity in the UK.  At zoom 4, it looks like `admin_level` 2 and 4 are shown but 3 is missing:

image

Then when we zoom in, the subdivisions in Ireland go away but the level 3 divisions in the UK (between Wales, England, and Scotland) suddenly appear:

image

So it seems like there's somethink wonky going on there with the zoom levels.

With regard to the UK, I'd suggest that `admin_level=3` should render at the same zoom level as 4 but with a thicker line.

Mashin6:
It is probably an OpenTileMap issue. `admin_level=3` is present in the tiles only from zoom 5 where you also see it appearing.  And there is definitely some problem with admin 4 and 3 for Ireland and France. Compare zoom 3 and 4, thats when it switches from `admin_level=4` to `admin_level=3`.

ZeLonewolf:
These boundaries look great.  Thanks for your hard work @Mashin6!

I've added an additional ticket, #40, to consider the need for `admin_level=5` and `admin_level=7` boundaries to be rendered as well.

ZeLonewolf (approved):

PR #36, opened 2021-07-13
Add support for bridges without layer tag

### LINKED ISSUE #34: Bridges not rendered at zoom 8 and lower
The recent changes in #26 to fix stacked bridges causes bridges to not render at zoom 8 and lower:
image

This is because OpenMapTiles suppresses the layer tag at lower zoom, so the style needs to ignore layer information at low zooms.

ZeLonewolf:
Fixes #34

This makes two changes to bridge layering:
1. The lowest level is now "bridges with no layer tag" which catches bridges where nobody bothered to set it, as well as bridges at lower zoom which are stripped of their layer tag by OpenMapTiles.
2. Bridges at layer 5 or above are now grouped together in a single layer.  It's doubtful that there's any cases in reality of a bridge stack of 6 or more, so this is a reasonable optimization.

Render sample from New Orleans at zoom 7:
image

### DIFF
--- a/style/js/util.js
+++ b/style/js/util.js
@@ -2,8 +2,13 @@
 
 //Limit the specified definition to a single numbered layer
 function restrictLayer(def, layer) {
+  return filteredClone(def, ["==", "layer", layer], "_layer_" + layer);
+}
+
+//Make a clone of a layer definition, with a filter added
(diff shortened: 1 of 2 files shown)


zekefarwell (approved):
looks good to me

PR #41, opened 2021-07-16
Add favico.ico

ZeLonewolf:
This adds a favico.ico to the style so we don't have to look at 404 errors anymore in the inspector :)

This is just a shrink of @zekefarwell's icon but feel free to optimize the for icon scale!


zekefarwell (approved):
Looks good to me.  Can optimize the icon later if needed but pretty low priority right now I'd say.

PR #46, opened 2021-07-20
Make sprite sheets work on relative URLs

### LINKED ISSUE #45: Sprite sheet doesn't load on gh-pages
The following sprite load code:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/7d38d24ad371e1c638d4d9ee5d9120b83a81a706/style/index.html#L84

Creates a root-referenced URL for sprites.  While this will work locally, it will break on gh-pages.  This needs to be converted into a javascript-generated absolute URL referenced to the current host name and path.

As noted in mapbox/mapbox-gl-js#9225 (and presumably proliferated into mapLibre, relative URLs are not supported.

ZeLonewolf:
Fixes #45 

This seems to work correctly on the root location.  Hopefully it works in gh-pages :)

### DIFF
--- a/style/index.html
+++ b/style/index.html
@@ -73,6 +73,9 @@
         style="position: absolute; top: 0; right: 0; border: 0; z-index: 100"
     /></a>
     <script>
+      var getUrl = window.location;
+      var baseUrl = getUrl.protocol + "//" + getUrl.host + getUrl.pathname;
+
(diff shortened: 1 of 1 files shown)


zekefarwell (approved):
Looks like it should work.  Lets give it a try!

PR #49, opened 2021-07-25
Add protected area rendering

ZeLonewolf:
This PR adds rendering for protected areas, specifically those areas rendered in the OpenMapTiles [park layer](https://openmaptiles.org/schema/#park).  Of note, this does not include `leisure=park` (that's in the land cover layer), however, since we're mostly working on the low zoom layers at this point, that can be dealt with separately.

Note that OpenMapTiles is limited to zoom 6+ for protected areas, but we would like to bring this as low as zoom 4-5.

Screen shots:

Zoom 6:
image

Zoom 7:
image

### DIFF
--- a/style/constants/color.js
+++ b/style/constants/color.js
@@ -1,3 +1,9 @@
+var colorTransparent = "hsla(0, 0%, 0%, 0)";
+
+var colorWaterFill = "hsl(211, 42%, 70%)";
+var colorWaterLine = "hsl(211, 73%, 78%)";
+var colorWaterIntermittent = "hsl(205, 89%, 83%)";
+
 var colorMotorway = "hsla(354, 71%, 40%, 1)";
 var colorMotorwayCasing = "hsla(354, 71%, 20%, 1)";
 var colorMotorwayTunnel = "hsla(354, 71%, 90%, 1)";
@@ -7,6 +13,6 @@ var colorBridgeCasing = "black";
 var colorBorder = "rgba(123, 119, 119, 1)";
 var colorBorderCasing = "rgba(224, 207, 232, 1)";
 
-var colorParkFill = "hsla(136, 41%, 89%, 1)";
-var colorParkOutline = "hsla(136, 41%, 79%, 1)";
-var colorParkLabel = "hsla(136, 71%, 29%, 1)";
+var colorParkFill = "hsl(136, 41%, 89%)";
+var colorParkOutline = "hsl(136, 41%, 79%)";
+var colorParkLabel = "hsl(136, 71%, 29%)";
--- a/style/constants/color.js
+++ b/style/constants/color.js
@@ -6,3 +6,7 @@ var colorBridgeCasing = "black";
 
 var colorBorder = "rgba(123, 119, 119, 1)";
 var colorBorderCasing = "rgba(224, 207, 232, 1)";
+
+var colorParkFill = "hsla(136, 41%, 89%, 1)";
+var colorParkOutline = "hsla(136, 41%, 79%, 1)";
+var colorParkLabel = "hsla(136, 71%, 29%, 1)";
--- a/style/index.html
+++ b/style/index.html
@@ -50,6 +50,7 @@
     <script type="text/javascript" src="layer/oneway.js"></script>
     <script type="text/javascript" src="layer/place.js"></script>
     <script type="text/javascript" src="layer/boundary.js"></script>
+    <script type="text/javascript" src="layer/park.js"></script>
 
     <script type="text/javascript" src="layers.js"></script>
   </head>
--- a/style/layer/park.js
+++ b/style/layer/park.js
@@ -3,7 +3,20 @@ var layerParkFill = {
   type: "fill",
   paint: {
     "fill-color": colorParkFill,
-    "fill-outline-color": colorParkOutline,
+  },
+  layout: {
+    visibility: "visible",
+  },
+  source: "openmaptiles",
+  metadata: {},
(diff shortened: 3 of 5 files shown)


zekefarwell:
Looks good.  We probably don't want to us this green fill for marine areas, though.  Right?

image

ZeLonewolf:
Lol, ok, that's not going to work.  We'll have to render the fill and outline separately, put the outline above the water layer, and the fill below the water layer.  That means the outline needs to work over water in addition to land.

matkoniecz:
How it will work with forest rendering?

One of thing I hate in Google Maps render is not showing forests and showing protected areas in forest rendering, making it unbelievably confusing and basically useless.

Though maybe in USA it makes sense? (my area has "landscape park" form with little protection covering huge areas, including urban areas)

ZeLonewolf:
> How it will work with forest rendering?
The method that we've discussed is fill -> land cover -> outline at zooms where we would want to render land cover.  That way you still preserve the outline because the boundary is important, and you still have a fallback render color in places where there isn't land cover available (i.e. most of the United States).

The outline-only approach used by Carto (as in: https://www.openstreetmap.org/#map=8/44.866/-112.406) does work well in the western US, which is dominated by protected areas, is thinly populated, and land cover is not well mapped, so we want to avoid that.

I would expect that we would want to render land cover at around the same zoom when hiking trails begin to appear, because outdoor activities is really the use case for which land cover information is the most useful.  At low zooms I find land cover to be more of an aesthetic decoration rather than actual useful information that should be conveyed to the map reader, but I could be convinced otherwise.

ZeLonewolf:
Below are imagery from this latest update.  I experimented with the water color in order to achieve acceptable contrast with protected area boundaries.  Please review with consideration of whether or not the water color is too strong.

Zoom 6:
image
image

Zoom 7:
image

Zoom 8:
image

Zoom 9:
image

Zoom 10:
image

Zoom 11:
image
image

Zoom 12:
image

zekefarwell (approved):
I don't think the water color is too strong.  It certainly could be lighter, but that kills the contrast with the boundary line.  Ideally I think a boundary line that is darker than the water would look better, but for that to work the water would have to very light...  and for very light water to contrast with the land properly it would need a darker outline...   and outlined water is not viable since it exposes seams between water areas.  

So I think this is good enough for now and we should merge it in.  If we come up with some better ideas later it can be improved.

zekefarwell:
For future reference, I think we may be able to style marine sanctuaries differently.  The [OpenMapTiles park layer](https://openmaptiles.org/schema/#park) sets the `class` field based on the OSM `protection_title` value.  The ones I looked at have `class` set as `national_marine_sanctuary`, `state_marine_conservation_area`, `state_marine_reserve`, `state_marine_sanctuary`, or something similar.

matkoniecz:
> For future reference, I think we may be able to style marine sanctuaries differently. The OpenMapTiles park layer sets the class field based on the OSM protection_title value. The ones I looked at have class set as national_marine_sanctuary, state_marine_conservation_area, state_marine_reserve, state_marine_sanctuary, or something similar.

Are there some relevant protection areas that are not marine sanctuary etc but including large chunk of sea/ocean/big inland lakes?

Note https://www.openstreetmap.org/relation/4524964 (not from USA, but I remembered this one)

screen

ZeLonewolf:
> Are there some relevant protection areas that are not marine sanctuary etc but including large chunk of sea/ocean/big inland lakes?

I'm sure they exist all over the US.  Here's an example of a protected area boundary crossing a pond:
https://zelonewolf.github.io/openstreetmap-americana/#15/41.639/-71.86887

zekefarwell:
Many protected area boundaries cross water and land, so the styling absolutely has to contrast well with both.  Using a different style for marine sanctuaries is not a solution to that problem.  However, since marine sanctuaries are entirely over water a different style could look much better especially at low zoom levels.  

On the [coast of Massachusetts](https://www.openstreetmap.org/#map=9/42.1522/-70.1010) for example, there are two large protected areas.  The boundary of Cape Cod National Seashore crosses land and water, but stays close to the coast.  At low zoom this is barely visible and the boundary line just looks like it's on the shore.  On the other hand, the boundary of Stellwagen Bank National Marine Sanctuary is entirely in the ocean.  The [coast of California](https://www.openstreetmap.org/#map=8/37.167/-122.283) also has many of these marine protected areas sticking far out into the ocean.

Using a blue outline where the boundary is entirely or mostly over water would be a nice cartographic effect and less distracting than a green outline in the middle of the ocean.  We could also decide to simply exclude marine sanctuaries on the basis that this map style focuses on land based features.

Screen Shot 2021-07-26 at 9 47 32 AM
Screen Shot 2021-07-26 at 9 47 53 AM

ZeLonewolf:
The effect I'm envisioning is something similar to how marinas currently render in openstreetmap-carto.  I don't think mapLibre has a string "contains" function, however, I would not be opposed to creating an exhaustive list of protection titles that equate to marine areas and using that in a giant case statement in the style.  Then, we could link it to the open issue in mapLibre expressing the need for a regex function.

Also, for the case of marine protected areas that include small bits of the shore or wetlands, I suspect it will be perfectly fine for the blue boundary to go partially over land.

Further, for render simplicity sake, perhaps we could simply render the marine boundary/label above the general protected area boundary styling, provided that we use the same or thicker-width line, so there will be no need to evaluate the mega-case-statement twice.

Since we'll need multiple copies of the mega-case-statement, this will be a good use case for separating it out into a separate variable in the javascript.

PR #56, opened 2021-08-02
Implement trunk rendering

ZeLonewolf:
This PR implements rendering for trunk roads.

image
image
image
image
image

Once trunk rendering is complete, the next step is to look across motorway and trunk to see how we can refactor out all the copy/paste in the road layer (after just two highway classes so far!)


zekefarwell (approved):
Looks good to me.  I'm definitely feeling the need for some refactoring as well and I've got some ideas about how to approach it.  I probably won't have time until next week though.

ZeLonewolf:
I've got a refactor branch in progress, recommend you hold off until I post it so we don't get cross-threaded.

PR #57, opened 2021-08-07
Road refactor

ZeLonewolf:
This PR is a massive refactor of the road layer.

Road styles will now live in `road.js`, with the style settings abstracted away.  Because the road layer is so repetitive, it makes sense to refactor out all of the copy/paste in a way that makes it easier for contributors to modify the style elements without having to parse through the piles of copypasta.

This also makes a handful of renames just for the sake of naming consistency.


zekefarwell (dismissed):
I very much like how this PR reduces code repetition and I think it's an improvement over the current state of things.  That said, the further abstracted away from the [MabLibre Style Specification](https://maplibre.org/maplibre-gl-js-docs/style-spec/) we get, the less approachable it feels.  Ideally I'd like to iterate and improve on this before merging in, but I don't feel I'm going to have the time to work on that in the short term and I don't want to hold things up.  Merge away if this works better for you.

I went through the motions of adding rendering for primary roads under this paradigm, and found it a bit confusing but it does make sense now that I've worked through it.  The layer generator functions feel a bit like a black box to me.  If you haven't added all right key names to the `roadMinZoom`, `roadWidth`, `roadFillColor`, and `roadCasingColor` objects then you get errors.  These errors come from MapLibre complaining about malformed style data (the output of the generator function), so it's not entirely clear where the mistakes you've made are.

I think to me it would make more sense to have an object for each road type and have the key names be for the attributes being styled, or maybe set up with class inheritance.  Perhaps something like this:

```
class Motorway {
  fill_color = "...";
  fill_width = "...";
  casing_color = "...";
  casing_width = "...";
}
class MotorwayBridge extends Motorway {
  casing_color = "...";
}
class MotorwayTunnel extends Motorway {
  fill_color = "..."
  casing_color = "...";
}
class MotorwayLink extends Motorway {
  fill_width = "...";
  casing_width = "...";
}
class MotorwayLinkBridge extends MotorwayLink {
  casing_color = "...";
}
class MotorwayLinkTunnel extends MotorwayLink {
  fill_color = "..."
  casing_color = "...";
}
```

ZeLonewolf:
@zekefarwell updated version based on your suggestions.  What do you think?

style/js/util.js
```diff
@@ -5,10 +5,19 @@ function restrictLayer(def, layer) {
   return filteredClone(def, ["==", "layer", layer], "_layer_" + layer);
 }
 
+function cp(obj) {
+  return JSON.parse(JSON.stringify(obj));
```
jleedev, line 9: For reference:

The lodash method `cloneDeep`: https://lodash.com/docs/4.17.15#cloneDeep

Coming soon to browsers near you: `structuredClone`: https://www.chromestatus.com/feature/5630001077551104

style/layer/road.js
```diff
@@ -0,0 +1,367 @@
...
+};
+let layoutRoadCase = {
+  "line-cap": "butt",
+  "line-join": "round",
+  visibility: "visible",
+};
+
+//Highway class styles
+class Motorway {
+  constructor(link) {
```
jleedev, line 30: Looks like this `link` parameter is not used in practice.

style/layer/road.js
```diff
@@ -0,0 +1,367 @@
+//At this zoom, render switches from unified to differentiated bridge/tunnel rendering
```
jleedev, line 1: Get in the habit of putting

```js
"use strict";
```

at the top of js files. (Or switch to ES modules / typescript).

jleedev (commented):
On the whole this is a positive refactor.

The main concerns are about maintainability, reducing copy/paste errors, etc. To that end I'd consider one or both of:

- Run this through some sort of build process, be it TypeScript or ESBuild, simply to catch basic goofs early on.
- Add tests with Jest, purely in the interest of basic checks such as "all of the road layers have casing and fill methods, which return objects with a couple key properties".

Never mind actually testing anything in particular, these can go a long way to verifying the most basic properties of "I'm importing what I think I'm importing", not clobbering variables or dropping everything in the global namespace, having to rely on naming conventions, ensuring that temporary variables aren't being unnecessarily exported, etc.

style/layer/road.js
ZeLonewolf: Just added this to all of the project's js files.  I wasn't sure what "switch to ES modules / typescript" means or which approach is preferred.

jleedev: Note that the bodies of classes are always in strict mode, but this directive enables it for the file as a whole. https://mdn.io/Strict_mode

As briefly as possible:

At the moment, the modules are all linked together via "bunch of script tags", and the dependencies are all implicitly via global variables; writing any sort of tests would be somewhat cumbersome as there's nothing to import as such.

So using some sort of module system is desirable. https://mdn.io/Modules

The main issue is javascript's historical lack of any module system. So 'CommonJS' was developed in the nodejs ecosystem, and build systems exist to port all of these `require()` calls into something that the browser can execute, and also combining code into one big file, and eliminating dead code, etc. Various compilers and compiler accessories exist (babel, gulp, grunt, rollup, webpack, etc.) but two worth mentioning are:

- TypeScript: Adds useful things at the compile step by adding static type annotations, has serious traction as a typed layer on top of javascript.
- ESBuild: Modern and very fast. Some of these compilers have been notoriously slow.

And now, javascript supports real modules (with `import`), and they can be natively supported by browsers without a build step. Which is cool. But a build step is still useful to have in order to get bundling, dead code elimination, type checking, etc. And you can get that with TypeScript or ESBuild as well. And your build target can still be ES modules (for modern browsers and toolchains) or commonjs for historical reasons.

So, if your source files are ES modules, then they're always in ES strict mode. And you really always want to be in strict mode. So, add the directive, or switch your source files to ES modules.

style/layer/road.js
```diff
@@ -0,0 +1,367 @@
...
+      6,
+      `hsl(${this.hue}, 60%, 50%)`,
+      minzoomBrunnel - 0.5,
+      `hsl(${this.hue}, 71%, 40%)`,
+    ];
+
+    this.fillColor = this.colorFillLowZoom.concat(
+      14,
+      `hsl(${this.hue}, 71%, 45%)`
+    );
```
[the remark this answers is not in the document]
ZeLonewolf: Cool!  I didn't even know javascript could do that.

style/layer/road.js
zekefarwell: I'm all for switching to ES modules 👍.  Would be nice to not have everything in the global namespace.

style/layer/road.js
```diff
@@ -0,0 +1,367 @@
...
+      layer.paint = tunnelCasePaint(this.casingColor, this.casingWidth);
+    } else {
+      layer.paint = roadPaint(this.casingColor, this.casingWidth);
+    }
+    return layer;
+  },
+};
+
+Object.setPrototypeOf(Motorway.prototype, Road);
+Object.setPrototypeOf(Trunk.prototype, Road);
```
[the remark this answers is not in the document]
ZeLonewolf: Road is a constant, not a class so I can't just extend it, at least the way it's currently implemented.

style/layer/road.js
ZeLonewolf: Oh, I was wondering if there was a better way to deal with the big pile of script tags.  That _does_ seem a lot easier.

style/layer/road.js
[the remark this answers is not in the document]
jleedev: Changing Road to a class ought to work just fine.

(There might be more complex cases that call for using mixins such as this, but it would be difficult to analyze statically.)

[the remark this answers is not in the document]
ZeLonewolf: Cool, that totally worked!

style/layer/road.js
ZeLonewolf: Removed useless constructor.

style/layer/road.js
ZeLonewolf: Okay, I've made the switch to ES modules.  Way cleaner now!

style/js/util.js
ZeLonewolf: Based on discussion on Slack, pulling in a 3rd party library just for the clone function is probably not worth the squeeze, but we should use a library clone function if we pull in one of these libraries for some other reason in the future.

ZeLonewolf:
New version ready for review.  I hope the change to ES modules doesn't overcomplicate the review too much, but it sure does improve overall readability.

style/americana.js
```diff
@@ -0,0 +1,164 @@
...
+  version: 8,
+  metadata: {
+    "mapbox:type": "template",
+    "maptiler:copyright":
+      "This style was generated on MapTiler Cloud. Usage outside of MapTiler Cloud requires valid OpenMapTiles Production Package: https://openmaptiles.com/production-package/ -- please contact us.",
+    "openmaptiles:version": "3.x",
+  },
+};
+
+var map = new maplibregl.Map({
```
jleedev, line 149: For debugging I might like `let map = window.map = new …`. The downside of modules is you don't implicitly get global variables, naturally.

(Technically this can be accomplished by making it an export but you'd need to write the correct import at the console.)

jleedev (approved):

zekefarwell (approved):
This is no longer just a road refactor.  It's an everything refactor 😀!  The switch to ES Modules looks great to me, and the structure for roads is much improved over the current state.

PR #61, opened 2021-12-06
Improve keyboarding

jleedev:
Focus the map canvas so that keyboard (arrows, +/-) work immediately
without having to click.


ZeLonewolf (approved):
Looks good and seems to be working here.

PR #62, opened 2021-12-20
Implement red/black road color scheme for trunk through tertiary

ZeLonewolf:
This is an initial implementation of a red/black roads color scheme for trunk through tertiary.  An initial roads implementation allows for highway shield development to proceed on lower-classification roads and helps promote community participation.

At the middle zooms, the scheme is:

`trunk`: thick red
`primary`: thick black
`secondary`: medium black
`tertiary`: thin black

At higher zooms, the lower classifications change to a white fill with black casing and inside labelling.  This PR implements that change as a progressive change, where primary through tertiary change to casing at progressively higher zooms.

**Unresolved issues**
* There is an issue with certain corner joins leaving undesired artifacts.  This is a somewhat complex problem and can't be easily solved without also causing undesired artifacts on bridge layering.  I'm leaving that issue as "good enough for now, can be resolved in a future bugfix.
* Bridges are undifferentiated for trunk through tertiary, and tunnels are not implemented.  This refinement can be worked in future PRs.
* At some zooms, we will show a combination of cased and stroked roads.  This is a somewhat unique approach as we weren't able to find examples of city-level maps that used both styles.  So we need to decide if this combination sufficiently meets an American design aesthetic, or if everything needs to flip over to cased at a single zoom level.  In the future, we would like to render "actual" road widths based on `width` and/or `lanes` tagging, but that data is not presently available in the vector tiles. 

### Screenshots

**Zoom 8**:
image

**Zoom 9**:
image

**Zoom 10**:
image
image
image

**Zoom 11**:
image
image
image

**Zoom 12**
image
image
image

**Zoom 13**
image

**Zoom 14**
image

**Zoom 15**
image

**Zoom 16**
image

### DIFF
--- a/style/americana.js
+++ b/style/americana.js
@@ -190,3 +190,4 @@ map.addControl(
   })
 );
 map.addControl(new maplibregl.NavigationControl(), "top-left");
+document.querySelector("#map canvas").focus();
--- a/style/americana.js
+++ b/style/americana.js
@@ -148,7 +148,9 @@ americanaLayers.push(
   lyrRoadLabel.primary,
   lyrRoadLabel.primaryHZ,
   lyrRoadLabel.secondary,
+  lyrRoadLabel.secondaryHZ,
   lyrRoadLabel.tertiary,
+  lyrRoadLabel.tertiaryHZ,
 
   lyrPark.label,
 
--- a/style/americana.js
+++ b/style/americana.js
@@ -51,30 +51,51 @@ americanaLayers.push(
 
   lyrRoad.motorway.casing(),
   lyrRoad.trunk.casing(),
+  lyrRoad.primary.casing(),
+  lyrRoad.secondary.casing(),
+  lyrRoad.tertiary.casing(),
 
   lyrRoad.motorwayLink.casing(),
   lyrRoad.trunkLink.casing(),
 
-  lyrRoad.motorway.fill(),
-  lyrRoad.trunk.fill(),
-
   lyrRoad.motorwayLink.fill(),
   lyrRoad.trunkLink.fill(),
+
+  lyrRoad.tertiary.fill(),
+  lyrRoad.secondary.fill(),
+  lyrRoad.primary.fill(),
+  lyrRoad.trunk.fill(),
+  lyrRoad.motorway.fill(),
 
   lyrOneway.road,
   lyrOneway.link
 );
 
 var bridgeLayers = [
-  lyrRoad.motorwayBridge.casing(),
-  lyrRoad.motorwayLinkBridge.casing(),
-  lyrRoad.motorwayBridge.fill(),
-  lyrRoad.motorwayLinkBridge.fill(),
+  lyrRoad.tertiaryBridge.casing(),
+  lyrRoad.tertiaryLinkBridge.casing(),
+  lyrRoad.tertiaryBridge.fill(),
+  lyrRoad.tertiaryLinkBridge.fill(),
+
+  lyrRoad.secondaryBridge.casing(),
+  lyrRoad.secondaryLinkBridge.casing(),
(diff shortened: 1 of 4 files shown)


adamfranco:
This is certainly something that could be tweaked later, but I find myself torn between thinking that a next level lower needs to be shown sooner (e.g. `primary` should show up at z=9 and secondary at z=10) and thinking that these zoom choices help highlight where we need more consistent network connectivity improvements to the data. 🤔

style/americana.js
```diff
@@ -104,6 +125,11 @@ bridgeLayers.forEach((layer) =>
 americanaLayers.push(
   //The labels at the end of the list have the highest priority.
```
jleedev, line 126: For my understanding of maplibre gl — just as later layers paint on top of previous ones and obscure them, the later text layers will take priority and suppress text/symbol layers before them?

style/americana.js
```diff
@@ -133,7 +159,8 @@ var style = {
   bearing: 0,
   sources: {
     openmaptiles: {
-      url: "https://api.maptiler.com/tiles/v3/tiles.json?key=" + mapTilerKey,
+      url: "http://localhost:8080/data/v3.json",
+      //      url: "https://api.maptiler.com/tiles/v3/tiles.json?key=" + mapTilerKey,
```
jleedev, line 163: Consider: https://www.npmjs.com/package/dotenv

So that you can put the localhost URL in your working directory `.env` file (and `.gitignore` it), and the deployment process can use the production URL.

style/americana.js
```diff
@@ -155,11 +182,15 @@ var map = new maplibregl.Map({
...
+});
+
 map.addControl(
   new maplibregl.AttributionControl({
     customAttribution:
       '<a href="https://openmaptiles.org/" target="_blank">&copy; OpenMapTiles</a> <a href="https://www.maptiler.com/copyright/" target="_blank">&copy; MapTiler</a> <a href="https://www.openstreetmap.org/copyright" target="_blank">&copy; OpenStreetMap contributors</a>',
   })
 );
 map.addControl(new maplibregl.NavigationControl(), "top-left");
-document.querySelector("#map canvas").focus();
```
jleedev, line 165: Deliberate?

Also: `map.getCanvas().focus()`.

style/americana.js
```diff
@@ -155,11 +182,15 @@ var map = new maplibregl.Map({
   zoom: 4, // starting zoom
   attributionControl: false,
 });
+
+map.on("styleimagemissing", function (e) {
+  Shield.missingIconLoader(map, e);
```
jleedev, line 187: Where is `Shield` defined?

jleedev (changes requested):

ZeLonewolf:
> I find myself torn between thinking that a next level lower needs to be shown sooner

I tend to agree.  I think the current zooms feel okay for what's prominent and labelled, but perhaps we could have thin/faint lines drawn at lower zooms to give an indication that "there's roads here, and zoom in for more detail on it".

style/americana.js
ZeLonewolf: Hmm, I would have assumed the earlier layers take priority since they're already placed.

style/americana.js
ZeLonewolf: Whoops, this was an accidental import from my highway shields branch.

jleedev (approved):

style/americana.js
ZeLonewolf: Looks like a merge error, I've reverted that delete.

zekefarwell (dismissed):
This looks good to me.  Plenty more refinement can be done I'm sure, but I think this is a good initial pass to get more of the road network showing up.

zekefarwell (approved):

PR #65, opened 2021-12-22
Lower-zoom roads

adamfranco:
Introduce roads at lower zooms to provide context. trunk-tertiary roads are introduced as soon as they are available in OMT, with widths tweaked so that they appear more fine than already displayed roads.

Before z=7:
road_layer-z7

After z=7:
lower-zoom-roads-z7


ZeLonewolf:
This auto-closed when I deleted the destination branch, but feel free to re-post the PR to main.

PR #66, opened 2021-12-25
Low zoom roads

adamfranco:
Introduce roads at lower zooms (when they first become available in OMT).

1. d2c21277 introduces roads at lower zooms, maintaining the same styling...
2. 80af93fc interpolates the luminosity of the casing color between a road's introduction and the zoom at which the next level is introduced. This lets the road fade in with less visual weight initially, then get fully black as the next level fades in.

Introducing `trunk` earlier is vital for remote regions like Canada, much of the rural US, and other places without a contiguous motorway network.

For lower levels, without this earlier introduction many zooms in rural areas show very few roads, not enough for usable context at the current zoom. See [this directory with screen-shots](https://drive.google.com/drive/folders/1-5L4eektpgZzPnQptwPJ3lIDza8qSVqP?usp=sharing) of both rural (northern New England) and urban (NYC) before at each of these two commits.

### Example, New England z5
Before:
NE-z5-A-before

After:
NE-z5-B-after

### Example, New England z9
Before:
NE-z9-A-before

After (single color):
NE-z9-B-after

After (with color interpolation):
NE-z9-C-after-interpolation

### Example, NYC z9
Before:
NYC-z9-A-before

After (single color):
NYC-z9-B-after

After (with color interpolati
NYC-z9-C-after-interpolation
on):

### DIFF
--- a/style/layer/road.js
+++ b/style/layer/road.js
@@ -243,7 +243,7 @@ class Trunk extends Road {
     this.link = false;
     this.hue = 0;
 
-    this.minZoomFill = 7;
+    this.minZoomFill = 5;
     this.minZoomCasing = 15;
 
     this.fillWidth = trunkFillWidth;
@@ -264,7 +264,7 @@ class Primary extends Road {
     this.hue = 0;
 
     this.minZoomFill = 14;
-    this.minZoomCasing = 10;
+    this.minZoomCasing = 7;
 
     this.fillWidth = Util.zoomMultiply(trunkFillWidth, 0.9);
     this.casingWidth = Util.zoomMultiply(trunkCasingWidth, 0.9);
@@ -284,7 +284,7 @@ class Secondary extends Road {
     this.hue = 0;
 
     this.minZoomFill = 15;
-    this.minZoomCasing = 11;
+    this.minZoomCasing = 9;
 
     this.fillWidth = Util.zoomMultiply(trunkFillWidth, 0.6);
     this.casingWidth = Util.zoomMultiply(trunkCasingWidth, 0.6);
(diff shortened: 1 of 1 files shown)


adamfranco:
Updated comment with link to a [directory with screen-shots](https://drive.google.com/drive/folders/1-5L4eektpgZzPnQptwPJ3lIDza8qSVqP?usp=sharing) for toggling back and forth between views.

ZeLonewolf (approved):
I think this looks great.  I've been experimenting with the unreleased OpenMapTiles 3.13 to see how we might finish out the low zoom scheme to z4.

Here is z4, showing only interstate motorways:
image

I think for the next zooms we want:

* **z5**: All motorway+trunk, however non-interstate motorway are styled identically to trunk.  This will fix the clutter that we currently have at z5.  I haven't written the code yet to do this so I don't have a sample render to share.
* **z6**: Probably the way it's composed in this PR.  However, based on experimentation, we might want to extend the z5 styling to z6.

In short -- in the absence of the new OMT 3.13 feature for filtering by network, this looks good to me with what's currently available in the public tiles and I see no reason not to merge it.

adamfranco:
> I think for the next zooms we want:
> 
> - z5: All motorway+trunk, however non-interstate motorway are styled identically to trunk. This will fix the clutter that we currently have at z5. I haven't written the code yet to do this so I don't have a sample render to share.
> - z6: Probably the way it's composed in this PR. However, based on experimentation, we might want to extend the z5 styling to z6.

This is a neat idea, Brian! At z5 I see connectivity as more important than road structure, especially with all the little spurs and bypasses.  With future OMT work do you see the possibility of getting select networks (e.g. Interstates and international equivalents) into the z4 or z3 tiles? I'd love to see the 2-digit Interstate network at continent scale. 😃

ZeLonewolf:
> With future OMT work do you see the possibility of getting select networks (e.g. Interstates and international equivalents) into the z4 or z3 tiles? I'd love to see the 2-digit Interstate network at continent scale.
Completely feasible from a technical perspective.  It would probably even look right at z4 because most of the 3-digit interstates are part of the balls of yarn around cities.  More challenging I think is figuring out the international equivalents on the OpenMapTiles side.

zekefarwell:
Didn't get a chance to look at this earlier, but I just want to say these changes look really good!  Thanks for working on this, Adam!

Issue #64, opened 2021-12-20
Convert URL and key to .env


ZeLonewolf:
Consider: https://www.npmjs.com/package/dotenv

So that you can put the localhost URL in your working directory `.env` file (and `.gitignore` it), and the deployment process can use the production URL.

_Originally posted by @jleedev in https://github.com/ZeLonewolf/openstreetmap-americana/pull/62#discussion_r772452753_

zekefarwell:
I'm a fan of this approach.  We should use it for the [mapTilerKey](https://github.com/ZeLonewolf/openstreetmap-americana/blob/ed6f0beb259e7df2c110272c06ae29b86cf7f53f/style/key.js#L11) as well.

ZeLonewolf:
I agree with those two values (URL and maptiler key).
I wasn't sure about the whole npm business and how we actually go about adding a package.

ZeLonewolf:
Fixed in #67

PR #67, opened 2021-12-26
Set up git ignored file for config variables

zekefarwell:
This is an alternative approach to resolving #64 that doesn't use [dotenv](https://www.npmjs.com/package/dotenv).

I started looking into dotenv, but it seems that a node.js server or build process would need to be added to the project if we want to use it.  At the moment neither of these is present and all JS is executing client side in the browser.  In this context, dotenv doesn't appear to work.

In the mean time I think we can get most of the same benefits without this library by setting up a git ignored JS config file.  This PR adds `style/config.default.js` as a template to be copied to `style/config.js` (a git ignored location) where you modify the variables.  This file is imported as a module and then variables can be accessed as `config.VAR_NAME`. 

Does this approach seem sufficient for now, or should we be looking into a build process so we can use the dotenv library?

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -20,8 +20,9 @@ jobs:
           sudo apt install -y xmlstarlet
           cd style
           npm install
-          echo 'var mapTilerKey = "hJFMGrd2JszhwfJx5M2k"; //This key only allows requests from zelonewolf.github.io' > key.js
+          sed 's/<your MapTiler key>/53iZvB2drcamS0Ge0xiD/g' config.default.js > config.js
           make sprites
+        # MapTiler key 53iZvB2drcamS0Ge0xiD only allows requests from zelonewolf.github.io
 
       - name: Deploy 🚀
         uses: JamesIves/github-pages-deploy-action@4.1.4
--- a/style/.gitignore
+++ b/style/.gitignore
@@ -3,4 +3,4 @@ sprites
 build
 rebusurance.zip
 .DS_Store
-
+config.js
--- a/style/CONTRIBUTE.md
+++ b/style/CONTRIBUTE.md
@@ -13,22 +13,22 @@ The repository is organized as follows:
 
 ## Config File
 
-Environment specific settings go in the untracked file `config.js`.  Copy the template
-`config.default.js` and rename it `config.js`.  The variables in this file can then 
+Environment specific settings go in the untracked file `config.js`. Copy the template
+`config.default.js` and rename it `config.js`. The variables in this file can then
 be changed without the risk of accidentally comitting to the main repo.
 
 ### MapTiler API Key
 
 By default this project is set up to use vector tiles provided by MapTiler.
-For this to work, you must create an account and obtain a free key from 
-[MapTiler Cloud][20]. This key should be pasted into the `MAPTILER_KEY` variable of 
+For this to work, you must create an account and obtain a free key from
+[MapTiler Cloud][20]. This key should be pasted into the `MAPTILER_KEY` variable of
 the `config.js` file.
 
 ### Custom OpenMapTiles URL
 
 For testing upcoming features of the [OpenMapTiles schema][21] or for fresher data than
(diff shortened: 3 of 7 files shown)


jleedev (dismissed):
At some point might end up with an esbuild process anyway for the purpose of having a releasable artifact, but this certainly works.

zekefarwell:
If we're going to go with this approach for the time being there should be be some changes to [style/CONTRIBUTE.md](https://github.com/ZeLonewolf/openstreetmap-americana/blob/main/style/CONTRIBUTE.md) about where to put the MapTiler key before this gets merged.

ZeLonewolf:
This looks good to me.  It separates the key stuff to its own file so there's no danger of committing a key with other changes, which was really the big issue.  It sounds like we just need to update the CONTRIBUTE.md to add the key to config.default.js instead of key.js?

zekefarwell:
Ok, I've updated CONTRIBUTE.md to explain how the config file works so this should be all set to merge.  I think the changes I made to [.github/workflows/deploy.yml](https://github.com/ZeLonewolf/openstreetmap-americana/pull/67/files#diff-28802fbf11c83a2eee09623fb192785e7ca92a3f40602a517c011b947a1822d3) should ensure the github pages deploy doesn't break, but it's hard to know without a way to test.

jleedev:
> ensure the github pages deploy doesn't break, but it's hard to know without a way to test.

Could make the action run on pull requests save for the deploy step. There are examples somewhere…

ZeLonewolf (approved):
Looks good to me.  If I understand correctly how `make` works, the `make config.js` won't overwrite a customized config.js.

zekefarwell:
>  If I understand correctly how make works, the make config.js won't overwrite a customized config.js
That's correct.  My thinking was that if someone runs `make run` without having copied `config.default.js` to `config.js` first.  This console error is a bit friendlier than just getting a 404 for `config.js`:

image

So have `make` copy the file for you if it's missing.

ZeLonewolf:
Only one way to see if this works.

zekefarwell:
Looks like the deploy didn't work properly.  I expected the config.js file to get committed back to the gh-pages branch.  It's not there though.
https://github.com/ZeLonewolf/openstreetmap-americana/tree/gh-pages

zekefarwell:
Must be because the file is git ignored.  So although it gets copied, it doesn't end up in the final deployed file set since the target is a git branch.  ugh.

ZeLonewolf:
Can we commit a different .gitignore to the gh-pages branch?

zekefarwell:
Maybe?  The whole idea of a git branch being the deploy target is a bit confusing to me because of things like this.

zekefarwell:
how are other ignored files handled?  like the whole sprites directory for example?

ZeLonewolf:
Looks like we need to add a `make config.js` to the github workflow here:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/c21037e3cf3c19365ec2fc3f9690afbb4b8c1d49/.github/workflows/deploy.yml#L19-L25

ZeLonewolf:
Oh wait, it's already there.

zekefarwell:
It looks like the way you handled the sprites is by manually committing them to the gh-pages branch despite them being ignored.
https://github.com/ZeLonewolf/openstreetmap-americana/tree/gh-pages/sprites
Since they are git ignored my guess is that will have to be done any time the sprites change even though the `make sprites` command is part of the deploy action.  Same with `config.js`.

zekefarwell:
Ok I manually committed the config.js file with the proper key to gh-pages branch and that appears to have fixed the production map
https://github.com/ZeLonewolf/openstreetmap-americana/commit/aea8d93735fd9b9683cb8c2b2d25a03f873d3f28

ZeLonewolf:
I think we can get rid of the export of MAPTILER_KEY, right?

zekefarwell:
Yup, that should be fine.  I don't think it's used anywhere outside of the config module so it doesn't really need to be exported.

PR #72, opened 2022-01-02
OpenMapTiles 3.13 Features: Highway shields, Interstate filtering

ZeLonewolf:
Fixes #6 
Fixes #69 
~~Blocked by openmaptiles/openmaptiles#1287~~ **Update:** OpenMapTiles 3.13 is released, MapTiler Cloud planet update is expected during the week of Jan 10, 2022.
Blocked by maplibre/maplibre-gl-js#188 with pending fix maplibre/maplibre-gl-js#716.  However, we can proceed with a development version of maplibre taken from that branch until it's released.

### Description

This branch contains work that is dependent on the upcoming release of OpenMapTiles 3.13.  This description will be updated over time with screen shots of the pending features.  PRs that are dependent on OMT 3.13 can be posted to this branch.  This PR will be kept in a "draft" status until OMT 3.13 is released.

Features included in this branch:

* Concurrent highway shields, with partial state-specific shield implementations for several states.  This change removes the import of shield graphics from rebusurance in favor of style-specific shields.
* Suppress non-interstate motorways from the lowest zoom
* Wider shields and/or text scaling for longer refs
* Ability to re-color a base shield, such as US:US:Historic shields or Georgia ADHS state routes

The following features are incomplete:
* Complete implementation of state-specific shields
* Highway shields outside of the US

### Caveats

* ~~This branch includes a development version of maplibre-gl-js taken from maplibre/maplibre-gl-js#716, which is incompatible with the license of openstreetmap-americana.  This must be resolved prior to merging.~~ (Resolved)
* Padding is needed between adjacent concurrent shields.  Due to issues with front-end rendering, this needs to be implemented as a perimeter pad on generated shield imagery.

### How to run this branch

1. Clone the current master of OpenMapTiles from https://github.com/openmaptiles/openmaptiles
2. Modify .env to set the desired zoom level to render to (defaults to 7, can go as high as 14)
3. Run `./quickstart <name of area>` to generate vector tiles (for example, `./quickstart rhode-island`)
4. Run `make start-tileserver` which starts a tile server on localhost:8080
5. In openstreetmap-americana, create a config.js that looks like this:
```
const OPENMAPTILES_URL = `http://localhost:8080/data/v3.json`;
export default {
  OPENMAPTILES_URL,
};
```
6. Run `make run` to run the style

### Screen shots

**Zoom 4 Interstate-only map of the US**
image

**Multiple-widths of highway shields (Interstates)**
image

**Highway shields, California**
image

**Highway shields, Minnesota**
image

[...]

PR #77, opened 2022-01-05
Build custom maplibre-gl-js on the fly

ZeLonewolf:
This PR builds a custom version of maplibre-gl-js which supports upright highway shields.  This is intended to ensure that #72 will be able to use a deployed maplibre-gl-js on the gh-pages branch.  This PR intentionally does not change the version of maplibre that index.html is pointing to, in case the github actions don't work properly - this is only intended to pre-stage the modified maplibre build to ensure that it can deploy.

### DIFF
--- a/style/Makefile
+++ b/style/Makefile
@@ -23,7 +23,7 @@ code_format:
 
 build/maplibre-js:
 	mkdir -p build/maplibre-js
-	git clone git@github.com:wipfli/maplibre-gl-js.git build/maplibre-js
+	git clone https://github.com/wipfli/maplibre-gl-js build/maplibre-js
 
 build/maplibre-js/dist/maplibre-gl.js: build/maplibre-js
 	cd build/maplibre-js; git checkout 270a47f473ba51375a7df4076ee271777cc04467; \
--- a/style/Makefile
+++ b/style/Makefile
@@ -1,5 +1,6 @@
 clean:
 	rm -rf sprites build icons/us_* rebusurance.zip
+	rm -rf build
 
 rebusurance.zip:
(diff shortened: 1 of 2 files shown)


style/Makefile
```diff
@@ -20,5 +21,16 @@ config.js:
 code_format:
 	npx prettier --write .
 
-run: sprites config.js
+build/maplibre-js:
+	mkdir -p build/maplibre-js
+	git clone git@github.com:wipfli/maplibre-gl-js.git build/maplibre-js
```
jleedev, line 26: This unnecessarily requires having an SSH key on file with github, which shouldn't be needed for an anonymous clone.
```suggestion
	git clone https://github.com/wipfli/maplibre-gl-js build/maplibre-js
```

style/Makefile
```diff
@@ -20,5 +21,16 @@ config.js:
...
+
+build/maplibre-js/dist/maplibre-gl.js: build/maplibre-js
+	cd build/maplibre-js; git checkout 270a47f473ba51375a7df4076ee271777cc04467; \
+          npm ci && npm run build-prod-min;
```
jleedev, line 30: I will note that this installs way too much crap that isn't needed to simply build the prod js and css, but this is of course temporary until the named commit is merged and released.

jleedev (changes requested):

style/Makefile
ZeLonewolf: That's just local/temporary, right?  I assume this doesn't deploy node_modules.

jleedev: Yes of course, just idly complaining.

(Java? Ah, for some diagramming tool in the docs.)

jleedev (approved):

ZeLonewolf:
Well, that didn't work 🤕

PR #82, opened 2022-01-19
Render Expressways in their own style

### LINKED ISSUE #22: Render expressways with a different style than standard highways
Highway maps commonly show expressways–also known as limited-access, controlled-access, or divided highways–with a distinct line style from standard two lane highways.  Beyond simply using a different color, expressway lines are usually thicker and have higher contrast casing.  Some examples:

conn mass legend
Kansas DOT 2009–2010 legend
legend-image
Oklahoma DOT legend
PennDOT legend
PennDOT highway map

Other than `highway=motorway` OSM `highway` classifications are functional, not physical, so there are both `trunk` and `primary` expressways (I don't think `secondary` expressways exist but I suppose it's possible).  So to indicate the physical characteristics of an expressway we have [a separate tag, `expressway=yes`](https://wiki.openstreetmap.org/wiki/Key:expressway).  It would be great for this style to be the first to render expressways distinctly based on this tag.  Below are some rendering ideas for how this could look.

<br>

### Low zoom where individually mapped carriageways merge together as one line

expressway-highway-low-zoom

<br>

### Higher zoom where individually mapped carriageways are visible

expressway-highway-high-zoom

All motorways are expressways, so `highway=motorway` implies `expressway=yes` and would always be rendered with the expressway line style.  trunk` and `primary` would have a standard highway line style and an expressway line style that would be used when `expressway=yes` is set.  

Currently the `expressway` tag is primarily used in North America, but the concept is applicable elsewhere as well.  In other parts of the world [the `motorroad` tag](https://wiki.openstreetmap.org/wiki/Key:motorroad) is far more common.  While `expressway` defines the physical construction of a highway, `motorroad` defines legal access restrictions (i.e. a road where only for motor vehicles are allowed).  Although the definitions are different, in practice `motorroad=yes` is often applied to the same or very similar kinds of limited access, divided highways as `expressway=yes`.  So for this map style we _**may**_ want to interpret `motorroad=yes` as implying `expressway=yes` when the tag is absent and render these highways with the expressway line style.  In the case of an explicit `expressway=no` the standard highway line style would be used.

**Update 2021-11-04:**
My thinking on the above has changed and I'm not sure if it is actually a good idea to assume that `motorroad=yes` implies `expressway=yes` when `expressway=no` is not present.  I more clearly understand now that the reason for the `motorroad=yes` tag is that in countries where it is used, access restrictions are not explicitly signed on these roads.  Instead there is a blue sign with an icon of a car which means "motor vehicles only".  Since the sign doesn't say exactly what modes of transport are prohibited it's hard for mappers to set all the right access tags (`horse=no`, `bicycle=no`, etc).  In North America on the other hand, access restrictions generally are explicitly signed (No pedestrians, No bicycles, etc) so it is easy for mappers to tag them explicitly and there is no need for the `motorroad` tag.  All this is to say that while `motorroad=yes` is tagged on many roads that physically qualify as expressways, it also is tagged on some that do not so I'm not sure we should use it for this rendering distinction.

adamfranco:
This PR fixes #22. Highways with `expressway=yes` are rendered with a heavy casing and a light fill to contrast with non-expressway (solid) roads. The fill width has also been narrowed slightly to allow the casing stroke to be perceived as heavier and more significant compared to the normal think black casing of non-expressway roads.

Here are some screen-shots of this rendering on `trunk`, `primary`, and `secondary` highways. `tertiary` is also supported, but I couldn't find an example to test with.

At low zooms, the casing and fill widths of trunks were chosen to put their visual weight halfway between non-expressway trunks and motorways:
Screen Shot 2022-01-19 at 12 22 55 AM

At medium zooms, expressways are clearly delineated from non-expressways: 
Screen Shot 2022-01-19 at 12 21 49 AM

At high zooms, the extra casing width (and light fill on trunks) distinguishes expressways:
Screen Shot 2022-01-19 at 12 25 50 AM


zekefarwell (approved):
I think this is a good starting point and we should go ahead and merge it in because it's great to be able to see where the `expressway=yes`'s are.  It introduces some weird rendering issues but I think a future PR can address them.

ZeLonewolf:
What is the bridge/tunnel situation, and can we live with it?

zekefarwell:
Bridges and tunnels definitely need work, but that work needs to be done for all roads, not just expressways.  Makes sense to look at holistically as mentioned in #84.  

On the topic of tunnels, I remember them having a dashed casing but I'm not seeing that now.  Was this an intentional change, accidental, or am I maybe misremembering?
Screen Shot 2022-01-19 at 11 27 05 PM

ZeLonewolf:
Also, is there anything preventing this from being merged into master?

zekefarwell:
`expressway` isn't available from MapTiler until they finish rebuilding the planet with v3.13, right?

ZeLonewolf:
> `expressway` isn't available from MapTiler until they finish rebuilding the planet with v3.13, right?

Yes, but it doesn't break anything to push these changes into main.  It would just render expressways like non-expressways until the new version shows up.  Whereas, putting the interstate zoom and highway shield code into main would break the existing zoom and shields.

zekefarwell:
I see.  Yeah maybe this should have been merged right into main.  I was thinking it might cause problems since regular, non-expressway road layers are filtering out the `expressway` key as well.  I didn't test it though. 
https://github.com/ZeLonewolf/openstreetmap-americana/blob/acef2f183be42938fedfdce4e341d11df52e1178/style/layer/road.js#L283

adamfranco:
It worked fine against `main` when I last tested, but there are currently merge conflicts when rebasing. If `openmaptilers-new-features-test` will be merged after the new data is available then just waiting for that merge is probably fine. If there is a further delay likely, then I can sort out merge conflicts to `main` if that is desired.

ZeLonewolf:
It was supposed to be out last week, so 🤷‍♂️.  Timeline unclear.

PR #83, opened 2022-01-19
Washington (State) highway shield

Pengor:
I've added an SVG and the Javascript reference for the Washington (State) highway shield for state routes.


ZeLonewolf:
Thanks @Pengor for working on this!

The design aesthetic that we've been using is to remove fill backgrounds from shields.  For example, Georgia State Route 38 is drawn on signage as image, but for the map, we've removed it to make things a bit cleaner and more consistent with other signs:



In general, we've found that a 40px height shield with a 2.5px stroke looks the best.

Given the unusual shape of Washington's head, I would also recommend that we develop a wider version that's friendlier to 3-digit routes, but that can be deferred to a future PR if you like.

Pengor:
Thanks for the feedback! I'll see about replacing the SVG and updating the padding to match. As for a wider shield design: I know on signage in WA there is only a single shield shape used for all 1, 2, and 3 digit routes. The only difference is how the text is scaled to fit within the "head" area. I'd be worried that a wider shield would result in the shield looking less like a head 😨

ZeLonewolf:
We'll

ZeLonewolf:
We'll have to see how it looks on the map and see if we're ok with the readability.  Some states do a wider variant of their shields in reality (see https://en.wikipedia.org/wiki/List_of_state_routes_in_Alabama for example) but that's not universal.  North Carolina and Michigan shields use diamonds, which are particularly difficult to read on 3-digit route numbers, so we'll likely need to come up with some way to stretch those out for this type of map  I'm not opposed to an initial implementation that we can improve upon later if we don't like the readability on larger `ref` values.

ZeLonewolf:
I am seeing a fair amount of text/shield collision as of fd5f70f:
image

Pengor:
Oops, I must not have properly understood how the text padding functions. I'll see if I can figure that out...

Pengor:
Okay after some experimentation I can see why there are different width shield variants in use. I haven't been able to come up with any padding values that both prevent clipping for 2 digit `ref` values and keep 3 digit `ref` values large enough to actually read (I'm on a 23" 1080p monitor).

I know you had mentioned "alternative text layout constraints" in the OSMUS Slack and that got me thinking that perhaps one solution could be to have different padding values for the different length `ref` values.

Any thoughts on how this PR should proceed for the time being @ZeLonewolf? I don't have any new commits since the text clipping that's currently occurring is at least better than having some illegible routes (see photo).

image

ZeLonewolf:
I would say the most recent screen shot is a sufficient starting position until we can get alternative text constraints in.  Then we can assess whether wider shield blanks are needed.

zekefarwell:
Wide boi 😄 
Screen Shot 2022-01-20 at 5 38 01 PM

A bit silly looking, but workable.

[rebusurance](https://github.com/1ec5/rebusurance) does something similar
Screen Shot 2022-01-20 at 5 39 25 PM

Pengor:
I saw the commits from earlier today so I thought I'd try adjusting the padding one more time and wound up with this, not too bad!

image

ZeLonewolf:
That looks pretty good!  I would maybe bring in the passing just a hair so we can get a pixel of separation from the shield blank.

I'm concerned that we're seeing a wide interstate shield on I-5, I thought I'd fixed those bugs 🤔

zekefarwell:
I'm seeing wide shields everywhere on 1x screens.  Opened issue #86

Pengor:
How's this look for padding?

image

ZeLonewolf:
Looks great! Thanks for working on this.

PR #89, opened 2022-01-26
Put npm back in github actions

ZeLonewolf:
Fix github actions breakage


zekefarwell (approved):
seems questionable, but I'll allow it 😄

PR #91, opened 2022-01-26
Add shields for US Business and PA Truck

jleedev:


ZeLonewolf (approved):
Ultimately I'd like to process banners in a generic way, but this seems fine to get these on the map, and we can polish the cannonball over time.

PR #100, opened 2022-01-26
Change place & park layers to prefer english names where available

zekefarwell:
This changes the park and place layers to prefer the `name:en` field whenever it is availabe rather than simply using the `name` field which will be in the local language of the place.  Since `name:en` is not always availabe the coalesce function is used to fall back to using `name:latin` or `name` in those cases.  

There has been discussion of adding a language switching feature (a major benefit of client side vector rendering), but for now it make most sense to have the map in English for the benefit of the primarily English speaking style developers.

### DIFF
--- a/style/layer/park.js
+++ b/style/layer/park.js
@@ -2,6 +2,14 @@
 
 import * as Color from "../constants/color.js";
 
+// Name fields in order of preference
+const name_en = [
+  "coalesce", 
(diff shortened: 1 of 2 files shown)


ZeLonewolf (approved):
Code looks right, and I'm seeing English labels everywhere on the map, including Eastern Hemisphere locations.

PR #103, opened 2022-01-26
Add maxsize parameter to shield definitons

### LINKED ISSUE #102: Allow a shield definition to specify a maximum font size
A shield definition should be able to specify a maximum font size, to prevent cases where a shield's text becomes undesirably large.  This is particularly an issue with circular shields:

image

ZeLonewolf:
Closes #102 

Shield definitions can now include a maximum font size parameter.  This PR also sets the max size of circle shield fonts to 32px.  Note the subtle difference between the two screen shot in the rendering of DE-1.  Please take special care to sanity check that this doesn't break under various screen/resolution conditions.

After this PR (circles set to max font size = 32):
image

Prior behavior:
image

### DIFF
--- a/style/js/shield.js
+++ b/style/js/shield.js
@@ -200,8 +200,8 @@ function drawShield(network, ref) {
 
   //If size-to-fill shield text is too big, shrink it
   if (shieldDef != null && typeof shieldDef.maxFontSize != "undefined") {
-    if (textLayout.fontPx > shieldDef.maxFontSize) {
-      let maxFontSize = shieldDef.maxFontSize * PXR;
+    let maxFontSize = shieldDef.maxFontSize * PXR;
+    if (textLayout.fontPx > maxFontSize) {
       var shrinkFactor = maxFontSize / textLayout.fontPx;
       var y0 = shieldBounds.height - padding.top - padding.bottom;
       var gap = y0 - textLayout.yBaseline + padding.top;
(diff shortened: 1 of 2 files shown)


zekefarwell:
Something isn't quite right here because the circles only have 20px diameter (the 1x, [CSS pixel](https://webplatform.github.io/docs/tutorials/understanding-css-units/) value).  Setting a 32px maxFontSize should have no effect since that's bigger than the canvas.  When I set `maxFontSize: 14` I see the expected font size at DPR 1:
Screen Shot 2022-01-26 at 8 22 03 PM

But a much smaller font at DPR 2:
Screen Shot 2022-01-26 at 8 22 17 PM

I think the 32px value is what we'd want to set if we were doing all the math in a DPR 2 context.  At this point all other values are in a DPR 1 context so it makes more sense to set `maxFontSize: 16` and then always multiply by `Gfx.getPixelRatio()` when making calculations.    

https://github.com/ZeLonewolf/openstreetmap-americana/blob/bd2411e9590e9f9cb28fd94abdabf0a18241a2d1/style/js/screen_gfx.js#L16-L24

zekefarwell:
I pushed up a small change that takes pixel ratio into account so the maxFontSize will for 1x contexts as well as 2x.  This also meant cutting the maxFontSize for circle shields in half from 32 to 16.

zekefarwell (dismissed):
Looks good to me.  I didn't notice any regressions, but this drawShield() function handles so many different cases it's hard to tell for sure....

zekefarwell:
Scratch that.  My commit created a regression. 
Screen Shot 2022-01-26 at 9 33 01 PM

zekefarwell:
Ok.  regression should be fixed.  
Screen Shot 2022-01-26 at 9 43 15 PM

zekefarwell (approved):

PR #107, opened 2022-01-27
Fix the recolored shields that I broke

zekefarwell:
While refactoring in the lead up to #72 getting merged, I removed something that appeared redundant.  This broke the recoloring of shields by adding an opaque background where it should be transparent.  This PR fixes that.

Current broken rendering:
Screen Shot 2022-01-27 at 4 55 56 PM

Fixed rendering:
Screen Shot 2022-01-27 at 4 57 22 PM


ZeLonewolf (approved):
Looks good!  I feel like there should be an easier way to do this but this works so let's ship it!

PR #108, opened 2022-01-28
Disable map rotation and pitch

zekefarwell:
The method of shield drawing we are using doesn't play well with rotating or pitching the map, so I'm suggesting we disable those features for now.  If and when there is more movement on https://github.com/maplibre/maplibre-gl-js/pull/716#issuecomment-995193918 then we can probably re-enable them.

Screenshot from 2022-01-27 19-59-10

Screen Shot 2022-01-27 at 8 58 40 PM

### DIFF
--- a/style/americana.js
+++ b/style/americana.js
@@ -233,7 +233,13 @@ var map = (window.map = new maplibregl.Map({
   center: [-94, 40.5], // starting position [lng, lat]
   zoom: 4, // starting zoom
   attributionControl: false,
+  pitchWithRotate: false,
+  dragRotate: false,
+  touchPitch: false,
 }));
+
+map.touchZoomRotate.disableRotation();
+map.keyboard.disableRotation();
 
 map.on("styledata", function () {
   ShieldDef.loadShields(map.style.imageManager.images);
--- a/style/index.html
+++ b/style/index.html
@@ -2,7 +2,7 @@
 <html>
   <head>
     <meta charset="utf-8" />
-    <title>Display a map</title>
+    <title>OSM Americana</title>
     <meta
       name="viewport"
       content="initial-scale=1,maximum-scale=1,user-scalable=no"


style/americana.js
```diff
@@ -233,8 +233,14 @@ var map = (window.map = new maplibregl.Map({
   center: [-94, 40.5], // starting position [lng, lat]
   zoom: 4, // starting zoom
   attributionControl: false,
+  pitchWithRotate: false,
+  dragRotate: false,
+  touchPitch: false,
```
ZeLonewolf, line 238: Do we need to disable pitch?  It seems like this should still be okay in a north-up configuration.

style/americana.js
```diff
@@ -233,8 +233,14 @@ var map = (window.map = new maplibregl.Map({
   center: [-94, 40.5], // starting position [lng, lat]
   zoom: 4, // starting zoom
   attributionControl: false,
+  pitchWithRotate: false,
+  dragRotate: false,
```
ZeLonewolf, line 237: Let's add comments that note that these disable lines can be removed once #111 is resolved.

ZeLonewolf (changes requested):

style/americana.js
zekefarwell: good call.  will do

style/americana.js
zekefarwell: We don't necessarily _need_ to disable rotate or pitch  They're just a bit sub-optimal at the moment so I think it gives a better first impression to not have them enabled.  Leaving pitch enabled certainly looks less bad than the upside down shields you get with rotating, but it still looks rather odd to have the shields skewed to the plane of the map while place names and icons remain sticking up vertically.  

Screen Shot 2022-01-28 at 1 48 05 PM

If you'd like to keep the pitch enabled that seems fine to me.  I'd actually be fine with leaving rotate enabled too and not bothering with any of this (unless the issue never gets addressed in MapLibre).  Folks in Slack were just mentioning that it didn't look that great and I figured we could disable temporarily.

ZeLonewolf: Oh, ok I didn't realize pitch was borked too.  I'll add that to #111.

zekefarwell: So you want to move ahead with disabling both?  Or should we leave it as is for better visibility of why we want to see changes in MapLibre?

ZeLonewolf: Let's pose it as a community question.

wipfli:
Did you try this combination?
```
'text-rotate-to-line': false,
'text-pitch-alignment': 'viewport',
```

zekefarwell:
> Did you try this combination?
> 
> ```
> 'text-rotate-to-line': false,
> 'text-pitch-alignment': 'viewport',
> ```

Oh...  haha!  Yes, that appears to handle it perfectly.

Screen Shot 2022-01-28 at 5 42 22 PM

I had misunderstood this comment (https://github.com/maplibre/maplibre-gl-js/pull/716#issuecomment-995943188) to mean more changes were needed to MapLibre to make this possible.  I guess we can close this PR and replace it with another that just adds these properties!

ZeLonewolf:
Closing in favor of properties identified by @wipfli that solves the issue!

zekefarwell:
I may have spoken a bit too soon.  Although the shields are kept upright as expected while the map is rotated or pitched, there is an unfortunate side effect.  When the map is flat, shields can look rather pixelated and bad.  @wipfli noted this as well (https://github.com/maplibre/maplibre-gl-js/pull/716#issuecomment-995943188), but I was unable to replicate the behavior at first.  Here is what it looks like on my DPR 1 display at 150% browser zoom.  

Flat map (bad)
Screen Shot 2022-01-28 at 8 30 32 PM

Pitched map (good)
Screen Shot 2022-01-28 at 8 30 17 PM

At 100% browser zoom on the same display I don't see this bad scaling behavior

Flat map (good)
Screen Shot 2022-01-28 at 8 29 39 PM

Pitched map (good)
Screen Shot 2022-01-28 at 8 28 45 PM

I also see no issue on my DPR 2 display.  It seems like the shields will look fine for anyone viewing the map with display scale at 100% (DPR 1) or 200% (DPR 2), but for scales in between like 125% (DPR 1.25) something is not working so well in the MapLibre graphics scaling.

ZeLonewolf:
So, if I understand this correctly, we have three options:

1. Implement the following style settings, which make the shields do what we want, but also makes them look crappy under certain display settings
```
'text-rotate-to-line': false,
'text-pitch-alignment': 'viewport',
``

2. Let the shields fly upside-down when rotated

3. Disable rotation/tilt entirely

zekefarwell:
Yes that is what it seems like to me.  I say lets go with option 1 and then see what we can do to get the graphcis scaling issue fixed in MapLibre.

zekefarwell:
Closing in favor of #124

PR #109, opened 2022-01-28
Add 1/2 and 3 digit Oregon highway shields

Pengor:
This would add two shield icons for Oregon routes: one for 1 and 2 digit route numbers and one for 3 digit route numbers.


ZeLonewolf:
Looks great!  Thanks for the contribution:
image

ZeLonewolf (approved):

Issue #96, opened 2022-01-26
Netherlands highway signs


501Ghost:
The Netherlands uses its own set of highway signs, which I think would look nice on the Americana map style. They're classified with the letters A, N, S and R. The most important ones, A & N, all have route relations and have red and yellow signs respectively. S & R are still a work in progress.

The full documentation can be found at https://wiki.openstreetmap.org/wiki/The_Netherlands_road_network, including links to all the route relations.

coolultra1:
Fork it and call it "openstreetmap-nethericana"

Mashin6:
A and N signs
![image](https://user-images.githubusercontent.com/5265707/151475955-020149f5-e0ef-46e1-b7a8-ae211ecec395.png)
![image](https://user-images.githubusercontent.com/5265707/151476070-ee85950b-2b25-4f80-bb05-771ff53caf44.png)

501Ghost:
That looks great!

ZeLonewolf:
@501Ghost what is S112 in the screen shot above?

ZeLonewolf:
It appears that there are some city routes also.  For example:
![image](https://user-images.githubusercontent.com/3254090/151638827-6afa73ff-d03a-462f-8114-5310bf942e37.png)

Seems to correspond with a network value of `NL:S:Amsterdam`, for example: https://www.openstreetmap.org/relation/13293820

Blijbol:
Besides S-roads there is also RING-roads (basically unnumbered S-roads) and R-roads. Aruba also has separate road numbers/shields but there are no route relations for Aruba yet.

Here is some reference rendering you can use: https://wiki.openstreetmap.org/wiki/Template:Ref-NL/doc

It would be very helpful if you could do prefix-matching for network tags, so we don't have to hardcode every single city with S-roads and R-roads in your source code.

1ec5:
> It would be very helpful if you could do prefix-matching for network tags, so we don't have to hardcode every single city with S-roads and R-roads in your source code.

Wildcard matching would have some drawbacks, like not being able to say definitively to taginfo that this project supports the specific values of `network` under that prefix. In case it’s of interest, in the U.S., there’s been [a push](https://lists.openstreetmap.org/pipermail/talk-us/2022-April/021671.html) to consolidate some regional networks that overlap numerically but are identically signposted and subject to the same numbering authority.

PR #110, opened 2022-01-28
Adding shields for Netherlands

Mashin6:
Adding shield definitions per #96.


ZeLonewolf (approved):
Looks great!  Thanks for the screen shots in the issue thread.

PR #112, opened 2022-01-29
Refactor max font size code to improve maintainability / extensibility

ZeLonewolf:
This PR refactors the code that enforces a maximum font size when drawing shield text.  The previous code included a bunch of math that I figured out on the back of a napkin that's essentially incomprehensible.  I removed this in favor of a more-understandable `Math.min()` function to limit the font size.  This is a needed step to support additional text layout improvements including vertical alignment and half-elliptical layout constraints, which will improve the text layout appearance of half-rounded shields such interstate and Virginia state route shields.

This PR should have no impact on rendering or performance.


ZeLonewolf:
Found a bug, will re-post as a comprehensive fix for #114

PR #113, opened 2022-01-29
Rename European icons per ISO 3166-1

1ec5:
Renamed the European shields to use ISO 3166-1 country codes instead of ad-hoc codes. The “eu” prefix is also unnecessary and confusing, since Switzerland is not an EU member. I retained “un” as a prefix for shields in UN-designated networks such as the E-road network.


ZeLonewolf (approved):
Looks good to me.  I'm thinking we should also take the "40" out of the file names while we're at it.  In the original go at this, I was thinking that we would have versions of the graphics at different widths and name them by px size, but we're handling that with _2 _3 etc prefixes that make more sense for this.

Mashin6:
Going forward, it might be good to think about some general .svg file naming convention. Assuming that the number of different icons for various things will be growing.
(btw "eu" was meant as a grouping mark to keep shields in Europe together, not as an EU code.)

1ec5:
Ah, thanks for the clarification. I was preparing to look into shields in Asia and didn’t want to make up a two-letter code for that continent or decide which countries go in Europe or Asia. (The E-road network spans both continents and overlaps with the Asian Highway network.) In any event, I suspect that the `us`-prefixed shields will dominate this folder for some time.

ZeLonewolf:
There is also an open question about whether we want to add simple shields to the sprite sheet or draw them on the fly in canvas code like we do for the white square/rectangle shields that we use as the default.  If a shield is just a rounded rectangle with a solid color, we don't really need to pack that into a sprite sheet, we could just have some definition code that declare what color a network's rectangle needs to be.

1ec5:
An optimization along these lines would be nice to have in conjunction with #118.

zekefarwell:
For the more generic cases maybe it makes more sense to name the files by shape and color (or just shape if we can recolor dynamically) rather than country or state.  There are a lot of common shapes that are used in a number of different countries.

PR #116, opened 2022-01-29
Half ellipse text constraint

### LINKED ISSUE #114: Half-ellipse shield text constraints
We currently support two types of shield text constraints: rectangles and ellipses.  A text constraint allows text placed on a shield to grow as large as possible without overflowing the rectangle or ellipse constraint, with the shape buffered by the supplied padding values.

In the current scheme, on shields that are roughly half-elliptical get a rectangular constraint, as follows.  The dashed magenta lines indicate where the text has been shrunk vertically to fit within the rectangle at a constant aspect ratio:
image

As the text gets longer, however, the vertical shrinking increases to allow the text to fit.  This creates an increasingly undesirable gap between the top of the text area and the text itself.  While we could address the vertical alignment, this would still result in text that's inappropriately small for the size of the shield area.
image

Instead, we could define the text area in terms of a half-ellipse.  In this example, we transcribe half an ellipse (green) into a padded text area (magenta), and allow the text to expand until it hits the edge of the half-ellipse.  This gives you the largest possible text size that fits in the shield, and still allows you to set the overall padding.
image

As a separate matter, we may want to consider a narrower font, which would allow the text to consume more vertical space and less horizontal space for the same text.  For example, this is how a narrower font (Arial Narrow in this case) would be expressed using a half-ellipse constraint:
image

ZeLonewolf:
Closes #114 

This PR implements a half-ellipse text constraint, for shields that are closer to a semi-circle in shape than a rectangle.  This allows for improved text scaling for variable-length shield refs.  This also simplifies the code for "max font size" constraints in the shield definitions, which was a pre-requisite for centralizing all of the text scaling code into shield_text.js functions.

Before/After Virginia:
image
image

Before/After I-H201:
image
image

### DIFF
--- a/style/js/shield_text.js
+++ b/style/js/shield_text.js
@@ -28,16 +28,24 @@ export function rectTextConstraint(spaceBounds, textBounds) {
  * @param {*} padding - top/bottom/left/right padding around text
  * @param {*} bounds - size of the overall graphics area
  * @param {*} textLayoutFunc - algorithm for text scaling
+ * @param {*} maxFontSize - maximum font size
  * @returns JOSN object containing (X,Y) draw position and font size
  */
-export function layoutShieldText(text, padding, bounds, textLayoutFunc) {
+export function layoutShieldText(
+  text,
+  padding,
+  bounds,
+  textLayoutFunc,
+  maxFontSize
+) {
   const PXR = Gfx.getPixelRatio();
-  var padding = padding || {};
+
   var padTop = padding.top * PXR || 0;
   var padBot = padding.bottom * PXR || 0;
   var padLeft = padding.left * PXR || 0;
   var padRight = padding.right * PXR || 0;
 
+  var maxFont = maxFontSize * PXR;
   //Temporary canvas for text measurment
   var ctx = Gfx.getGfxContext(bounds);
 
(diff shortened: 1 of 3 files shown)


style/js/shield_text.js
```diff
@@ -102,17 +188,10 @@ export function drawShieldText(ctx, text, textLayout) {
...
-    {
-      left: 0,
-      right: 0,
-      top: 0,
-      bottom: 0,
-    },
-    { width: ctx.canvas.width, height: ShieldDef.bannerSizeH },
-    rectTextConstraint
-  );
+  var textLayout = layoutShieldTextFromDef(text, null, {
```
zekefarwell, line 191: I see why it's done this way rather than calling layoutShieldText, but it's a bit counterintuititve to call a function named layoutShieldTextFrom**Def**, and then pass the **def** parameter as `null`.  Perhaps the the function names can be made clearer in the future.

style/js/shield_text.js
```diff
@@ -69,7 +102,21 @@ export function layoutShieldText(text, padding, bounds, textLayoutFunc) {
...
+      yBaseline = padTop + (availHeight - textHeight) / 2;
+      break;
+    case VerticalAlignment.Top:
+      yBaseline = padTop;
+      break;
+    case VerticalAlignment.Bottom:
+      yBaseline = padTop + availHeight - textHeight;
+      break;
+    default:
+      return null; //Code error, should never happen
```
zekefarwell, line 118: This seems fragile.  Maybe it would make sense to throw an exception here and add error handling at the top level (the `missingIconLoader()` function.  This could log a helpful message to the console when this code error does happen.  It's hard to say what sort of error returning `null` here might produce elsewhere in the code, but a custom error message in that unlikely case would probably be clearer.

zekefarwell (approved):
I think this all makes sense.  I haven't managed to wrap my head around exactly how and why the space bounds, text bounds, available height/width, and various layout contraint functions work.  However, they certainly seem to work as advertised so I'm ok with that 😄.

I left a couple comments as suggestions for future improvement.  Nothing I feel should hold up this PR though.

Issue #117, opened 2022-01-29
Support variable-width rectangular & oval signs


ZeLonewolf:
For cases where signs are plain squares or rectangles, we've drawn white squares with black borders.  This has covered states like CT/RI/MA/ME which use state highway shields that look similar.  However, in most cases, 3+ digit signs are drawn as rectangles:
![image](https://user-images.githubusercontent.com/3254090/151674898-e9590362-30b7-4e0b-be61-7c04a65e1304.png)

while shorter route numbers are drawn as squares:
<img src="https://user-images.githubusercontent.com/3254090/151674954-83da635f-2968-45f6-8e0b-6836d45c4159.png" width=100>

Thus, we should make these generically-drawn signs variable-width, perhaps up to a 3:2 or 2:1 ratio rather than squeezing the text into a square.

1ec5:
Same goes for circular shields, which don’t fit longer refs very well. In fact, Kentucky has many four-digit-long refs, but they use rounded rectangles instead of ellipses.

ZeLonewolf:
Though it's not quite a rounded rectangle, how would we feel about using ellipses as the generalized expansion of round shields?

![image](https://user-images.githubusercontent.com/3254090/151701936-d8915ab0-86a0-4f83-9a10-425da18979cc.png)

zekefarwell:
The ellipses certainly look better than shrinking long refs into circles.
<img width="295" alt="Screen Shot 2022-01-30 at 1 10 58 PM" src="https://user-images.githubusercontent.com/281482/151711866-1f5af176-bb1f-4a48-848f-7a859d6c7f86.png">

The [real world elliptical shields](https://commons.wikimedia.org/wiki/Category:Elliptical_route_shields) seem rounder to me though.  Maybe they don't end up being as wide due to narrow text with 3 digits.
![image](https://user-images.githubusercontent.com/281482/151711936-d7dfffe7-c46b-4692-89a9-3cc022c70d75.png)

At the small size we're displaying these shields, I think rounded rectangles (elongated circles) will probably look better.  It may not make too much difference for a 3 digit ref, but I think it will look quite a bit better for longer refs.
![image](https://user-images.githubusercontent.com/281482/151711995-3eb6212c-1a33-46e4-8239-2f88a609e868.png)

[...]

PR #119, opened 2022-01-30
Shields of Kentucky

1ec5:
Added shields for Kentucky state highways and parkways.

BG at KY 555

For now, the state highway shield is a circle for simplicity. As tail work, it should be replaced by an ellipse or the rounded rectangle that KYTC posts on signs: https://github.com/ZeLonewolf/openstreetmap-americana/issues/117#issuecomment-1025013060.

The parkways are more complex. Each of the parkways uses a standardized shield that bears the state tourism logo and the road’s entire official name. Unfortunately, most screens don’t come with enough pixel density to legibly display the namesake’s middle initial in a small shield. Each parkway has an official two-letter initialisms and four-digit route number that is used only in KYTC publications, so we can’t tag them as refs in OSM. Even so, I think users would recognize the initialisms on a map, because they’re derived from the road name’s initials (or the namesake’s first and last initials). Some of the routes, like the [Western Kentucky Parkway](https://www.openstreetmap.org/relation/1648091) (WK), used to bear these initialisms on their shields.

[Western Kentucky Parkway](https://commons.wikimedia.org/wiki/File:BluegrassPkwyWest-Exit1AB-Int65nsSplit_%2838035267226%29.jpg)
_The shield is so poorly designed that KYTC has to stick alt text next to it._

In order to display shields for the parkways, I had to plumb the way name through to the shield definition and map way names to hard-coded display refs. In general, it’s a very bad idea to conflate route shields with road names, but I don’t think we really have a choice with Kentucky’s parkway system. To head off abuse of this mechanism, I’ve made it so that it can only map way names to refs, not background images. In theory, we could hard-code logic in OpenMapTiles to expose the Kentucky parkway route relations’ `short_name`s in the `route_*` properties. However, it’s my understanding that OpenMapTiles would prefer to avoid such regionally targeted logic.

As a special case, the AA Highway has a shield that resembles the parkway shield, but it does have a signposted ref of “AA” along with a unique `network` tag. The `network=US:KY:AA` tag was introduced [less than a month ago](https://www.openstreetmap.org/changeset/115738030), so it hasn’t propagated to OpenMapTiles yet. For now, it’ll render as a lettered state highway shield, but here’s a jury-rigged preview:

AA before AA after

### DIFF
--- a/style/icons/shield40_us_ky_parkway.svg
+++ b/style/icons/shield40_us_ky_parkway.svg
@@ -1,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   width="20"
+   height="20"
+   version="1.0"
+   id="svg1113"
+   sodipodi:docname="shield40_us_ky_parkway.svg"
+   inkscape:version="1.1.1 (c3084ef, 2021-09-22)"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:xlink="http://www.w3.org/1999/xlink"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg">
+  <defs
+     id="defs1117" />
+  <sodipodi:namedview
+     id="namedview1115"
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1.0"
+     inkscape:pageshadow="2"
+     inkscape:pageopacity="0.0"
+     inkscape:pagecheckerboard="0"
+     showgrid="false"
+     inkscape:zoom="31.35"
+     inkscape:cx="11.897927"
+     inkscape:cy="10"
+     inkscape:window-width="1383"
+     inkscape:window-height="855"
+     inkscape:window-x="0"
+     inkscape:window-y="23"
+     inkscape:window-maximized="0"
+     inkscape:current-layer="svg1113" />
+  <rect
+     width="20"
+     height="20"
+     rx="1.25"
+     ry="1.25"
+     fill="#003478"
+     stroke-width="0.625"
+     id="rect1107"
+     x="0"
+     y="-20"
+     transform="scale(1,-1)"
+     style="fill:#003f87" />
+  <use
+     transform="matrix(0.92938,0,0,-0.92938,19977,4810.8)"
+     width="300"
+     height="240"
+     fill="#ffffff"
+     stroke="#000000"
+     stroke-linejoin="round"
+     stroke-width="309"
+     xlink:href="#state_outline"
+     id="use1109"
+     x="0"
+     y="0" />
+  <rect
+     style="fill:#ffffff;stroke-width:0"
+     id="rect1727"
+     width="18"
+     height="14"
+     x="1"
+     y="1" />
+</svg>
--- a/style/js/shield_defs.js
+++ b/style/js/shield_defs.js
@@ -194,6 +194,20 @@ export function loadShields(shieldImages) {
       bottom: 6,
     },
   };
+  shields["US:KY:Parkway"] = Object.assign(
+    {
+      refsByWayName: {
+        "Audobon Parkway": "AU",
+        "Bluegrass Pkwy": "BG",
+        "Cumberland Pkwy": "LN",
+        "Hal Rogers Pkwy": "HR",
+        "Mountain Pkwy": "MP",
+        "Purchase Pkwy": "JC",
+        "Western Kentucky Pkwy": "WK",
+      },
+    },
+    shields["US:KY:AA"]
+  );
 
   shields["US:MI"] = diamondShield;
 
--- a/style/js/shield_defs.js
+++ b/style/js/shield_defs.js
@@ -184,6 +184,16 @@ export function loadShields(shieldImages) {
   };
 
   shields["US:KY"] = circleShield("white", "black");
+  shields["US:KY:AA"] = {
+    backgroundImage: shieldImages.shield40_us_ky_parkway,
+    textColor: "#003f87",
+    padding: {
+      left: 2,
+      right: 2,
+      top: 2,
+      bottom: 6,
+    },
+  };
 
   shields["US:MI"] = diamondShield;
 
--- a/style/js/shield_defs.js
+++ b/style/js/shield_defs.js
@@ -183,6 +183,8 @@ export function loadShields(shieldImages) {
     },
   };
 
+  shields["US:KY"] = circleShield("white", "black");
+
   shields["US:MI"] = diamondShield;
 
   shields["US:MN"] = {
--- a/style/layer/highway_shield.js
+++ b/style/layer/highway_shield.js
@@ -6,7 +6,7 @@ function routeConcurrency(num) {
   return [
     "case",
     ["!=", ["get", "route_" + num], null],
-    ["image", ["concat", "shield_", ["get", "route_" + num]]],
+    ["image", ["concat", "shield\n", ["get", "route_" + num]]],
     ["literal", ""],
   ];
 }
--- a/style/layer/highway_shield.js
+++ b/style/layer/highway_shield.js
@@ -6,7 +6,10 @@ function routeConcurrency(num) {
   return [
     "case",
     ["!=", ["get", "route_" + num], null],
-    ["image", ["concat", "shield\n", ["get", "route_" + num]]],
+    [
+      "image",
+      ["concat", "shield\n", ["get", "route_" + num], "\n", ["get", "name"]],
+    ],
     ["literal", ""],
   ];
 }
--- a/style/js/shield.js
+++ b/style/js/shield.js
@@ -227,7 +227,7 @@ export function missingIconLoader(map, e) {
     return;
   }
 
-  var network_ref = id.split("_")[1];
+  var network_ref = id.split("\n")[1];
   var network_ref_parts = network_ref.split("=");
   var network = network_ref_parts[0];
   var ref = network_ref_parts[1];
--- a/style/js/shield.js
+++ b/style/js/shield.js
@@ -109,7 +109,7 @@ function textColor(shieldDef) {
   return "black";
 }
 
-function drawShield(network, ref) {
+function drawShield(network, ref, wayName) {
   var shieldDef = ShieldDef.shields[network];
   var ctx = null;
   var bannerCount = 0;
@@ -136,6 +136,10 @@ function drawShield(network, ref) {
   } else {
     bannerCount = ShieldDef.getBannerCount(shieldDef);
     padding = shieldDef.padding;
+
+    if (shieldDef.refsByWayName) {
+      ref = shieldDef.refsByWayName[wayName];
+    }
 
     var shieldArtwork = getRasterShieldBlank(network, ref);
     var compoundBounds = null;
@@ -231,10 +235,11 @@ export function missingIconLoader(map, e) {
   var network_ref_parts = network_ref.split("=");
   var network = network_ref_parts[0];
   var ref = network_ref_parts[1];
+  var wayName = id.split("\n")[2];
 
   var colorLighten = ShieldDef.shieldLighten(network, ref);
 
-  var ctx = drawShield(network, ref);
+  var ctx = drawShield(network, ref, wayName);
 
   if (ctx == null) {
     //Does not meet the criteria to draw a shield


style/layer/highway_shield.js
```diff
@@ -6,7 +6,10 @@ function routeConcurrency(num) {
   return [
     "case",
     ["!=", ["get", "route_" + num], null],
-    ["image", ["concat", "shield_", ["get", "route_" + num]]],
+    [
+      "image",
+      ["concat", "shield\n", ["get", "route_" + num], "\n", ["get", "name"]],
+    ],
```
1ec5, line 12: The downside of introducing `name` into the image name is that the sprite sheet gets very fragmented by routes that travel over roads with many different names. The Kentucky parkways have uniform road names, but that certainly isn’t the case for other long-distance routes. We could conditionalize the addition of `name` on `route` equaling `US:KY:Parkway=`.

style/icons/shield40_us_ky_parkway.svg
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
ZeLonewolf, line 1: Please save as "optimized svg" so we don't commit all the extra Inkscape stuff into the repo.

style/layer/highway_shield.js
1ec5: I made it a `match` expression so it’s easy to add more special cases as necessary.

style/js/shield_defs.js
```diff
@@ -183,6 +183,32 @@ export function loadShields(shieldImages) {
...
+  };
+  shields["US:KY:Parkway"] = Object.assign(
+    {
+      refsByWayName: {
+        "Audobon Parkway": "AU",
+        "Bluegrass Pkwy": "BG",
+        "Cumberland Pkwy": "LN",
+        "Hal Rogers Pkwy": "HR",
+        "Mountain Pkwy": "MP",
+        "Purchase Pkwy": "JC",
```
1ec5, line 205: I extracted these initialisms from KYTC’s internal alphanumeric route numbers. Some of them are pretty recognizable, but some like “JC” are more obscure. (It’s from the namesake in the official name.) For the benefit of non-locals who don’t remember the older, more usable shields, I’d be open to changing it to something more intuitive, like “P”, but such ad-hoc abbreviations definitely shouldn’t go into OSM.

1ec5:
#121 implements Kentucky state highway shields, which would limit this PR to just the AA Highway and parkway shields.

ZeLonewolf:
It seems like we have three options here.

1. Render all of these KY parkways with a common shield
2. Render parkways with unsigned internal KY initialisms
3. Render parkways with unsigned internal KY initialisms with an exception for the JC one so it's more sane to map readers

A separate question is whether these are tagged correctly, as we do not want to do an end run around correct tagging if there's a problem in the map.

Because this series of parkways has a common shield background, I think it's correct that they share a common network.  I can't think of any case where road networks with a common shield series have different network values which we could use as an comparable case.  In New Jersey, we tagged their parkways with different networks because each one had completely different shield designs.  In New York, the consensus (possibly not yet implemented at this point) was that each of their parkway shield series (the green ones vs the white Long Island ones) should have separate network values.  So I think the patter of common shield design = common network value is correctly preserved in NJ, NY, and KY cases.

I also agree that the long text names in the KY shields is not a ref, which is reserved for coding schemes.  These are clearly names.  Additionally, the KY internal 2-letter codes are not signed (with the exception of AA), so it is probably appropriate to tag those `unsigned_ref=XX`.  We could promote those to `ref`, but that would be tagging for this renderer, which I don't think we want to make a precedent for.

So this leaves us with the decision of what to draw on the map.  I think drawing the shield blank with no text would be somewhat lacking, and drawing a real-world shield with unreadable text would just be graphic noise.  The use of initials seems like a reasonable compromise that gets recognizable, differentiated shields on that map that map users are likely to understand.  By that rationale, the Purchase Parkway code should be "PP" to be consistent with the "MP" initials of Mountain Parkway.  I vote map user understandability over KYDOT coding schemes.

I would like to hear additional opinions, but so far I have not heard a better way to handle it, nor a consensus to tag these roads differently.

zekefarwell:
Fourth option:  Don't render shields for these since they don't  have ref values.  Instead figure out how to render the names prominently at the zoom level where a shield would start displaying.  Under this option, AA Parkway would be an exception since it is signed in a way where "AA" makese sense as a ref value, so it would gets a shield.  This is [what Apple Maps does](https://duckduckgo.com/?q=Alexandria%2C+KY&t=h_&ia=web&iax=about&iaxm=maps&strict_bbox=0&bbox=39.14543179265603%2C-84.59895666630086%2C38.72347908892468%2C-83.97822912723836&metatoken=0).

ZeLonewolf:
> Fourth option: Don't render shields for these since they don't have ref values.

The objection I have to this idea, is that these roads have shields in the real world, so I would expect to have shields in the map.

zekefarwell:
>  these roads have shields in the real world

Well, I guess that get's into the philosophical question of what makes a sign a "shield" vs a sign with a name on it.  I think we can all agree that this is just a sign with a name on it, not a shield:

image

To me, the Kentucky Parkway signs don't seem very different, though perhaps the design is a bit more unique:
Screen Shot 2022-02-05 at 1 00 18 PM

Does that make them "shields"?  Can't say I can answer that question, and rather than attempting to, it seems simpler to make the criteria for whether we display a shield on the map be that there is a signed shortcode we can sensibly call a ref.  In the case of AA parkway, there is.  In the case of the others, there isn't.

1ec5:
Yes, the parkway signs are shields, however unwieldy. Over the years, they have evolved from (much more attractive) ad-hoc signs similar to the ones used in New Jersey, to a color-coded scheme, to the more uniform shields today. Here are some of their previous shields for comparison:

[Audubon Parkway](https://commons.wikimedia.org/wiki/File:Audubon_Parkway_Shield.png) [Bert T. Combs Mountain Parkway](https://en.wikipedia.org/wiki/File:Bert_T._Combs_Mountain_Parkway_logo.png) [Martha Layne Collins Bluegrass Parkway](https://commons.wikimedia.org/wiki/File:MLC_Bluegrass_Parkway_Shield.png) [Louie B. Nunn Cumberland Parkway](https://commons.wikimedia.org/wiki/File:Louie_B_Nunn_Cumberland_Parkway.png) [Hal Rogers Parkway](https://commons.wikimedia.org/wiki/File:Hal_Rogers_Parkway.png) [Purchase Parkway](https://commons.wikimedia.org/wiki/File:Julian_M_Carroll_Purchase_Parkway_Shield.png) [Wendell H. Ford Western Kentucky Parkway](https://commons.wikimedia.org/wiki/File:Wendell_H_Ford_Western_KY_Parkway_Shield.png) [Edward T. Breathitt Pennyrile Parkway](https://commons.wikimedia.org/wiki/File:Edward_T_Breathitt_Pennyrile_Parkway_Shield.png) [William H. Natcher Parkway](https://commons.wikimedia.org/wiki/File:William_H_Natcher_Pkwy_Shield.svg)

[WK](http://www.billburmaster.com/rmsandw/kentucky/parkway/wkpkwy.html) [P](http://www.billburmaster.com/rmsandw/kentucky/parkway/pennypile.html)

Today’s signs continue to function as shields, appearing everywhere the MUTCD calls for shields to be used on a numbered highway: on guide signs, trailblazer assemblies, and reassurance markers, replete with cardinal direction auxiliary plates. The “alt text” I referred to above is standard KYTC practice [on any named road](https://wiki.openstreetmap.org/wiki/User:Minh_Nguyen/destination:street), because many state highways are better known by their road names.

zekefarwell:
Fair enough.  I'm convinced these are indeed shields, so disgregard the fourth option I proposed.  I still feel that the best path forward here is to have the initialisms in an OSM field rather than hard coded into this style, though.  I get that they aren't technically refs so it would be better not to use that field.  How about working on including short_name in the OpenMapTiles transportation layer?  I doubt this will be the last case where that field is useful.

1ec5:
> How about working on including short_name in the OpenMapTiles transportation layer? I doubt this will be the last case where that field is useful.

The good news is that `short_name` is already tagged on these route relations. For these parkways, it would make a lot of sense for OpenMapTiles’ `route_*` properties to fall back from `{network}={ref}` to `{network}={short_name}` if the route relation has a `network` but no `ref`. However, we should probably think about the applicability of this approach beyond Kentucky, since OpenMapTiles is pretty lean on regional special cases.

I think `short_name` could be justified for other cases where an initialism isn’t signposted but would still be useful for rendering. For example, mappers routinely put [initialisms](https://www.openstreetmap.org/relation/53754) in the `ref`s of bicycle route relations, just so the routes get labeled by renderers such as OpenCycleMap. This makes it more difficult to tell when the `ref` [should actually appear on a shield](https://www.openstreetmap.org/relation/1738747). Overloading `ref` in this manner will complicate any future work to render recreational route shields in Americana.

ZeLonewolf:
The `short_name` coalesce that Minh suggested would be easy to implement in OMT, technically, but I think it would be hard to justify if Kentucky Parkways is the only case where it's needed.  I would think we'd need multiple examples to rationalize extra processing globally.

zekefarwell:
It's arguable that `short_name` would be more appropriate for initialisms of parkways and toll roads like many in New York, New Jersey, and Texas.  Most of these use the `ref` field currently.  Like FBP and SHT (lol 💩) here:

Screen Shot 2022-02-05 at 5 47 09 PM

Conversely if it's not a problem for these parkways to have their initials in the `ref` field, I think it would be fine to do the same for the Kentucky parkways.

1ec5:
Texas is more complicated. Some of the toll operators use the short names as conventional `ref`s too. For example, the seven or so named toll roads operated by the North Texas Tollway Authority use reassurance markers with spelled-out road names [Sam Rayburn Tollway](https://commons.wikimedia.org/wiki/File:Toll_Texas_Sam_Rayburn_Tollway.svg) but initialisms as part of shields on guide signs and trailblazer assemblies [SRT](https://commons.wikimedia.org/wiki/File:Toll_Texas_SRT_new.svg). I think it’s appropriate for the NTTA route relations to be tagged with identical `short_name`s and `ref`s.

On the other hand, I would argue that the Houston-area toll roads only have initialisms of convenience, not actual reference codes. They should be tagged `short_name`, if not a one-off `network` value, whereas tagging `ref` would be optimizing for a particular renderer.

zekefarwell:
Here are a couple of examples from Toronto that seem like a similar situation.
https://en.wikipedia.org/wiki/Don_Valley_Parkway
https://en.wikipedia.org/wiki/Gardiner_Expressway

The generic case seems to be routes where:
- Real world shields display a full name rather than a number, code, or abbreviation
- Real world shield design without the name is not unique or recognizable enough 
- Name is too long to fit on a map sized shield

These factors make an initialism a reasonable choice.  Although I've yet to see any examples like this outside of North America, it doesn't seem to me like these Kentucky parkways are entirely unique.  Personally I have no issue bending the meaning of `ref` to include initialisms of convience (as many mappers have), but if you feel strongly that this is the wrong thing to do, I'll help promote `short_name` as the proper place for them.  I think we'd be able to come up with enough of these cases that the `short_name` coalesce is justified in OMT.

1ec5:
I think your synopsis of the issue is correct, although I think we need to keep digging for situations similar to the Kentucky parkway situation:

* The New York City and Long Island parkway shields now rely on letters.
* The New York State parkways are plain-text signs, but at least they do suggest initialisms using drop caps.
* The New York Palisades parkway shields rely on inscrutable text for distinguishing one route from another. ✔️
* The Houston-area Harris County Toll Road Authority (HCTRA) shields rely on inscrutable text for distinguishing one route from another. ✔️
* The Toronto-area shields are identifiable by colors (similar to the Pittsburgh-area color belts).

Basically, there are a lot of ways to make shields usable, some of which used to be used in Kentucky, but KYTC has doubled down on unreadable shields for some reason. The Palisades and HCTRA shields give a logo more prominence than the road name. For those routes, could ignore the name entirely, giving every route within the network the same exact shield with the logo in the middle. But for the Kentucky parkways, a blank box would look like a bug, especially since “BG” and “WK” would be the common-sense things to put in there, based on the old shields.

Even though I’m quite certain that the Kentucky parkway shields are shields, I wouldn’t completely rule out replacing them with text labels, as the KYTC paper maps do, if we can find a way to do that reliably. Text labels would collide out a lot of nearby shields for intersecting routes, but I don’t think that’s a huge problem for most of these roads.

zekefarwell:
> The New York State parkway shields bear the full name, but at least they do suggest initialisms using drop caps, so `ref` is OK

To me there is no distinction between this situation and the Kentucky shields.  I find it quite a stretch to say that it's definitely wrong to put an initialism in a `ref` tag under normal circumstances, but if the initialism is suggested by drop caps then it's perfectly OK 😁.

zekefarwell:
Looks like the [Oklahoma Turnpikes](https://en.wikipedia.org/wiki/Turnpikes_of_Oklahoma) shields are similar to the Kentucky Parkways.  They all use the same shield design, with only the name varying, and no ref.  The difference is that most of them are concurrent with a numbered Interstate, US, or State route, but a few are not.

ZeLonewolf:
There is also the MassPike:
https://en.wikipedia.org/wiki/Massachusetts_Turnpike#/media/File:Mass_Pike_shield.svg

It's not currently mapped at all, and it's concurrent with I-90 in Massachusetts.

1ec5:
The Oklahoma turnpikes are similar to the Palisades parkways and HCTRA in that the text is a minor part of the design, so the second point in https://github.com/ZeLonewolf/openstreetmap-americana/pull/119#issuecomment-1030727100 may apply. For these networks, we have a choice as to whether to remove all text from the shield for simplicity (effectively matching what drivers perceive on the ground) or to emphasize the text more than the actual shield does (so that the user can more easily distinguish different routes in the same network).

The Massachusetts Turnpike has a one-off shield. The second point in https://github.com/ZeLonewolf/openstreetmap-americana/pull/119#issuecomment-1030727100 definitely doesn’t apply in that case.

zekefarwell:
Here's a abstract rendition of the Kentucky parkway shield design that is a bit more visually interesting than a blank rectangle.  This could be used if we were to choose option 1 (render all KY parkways with a common shield).
shield40_us_ky_parkway_2
Kind of a joke, but maybe it actually would work? 😄 I don't know.  I'd say it is a reasonable approximation of what drivers perceive from a distance.

ZeLonewolf:
Some more terrible ideas:

ky_outline

kentucky_horse

1ec5:
Out of context, those terrible ideas look a lot more attractive than what the Kentucky parkways actually use. Wonder if they’re soliciting redesigns. 😏

To recap, here are the options proposed so far for Kentucky parkways:

1. Hard-code initialisms in the style as implemented in this PR.
2. Introduce logic to OpenMapTiles that coalesces `ref` with `short_name` when calculating `route_*` properties.
3. Retag the route relations with `ref`, holding our noses while tagging incorrectly for this renderer.
4. Design a common shield for the entire parkway system that doesn’t distinguish individual routes, taking some artistic liberty.
5. Add a special case to label motorways from z7 or z8 instead of z10 if the [index of](https://maplibre.org/maplibre-gl-js-docs/style-spec/expressions/#index-of) `US:KY:Parkway` within `route_1` is 0.

Options 1, 4, and 5 are all solvable purely within this repository. Each of the options except 4 allows the user to discern the individual routes at a glance.

ZeLonewolf:
So when I look at the green-style NY parkways:
image

...I agree that equating the larger font of the first letter of each word as the "ref" is a load of 💩 that we probably just did to make it show up on osm-carto.  Extending this to doing the same on the KY parkways is barely different.  If the next rev of the KY parkway signage switches to a style that emphasizes the first letter of the words -- does that really suddenly make it a ref?

If we're going to make 🐮💩 refs in NY,  I think we can make them in KY.  Or vice versa, if we went with the short_name versions here, that should apply in the other places too.

Can anyone articulate what the negative consequence of ref initialisms in KY would be on data consumers that isn't already a problem on NY parkways that are doing the same thing?

zekefarwell:
Here's a closer crop of the horses head, modified for better pixel alignment.  I think I turned it into a dragon. 🤨
1x: shield40_us_ky_parkway_horse 2x: shield40_us_ky_parkway_horse_2x

1ec5:
> Can anyone articulate what the negative consequence of ref initialisms in KY would be on data consumers that isn't already a problem on NY parkways that are doing the same thing?

Data consumers should be able to have confidence that whatever appears in `ref` is something that can be announced or displayed on its own if necessary. Overloading `ref` for initialisms of convenience can have an adverse effect on the guidance instructions generated by routers. For example, when a route turns onto [this way](https://www.openstreetmap.org/way/204718817) tagged `name=Lake Ontario State Parkway` `ref=LOSP`, part of [this relation](https://www.openstreetmap.org/relation/408496) tagged similarly, a router would reasonably say “Turn right onto Lake Ontario State Parkway, L&nbsp;O&nbsp;S&nbsp;P”.[^osrm] In a more urban, higher-stress parkway context, that amusing redundancy can cause confusion and missed turns. In fact, on `motorway`s and `expressway`s, a router is likely to skip the `name` in favor of the `ref` under the assumption that `ref` is just as meaningful but more concise: Project-OSRM/osrm-text-instructions#51.

> If we're going to make manufactured refs in NY, I think we can make them in KY. Or vice versa, if we went with the short_name versions here, that should apply in the other places too.

It’s a stretch to use `ref` for the initialism of a standard green New York parkway shield like the Korean War Veterans Parkway, but at least “KWVP” obviously visually relates to the shield, which is why I haven’t argued against it so far.

[](https://www.alpsroads.net/roads/ny/sawmill/)

A much better case for `ref` is the NTTA toll roads in https://github.com/ZeLonewolf/openstreetmap-americana/pull/119#issuecomment-1030711367, which happen to use the initialism as if it’s a route number. If a router skips “President George Bush Turnpike” in favor of “PGBT”, great.

[^osrm]: This is what [OSRM](https://map.project-osrm.org/?z=18&center=43.345383%2C-77.854775&loc=43.345112%2C-77.855371&loc=43.345572%2C-77.854174&hl=en&alt=0&srv=0) does in its voice instructions. OSRM only uses way refs, but the same challenge would remain if it were to incorporate relation refs into guidance instructions.

ZeLonewolf (approved):
Based on copious Slack discussions, I believe the consensus is:

1. We should not tag `ref` with fictional values to support shield rendering.  Instead these should go in other tags appropriate for initialisms.
2. The approach here is probably performant enough that it scales to the likely number of special cases that we might encounter worldwide.
3. The render samples look good and the author has local knowledge that suggests that the shields are likely to be locally understood.

A technical note - planetiler does not currently do road suffix abbreviations, so `Pkwy` would get expanded to `Parkway` if we transition to a planetiler-based tile source (and planetiler doesn't implement that functionality in the meantime).  In that case, we would need to add the fully-spelled out versions to the special case list as well.

zekefarwell (approved):

ZeLonewolf:
To add further to this - the solution in this PR is understood to be a temporary fix until other tagging can be added to the vector tile source (such as `short_name`, `colour`, or some other initialism/abbreviation tagging) which can be used to more properly construct these shields in a systematic way.

style/CONTRIBUTE.md
```diff
@@ -221,3 +221,23 @@ Shields should target 8-12px text actual-size character heights for readability:
...
+- **`norefImage`** – A reference to an alternative image file used when there is no `ref`. This is appropriate if some routes in the network have a `ref` tag and others do not.
+- **`notext`** – By default, a relation missing a `ref` tag will not appear as a shield. Set this property to `true` to display a shield even if it has no `ref`. This is appropriate for one-off shield networks, which are common for toll roads and touristic routes.
+- **`padding`** – An object that specifies the amount of padding on each side of the inscribed text relative to the background image.
+- **`textColor`** – The color of the inscribed text to superimpose on the background.
+- **`textLayoutConstraint`** – A strategy for constraining the text within the background image, useful for shields of certain shapes.
+
+Additionally, **`refsByWayName`** is an object mapping way names to text that can be superimposed on the background as a fallback for a missing `ref` value. (`refsByWayName` implies `notext`.) This temporary fallback is designed for use in [limited situations](https://wiki.openstreetmap.org/wiki/United_States/Unusual_highway_networks) that meet each of the following criteria:
```
1ec5, line 237: I took a pass on instructions for crafting a shield definition object. As part of these instructions, I detailed the rationale for `refsByWayName` with enough conditions that hopefully it won’t open the floodgates to any end-runs around OSM.

style/CONTRIBUTE.md
```diff
@@ -221,3 +221,23 @@ Shields should target 8-12px text actual-size character heights for readability:
...
+
+- **`backgroundImage`** – A reference to the image file used as the shield background, based on the name of the file in style/icons/. To use a different image depending on the length of the inscribed text, specify an array of images.
+- **`colorLighten`** – The color to recolor the background image by.
+- **`norefImage`** – A reference to an alternative image file used when there is no `ref`. This is appropriate if some routes in the network have a `ref` tag and others do not.
+- **`notext`** – By default, a relation missing a `ref` tag will not appear as a shield. Set this property to `true` to display a shield even if it has no `ref`. This is appropriate for one-off shield networks, which are common for toll roads and touristic routes.
+- **`padding`** – An object that specifies the amount of padding on each side of the inscribed text relative to the background image.
+- **`textColor`** – The color of the inscribed text to superimpose on the background.
+- **`textLayoutConstraint`** – A strategy for constraining the text within the background image, useful for shields of certain shapes.
+
```
ZeLonewolf, line 236:
```suggestion
- **`textLayoutConstraint`** – A strategy for constraining the text within the background image, useful for shields of certain shapes.  By default, the text will expand to fill a rectangle bounded by the specified padding while maintaining the same aspect ratio.

```

style/CONTRIBUTE.md
```diff
@@ -221,3 +221,23 @@ Shields should target 8-12px text actual-size character heights for readability:
...
+Additionally, **`refsByWayName`** is an object mapping way names to text that can be superimposed on the background as a fallback for a missing `ref` value. (`refsByWayName` implies `notext`.) This temporary fallback is designed for use in [limited situations](https://wiki.openstreetmap.org/wiki/United_States/Unusual_highway_networks) that meet each of the following criteria:
+
+- Each route in a network has a distinct shield that is dominated by the road name as opposed to a glyph or logo.
+- Each shield in the network differs by only the road name, not a number, initialism, or color.
+- Each route would be recognizable by an initialism, even though it is not signposted.
```
ZeLonewolf, line 241:
```suggestion
- Each route in a network has a distinct shield that is dominated by the road name as opposed to a glyph or logo.
- Each shield in the network has a common thematic design, and differs by only the road name, not a number, initialism, or color.
- Each route would be recognizable by an initialism, even though it is not signposted.
```

style/js/shield_defs.js
```diff
@@ -244,6 +244,37 @@ export function loadShields(shieldImages) {
...
+      bottom: 6,
+    },
+  };
+  shields["US:KY:Parkway"] = Object.assign(
+    {
+      refsByWayName: {
+        "Audobon Parkway": "AU",
+        "Bluegrass Parkway": "BG",
+        "Bluegrass Pkwy": "BG",
+        "Cumberland Parkway": "LN",
```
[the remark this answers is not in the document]
ZeLonewolf: Let's include this note as a comment here or open a new ticket so we don't lose track of it.

ZeLonewolf (approved):

PR #121, opened 2022-01-30
Allow circular and square shields to expand horizontally

### LINKED ISSUE #117: Support variable-width rectangular & oval signs

ZeLonewolf:
Closes #117

This PR allows canvas-drawn circle and square shields to expand in the horizontal direction dynamically as a function of ref length, up to a maximum of a 3:2 aspect ratio.  For circles I implemented this as an ellipse, though we may consider in the future changing this to a rounded rectangle to be more similar to wide-circular signs in places like Kentucky.

This also adds state highway shields for Kentucky (`network=US:KY`).

Rhode Island, 2,3, and 4-digit rectangular shields:
image

Kentucky 1,2,3, and 4-digit shields:
image


style/js/shield_text.js
```diff
@@ -105,17 +105,16 @@ function layoutShieldText(text, padding, bounds, textLayoutFunc, maxFontSize) {
...
       yBaseline = padTop;
       break;
     case VerticalAlignment.Bottom:
       yBaseline = padTop + availHeight - textHeight;
       break;
     default:
-      return null; //Code error, should never happen
+    case VerticalAlignment.Middle:
+      yBaseline = padTop + (availHeight - textHeight) / 2;
+      break;
```
zekefarwell, line 117: I guess I shouldn't be surprised that Javascript allows the `default:` case to be in the middle rather than last.  Unless there's some good reason I'm not grasping for the exact order as is, can we flip it around so the default case is last?
```
    case VerticalAlignment.Middle:
    default:
      yBaseline = padTop + (availHeight - textHeight) / 2;
      break;
```
Seems a bit clearer and the convention I'm used to seeing, even if JavaScript does allow just randomly chucking it in wherever 🤠.

zekefarwell (changes requested):
This all looks good to me except a few things I think should change about how the pixel ratio is handled with the min and max width values.

zekefarwell:
Oops, I was so focused on ovals that I forgot to look at rectangles.  Looks like something is slightly off with their math.  I'm seeing a missing border on one side in some cases (on a 1x screen).

Screen Shot 2022-01-30 at 4 11 18 PM

On a 2x screen the border is much thinner than other shields.

Screen Shot 2022-01-30 at 4 14 21 PM

zekefarwell (approved):

PR #127, opened 2022-01-31
Implement rounded rectangles as canvas drawing code

### LINKED ISSUE #123: Support canvas-drawn rounded rectangles
We should support canvas-drawn rounded rectangles, which generate arbitrary-width rounded-rectangle shields.  This will allow us to generate `US:KY` wide-length shields and can eliminate a number of European shields that are currently sprite graphics but don't need to be.  Additionally, we can re-implement the current rectangle shields as rounded rectangles with tiny radius, which is how they're generally drawn in reality.

_Originally posted by @1ec5 in https://github.com/ZeLonewolf/openstreetmap-americana/issues/117#issuecomment-1025013060_

Same goes for circular shields, which don’t fit longer refs very well. In fact, Kentucky has many four-digit-long refs, but they use rounded rectangles instead of ellipses.

KY 3076 KY 3076

ZeLonewolf:
Closes #123 

This PR implements rounded rectangles as draw code, and eliminates several shields that were previously drawn as raster (from NL, Slovakia, Austria).  This simplifies future shields that are just numbers/letters on a solid background, and reduces the size of the sprite sheet.  In addition, the rounded rectangles will size to fit rather than be reliant on a set of fixed size shield blanks.

image
image


zekefarwell (approved):
Looks good to me.  Tested and didn't see any issues in the places I looked.

Mashin6:
There were few things. Some of the text was too small compared to other shields so I increased the calculated width. Also the height was smaller than 40px(20px), because line extends the object only by 1/2 of its thickness.
(before-after)
image
image

Shields in some countries have extended padding on right left so I added fixed-width option for the shield (maybe it should rather have been custom min-width).
(befire-after)
image
image

And I also added option for custom outline width.

Mashin6 (approved):

ZeLonewolf:
Thanks!  I'm going to test this is a few other spots in the US to make sure it still works before merging.

Regarding the custom thicknesses, we should probably have a separate style discussion to decide if we actually want variability in outer stroke width.  For example, do we want thicker borders on CT shields and thinner ones on RI/MA that promotes more realism, or do we want a simpler style choice that promotes stylistic consistency?  This is a separate question from this PR but one we should address.

Mashin6:
> do we want thicker borders on CT shields

Yesss!! It's essential! 😂

here is how it looks in Apple maps
image
image

ZeLonewolf:
Well, Apple Maps has the luxury of designing for garaunteed high pixel density displays, whereas we still need to support 1x.  They can get away with much fiber details.

Mashin6:
Yeah, it needs to be tested.

zekefarwell:
I've looked at Apple maps (via DuckDuckGo) on a 1x display and the thicker borders look fine.  They custom design the 2x and 1x versions separately though.  The 2x versions include extra detail that is missing from the 1x versions.  Not sure we'd want to take on that doubling of effort.  I think we can certainly make thick and thin borders work at 1x, it's more of a design decision.  i.e. do we want to lean closer toward the real world sign designs like Apple Maps does, or more towards simplified, representational designs like Google Maps and MapBox Streets does.  I'm in favor representational simplicity, but I can get behind a more real world approach if that's what the community is more excited about.

1ec5:
I don’t know the full history behind the Connecticut shield, but the black part of it seems to me like an extra bit of background to contrast the signs from ordinary regulatory signs. [A proper border](https://mutcd.fhwa.dot.gov/SHSm/design.pdf#page=2) would be inset from the sign edge according to the MUTCD, and modern guide signs don’t feature the black border. We’ve been omitting the black background from shields for simplicity.

[guide sign](https://commons.wikimedia.org/wiki/File:IMAG0685_%285175363101%29.jpg)

That said, some countries do distinguish route signs by their borders, so the ability to make this distinction would be beneficial. For example, Finland distinguishes link roads by making the border dashed:

F29-1 F35-2-1 F30-40 F35-3-40 F28-4 F35-1-4 F31-100 F35-4-100

Mashin6:
No idea either. I definitely saw the thick background on a standalone route signs, not sure about those big information boards.
Found this document about signs. (pg. 72, 80, 82, 88)
https://portal.ct.gov/-/media/DOT/documents/dtrafficdesign/SignCatalogpdf.pdf

PR #129, opened 2022-01-31
Add Idaho highway shields

Pengor:
This shield is bit tricky with the padding since the number goes in the upper right but there are two shield sizes (1/2 digit routes & 3 digit routes). I opted to set the padding in such a way that the 3 digit route shield is always generated so at least all the route numbers are legible. I played around with the padding for quite a while but wasn't able to come up with any values that properly drew the 1/2 digit shield and also didn't have any digits overlap with the Idaho outline.


Pengor:
Some renderings:
image image image

ZeLonewolf:
Well, I guess it was a matter of time before we got to the difficult state shields, and I think this is going to be the first in a series of challenging state shields that we need to convert into digital-map-appropriate rendering.  Rendering below.  Readability is passable, but not great with the small text size on my display.  I think it'll be especially hard to read at 1x resolution.  I'm not convinced that a 1:1 recreation of the state shield is the right approach here given the space taken up by the black Idaho outline.

Would we consider a portrait-oriented Idaho shape with text in the bottom part?

image
image

Pengor:
Perhaps it's time to come up with a formal style guide for the project that identifies what should be prioritized in generating shield graphics? I'm open to making concessions to legibility rather than try and create a 1-to-1 likeness for shields, I certainly understand this is a different domain than freeway signage. My only concern is deviating from the "authentic" shields used by government agencies opens things up to more subjective opinions on design.

ZeLonewolf:
> Perhaps it's time to come up with a formal style guide for the project that identifies what should be prioritized in generating shield graphics?

I'm all for it!  So far I've been somewhat inspired by https://github.com/1ec5/rebusurance and shooting for a simplified design aesthetic rather than a 1:1 sign replication.

Pengor:
Awesome, I'll rework Idaho shield(s) for enhanced legibility once the guidelines from #133 get merged.

Pengor:
I think we're in business now! The Idaho outline didn't need to be made _too_ much smaller to find a nice balance between the actual highway shield and legibility. 

Here's 1x:
image

And 2x:
image

For whatever reason I wasn't able to devise a padding to get the 2 digit route numbers to use the 1/2 digit shield, but I don't think this is a major issue. It might just not play nicely with how padding works when the text is placed this far into a corner.

ZeLonewolf (approved):
Looks good, thanks for the contribution!

PR #130, opened 2022-01-31
Fix the map crashing when zoomed into Pittsburgh

jleedev:
Add handling for Purple Belt, color from Wikimedia

Note that the shield style for this should technically be different, and
of course it would be good if shield crashes could be recovered from.


ZeLonewolf (approved):
Code looks good, but #bd0063 seems closer to magenta?  But if you say that's the color, who am I to argue...

jleedev:
Stylistically the color comes from https://commons.wikimedia.org/wiki/Category:Pittsburgh_Wayfinder_System, and the name "Purple" is more of a passing reference to the county color belts.

PR #131, opened 2022-01-31
Recover from exceptions loading shields

jleedev:
Log the exception to the console, but allow the tile to continue
rendering without the missing image.


ZeLonewolf (approved):
Looks good.

Today I learned that `null.foo();` is perfectly valid javascript.

PR #133, opened 2022-02-01
Add shield guidelines to contributor's guide

ZeLonewolf:
This is a documentation update to give better guidelines to contributors working on highway shields.


style/CONTRIBUTE.md
```diff
@@ -124,3 +124,27 @@ restart the web server.
 
 Please prettify all files prior to submission. Run `make code_format` to format all
 code files with js-prettifier.
+
+## Highway Shield Contributor's Guide
+
+Highway shields are a key feature of the OpenStreetMap Americana style. This guide describes some of the style principles that contributors of highway shield artwork should consider when submitting new shields. The required elements are as follows:
+
+1. Shields are in .svg format, saved in "Optimized SVG" format in Inkscape
+2. Shields are 20px on the smallest dimension (this will be rasterized to 20px or 40x depending on display resolution)
```
1ec5, line 133:
```suggestion
2. Shields are 20px on the smallest dimension (this will be rasterized to 20px or 40px depending on display resolution)
```

style/CONTRIBUTE.md
```diff
@@ -124,3 +124,27 @@ restart the web server.
...
+3. Shields are license-compatible (public domain or CC0)
+
+In addition, the following style guidelines are recommended:
+
+1. Use a 1.25px stroke size for line elements.
```
1ec5, line 138: I’ve been wondering about this stroke size: I would’ve thought it would prevent pixel alignment, potentially making the image blurry on 1× displays. A hairline stroke would be great for the edge of a sign that’s meant just for contrast rather than as a border in its own right, but I’m not sure about other line elements.

style/CONTRIBUTE.md
```diff
@@ -124,3 +124,27 @@ restart the web server.
...
+2. Complex or intricate graphic elements should be simplified
+3. Background shading should be removed
+4. Route numbers should be large enough to be legible
```
1ec5, line 141: Another point to consider is that shields should consistently use squared or rounded corners, but not a mix of both.

style/CONTRIBUTE.md
```diff
@@ -124,3 +124,27 @@ restart the web server.
...
+| ---------------------- | ---------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------- |
+| NJ Turnpike            | <img src="https://upload.wikimedia.org/wikipedia/commons/a/af/New_Jersey_Turnpike_Shield.svg" width="40"/> | <img src="https://raw.githubusercontent.com/ZeLonewolf/openstreetmap-americana/main/style/icons/shield40_us_nj_njtp_noref.svg" width="40"/> |
+| Washington State Route | <img src="https://upload.wikimedia.org/wikipedia/commons/5/52/WA-blank.svg" width="40"/>                   | <img src="https://raw.githubusercontent.com/ZeLonewolf/openstreetmap-americana/main/style/icons/shield40_us_wa.svg" width="40"/>            |
+
+More complicated shields may be more challenging to simplify. Consider taking some inspiration from the [rebusurance](https://github.com/1ec5/rebusurance) project, which effectivly simplifies a number of complex state shield designs by stretching, compressing, omitting, or simplifying graphic elements.
```
1ec5, line 150: The very first step I did for the Rebusurance project was to adopt a unified color palette for every shield. Rebusurance uses the same standard [MUTCD colors](https://commons.wikimedia.org/wiki/Commons:MUTCD_colors) as Wikimedia Commons. These colors are effectively Americana’s color palette at the moment, because many of the shields were derived from Commons images, but we can choose a different color scheme that aligns with the rest of the map. Internal consistency is important because even slightly different shades of blue or green are easily perceptible on some screens. Shields from other countries should probably share the same color palette, with same exceptions for unusual colors like sky blue.

We could check in an [Inkscape color palette](https://commons.wikimedia.org/wiki/Commons:WikiProject_U.S._Roads/Shields#Installing_the_MUTCD_color_palette) for contributors to install. Maybe there’s even a tool out there for listing SVGs for compliance with this color palette that we could run on CI.

Pengor:
Just an idea that popped into my head: would a minimum allowable font size be a good guideline to include? Perhaps something like "Shields should allow for rendering a digit at least 10 pixels in height on a 1x display" (rough wording and not sure what value to actually set as a limit, but hopefully you get my point). I assume a height since character width is variable I believe.

Also maybe even some basic things like:
* Shield text padding should allow for at least 1 pixel between text and elements of a similar color.
* Shields with multiple-size variants should render with similar spacing between shield artwork and text.

I'm a graphics novice, but these would help guide me I'd think.

style/CONTRIBUTE.md
ZeLonewolf: This makes sense to me.  Other digital maps don't try to perfectly match shield colors so I would support a standard palette.  Also, the Inkscape color palette is a great idea.  However, I feel wildly incompetent in graphic design and would welcome help in the best way to document and provide such a palette.

style/CONTRIBUTE.md
ZeLonewolf: When the SVG size was 40px, I just eyeballed 2.5px as a nice looking size by eye, so I just divided by two to get 1.25px.  However, I'm not married to that line width.

ZeLonewolf: When we go through the SVG->raster pipeline, is there a way to ensure that those vertical/horizontal lines remain aligned to the pixel grid?  Are there things we can turn on in Inkscape to ensure that alignment?  Clearly, a 1px horizontal line that's straddling two pixel rows won't achieve a nice result.

zekefarwell: That is something the designer of the SVG has to do.  The screenshots above are from [Affinity Designer](https://affinity.serif.com/en-us/designer/) with gridlines turned on and set to 1px.  It looks like Inkscape has a similar grid feature but I'm less familiar with Inkscape.

ZeLonewolf: Fixed in c1e7753

ZeLonewolf:
An additional thread discussing whether or not have have a standard line thickness occurred here:

https://github.com/ZeLonewolf/openstreetmap-americana/pull/127#issuecomment-1027119758

Note that this PR describes more shield consistency (standard line width, standard palettes) than is currently implemented in the map, and implies that the existing artwork should be adjusted to be consistent with these guidelines.

ZeLonewolf:
I've updated the color palette documentation and created an importable file for Inkscape.

Should we set a minimum font size for legibility?

Pengor:
I looked into minimum font size guidelines for different platforms and came up with the following in no particular order:
| Platform | Minimum font size |
| -------- | ----------------- |
| Google (Material Design) | 10sp[^1] |
| Apple (iOS) | 11 points[^2] |
| Apple (macOS) | 10 points[^3] |
| Microsoft (Windows apps) | 12 px[^4] |

[^1]: [The type system - Material Design](https://material.io/design/typography/the-type-system.html#type-scale)
[^2]: [Typography - Visual Design - iOS - Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography/)
[^3]: [Typography - Visual Design - macOS - Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/macos/visual-design/typography/)
[^4]: [Typography in Windows Apps](https://docs.microsoft.com/en-us/windows/apps/design/style/typography)

zekefarwell:
The latest comments on color in #139, make realize that while it will be good to adopt a standard palette, perhaps limiting shields to only one shade of each main color (green, blue, etc) may be a bit too strict.  Perhaps the palette defines the allowed hues, and then we define a couple different standard variations of saturation and lightness within each hue.  So we might have dark and light variants of blue and green for example.

zekefarwell:
Yeah we should probably just move ahead with this color palette for now.  We can consider expanding it once the map includes a more complete set of features.

Issue #135, opened 2022-02-03
Add an OpenMapSamples control to the demo


adamfranco:
I’m happy to report that I’ve now built a prototype MapLibre control that can temporarily replace the data sources used by a map style with dynamically generated GeoJSON:
* https://github.com/adamfranco/OpenMapSamples-MapLibre
* https://github.com/adamfranco/OpenMapSamples

A demo is available at https://adamfranco.github.io/OpenMapSamples-MapLibre/

The reason I wanted to dynamically generate the data rather than hand-coding it can be seen in the definition of the [HighwayAttributes sample](https://github.com/adamfranco/OpenMapSamples/blob/main/samples/OpenMapTiles/HighwayAttributes.js). A simple two-dimensional array of tag combinations is all that’s needed to generate sample data that is spaced out to be visible at 16 different zoom levels. To hand-map this one sample’s examples would have required mapping 2016 individual ways. :wink:

The next step is to iron out some more example data to add labels and improve coverage. The [OpenMapTiles schema documentation](https://openmaptiles.org/schema/) isn’t quite sufficient to figure out what is needed for a more realistic mock — it doesn’t note data types or full property formatting and seems a bit stale as well. 

Once this control and the samples are more fleshed out we can wire up this control in the Americana demo so that we can jump around to different examples as we work on style definition.

One thing I ran into in a first attempt at including this control in Americana is that the current `make` build process doesn't do any bundling step, so references to dependent `node_modules/` don't get resolved. #134 will help with that, but we may need to bundle the demo with either [Parcel](https://parceljs.org/) (what I've used for my projects), or [WebPack](https://webpack.js.org/). See https://levelup.gitconnected.com/parcel-vs-webpack-2021-64c347bcb31 for a discussion of the pros & cons of each.

adamfranco:
I've now successfully gotten the OpenMapSamples-Maplibre control pulled into the Americana demo:
![Screen Shot 2022-02-07 at 12 14 56 PM](https://user-images.githubusercontent.com/25242/152838783-825df521-4aa7-43f8-aa98-013cb7500d36.png)
[...]

Issue #132, opened 2022-01-31
`npm i` fails if node-mapnik binaries aren't available for arm/arm64/win-x86


adamfranco:
I'm trying to set up americana on a new M1/arm64 Macbook, but found that installation is failing because the `mapnik-node` binary doesn't exist for the arm64 architecture.

The x64 binary exists: 
https://mapbox-node-binary.s3.amazonaws.com/mapnik/v4.5.8/Release/darwin-x64.tar.gz
but the arm64 binary doesn't exist and triggers a custom build of : 
https://mapbox-node-binary.s3.amazonaws.com/mapnik/v4.5.8/Release/darwin-arm64.tar.gz

Upstream issue: mapnik/node-mapnik#945

`npm i` fails with this error:
```
npm ERR! code 1
npm ERR! path /Users/afranco/OSM-Dev/openstreetmap-americana/style/node_modules/mapnik
npm ERR! command failed
npm ERR! command sh -c node-pre-gyp install --fallback-to-build
npm ERR! Failed to execute '/usr/local/bin/node /usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js configure --fallback-to-build --module=/Users/afranco/OSM-Dev/openstreetmap-americana/style/node_modules/mapnik/lib/binding/mapnik.node --module_name=mapnik --module_path=/Users/afranco/OSM-Dev/openstreetmap-americana/style/node_modules/mapnik/lib/binding --napi_version=8 --node_abi_napi=napi --napi_build_version=0 --node_napi_label=node-v93' (1)
npm ERR! node-pre-gyp info it worked if it ends with ok
npm ERR! node-pre-gyp info using node-pre-gyp@1.0.5
npm ERR! node-pre-gyp info using node@16.13.2 | darwin | arm64
npm ERR! node-pre-gyp info check checked for "/Users/afranco/OSM-Dev/openstreetmap-americana/style/node_modules/mapnik/lib/binding/mapnik.node" (not found)
npm ERR! node-pre-gyp http GET https://mapbox-node-binary.s3.amazonaws.com/mapnik/v4.5.8/Release/darwin-arm64.tar.gz
npm ERR! node-pre-gyp ERR! install response status 403 Forbidden on https://mapbox-node-binary.s3.amazonaws.com/mapnik/v4.5.8/Release/darwin-arm64.tar.gz 
[...]

PR #136, opened 2022-02-04
Replace Makefile with npm scripts

### LINKED ISSUE #134: Replace Make targets with NPM scripts
We’re already using NPM for package management, so using it also for the build system would simplify the development environment, especially for Windows users who may not have `make` installed. Each of the Make targets can be replaced by a `script` in the package.json file, so developers can run for instance `npm run code_format`.

adamfranco:
This PR fixes #134, replacing the Makefile with npm scripts defined in the `package.json`.

I'm submitting this PR as a stand-alone without further changes to the build process. As noted in #135, it will likely be useful with either [Parcel](https://parceljs.org/) (what I've used for my projects), or [WebPack](https://webpack.js.org/) to minify the JS, combine CSS, and resolve paths to dependencies in node_modules. I'll submit a separate PR that builds on this one and uses parcel to do the bundle/build. In the event that WebPack is preferred or a different build process is desired, this PR can be used alone.


ZeLonewolf (commented):
This looks good to me.  I would like @Mashin6 to review also to confirm that this all works as expected on Mac.

adamfranco:
More review is welcome. I'm pretty new to npm myself and am sure there are more canonical ways of doing things than I came up with (such as calling scripts from scripts without just chaining them in a shell with `&&` 🤷 ).

FWIW I'm put this together on an x64 Mac running Catalina and have tested on an m1 Mac running Big Sur after going the steps in [#132](https://github.com/ZeLonewolf/openstreetmap-americana/issues/132#issuecomment-1028271663).

zekefarwell:
This project was the the first time I encountered a Makefile in a web development project.  I think it was something @akadouri set up early on.  I thought it was a bit weird at first and wondered why we weren't just using npm scripts, which I was more familiar with.  However, since then I've come to appreciate the flexibility of the Makefile syntax and I've switched over a couple other projects to using Makefiles instead of npm scripts.  I can't say I see what the need to make this change really is, but if everyone else is on board it's fine with me.  I'm comfortable with either tool.

Mashin6:
I managed to run it on my arm64 Mac fine. I can't comment on the code itself since I know nothing about javascript or node.js

akadouri:
Those were quite the early days @zekefarwell. I didn't really have any specific reason for going that way instead of npm scripts, at work we stick to Makefiles for consistency across projects no matter the language/framework. The important thing is making sure commands people need are committed and easy enough to find.

Mashin6 (approved):

ZeLonewolf:
Looks good!

PR #139, opened 2022-02-05
Shields for European countries

Mashin6:
Germany:
image

Protugal, Spain:
image

France:
image

Iceland:
image

Netherlands S-shields taking care of #96:
image

Poland, Lithuania, Latvia, Estonia, Finland, Belarus:
image

Bulgaria:
image

Bosnia and Herzegovina:
image

Ukraine, Russia:
image

Greece:
image


Dimitar5555:
The colour for Bulgarian highways is too dark. It should be the same as the lighter green, used in `E 79`. Also can you show https://www.openstreetmap.org/way/41196732 (both shields should be rendered in blue) and https://www.openstreetmap.org/relation/5650874 (not sure what colours would be used here. We don't have any signs that reference municipal roads).

> Bulgaria: image

1ec5:
In #133, we’re considering some shield design guidelines for consistency within the style, including a standard color palette. These guidelines may end up requiring some shields to have a slightly different appearance than in reality, in the name of consistency and legibility on a map. In particular, some sort of stroke around he shield would be desirable to ensure contrast with the map. However, we would also ensure that any two green shields use the same shade of green.

Mashin6:
So this color scheme is not used any more? https://en.wikipedia.org/wiki/Trakia_motorway

I personally don't think it's a good idea to have E-shields in the same color as national highway shields. In places with several concurrencies this gets really hard to visually parse e.g. Italy right now.

ZeLonewolf:
> In places with several concurrencies this gets really hard to visually parse e.g. Italy right now.

The concurrencies being visually hard to parse in Italy is an accurate reflection of reality:
image

Mashin6:
Exactly, and not only there. That's why it would be good to make it less eye straining. 

Also I don't understand why do countries place their shields on a background of the exact color. If you wanted to hide something you couldn't do it better.

Dimitar5555:
> So this color scheme is not used any more? https://en.wikipedia.org/wiki/Trakia_motorway

I'm not even sure from where they got that shade of green. This is the one that's used (https://www.google.com/maps/@42.7657364,23.2963817,3a,41.6y,287.13h,99.85t/data=!3m6!1e1!3m4!1stmOobM21MQs9JrakqHGpxA!2e0!7i16384!8i8192).


> I personally don't think it's a good idea to have E-shields in the same color as national highway shields. In places with several concurrencies this gets really hard to visually parse e.g. Italy right now.

It all depends on font weight, size and colour combinations. Personally I don't think that our signs are hard to visually parse but it might be because I'm used to them.

ZeLonewolf:
> @ZeLonewolf discussion seems dead now

From that discussion, it seems clear to me that there isn't a consensus in the Germany mapping community as to whether or not road route relations should even be mapped, never mind what the scheme should be on classifying multiple levels of route.  Given the uncertainty regarding those road relations, the apparent lack of consensus on the issue, I feel it would be inappropriate for this project, as a route relation shield renderer, to decide that the current tagging is the winner and provide that as mapper feedback.  By doing so, we would be reinforcing the current scheme which does not appear to have consistent community support.

In the future, if the tagging/mapping scheme for route relations in Germany becomes more established, acceptance by the community emerges, and/or other data consumers begin supporting it, at that point it would be appropriate to add them to the style.

coolultra1:
Couldn't we just render the ref as it is and if Germans get mad, they'll
change it?

ZeLonewolf:
> Couldn't we just render the ref as it is and if Germans get mad, they'll change it?

I would prefer not to maintain a library of workarounds for issues that are solvable in tagging.

zekefarwell:
> Couldn't we just render the ref as it is and if Germans get mad, they'll change it?

We are currently rendering the refs exactly as they are, just no custom shield design.  I see no problem with this given the non-standard network value.

Screen Shot 2022-02-19 at 1 11 45 PM

style/js/shield_defs.js
```diff
@@ -1191,6 +1203,209 @@ export function loadShields(shieldImages) {
...
+  shields["BAB"] = {
+    backgroundImage: shieldImages.shield40_de_2,
+    textColor: "white",
+    padding: {
+      left: 3,
+      right: 3,
+      top: 4,
+      bottom: 5,
+    },
+  };
```
JannikGM, line 1237: German "BAB" shields contain "A 7" in `ref`, but should be displayed as "7" - the design implies that it's an Autobahn ("A").
So I'd propose to check if the `ref` starts with "A #" and if it is, then use "#", otherwise warn and use the full ref?

ZeLonewolf: Please see the discussion on this topic on #149

In short, I feel this would be a workaround for something that's essentially a mapping problem.  If the "A" is not included in the highway shield, then it should not be included in the `ref` tag on the road's highway relation.

JannikGM: >  If the "A" is not included in the highway shield, then it should not be included in the ref tag on the road's highway relation

While it's not part of the sign, it's part of the name of the road, in addition to being an Autobahn.
In spoken communication (including radio) you say "Autobahn A 1". On some roadside safety markers it apparently says "A1" or it might just say "1" - there seem to be changes over time or regionally.

So, I have a hard time deciding this (and so does the German community apparently).

What makes this worse is that OSM seems to differentiate between references for [relations](https://wiki.openstreetmap.org/wiki/Key:ref#Examples_on_relations) and [ways](https://wiki.openstreetmap.org/wiki/Key:ref#Examples_on_ways):
- Relations: Machine readable and meant to be used together with `network`; "bare number"
- Ways: Human readable ("What's on the road")

Regardless of the above, with this being a `ref[erence]`, I think "A 7" is most clear & descriptive.

I also don't think bare numbers will work globally and as long as it's consistent within the network, it'll be fine.
If we define `network=DE:BAB` to have the "A " prefix in `ref`, that's probably not too bad?
The consumer of the data must be aware of the fact that "DE:BAB" means "A " prefix regardless (either to remove it, or to add it).
There's special code required to handle this shield in Americana anyway.

(Note, however, that I'd agree that `traffic_sign=DE:405:[number]` should not have the "A " prefix, as it's specifically meant for traffic-signs and shields)

ZeLonewolf: You're mixing up two separate taggings.  On ways, network and ref information are packed into the ref tag.  I.e. `ref=A1`.  The "A" is the network, and the "1" is the route number.  It's the same in the US, ways on US interstate highways have `ref=I-95` for Interstate 95.  Route relations are different because network and ref are separately tagged, so we have `network=US:I` and `ref=95` for the same road.  Note the difference in ref tagging.

IMO, what would make sense in Germany is to have `ref=A1` on the ways, and `network=DE:BAB` + `ref=1` on the relation.  That would be consistent with the preponderance of global usage.  We're waiting to see what the German mapping community comes up with, but at this point, prominent members of the German mapping community are even going so far as to argue against having route relations at all.  So in that light, it would be inappropriate for us to craft workarounds just to make the shields render in a certain way while the community is still deciding how to handle route tagging, and whether to even tag them at all.

Blijbol:
Perhaps move Germany into a separate pull request, so that the other countries can be merged?

ZeLonewolf:
> Perhaps move Germany into a separate pull request, so that the other countries can be merged already?

This is what I'm planning to do, based on what we've done in #309 and #310

ZeLonewolf:
Fixes #347

claysmalley:
One question remains: do we want to standardize on a 1px border where the real-life shield may not have a border? I think it looks awkward to give some shields a border and leave it off others.

I think the European heraldry principle mentioned by @Blijbol in #360 is relevant here, where yellow and white may touch other colors (including black), but yellow shouldn't touch white and the other colors shouldn't touch each other either. 

Screenshot from 2022-06-10 19-35-59 Screenshot from 2022-06-10 19-35-23
Screenshot from 2022-06-10 19-36-51 Screenshot from 2022-06-10 19-37-05

ZeLonewolf:
My vote would lean towards "a border, always" and add that to the style guide for shields.  I think the consistency outweighs any on the ground replication principle here.

zekefarwell:
I'm also in favor of always having a border for better readability and consistency

ZeLonewolf (approved):
Looks good to me.  I scrolled all over Europe and no obvious problems.  I like the Moldovan shields the best out of all of these :)

PR #140, opened 2022-02-05
Add Arizona Shields

amcginni:
This pull request adds the Arizona state highway shield for networks `US:AZ` and `US:AZ:Loop`.  A rendering of the Phoenix area with the shields is shown.
image


style/js/shield_defs.js
```diff
@@ -1,619 +1,634 @@
...
- * @param {*} shield - Shield definition
- * @returns the number of banner placards that need to be drawn
- */
-export function getBannerCount(shield) {
-  if (shield == null || typeof shield.modifiers == "undefined") {
-    return 0; //Unadorned shield
-  }
-  return shield.modifiers.length;
-}
+"use strict";
```
1ec5, line 1: Looks like this PR changed the file to use Windows-style line endings instead of Unix-style line endings. Run `npm run code_format` to change it back and apply any other formatting fixes.

1ec5 (commented):
Thanks – these Arizona shields look great.

style/js/shield_defs.js
```diff
@@ -1,619 +1,634 @@
...
+    ],
+    textColor: "black",
+    padding: {
+      left: 4,
+      right: 3,
+      top: 3,
+      bottom: 4,
+    },
+  };
+  shields["US:AZ:Loop"] = shields["US:AZ"];
```
1ec5, line 174: This causes the Loop shields to look identical to the normal state route shields. Since the loop routes numerically overlap, it may be important for users to know that a loop route is a loop route. To display the word “Loop” on top of the state shape, set the object’s `modifiers` property to an array containing the string `Loop`, as with the `US:US:Business` definition above. To keep the `US:AZ:Loop` definition from affecting the `US:AZ` definition, you’ll need to either duplicate the whole object or call [`Object.assign()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign).

ZeLonewolf: I would suggest making the loop modifer `LOOP` in all-caps.

style/js/shield_defs.js
amcginni: Thanks for the advice!  This is addressed in bd0753125d96e3d5af1cb12e0c7e249e7dd3bd38

ZeLonewolf:
Hmm, I'm not sure how it let me commit to your branch, but I just updated the 2-digit shield to align the artwork to the pixel grid and square the corners per below.  Please verify that the shield still looks good.

image

style/js/shield_defs.js
```diff
@@ -158,6 +158,21 @@ export function loadShields(shieldImages) {
...
+
```
ZeLonewolf, line 175:
```suggestion
  shields["US:AZ:Loop"] = banneredShield(shields["US:AZ"], ["LOOP"]);

```

amcginni:
> Hmm, I'm not sure how it let me commit to your branch, but I just updated the 2-digit shield to align the artwork to the pixel grid and square the corners per below. Please verify that the shield still looks good.

Looks good to me.

style/js/shield_defs.js
amcginni: While I don't see it as strictly necessary, as the loop routes currently don't numerically overlap with other state routes, I don't see anything wrong with displaying the LOOP banner either.

ZeLonewolf:
These shields look great!

I'm not seeing the loop routes show up for some reason, I'm investigating to see what might be going on here.
image

amcginni:
I think this was forked before the banneredShield function was added.  After adding that function to my local branch, the shields display as intended.  I would assume that this would work after merging back into main.

ZeLonewolf:
Ah yes.  And because this is javascript, you can just call imaginary functions and the interpreter thinks this is all perfectly fine...

1ec5:
> Hmm, I'm not sure how it let me commit to your branch

By default, when you open a pull request from a fork, GitHub lets the upstream repository’s maintainers commit directly to the PR branch on your fork.

> I would assume that this would work after merging back into main.

It would be a good idea to merge main into this branch or rebase this branch onto main, just to make sure it works as intended.

zekefarwell:
I merged in `main` so should be up to date now

amcginni:
> 

That works too.  Thanks!
image

PR #137, opened 2022-02-04
Bundle with Parcel

adamfranco:
This is a follow-on to #136 which also uses [Parcel](https://parceljs.org/) to bundle the project to a single minified JS file with all dependencies included. This isn't strictly needed currently, but will reduce the number of HTTP requests  from clients and allow addition node_module dependencies to be referenced as noted in #135.

PR #144, opened 2022-02-07
Add OpenMapSamples

### LINKED ISSUE #135: Add an OpenMapSamples control to the demo

adamfranco:
This is a PR to add the [OpenMapSamples-Maplibre](https://github.com/adamfranco/OpenMapSamples-MapLibre) control to the Americana demo. Will fix #135.

Still to do:
- [x] Complete and merge #137 so that this control and its dependencies can be included.
- [x] Shift the samples button to not conflict with the MapTiler attribution.


adamfranco:
More samples are now ready for use. 🎉 

Screen Shot 2022-02-08 at 6 43 57 PM
Screen Shot 2022-02-08 at 6 42 06 PM
Screen Shot 2022-02-08 at 6 41 50 PM
Screen Shot 2022-02-08 at 6 41 35 PM

adamfranco:
Ok, this PR is now installing reliably now that I've figured out how to force the `openmapsamples-maplibre` dependency to use our forked `maplibre-gl` instead of any other version. This require npm 8.30+ to get the new `overrides` feature, but that is the current stable version, so that shouldn't be an issue.

style/package.json
```diff
@@ -10,16 +10,25 @@
...
   },
   "devDependencies": {
-    "browser-sync": "^2.27.7",
+    "maplibre-gl": "https://github.com/ZeLonewolf/maplibre-gl-js/releases/download/v2.0.5/maplibre-gl.tar.gz",
+    "openmapsamples-maplibre": "github:adamfranco/OpenMapSamples-MapLibre",
+    "parcel": "^2.3.0",
     "prettier": "2.3.2"
+  },
+  "engines": {
+    "npm": ">=8.3.0"
```
1ec5, line 29: Can you add an .nvmrc file containing something like `stable` for those who use `nvm` to manage their Node versions? Looks like `engines` doesn’t enforce the version that `nvm` uses, unfortunately: nvm-sh/nvm#651.

adamfranco: We could do that, but it seems that `stable` is a deprecated alias: https://stackoverflow.com/questions/68357267/nvm-use-stable-sets-node-version-to-older-than-version-installed-by-nvm-insta

We also probably don't want to hard-code to a single Node version, just provide a minimum version, but I'm open to suggestions. :-)

adamfranco:
I broke the build and deploy action, but #152 should fix it. Once that is in place this should merge and deploy cleanly. 🤞

ZeLonewolf:
An `npm start` failed out of the box for me:

image

adamfranco:
Did you run `npm i --include=dev` again after checking out the branch?

ZeLonewolf:
That gives me:
```
brian@laptop:~/git/openstreetmap-americana/style$ npm i --include=dev
npm ERR! code ENOTSUP
npm ERR! notsup Unsupported engine for openstreetmap-americana@0.1.0: wanted: {"npm":">=8.3.0"} (current: {"node":"14.18.2","npm":"6.14.15"})
npm ERR! notsup Not compatible with your version of node/npm: openstreetmap-americana@0.1.0
npm ERR! notsup Not compatible with your version of node/npm: openstreetmap-americana@0.1.0
npm ERR! notsup Required: {"npm":">=8.3.0"}
npm ERR! notsup Actual:   {"npm":"6.14.15","node":"14.18.2"}

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/brian/.npm/_logs/2022-02-11T14_50_05_834Z-debug.log
```

ZeLonewolf:
Ideally we should be able to hide this low-level setup stuff behind top-level run commands.

ZeLonewolf:
Okay, apparenly I'm getting this now regardless of branch.  Really annoying how npm seems to always end up in some kind of version hell that you can't escape from.

adamfranco:
Are you using nvm to install node/npm or a global install of node?

https://github.com/nvm-sh/nvm

adamfranco:
I hadn't heard of nvm until fighting with getting the right versions set up
on Apple m1. It lets you install many versions of node and switch which one
is active. We should probably update the install instructions.

style/package.json
zekefarwell: 8.3 here is also the [NPM version](https://github.com/npm/cli/releases) not the [Node version](https://nodejs.org/en/download/releases/).   I don't think NVM helps with manageing NPM versions, but I could be wrong. NPM can update itself to the latest version with `npm update -g npm`.

zekefarwell:
> We should probably update the install instructions.

I can do this later today.  I like writing documentation.

zekefarwell (approved):
Seems to work fine for me.  I tested with NPM 8.4 and 8.5 running on Node 14 and 16.

PR #145, opened 2022-02-07
Remove uneeded map setup code

ZeLonewolf:
This PR removes unneeded/duplicated map setup code, and deletes duplicated attribution text.


adamfranco (approved):
This looks fine to me. I'll open an issue to detail some more refactoring that I think would be useful, but this is a good first step.

PR #150, opened 2022-02-09
Minor roads

adamfranco:
Add minor and service roads.

Before:
Screen Shot 2022-02-09 at 10 10 32 AM

After:
Screen Shot 2022-02-09 at 3 17 13 AM

Minor roads (unclassified & residential) are introduced at z12 slightly smaller than tertiary. 

Service roads are broken into two classes:
1. Plain service and `service=alley`. (introduced at z=13)
2. `service=driveway` and `service=parking_aisle` (Smaller, introduced at z15)

All 3 layers use the same minZoomFill of 16 as does tertiary, so that tertiary on down switch to fill at the same time.

z15
z16
z19

### DIFF
--- a/style/americana.js
+++ b/style/americana.js
@@ -64,6 +64,7 @@ americanaLayers.push(
   lyrRoad.secondary.casing(),
   lyrRoad.tertiaryExpressway.casing(),
   lyrRoad.tertiary.casing(),
+  lyrRoad.minor.casing(),
 
   lyrRoad.motorwayLink.casing(),
   lyrRoad.trunkLink.casing(),
@@ -71,6 +72,7 @@ americanaLayers.push(
   lyrRoad.motorwayLink.fill(),
   lyrRoad.trunkLink.fill(),
 
+  lyrRoad.minor.fill(),
   lyrRoad.tertiary.fill(),
   lyrRoad.tertiaryExpressway.fill(),
   lyrRoad.secondary.fill(),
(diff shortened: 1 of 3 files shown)


Mashin6:
Very nice! 
Can do something about the street names so they are more readable? Maybe thicker casing or font, change of color?

adamfranco:
@Mashin6, Do you mean changing the street names for minor/service roads? At particular zooms? Or all road labels?

If just minor roads, then it might make sense to keep using the `offsetLabelStyle` instead of switching to the [`centerLabelStyle` at z17](https://github.com/ZeLonewolf/openstreetmap-americana/pull/150/files#diff-e92f21e3a4442a90897223492258e986deb56b777ef539dc8f5b578aef8d8e65R163) since even with fill, these lines are very narrow.

zekefarwell (dismissed):
Looks good to me.  Since these roads currently aren't rendered at all I think we should just go ahead and merge this in.  There's still plenty of room for refinement with the whole road network, but that can come in future PRs.

Mashin6:
@adamfranco I originally meant it for the your screenshot example. The street name labels get criss-crossed by the minor roads and because the font has the same color and also thickness I can barely read some of them. But since this problem is also caused by overlapping with secondary/ tertiary roads to some degree, it seems it might be better as a separate PR.

ZeLonewolf (approved):
Below is a commercial area with lots of parking aisles and they're barely discernable from the residential roads around it.

image

There is probably some tweaking we could do along these lines, but I think in general it looks pretty good and it's fine as a starting point to get these features on the map so we can start using the openmapsamples controls.

adamfranco:
I've improved the differentiation between minor/residential and service roads (as well as big and small) by fixing the color on zoom. Need a re-review so that this can be merged. :-)

ZeLonewolf (approved):
Looks good, thanks!

PR #151, opened 2022-02-10
Add shields for US ALT, BYP, and SPUR

arch0345:


ZeLonewolf (approved):
Probably should get rid of that code duplication at some point, but this looks good to me.

PR #152, opened 2022-02-11
Fix build and deploy action with parcel by updating Node version.

adamfranco:
After merging #137 I saw that [build and deploy](https://github.com/ZeLonewolf/openstreetmap-americana/runs/5151994151?check_suite_focus=true) failed due to using Node 16.x rather than Node 17 (with npm 8.3.0+) that we are now requiring. 
Screen Shot 2022-02-11 at 12 27 09 AM
 
I'm hoping that these changes to the workflow will fix the deployment.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -11,10 +11,10 @@ jobs:
     steps:
       - name: Checkout 🛎️
         uses: actions/checkout@v2.3.1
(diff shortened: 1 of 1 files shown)


adamfranco:
Also, go ahead and merge right away after marking as reviewed.

ZeLonewolf (approved):

PR #153, opened 2022-02-11
Remove duplication in banner definitions

ZeLonewolf:
This PR eliminates the duplication in defining "bannered" routes by adding a `banneredShield()` function that creates a bannered version of a base shield.

Seems to still be working after this change:
image
image

### DIFF
--- a/style/js/shield_defs.js
+++ b/style/js/shield_defs.js
@@ -3,6 +3,7 @@
 import * as ShieldDraw from "./shield_canvas_draw.js";
 import * as ShieldText from "./shield_text.js";
 import * as Gfx from "./screen_gfx.js";
+import * as Util from "./util.js";
 
 //Height of modifier banners
 export const bannerSizeH = 8 * Gfx.getPixelRatio();
@@ -52,6 +53,13 @@ function roundedRectShield(
     maxFontSize: 16,
     textColor: textColor,
   };
+}
+
+function banneredShield(baseDef, modifiers) {
+  let newShield = Util.cp(baseDef);
+  newShield.backgroundImage = baseDef.backgroundImage;
+  newShield.modifiers = modifiers;
+  return newShield;
 }
 
 export function loadShields(shieldImages) {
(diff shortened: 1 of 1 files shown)


style/js/shield_defs.js
```diff
@@ -54,6 +55,13 @@ function roundedRectShield(
   };
 }
 
+function banneredShield(baseDef, modifiers) {
+  let newShield = Util.cp(baseDef);
+  newShield.backgroundImage = baseDef.backgroundImage;
```
jleedev, line 60: Redundant?

ZeLonewolf: So `backgroundImage` in this case is a pointer to the sprites object in maplibre and doesn't survive a stringify copy - that's why I put that there.  Is there a better way perhaps to change that pointer to a string that can be a lookup into the sprite sheet maybe?  🤔

jleedev: Ah, I see there's a Uint8Array in there.

Sure, you could change the backgroundImage to be an array of string keys into shieldImages.

As is would be fine with a comment.

jleedev: I think you're probably also clobbering textLayoutConstraint.

ZeLonewolf: Ah yes that's true, and not apparently because no bannered route uses an unusual text layout constraint.  I'm wondering if the shield definition objects should be more portable or if we don't really care.

jleedev: Same suggestion for the constraints applies, either reference them by name to allow copying in this manner, or comment that they need to be reinstated.

Or switch to TypeScript and let the compiler do my job for me :)

ZeLonewolf: > Or switch to TypeScript and let the compiler do my job for me :)

Uhh... 😅

Pengor:
This may be a bit outside the scope of this PR but it's at least closely related. Should a `FUTURE` banner be added for US Routes? I just happened to encounter this signage in Spokane, WA:
image
[image source](https://www.mapillary.com/app/?pKey=2347391622071374)

It looks like this particular section of road is tagged `ref=US 395 Fut` but I noticed [this highway tagging page on the wiki](https://wiki.openstreetmap.org/wiki/United_States_roads_tagging/Routes#Banners) suggests using `ref=* Future` rather than `Fut` so I may retag it if the banner is added to Americana.

style/js/shield_defs.js
```diff
@@ -54,6 +55,17 @@ function roundedRectShield(
...
+  let newShield = Util.cp(baseDef);
+  newShield.modifiers = modifiers;
+
+  //Restore clobbered pointers
+  newShield.backgroundImage = baseDef.backgroundImage;
+  newShield.textLayoutConstraint = baseDef.textLayoutConstraint;
+
+  return newShield;
+}
+
```
zekefarwell, line 68: I think a shallow clone is all that's needed here.  i.e:
```
function banneredShield(baseDef, modifiers) {
  return {
    ...baseDef,
    modifiers: modifiers,
  };
}
```

zekefarwell (dismissed):
Looks good to me

style/js/shield_defs.js
ZeLonewolf: Gonna try this and see if it still works.  I like the sound of the shallow clone better than some mysterious clobber comments that nobody will understand.

ZeLonewolf: Yup... this is the way.

Ready for reapproval.

PR #156, opened 2022-02-11
OpenMapSamples redux

adamfranco:
Undo the revert of openmapsamples merge. Needs some npm install fixup to work on a clean node_modules directory.

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -1,4 +1,4 @@
-name: Lint
+name: Test Build
 
 on:
   pull_request:
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -1,0 +1,23 @@
+name: Lint
+
+on:
+  pull_request:
+    branches: [main]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout 🛎️
+        uses: actions/checkout@v2.3.1
+      - name: Use Node.js 17.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 17.x
+      - name: Install and Build 🔧
+        run: |
+          cd style
+          npm install --include=dev
+          sed 's/<your MapTiler key>/53iZvB2drcamS0Ge0xiD/g' config.default.js > config.js
+          npm run build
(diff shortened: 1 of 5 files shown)


adamfranco:
Ok, this PR is now ready for review. I've added a **Test Build** check to on pull requests to ensure that the PR will not break the build.

The big changes upstream in OpenMapSamples are that I moved the demo site to its own package at [OpenMapSamples-examples](https://github.com/adamfranco/OpenMapSamples-examples) to remove as many dependencies as possible from the [OpenMapSamples](https://github.com/adamfranco/OpenMapSamples) library itself and the [OpenMapSamples-MapLibre](https://github.com/adamfranco/OpenMapSamples-MapLibre) control. It turns out that the demo had a dependency on the `dotenv` package which was breaking for some reason after an update and then causing the Americana build to fail. With these extra dependencies removed, builds should be much more stable.

I also added a[ Places sample](https://adamfranco.github.io/OpenMapSamples-examples/?omss=places&omsv=8#8.02/-0.084/-0.012) with a grid of points for each place:
Screen Shot 2022-02-14 at 7 57 14 PM

Places in Americana:
Screen Shot 2022-02-14 at 7 57 51 PM

adamfranco:
PS, the prettier check seems to be failing due to an unrelated CDN error with: http://dl-cdn.alpinelinux.org/alpine/v3.9/main/x86_64/APKINDEX.tar.gz

ZeLonewolf:
Awesome!  I think rank only goes up to 10 in OpenMapTiles?

adamfranco:
https://openmaptiles.org/schema/#place says that "the rank field for cities
ranges from 1 to 10 for the most important cities and continues from 10
serially based on the local importance of the city (derived from population
and city class)."

I have no idea where it actually ends.

ZeLonewolf:
Oh yeah that makes sense.  There's code in OpenMapTiles that uses PostGIS's partition function on a per-tile basis.  So it'll rank cities that fall within a tile from 10 and up.

ZeLonewolf:
It looks like the style for motorway and trunk tunnels are swapped:
image

However, the integration (which is what this PR does) looks good to me!

ZeLonewolf (approved):

zekefarwell (approved):
Looks good to me.  For real this time! 😄 I deleted my node_modules folder before running `npm install` just to be sure.

Issue #146, opened 2022-02-07
Refactor project to allow use as a dependency


adamfranco:
This project has started pretty organically and now that #137 is getting close, it may be time to think about refactoring the code for clarity for contributors.

Broadly, there are two ways that this project can be used:
1. As a stand-alone map rendering/demo.
2. As a dependency of another map, a style that another application would want to use as a basemap or to render its data.

The stand-alone map/demo case is currently what we've been focused on, but the inclusion of the style as a dependency of other projects shouldn't be too far off.

Here are some steps that might get us there, though I'm probably missing a few and these should be discussed:
1. Move `package.json` to the root of the repository if there isn't a good reason to bury it in `style/`.
2. Move `index.html` out of the `style/` directory either to the root or to a `demo/` or `site/` subdirectory.
3. Move the config and maplibre calls out of `americana.js` into an `index.js` or `demo.js` that goes with the demo's `index.html`. This should leave `americana.js` exporting a style object with all layers that is as complete as possible and which can be passed added in the demo or other map renderings after having the source tweaked to add the access key or change the source url.
4. Any dependencies for the demo should be included when running `npm i --include=dev`. Any dependencies that are needed only when used as a library should be included by default.
5. If the style needs to distribute built artifacts for consumption by downstream users, these should be added as addition tasks and packaged as releases.

Thoughts? What did I miss?

ZeLonewolf:
See also #104, #70

ZeLonewolf:
Not mentioned is also a `dev` folder, which I envision could be a collection of useful developer files and utilities, such as the Inkscape color palette file envisioned by #133 or perhaps @jleedev 's shield generator.

adamfranco:
If any useful developer tools don't specifically rely on the style or vice-versa then another would be to factor them out into their own repositories and include them as `--save-dev` npm dependencies. This would let them be used in other projects without this whole style coming along for the ride. This might not apply to these tools, but was what I had in mind for OpenMapSamples & OpenMapSamples-Maplibre -- they are very much built to support of this project, but can also be used in other styles and map rendering projects.

ZeLonewolf:
I agree with that approach.  I'm thinking specifically of dev tools that are specific to this style.

zekefarwell:
[...]

PR #158, opened 2022-02-12
Update and revise contributing guide

zekefarwell:
I updated the contributing guide to better match the state of the project.  Removed out of date info, simplified some things, expanded on some other things.  Hopefully it's an overall improvement.

### DIFF
--- a/style/CONTRIBUTE.md
+++ b/style/CONTRIBUTE.md
@@ -11,74 +11,91 @@ The repository is organized as follows:
 - **americana.js** - OpenMapTiles loader with layer ordering
 - **index.html** - Demonstration map HTML page
 
-## Install Pre-requisites
+## Install Node.js
 
-These development tools are required and must be installed manually on your system:
+Any [currently supported version of Node.js][31] should work. The current LTS
+release is recommended. Installation options:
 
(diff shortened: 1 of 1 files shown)


ZeLonewolf (approved):
Looks good to me.  I made some adjustments.  Thanks Zeke for taking this on!

adamfranco:
Sorry for being late to reviewing -- the updated `CONTRIBUTING.md` says to use `npm install`, but the dependencies are currently dev-dependencies, which require `npm install --include=dev`. We could make these normal dependencies if the intention is to not allow usage of Americana code by other projects(see #146), or we should update the docs to add the `--include=dev`. I could go either way.

zekefarwell:
In my experience `npm install` installs both dependencies and devDependencies unless running in a production context.  So adding `--include=dev` shouldn't be necessary when running the project for local development.

From [the NPM Docs](https://docs.npmjs.com/cli/v8/commands/npm-install):
 
> By default, npm install will install all modules listed as dependencies in [package.json](https://docs.npmjs.com/cli/v8/configuring-npm/package-json).

> With the --production flag (or when the NODE_ENV environment variable is set to production), npm will not install modules listed in devDependencies. To install all modules listed in both dependencies and devDependencies when NODE_ENV environment variable is set to production, you can use --production=false.

PR #160, opened 2022-02-14
Add photon search capability

ZeLonewolf:
This PR adds a search box to the demo map.  This basic capability is intended to make navigation more convenient to contributors, in order to avoid manually panning and zooming the map to locations of interest.  If anyone would like to improve upon my amateur attempts, help is welcome :)

Remaining issues:

- [x] Map loses focus after re-centering to search result
- [x] Make search results clickable so it works with mouse/mobile, not just keyboard

image


ZeLonewolf:
PR is ready for review.  Not sure if this is actually the best way to do this or if the search code should go in a subfolder.  Thoughts?

style/index.html
```diff
@@ -45,5 +47,16 @@
...
+    <div id="geocoder-search-panel">
+      <form>
+        <input
+          id="geocoder-search-input"
+          type="text"
+          placeholder="Search"
+          autocomplete="off"
+        />
+      </form>
+      <div id="geocoder-live-results" />
```
jleedev, line 59: This is incorrect for elements which are not self-closing in HTML5.

```suggestion
      <div id="geocoder-live-results"></div>
```

style/search.js
```diff
@@ -0,0 +1,203 @@
...
+    encodeURIComponent(e.target.value) +
+    "&lat=" +
+    position.lat +
+    "&lon=" +
+    position.lng +
+    "&limit=3";
+
+  fetch(searchQuery)
+    .then((response) => response.json())
+    .then((data) => geocoderResponse(data));
```
jleedev, line 163: I'd suggest writing this with async/await, but it also needs updated to cancel all but the most recent request.

style/search.js
```diff
@@ -0,0 +1,203 @@
...
+}
+
+function arrowNavigate(e) {
+  e = e || window.event;
```
jleedev, line 167: What needs this?

style/search.css
```diff
@@ -0,0 +1,53 @@
...
+  padding: 2px;
+}
+.gc-result-item-selected {
+  border: 1px solid lightgray;
+  border-color: lightgray;
+  border-width: 1px 1px 0px 1px;
+  background-color: lightgrey;
+  padding: 2px;
+  cursor: pointer;
+}
```
jleedev, line 36: Suggest having `.gc-result-item` on always and toggling the class `.gc-result-item-selected` only the values to change (background-color).

style/search.js
```diff
@@ -0,0 +1,203 @@
...
+    resultSelectIndex = -1;
+  }
+
+  if (priorIndex >= 0) {
+    liveResults.children.item(priorIndex).className = "gc-result-item";
+  }
+  if (resultSelectIndex >= 0) {
+    liveResults.children.item(resultSelectIndex).className =
+      "gc-result-item-selected";
+  }
```
jleedev, line 202:
```suggestion
  if (priorIndex >= 0) {
    liveResults.children[priorIndex].classList.remove("gc-result-item-selected");
  }
  if (resultSelectIndex >= 0) {
    liveResults.children[resultSelectIndex].classList.add("gc-result-item-selected");
  }
```

jleedev (changes requested):
Bugs:

* When the keyboard opens and closes, the map resizes. Causes it to fly to a squished bbox.
* (Firefox Android) Tapping a result from the list causes the list to hide and reappear.
* Typing quickly causes the results to race, settling on search results for not the current query.
* Up and down arrow keys cause the text cursor to jump to the beginning/end of the search box.

Do you want to implement this as a [control](https://maplibre.org/maplibre-gl-js-docs/api/markers/#icontrol)? Will probably be better than absolutely positioning it around the other controls, it will appear simultaneously with the other controls instead of earlier, it doesn't have to access the `map` as a global variable.

style/search.js
```diff
@@ -0,0 +1,203 @@
...
+    .then((response) => response.json())
+    .then((data) => geocoderResponse(data));
+}
+
+function arrowNavigate(e) {
+  e = e || window.event;
+
+  let priorIndex = resultSelectIndex;
+
+  switch (e.keyCode) {
```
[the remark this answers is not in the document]
ZeLonewolf: Looks like IE is the only outlier that doesn't support the replacement `code`, which has just [0.45% market share](https://gs.statcounter.com/browser-market-share).  I'm pretty comfortable using the newer variant.

jleedev:
> Not sure if this is actually the best way to do this or if the search code should go in a subfolder. Thoughts?

Ideally there would be some distinction between "the demo map" and "the style, which is also available standalone", but the latter really doesn't exist yet as a thing, so there's not a lot of reason to reorganize the code as such.

jleedev:
The implementation as a control is brittle as it depends on markup existing in the document. Would be more proper to create the dom nodes in the onAdd method.

style/search.js
```diff
@@ -0,0 +1,263 @@
...
+
+  switch (e.code) {
+    case "ArrowUp":
+      // up arrow
+      resultSelectIndex--;
+      break;
+    case "ArrowDown":
+      // down arrow
+      resultSelectIndex++;
+      break;
```
jleedev, line 199: Prevent up/down going to beginning/end of text box.

```suggestion
    case "ArrowUp":
      // up arrow
      resultSelectIndex--;
      e.preventDefault();
      break;
    case "ArrowDown":
      // down arrow
      resultSelectIndex++;
      e.preventDefault();
      break;
```

jleedev (approved):
Pretty good!

Some minor issues remain, reiterating here:

- When the keyboard opens and closes, the map resizes. Causes it to fly to a squished bbox.
- (Firefox Android) Tapping a result from the list causes the list to hide and reappear.
- The use of `id` attributes in what should technically be an instanceable control is invalid if you end up with two of them, but those ids are only used for CSS purposes, so nothing should be broken.

I'm willing to ship this in its current state.

Issue #12, opened 2021-05-25
Support for highway exits


ZeLonewolf:
Labelled or icon-indicated highway exits are a common feature on American-style maps, either as a label or often with a white square or diamond to indicate an exit.  Highway exits are tagged with `highway=motorway_junction` in OSM.  However, these objects are not currently available in openmaptiles.

Experiments with using iconified `highway=motorway_link` as a proxy for highway exits have shown that the icon placement is too erratic.

This is related to openmaptiles/openmaptiles#1114 which proposes to add this tag to OpenMapTiles.

If support is not added in OpenMapTiles, a custom vector layer is needed.

ZeLonewolf:
Highway exits have been added to OpenMapTiles, however, it is not yet available on MapTiler.

ZeLonewolf:
Closed via #172

PR #172, opened 2022-02-18
Render highway exits

jleedev:
#12 

At z14 and higher, show motorway_junctions having a nonempty ref with a simple yellow font with black stroke. This is unobtrusive enough to not be confused with a shield.

Screenshots of this implementation:

Screenshot 2022-02-18 08 07 14
Screenshot 2022-02-18 08 06 17
Screenshot 2022-02-18 08 05 18
Screenshot 2022-02-18 08 04 29

---

OMT has exit numbers from z9, but z14 is the lowest zoom level I'm comfortable using symbols on the map without any further style tweaking. At z13, for instance, the I-376/US-22/US-30 shield complex is entirely hidden by exit numbers with the map's default behavior.

At lower zoom levels, I experimented with showing small diamond icons (first commit on this branch), but I think they need more work:

Also, junctions lacking a ref ought to be rendered in a way that's harmonious alongside these refs.

Screenshot 2022-02-17 22 09 15
Screenshot 2022-02-17 22 05 56

### DIFF
--- a/style/americana.js
+++ b/style/americana.js
@@ -15,6 +15,7 @@ import * as lyrPlace from "./layer/place.js";
 import * as lyrRoad from "./layer/road.js";
 import * as lyrRoadLabel from "./layer/road_label.js";
 import * as lyrWater from "./layer/water.js";
+import * as lyrHighwayExit from "./layer/highway_exit.js";
 
 import * as maplibregl from "maplibre-gl";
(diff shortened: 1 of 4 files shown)


ZeLonewolf (approved):
I think the font size looks good, and z14 is definitely a safe zoom to render this at.  We may need to revisit the color once there's more things on the map, but this is a reasonable appearance given the current feature set.

PR #173, opened 2022-02-18
Add building layer

### LINKED ISSUE #170: Render buildings
At high zooms, Americana should render building. Thoughts?

Here are some ideas with plain gray fill:
Screen Shot 2022-02-17 at 4 48 58 PM

Most topo-maps use solid black, but this is a heavy treatment that may be too dark.

adamfranco:
PR to fix #170 .

Extrusion is fixed for now to keep the map mostly flat-looking.

Color (gray 80%) was picked to just barely stand out from background. 
Might need to be darkened slightly if landuse/landcover is much darker 
than 85% as the contrast between the may be too low.

85% opacity alows tunnels and highways to show through just enough for 
continuity.

Light intensity was chosen to make the extrusion subtle, but still 
visible at pedestrian scales.

z13 - Newark, NJ:
z13 Newark

z16 - Lincoln Tunnel:
z16 Lincoln Tunnel

z13 Boston:
z13 Boston

Buildings on parks, to show the color-consistency even with a bit of opacity.
buildings on parks

### DIFF
--- a/style/americana.js
+++ b/style/americana.js
@@ -15,6 +15,7 @@ import * as lyrPlace from "./layer/place.js";
 import * as lyrRoad from "./layer/road.js";
 import * as lyrRoadLabel from "./layer/road_label.js";
 import * as lyrWater from "./layer/water.js";
+import * as lyrBuilding from "./layer/building.js";
 
 import * as maplibregl from "maplibre-gl";
 import "maplibre-gl/maplibre-gl.css";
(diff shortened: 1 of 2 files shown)


ZeLonewolf:
Wow, these look better and better.  I'm definitely in favor of keeping buildings subtle, at least until high zoom.  My main complaint with osm-carto is the building layer causing too much clutter at lower zooms but this is not bad.

ZeLonewolf (approved):
Looks good to me here.

ZeLonewolf (approved):

PR #175, opened 2022-02-19
Building extrusion adjustment

zekefarwell:
The extruded buildings intruduced in #173 were looking a little odd to me so I played with the extrusion properties.  Currently with `fill-extrusion-base: 3` when you tilt the map the buildings look like they have walls and no roofs.

Screen Shot 2022-02-19 at 11 11 52 AM

Changing it to `fill-extrusion-height: 3` gives the result I expect we want.

Screen Shot 2022-02-19 at 11 12 13 AM

### DIFF
--- a/style/layer/building.js
+++ b/style/layer/building.js
@@ -13,7 +13,7 @@ export const building = {
       16,
       `hsl(0, 0%, 80%)`,
     ],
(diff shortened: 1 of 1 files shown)


ZeLonewolf (approved):
Looks great!

PR #174, opened 2022-02-19
Ohio county road shields

1ec5:
Added shields for all the counties in Ohio that signpost their county roads with shields, other than the ones that post black text on a white rectangle, which is already the fallback shield.

[Mahoning](https://zelonewolf.github.io/openstreetmap-americana/#12.04/41.11198/-80.82983)<br>_Mahoning County_

[Medina–Wayne](https://zelonewolf.github.io/openstreetmap-americana/#12.12/40.9974/-82.08908)<br>_Medina County (top) and Wayne County (bottom)_

[Vinton–Athens](https://zelonewolf.github.io/openstreetmap-americana/#13.99/39.25122/-82.26493)<br>Vinton County (left) and Athens County (right)_

[Ashland–Wayne](https://zelonewolf.github.io/openstreetmap-americana/#15.12/40.778914/-82.126555)<br>_Ashland County (left) and Wayne County (right)_

[Scioto](https://zelonewolf.github.io/openstreetmap-americana/#16.05/38.692246/-82.856904)<br>_Scioto County_

[Williams](https://zelonewolf.github.io/openstreetmap-americana/#12.33/41.61831/-84.61263)<br>_Williams County_

Issue #40, opened 2021-07-16
Add render support for admin_level 5 and 7


ZeLonewolf:
#35 added administrative boundaries for 2, 3, 4, 6, and 8, as these are the most commonly used ones.  However, level 5 and 7 are in use also, for examples:

* level 5 is used for New York City
* level 7 is used for Illinois townships

Level 5 is also used for Ireland provinces, but this tagging is questionable.  However, it may be a convenient test area to inspect render support.

ZeLonewolf:
Duplicated by #481, but #481 has the more comprehensive description.

PR #161, opened 2022-02-14
Add American bannered shields

arch0345:
This PR adds all missing bannered shields for the US except for `US:US:Scenic` as I couldn't think of a good abbreviation

Issue #75, opened 2022-01-05
Create key/legend for Highway Shields


akadouri:
Map legends may not be the most popular on web maps, but they offer the opportunity to showcase the rich information held within the vector data being shown.

A floating/sidebar legend could display the currently visible highway shields (via querying the visible roads and determining the shields from there).

Here is an example for querying the rendered features with MapboxGlJS https://docs.mapbox.com/mapbox-gl-js/example/filter-features-within-map-view/

See thread https://osmus.slack.com/archives/C01V02K52UX/p1641410558052000
Related to https://github.com/ZeLonewolf/openstreetmap-americana/issues/76

PR #177, opened 2022-02-20
Ohio township road shields

1ec5:
Hot on the heels of #174, this PR adds shields for [many townships in Ohio](https://commons.wikimedia.org/wiki/Category:Diagrams_of_Ohio_township_route_markers) that are known to signpost their township roads with shields. It includes some nonstandard shields that double as street name signs or township limit signs, but it excludes “blade”-style street name signs, such as sign D3-H6b in the [OMUTCD](https://wiki.openstreetmap.org/wiki/MUTCD/Ohio).

Townships that post the standard black text on white rectangle have also been omitted because that is already the fallback for unrecognized networks. However, if the county posts a sign with the same shape and colors as the surrounding county, it is included in this PR with a “TWP” banner on top. This is common in some counties because the OMUTCD specifies nearly identical shields for county routes (M1-H6a) as for township routes (M1-H6b), but it also happens in some counties that have chosen other background colors.

[Salt Creek–Salt Creek](https://zelonewolf.github.io/openstreetmap-americana/#14.1/40.66604/-81.8229)<br>_Salt Creek townships in Wayne (top) and Holmes (bottom) counties_

[Perry](https://zelonewolf.github.io/openstreetmap-americana/#13/40.85431/-82.13199)<br>_Perry Township in Ashland County (left) and Chester Township in Wayne County (right)_

There may well be more townships that post these signs or their own unique creations. We can easily add them over time as we discover more in Mapillary and KartaView’s superb coverage of rural county and township roads. Not all of these township networks have been mapped with route relations yet, but the `network` pattern is predictable enough that we can add support for them ahead of time as an incentive to mappers.

In general, township roads in Ohio are tagged `highway=unclassified` or `highway=residential`, which requires a new layer for shields on `minor` roads. Fortunately, the townships that post shields tend to be rural and sparse, so clutter isn’t as much of a concern. This layer also exposes quite a bit of [cargo cult tagging](https://en.wikipedia.org/wiki/Cargo_cult_programming) in Ohio, in which mappers [assumed](https://www.openstreetmap.org/changeset/74104851) every road number in TIGER data represents a signposted route, even municipal alley inventory numbers. This layer will therefore help us track the deletion of these spurious route relations.

[Rio Grande](https://zelonewolf.github.io/openstreetmap-americana/#16.05/38.880495/-82.377448)<br>_Rio Grande_

Neighboring townships with different shield designs can be slightly disorienting, since there tend to be similar county road shields in the same area. I think the distribution of shield types will become more intuitive once township boundaries render on the map: #40.

[Washington–Harrison](https://zelonewolf.github.io/openstreetmap-americana/#13.99/40.39749/-83.86577)<br>_Washington (left) and Harrison (right) townships in Logan County_

### DIFF
--- a/style/js/shield_defs.js
+++ b/style/js/shield_defs.js
@@ -422,6 +422,8 @@ export function loadShields(shieldImages) {
   shields["US:OH:Bypass"] = banneredShield(shields["US:OH"], ["BYP"]);
   shields["US:OH:Business"] = banneredShield(shields["US:OH"], ["BUS"]);
 
+  // Ohio county and township roads
+
   ["COL", "JEF", "MAH", "OTT", "SEN", "STA", "SUM", "TUS"].forEach(
     // Yellow on blue pentagon
     (county) => (shields[`US:OH:${county}`] = usMUTCDCountyShield)
@@ -437,10 +439,24 @@ export function loadShields(shieldImages) {
     "WAS",
     "WIL",
     "WYA",
+    "COS:Jackson",
+    "FAI:Greenfield",
+    "JEF:Knox",
+    "LOG:Bokescreek",
+    "LOG:Pleasant",
+    "LOG:Washington",
+    "MED:Sharon",
+    "MRG:York",
+    "MRW:Bennington",
+    "MRW:Chester",
+    "MRW:Franklin",
+    "PER:Bearfield",
+    "PER:Hopewell",
+    "WAY:East_Union",
   ].forEach(
     // White on green rectangle
-    (county) =>
-      (shields[`US:OH:${county}`] = roundedRectShield(
+    (countyOrTownship) =>
+      (shields[`US:OH:${countyOrTownship}`] = roundedRectShield(
         "#006747",
         "white",
         "white",
@@ -449,10 +465,10 @@ export function loadShields(shieldImages) {
         null
       ))
   );
-  ["MED", "NOB"].forEach(
+  ["MED", "NOB", "WAY:Paint", "WAY:Salt_Creek"].forEach(
     // White on blue rectangle
-    (county) =>
-      (shields[`US:OH:${county}`] = roundedRectShield(
+    (countyOrTownship) =>
+      (shields[`US:OH:${countyOrTownship}`] = roundedRectShield(
         "#003f87",
         "white",
         "white",
@@ -461,10 +477,10 @@ export function loadShields(shieldImages) {
         null
       ))
   );
-  ["TRU", "VIN"].forEach(
+  ["TRU", "VIN", "COS:Adams"].forEach(
     // Black on yellow rectangle
-    (county) =>
-      (shields[`US:OH:${county}`] = roundedRectShield(
+    (countyOrTownship) =>
+      (shields[`US:OH:${countyOrTownship}`] = roundedRectShield(
         "#ffcd00",
         "black",
         "black",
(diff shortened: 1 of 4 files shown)


1ec5:
[This build](https://github.com/ZeLonewolf/openstreetmap-americana/runs/5262048375?check_suite_focus=true) crashed:

```
dist/index.html              1.02 KB    7.28s
dist/index.52868c16.js     747.38 KB    6.93s
dist/index.c4fcd4f8.css     68.81 KB    7.22s
Segmentation fault (core dumped)
Error: Process completed with exit code 139.
```

style/js/shield_defs.js
```diff
@@ -458,6 +474,86 @@ export function loadShields(shieldImages) {
...
+    ["LOG", "Miami"],
+    ["LOG", "Monroe"],
+    ["LOG", "Perry"],
+    ["LOG", "Richland"],
+    ["LOG", "Rushcreek"],
+    ["LOG", "Stokes"],
+    ["LOG", "Union"],
+    ["MRW", "Canaan"],
+    ["MRW", "Harmony"],
+    ["MRW", "South_Bloomfield"],
```
1ec5, line 545: South Bloomfield Township posts both white-on-black and black-on-white shields. A few other townships are similarly inconsistent. I went for the path of least resistance for these townships.

ZeLonewolf:
Is the township banner actually needed to disambiguate overlapping numbering schemes?  I'd like to explore the implications that this may have on other areas, for example Texas.  Are we affirming this as a general-purpose cartographic practice for disambiguating similar shield designs within a common geographical area?

1ec5:
> Is the township banner actually needed to disambiguate overlapping numbering schemes?

In some counties, there is a potential for numeric overlap. When part of a state route is decommissioned, that part becomes a county road, often keeping the same number. Likewise, when part of a county route is decommissioned, that part often becomes a township road with the same number. For example:

[](https://zelonewolf.github.io/openstreetmap-americana/#15/40.50032/-83.63607)<br>_[Rushcreek Township Road 40](https://www.openstreetmap.org/relation/3581437) (left) and [Logan County Road 40](https://www.openstreetmap.org/relation/2597802) (right) in Rushcreek Township, Logan County_

[](https://zelonewolf.github.io/openstreetmap-americana/#20/40.4902431/-83.7655076)<br>_[Richland Township Road 49](https://www.openstreetmap.org/relation/3581867) (top) and [Logan County Road 49](https://www.openstreetmap.org/relation/2597751) (bottom) in Richland Township, Logan County_

(I’m not sure why this particular shield has the text in the wrong place. It reminds me of the corruption in https://github.com/ZeLonewolf/openstreetmap-americana/pull/161#discussion_r810437534.)

Even in counties where there isn’t overlap, omitting any distinction between county and township roads would defeat the purpose of showing shields. The shields would convey less information than simple road name labels.

> I'd like to explore the implications that this may have on other areas, for example Texas. Are we affirming this as a general-purpose cartographic practice for disambiguating similar shield designs within a common geographical area?

I went with a banner mostly out of expediency. A more faithful depiction would have “TWP” inside the shield, or “TR” to the left of the number (or “TH”/“T-” depending on the county), but that would’ve required special artwork or canvas drawing code. Plus, there simply isn’t room for any more text inside the shape of Ashland County.

County roads are the implied default when there isn’t a banner above a white rectangle. Unfortunately, the “TWP” gives more prominence to the township roads than the county roads. I considered also adding “CO” above the county route shields, but then “CO” would appear unnecessarily in some townships that don’t need that disambiguation.

I think Texas could be handled the same way, since it uses a white rectangle for multiple networks. Incorporating “FM” or “BW” or “NASA” into the shield would piggyback on Texans’ familiarity with freeway guide signs, but placing it above the shield wouldn’t be very jarring at all in the context of this style’s other shields. State highways could be implied in the absence of a banner, or we could add “TX” or “SH” above.

zekefarwell:
It looks like in both of these cases the township road is a continuation of the county road of the same number, not diverging routes with the same number (as happens with business routes).  So it doesn't seem to me that the differentiation is needed.  A very detail oriented person may be interested that they are now on the township maintained portion of route 40 vs the county maintained portion of route 40, but a map that just shows the whole.thing as route 40 isn't ambiguous.  I'd vote for no distinction here as its really getting into the weeds.

ZeLonewolf:
For the sake of argument, is this case of differentiation really that different from, say, when RI-138 crosses the state line to become CT-138?

image

1ec5:
> For the sake of argument, is this case of differentiation really that different from, say, when RI-138 crosses the state line to become CT-138?

Yes, because there are no Connecticut routes in Rhode Island or vice versa.

> It looks like in both of these cases the township road is a continuation of the county road of the same number, not diverging routes with the same number (as happens with business routes). So it doesn't seem to me that the differentiation is needed. A very detail oriented person may be interested that they are now on the township maintained portion of route 40 vs the county maintained portion of route 40, but a map that just shows the whole.thing as route 40 isn't ambiguous. I'd vote for no distinction here as it's really getting into the weeds.

Eliding the distinction between county and township roads would be a tradeoff of comprehensibility versus functionality. I probably could dig up examples of numeric conflict that isn’t contiguous, but as with anything about local routes in Ohio, it would vary from one place to another, so I’m not sure that’s a reliable deciding factor. I think a more important consideration is that we already are making visual distinctions for some townships but not others.

In the screenshot below, Washington and Harrison townships are both in Logan County. Washington Township Road 33 is visually distinct from Logan County Road 13, but Harrison Township Road 31 is not. That may not be technically confusing in a turn-by-turn navigation context, but for someone browsing the map, the rules for when something is visually distinct aren’t as obvious.

[Washington–Harrison](https://zelonewolf.github.io/openstreetmap-americana/#13.99/40.39749/-83.86577)

Eventually, if a map legend dynamically defines the shields in the current viewport, as proposed in #75, it would be confusing for multiple white rectangles to represent multiple distinct systems in the same geographic area.

That said, the road name labels still appear for minor roads if you zoom in further, so there’s still a potential for disambiguation, just not a lot of clarity at zoom levels where multiple townships are visible.

ZeLonewolf:
I'm guessing that when we get around to Texas farm-to-market roads, we're gonna stick a "FARM" on top of a ordinary rectangle.  I could be wrong, maybe there's a way to distort the Texas shape, but I'm thinking that's how it's gonna go.  And if so, it's the same situation -- a route network that needs to be differentiated from surrounding route network.  So if we do that in Texas, it's reasonable to do the same thing in Ohio.  One thing that we should decide on, is how do you determine which of multiple route networks gets the banner and which ones are unadorned?  Is it just the least significant one that gets the banner?  Or the lowest admin_level owner of it?  Or just whichever one looks more cartographically pleasing?  It would probably be useful to document our rationale and method for "using banner to differentiate multiple route networks in a common geographical area with common symbology".

image

1ec5:
The Texas-shaped FM shield is tricky to work with, but TxDOT already has a well-established design of white rectangles for all the statewide networks, including FM roads. Like the Ohio county and township road shields, these rectangles disambiguate the network using a text legend, so the banner treatment would be consistent.

[Loop 340 / SH 6 / FM 3051](https://commons.wikimedia.org/wiki/File:Texas_guide_sign_SL_SH_FM.jpg)

If any network goes without a banner, it should be the most important network – county routes in Ohio, state highways in Texas. Maybe there should be a way to shrink the shield a little so that the bannered shields don’t appear more pronounced than the more important network’s shields.

ZeLonewolf (approved):
Given #186, I agree with this approach.  The map nerds at the Ohio DOT, who exist only to confuse and frustrate the whims of open source shield renderers, have created a situation in which two otherwise identical shield symbiologies need to be differentiated in the map.  I agree with giving the "township" shields the banner given that it is less significant than the county routes.  When the Ohio DOT abolishes ye olde fashioned methode of assigning shields to their routes, we can all raise a glass in collective joy and simultaneous sadness at the departure of a true corner of Americana.

It is the way.

1ec5:
The Graphic Design Section of the Salem Township Public Works Department is available for a nominal consulting fee.

PR #179, opened 2022-02-20
Generate taginfo project file

### LINKED ISSUE #178: Publish taginfo project file
Americana should publish a [taginfo project file](https://wiki.openstreetmap.org/wiki/Taginfo/Projects) listing the tags that are supported in some way. For example, the file can list each of the `network` values we support, just like [the original osm-shields documentation](http://web.archive.org/web/20210515212718/https://elrond.aperiodic.net/shields/supported.html) but more machine-readable.

Once taginfo is configured to point to Americana’s project file, it’ll list Americana under the Projects tab of any key or tag value, as well as in an index of projects. Publicizing our support for individual tags through taginfo will incentivize a mapper to finally get around to tagging something they had been putting off due to an apparent lack of software support.

Americana is evolving rapidly enough that a project file could quickly get outdated. We should document the need to update the project file in the CONTRIBUTING.md guide. Better yet, we can automate some parts of the project file as a build step.

1ec5:
Added a script that generates a taginfo project file and saves it to the dist/ folder. The script is hooked up to `npm run build`, but you can also run it independently using `npm run taginfo`.

The file is based on a boilerplate JSON file. Contributors can feel free to manually add anything notable there. However, the proper place for most tags would be OpenMapTiles’ project file: openmaptiles/openmaptiles#1269. The final generated file primarily lists the `network` values Americana supports, because OpenMapTiles just forwards the raw `network` values to the style for style-specific processing. Here’s what the `network` entries look like:

```json
{
  {
    "key": "network",
    "value": "US:US",
    "object_types": [
      "relation"
    ],
    "description": "Roads carrying routes in this network are marked by shields.",
    "icon_url": "https://raw.githubusercontent.com/ZeLonewolf/openstreetmap-americana/main/style/icons/shield40_us_us_2.svg"
  },
  {
    "key": "network",
    "value": "US:US:Truck",
    "object_types": [
      "relation"
    ],
    "description": "Roads carrying routes in this network are marked by shields modified by TRK banners.",
    "icon_url": "https://raw.githubusercontent.com/ZeLonewolf/openstreetmap-americana/main/style/icons/shield40_us_us_2.svg"
  }
}
```

That will look something like this in taginfo:

Tag | Objects | How this key/tag is used in the project
----|:--:|----
[network](https://taginfo.openstreetmap.org/keys/network)=[US:US](https://taginfo.openstreetmap.org/tags/network=US:US) | yes |  Roads carrying routes in this network are marked by shields.
[network](https://taginfo.openstreetmap.org/keys/network)=[US:US:Truck](https://taginfo.openstreetmap.org/tags/network=US:US:Truck) | yes |  Roads carrying routes in this network are marked by shields modified by TRK banners.

Fixes #178. Once this PR is merged, the [taginfo project list](https://github.com/taginfo/taginfo-projects/blob/master/project_list.txt) needs to be updated.

### DIFF
--- a/style/js/screen_gfx.js
+++ b/style/js/screen_gfx.js
@@ -20,5 +20,5 @@ export function getGfxContext(bounds) {
  * So this function will always return 1 or 2 as appropriate based on DPR.
  */
 export function getPixelRatio() {
-  return window.devicePixelRatio > 1 ? 2 : 1;
+  return (typeof window !== "undefined" && window.devicePixelRatio) > 1 ? 2 : 1;
 }
--- a/style/scripts/taginfo.js
+++ b/style/scripts/taginfo.js
@@ -1,0 +1,59 @@
+"use strict";
+
+import * as fs from "fs";
+import * as ShieldDef from "../js/shield_defs.js";
+
+/**
+ * Adds documentation about network=* tags to a project description object, modifying it in place.
+ *
+ * @param {*} project - The project description object to modify.
+ * @param {*} sprites - Sprite metadata object parsed from a JSON spritesheet.
+ */
+function addNetworkTags(project, sprites) {
+  // Inject a map of each sprite ID to an absolute image URL instead of the usual sprite metadata.
+  let shieldImageURLs = Object.fromEntries(
+    Object.keys(sprites).map((sprite) => [
+      sprite,
+      `https://raw.githubusercontent.com/ZeLonewolf/openstreetmap-americana/main/style/icons/${sprite}.svg`,
+    ])
+  );
+  let shields = ShieldDef.loadShields(shieldImageURLs);
+
+  // Convert each shield's rendering metadata to an entry that taginfo understands.
+  let tags = Object.entries(shields).map((entry) => {
+    let network = entry[0],
+      definition = entry[1];
+
+    let description = `Roads carrying routes in this network are marked by shields`;
+    if (definition.modifiers && definition.modifiers.length > 0) {
+      description += ` modified by ${definition.modifiers.join(", ")} banners`;
+    }
+    description += ".";
+
+    let icon = definition.backgroundImage;
+    if (Array.isArray(icon)) {
+      icon = icon[0];
+    }
+
+    return {
+      key: "network",
+      value: network,
+      object_types: ["relation"],
+      description: description,
+      icon_url: icon,
+    };
+  });
+  project.tags.push(...tags);
+}
+
+let project = JSON.parse(
+  fs.readFileSync(`${process.cwd()}/scripts/taginfo_template.json`)
+);
+let sprites = JSON.parse(
+  fs.readFileSync(`${process.cwd()}/sprites/sprite.json`)
+);
+addNetworkTags(project, sprites);
+fs.writeFileSync(
+  `${process.cwd()}/dist/taginfo.json`,
+  JSON.stringify(project, null, 2)
+);
--- a/style/scripts/taginfo.js
+++ b/style/scripts/taginfo.js
@@ -50,7 +50,7 @@ let project = JSON.parse(
   fs.readFileSync(`${process.cwd()}/scripts/taginfo_template.json`)
 );
 let sprites = JSON.parse(
-  fs.readFileSync(`${process.cwd()}/sprites/sprite.json`)
+  fs.readFileSync(`${process.cwd()}/dist/sprites/sprite.json`)
 );
 addNetworkTags(project, sprites);
 fs.writeFileSync(
--- a/style/scripts/taginfo_template.json
+++ b/style/scripts/taginfo_template.json
@@ -1,0 +1,19 @@
+{
+  "data_format": 1,
+  "project": {
+    "name": "OpenStreetMap Americana",
+    "description": "A quintessentially American map style.",
+    "project_url": "https://github.com/ZeLonewolf/openstreetmap-americana/",
+    "doc_url": "https://wiki.openstreetmap.org/wiki/OpenStreetMap_Americana",
+    "icon_url": "https://raw.githubusercontent.com/ZeLonewolf/openstreetmap-americana/main/doc-img/osm-americana-logo.png",
+    "contact_name": "",
+    "contact_email": ""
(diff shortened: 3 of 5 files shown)


ZeLonewolf:
Background information:
https://wiki.openstreetmap.org/wiki/Taginfo/Projects

style/scripts/taginfo_template.json
```diff
@@ -0,0 +1,19 @@
+{
+  "data_format": 1,
+  "project": {
+    "name": "OpenStreetMap Americana",
+    "description": "A quintessentially American map style.",
+    "project_url": "https://github.com/ZeLonewolf/openstreetmap-americana/",
+    "doc_url": "https://wiki.openstreetmap.org/wiki/OpenStreetMap_Americana",
+    "icon_url": "https://raw.githubusercontent.com/ZeLonewolf/openstreetmap-americana/main/doc-img/osm-americana-logo.png",
+    "contact_name": "",
+    "contact_email": ""
```
ZeLonewolf, line 10: Based on https://wiki.openstreetmap.org/wiki/Taginfo/Projects these are required fields 🙄 so I guess that's me for now.
```suggestion
    "contact_name": "Brian Sperlongano",
    "contact_email": "zelonewolf@gmail.com"
```

style/scripts/taginfo.js
```diff
@@ -0,0 +1,59 @@
...
+    let network = entry[0],
+      definition = entry[1];
+
+    let description = `Roads carrying routes in this network are marked by shields`;
+    if (definition.modifiers && definition.modifiers.length > 0) {
+      description += ` modified by ${definition.modifiers.join(", ")} banners`;
+    }
+    description += ".";
+
+    let icon = definition.backgroundImage;
```
1ec5, line 33: For programmatically generated background images, such as calls to `circleShield()` and `roundedRectShield()`, the `icon_url` property ends up getting omitted from the JSON file because a function can’t be serialized to JSON. Ideally, we would save off the icon in a separate image in addition to the spritesheet so that we can link to it. But I think that can be tail work, because the icon is mostly decorative. Mappers don’t really need to know what the shield will look like, just that it’ll be present.

style/scripts/taginfo_template.json
```diff
@@ -0,0 +1,19 @@
+{
+  "data_format": 1,
+  "project": {
+    "name": "OpenStreetMap Americana",
+    "description": "A quintessentially American map style.",
+    "project_url": "https://github.com/ZeLonewolf/openstreetmap-americana/",
+    "doc_url": "https://wiki.openstreetmap.org/wiki/OpenStreetMap_Americana",
+    "icon_url": "https://raw.githubusercontent.com/ZeLonewolf/openstreetmap-americana/main/doc-img/osm-americana-logo.png",
```
ZeLonewolf, line 8: The taginfo projects page [says](https://wiki.openstreetmap.org/wiki/Taginfo/Projects):
> project logo, should work in 16x16 pixels on white and light gray backgrounds

Do we need to provide a 16x16 version or does taginfo shrink the image down to size?  We do also have a favico.ico that can potentially be used.

adamfranco:
The networks totally make sense to generate programmatically, but other tags get translated through OMT and may need to be manually itemized. I'm thinking about `surface=*` and `expresssway=*`.

style/scripts/taginfo_template.json
1ec5: It seems to generate its own copy of the image at 16×16. I’m not sure if it supports .ico format. It only seems to copy the project image, not the individual tag icons. It embeds those directly, using CSS to control the size.

1ec5:
openmaptiles/openmaptiles#1269 tracks a taginfo file upstream that would be more maintainable in terms of saying which OSM tags are translated into which layers and properties. I wouldn’t necessarily rule out mentioning some notable tags here too, if this style uses them in an unusual manner that might not be apparent from a lower-level description. The idea is for lay mappers to get a sense of how the tag is used in practice, so we might need to help with connecting the dots. At the same time, there’s no need to be comprehensive if it would create a maintenance burden.

style/js/screen_gfx.js
```diff
@@ -20,5 +20,5 @@ export function getGfxContext(bounds) {
  * So this function will always return 1 or 2 as appropriate based on DPR.
  */
 export function getPixelRatio() {
-  return window.devicePixelRatio > 1 ? 2 : 1;
+  return (typeof window !== "undefined" && window.devicePixelRatio) > 1 ? 2 : 1;
```
ZeLonewolf, line 23: This change seems unrelated?

ZeLonewolf (approved):
Seems to work for me and looks essentially correct.  Please confirm the getPixelRatio() change is supposed to be there.

style/js/screen_gfx.js
1ec5: Without this change, the script throws an error:

```
file:///path/to/openstreetmap-americana/style/js/screen_gfx.js:23
  return window.devicePixelRatio > 1 ? 2 : 1;
  ^

ReferenceError: window is not defined
    at Module.getPixelRatio (file:///path/to/openstreetmap-americana/style/js/screen_gfx.js:23:3)
    at file:///path/to/openstreetmap-americana/style/js/shield_canvas_draw.js:10:24
    at ModuleJob.run (node:internal/modules/esm/module_job:197:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:337:24)
    at async loadESM (node:internal/process/esm_loader:88:5)
    at async handleMainPromise (node:internal/modules/run_main:61:12)
```

It’s because of this function call being exported as part of a module that gets imported:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/21916204118f3da81e2cafe51185b53bd17d33f6/style/js/shield_canvas_draw.js#L10

ZeLonewolf:
Looks good to me, thanks!

1ec5:
> Once this PR is merged, the [taginfo project list](https://github.com/taginfo/taginfo-projects/blob/master/project_list.txt) needs to be updated.

taginfo/taginfo-projects#127

PR #180, opened 2022-02-20
Disable network=motorway rendering

ZeLonewolf:
This PR temporarily disables the rendering of network-specific highway shields for `network=motorway`, which is currently tagged on Swiss highway networks following discussion on #149.  This is intended to allow for the project to take a "wait and see" approach to determine whether this generic value is adopted by other route relation data consumers as the tagging specific to Switzerland.  In the meantime, Swiss highway networks will continue to be rendered with generic white rectangles.

### DIFF
--- a/style/js/shield_defs.js
+++ b/style/js/shield_defs.js
@@ -619,17 +619,24 @@ export function loadShields(shieldImages) {
     },
   };
 
+  /**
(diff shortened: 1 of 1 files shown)


zekefarwell (approved):
Seems entirely appropriate

PR #181, opened 2022-02-21
Make tile server, source, and attribution configurable

### LINKED ISSUE #143: Make tile server attribution configurable
Currently, the MapTiler attribution is hard-coded in the map.  However, this should only appear when using a MapTiler tile server.  Thus, we should add an additional parameter to config.js which allows the attribution to be specified on a per-source basis.

ZeLonewolf:
Closes #143

This PR moves the configuration of the vector tile server and attribution to be displayed on the demo into configuration files.  Additionally, this adds a configuration file with the current address of the planetiler vector tile source demo.  This setup allows for the MapTiler attribution and logo to be shown only when using a MapTiler service.  This is an initial step towards supporting a self-hosted tile server.


jleedev:
FYI, there's some chatter about the logo not displaying by default https://github.com/maplibre/maplibre-gl-js/pull/786

style/configs/config.maptiler.js
```diff
@@ -0,0 +1,38 @@
...
+  width="149"
+  height="149"
+  src="https://github.blog/wp-content/uploads/2008/12/forkme_right_green_007200.png?resize=149%2C149"
+  class="attachment-full size-full"
+  alt="Fork me on GitHub"
+  data-recalc-dims="1"
+  style="position: absolute; top: 0; right: 0; border: 0; z-index: 100"
+/></a>`;
+const ATTRIBUTION_TEXT =
+  '<a href="https://www.maptiler.com/copyright/" target="_blank">&copy; MapTiler</a>';
```
zekefarwell, line 32: All this html feels overloaded for a config file.  Why not just hava boolean variable call `MAPTILER_ATTRIBUTION` or something?  Then all this html could live somewhere else inside an if statement that checks that variable.

ZeLonewolf: Oh, this is only supposed to have the maptiler stuff, not the Github stuff.

ZeLonewolf: 6bcd0c should have the intended version

zekefarwell: Still feels weird to have html in a config file, but that is definitely more manageable.

style/configs/config.planetiler-sf.js
```diff
@@ -0,0 +1,12 @@
+"use strict";
+
+/*
+Planetiler tile server, hosted by StreetFerret
+*/
+const TILESOURCE = [
+  "http://vtile.streetferret.com:8080/data/v3/{z}/{x}/{y}.pbf",
+];
```
zekefarwell, line 8: Why a direct pbf url here?  I expected `http://vtile.streetferret.com:8080/data/v3.json`

ZeLonewolf: Actually that one can be.  I've having trouble getting the tileserver to respond on the domain name it came from.  The bigger issue is the HTTPS proxied version, e.g. take a look at the pbf url present in: https://6ug7hetxl9.execute-api.us-east-2.amazonaws.com/data/v3.json

By supporting both a direct tile source as well as a JSON config, you have the possibility of specifying either.

ZeLonewolf: See f882610 which is part of my motivation for this PR - getting a working https-backed tile source that we can deploy.

zekefarwell: Does the direct tile source behave differently than the JSON config?  Is there a need for it?

ZeLonewolf: As far as I can tell, the only thing the JSON config provides of value is the pbf direct link, the OpenMapTiles attribution (which I have just hard-coded in by default now), and a bunch of layer definition stuff that we don't use.

jleedev: > Actually that one can be. I've having trouble getting the tileserver to respond on the domain name it came from.

Are you passing `--public_url https://6ug7hetxl9.execute-api.us-east-2.amazonaws.com/` to tileserver?

It seems to respect Host but not X-Forwarded-Host, so maybe you can ask your Lambda to send a custom host header instead.

ZeLonewolf: Oh yeah now we're cooking with gas.

zekefarwell (approved):
This all looks good to me.  I was thinking it was more confusing having multiple config files to choose from, but I definitely see how that makes things simpler for changing which config file gets used for deployment.

Looks like the deploy is still set to use the maptiler config.  Did you want to swap over to the AWS server, or hold off on that for a bit longer?

ZeLonewolf:
> Looks like the deploy is still set to use the maptiler config. Did you want to swap over to the AWS server, or hold off on that for a bit longer?

Let's do that in a separate step once this is definitely all working.

PR #182, opened 2022-02-21
Remove extra padding from attribution logo

ZeLonewolf:
This PR removes the extra vertical padding that currently appears under the attribution logo, due to a repeated CSS id.


adamfranco (approved):
I'm not seeing much difference, but nothing looks broken. 👍

PR #183, opened 2022-02-21
Document more tags for taginfo

1ec5:
Added several more tags to the taginfo project file that this style does something notable with.

### DIFF
--- a/style/scripts/taginfo_template.json
+++ b/style/scripts/taginfo_template.json
@@ -27,6 +27,26 @@
       "icon_url": "https://raw.githubusercontent.com/ZeLonewolf/openstreetmap-americana/main/style/icons/star_state_capital.svg"
     },
     {
+      "key": "claimed_by",
+      "object_types": ["way", "relation"],
+      "description": "Hides the boundary.",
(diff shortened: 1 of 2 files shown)


style/scripts/taginfo_template.json
```diff
@@ -10,10 +10,59 @@
...
       "object_types": ["relation"],
-      "description": "Populates shields that mark roads carrying the route relation."
+      "description": "Populates shields that mark roads carrying the route relation.",
+      "doc_url": "https://openmaptiles.org/schema/#route_1"
+    },
+    {
+      "key": "surface",
+      "object_types": ["way"],
+      "description": "Unpaved roads have an alternating dash pattern.",
+      "doc_url": "https://openmaptiles.org/schema/#surface"
```
1ec5, line 65: This documentation lists several specific values of `surface` that are considered “unpaved”. I refrained from hard-coding them here because the list could change out from under us.

ZeLonewolf (approved):
Thanks for improving the project's documentation!

ZeLonewolf (approved):

Issue #169, opened 2022-02-17
Distinguish highway bridges from non-bridges


adamfranco:
While extra casings for bridges add visual clutter, it was noted by Minh in Slack that without them it can be difficult or impossible to disambiguate between an intersection and a flyover.

If we'd like to keep bridges subtle, then maybe just a slight extra width to the casing could be added. We could potentially layer two levels of casing (widest dark bottom, then a lighter slightly narrower, both under the normal road casing) to visually offset an outer bridge line from the normal casing. This technique could be especially useful for the dark-stroke style at lower zooms. 

Maybe something along these lines:
![Screen Shot 2022-02-17 at 4 28 21 PM](https://user-images.githubusercontent.com/25242/154574542-8791dc76-2870-402e-832f-6625f070aa62.png)

This sort of treatment could work with both case+fill style as well as stroked style and shouldn't conflict with expressway rendering.

At some point features will become too small to make differentiating bridges useful, maybe somewhere around z12

1ec5:
Many state-scale road maps show “guardrails” similar to your mockup above, though the casing is thinner and the gap larger:

![Clermont County Engineer (Engels) 2004](https://user-images.githubusercontent.com/1231218/154591354-9ca3e067-6856-4c9b-9f53-3e9bb445f715.png)
<img src="https://user-images.githubusercontent.com/1231218/154589344-f4f9eb45-d952-490f-965e-e0ae0fd571fa.png" width="250" alt="American Map NYC Eastern Long Island">
<img src="https://user-images.githubusercontent.com/1231218/154589577-3fabf0ed-e6ea-4547-a18f-bc9fda7583cb.png" width="250" alt="Alabama DOT 1987–88 strip map">

Mostly only show this detail when there’s a grade separation with no connecting ramps; otherwise, the interchange is generalized as a square or diamond, which is understood to be a cloverleaf, trumpet, etc.:

<img src="https://user-images.githubusercontent.com/1231218/154588805-e688794c-a1a3-4393-a546-6137d80d0ab5.png" width="250" alt="OklaDOT">
<img src="https://user-images.githubusercontent.com/1231218/154589854-fb25b643-6421-41ac-a619-02ea0e23db38.png" width="250" alt="Iowa DOT 2010">
<img src="https://user-images.githubusercontent.com/1231218/154590001-2845dd02-2714-4574-90f1-e4128842142d.png" width="250" alt="West Virignia DOT">
[...]

Issue #84, opened 2022-01-20
Improve relative highway classification prominence at z13+


ZeLonewolf:
#62 provided an initial implementation of the road system through tertiary, and #82 provided an initial implementation of expressway rendering.    The high-zoom road network needs to be holistically considered to ensure that it's appropriately conveying relative prominence at these zoom levels.

Additional refinement is needed to ensure that:

1. More prominent roads are rendered more prominently.  Currently, the progressive switch from stroked to cased roads has the side effect of making the cased roads look less important that the thick stroked roads.  For example:

![image](https://user-images.githubusercontent.com/3254090/150269078-d4e86565-0f95-4368-b8a0-b75d5672b409.png)

2. Expressways are visually understood to be more prominent than non-expressway versions of the same highway class.

![image](https://user-images.githubusercontent.com/3254090/150269414-4465fc75-e9aa-4e25-937c-94f7ec736e36.png)

3. Expressway styling is not confused with tunnel rendering.  For example:

![image](https://user-images.githubusercontent.com/3254090/150269285-1a26cf68-405e-4077-8df4-213fae450c4e.png)

adamfranco:
Currently `primary` switches from black stroke to cased-style at z14 and all lower classes at z16.

Currently the [fully-stroked look at z13](https://zelonewolf.github.io/openstreetmap-americana/?omss=highway-attributes&omsv=13#13/0/0) makes relative importance intelligible based on width:
![z13](https://user-images.githubusercontent.com/25242/154567774-6c132ce9-136c-45ac-aa07-ca6cc73c8d25.png)

Similarly, the [fully-case/filled look at z16](https://zelonewolf.github.io/openstreetmap-americana/?omss=highway-attributes&omsv=16#16/0/0) makes relative importance understandable.
![z16](https://user-images.githubusercontent.com/25242/154568064-a68e42e0-eb11-4998-b827-ce9158f4e25a.png)

The problem is in between those levels where there are mixed styles like [z15](https://zelonewolf.github.io/openstreetmap-americana/?omss=highway-attributes&omsv=15#15/0/0):
![z15](https://user-images.githubusercontent.com/25242/154568231-e4f6d5cb-4d9a-4802-9796-b230ded0d402.png)

**Shall we switch all lower classes to cased/fill-style at the same zoom?** Doing so would use width to maintain the relative prominence without struggling with the different visual weight between heavy black and light cased/fill styles.

[...]

PR #184, opened 2022-02-21
Road class and bridge/junction disambiguation

adamfranco:
This PR disambiguates highway (#84) levels by always showing primary and lower highways with either dark or light treatments rather than a mix of both. The current dark/light switch happens at z15, but it could be changed earlier or later.

Additionally, the dark treatment from highways changes from from using the fill layer to draw "strokes" (without a casing layer) to always drawing both the casing and fill layers, but using a light casing with dark fill and a dark casing with light fill at the different zooms of primary and down. This partially or fully addresses #169 by allowing the the dark treatment's lighter casing to disambiguate between junctions and bridges/overpasses.

LA ([localhost](http://localhost:1776/#14.86/34.05333/-118.25987), [demo](https://zelonewolf.github.io/openstreetmap-americana/#14.86/34.05333/-118.25987))
Screen Shot 2022-02-21 at 3 48 00 PM

We may still wish to add additional rails as suggested in #169 , but the dark-style highway rendering used at the lower zooms is now not as problematic.

Another potential direction opened up by this change is using the dark style for all zooms and not switching to light at all, but that can be left for a future PR.


adamfranco:
LA before:
la before

LA after:
la after

1ec5 (approved):
I think the hairline casing works really well for distinguishing roads from each other. We’ll have to keep an eye on how it interacts with other backgrounds for landuse, but for now I’m not even worried about how it looks over water, because it’s thin enough to look like a normal road casing without appearing to be the color of land.

ZeLonewolf (approved):
Looks good to me.  Styling over water seems OK as an initial step.  I'm glad to have everything switch at a common zoom level, this is definitely an improvement.

image

PR #186, opened 2022-02-22
Add special case policy for ref parsing and banner employment

ZeLonewolf:
This PR proposes a written policy for determining when this style should support special cases, including banners.  By establishing consensus on when special cases are appropriate and documenting our rationale, we make it easier to adjudicate PRs proposing special case code, including the currently open #177 and #205.

I consider this a starting point and encourage the other maintainers to weigh in on crafting this policy so that we can have a decision-making framework for special cases now and in the future.


style/CONTRIBUTE.md
```diff
@@ -248,3 +248,25 @@ Additionally, **`refsByWayName`** is an object mapping way names to text that ca
...
+
+<img src="https://upload.wikimedia.org/wikipedia/commons/8/83/Business-alternate-truck_plate.svg" width=40 /><br/><img src="https://upload.wikimedia.org/wikipedia/commons/e/ec/US_30.svg" width=40 />
+
+Banners should be specified in the following cases:
+
+- When a route represents a variant of a main route with which it shares a common shield design. The banner ensures that the variant route information, which is an important component of the route, is visually displayed.
+- When two or more routes from different networks share a common symbology in the map within a common geographical area. Shields which are very similar may be drawn using common graphics for simplicity and readability, for example, when the networks differ only by a difference in text. In these cases, the most significant network should be drawn with no banner, and each of the less significant networks should be drawn with a banner.
+
+In all cases, banner text should be no more than **4** characters in length.
```
1ec5, line 263: Is this limit due to space constraints or aesthetic concerns? So far we’ve gotten lucky about finding reasonably intuitive abbreviations for standard banners, but we might run into trouble with “SCENIC”.

style/CONTRIBUTE.md
```diff
@@ -248,3 +248,25 @@ Additionally, **`refsByWayName`** is an object mapping way names to text that ca
 - Each route would be recognizable by an initialism, even though it is not signposted.
```
1ec5, line 248: This list is redundant to the explanation about Kentucky parkways below. Should we keep it?

style/CONTRIBUTE.md
```diff
@@ -248,3 +248,25 @@ Additionally, **`refsByWayName`** is an object mapping way names to text that ca
...
+
+In all cases, banner text should be no more than **4** characters in length.
+
+### Special Cases
+
+This style strives to draw representative highway shields wherever they are tagged on road route relations consistently with international norms. This style operates on the expectation that the `network` value on a route relation corresponds to the shield design that will be drawn, and the `ref` value will contain the text which is drawn on the shield. In order to give appropriate mapper feedback, this style will add support for special cases only when the complexity of the route network and shield styling cannot be adequately expressed via `network` and `ref` alone. These special cases should be exceptionally rare and documented in the list below. PRs to add special case code should also add an entry below justifying its inclusion.
```
1ec5, line 267: Each of these special cases is implemented in a different way in a different part of the codebase. Are they adequately documented in place, or should we add documentation comments pointing to this list?

1ec5 (approved):
This writeup includes some expectations about tagging that should probably make their way back to the OSM Wiki at some point, if only as an observation of data consumers’ expectations.

style/CONTRIBUTE.md
ZeLonewolf: 4 characters is the limit after which the text has to shrink in the vertical direction to fit within 20px.  If we put in a longer string, the code would happily shrink it to fit.

adamfranco (commented):
I'm generally staying out of the shields work as a number of you seem to have it well covered. That said, these suggestions seem like a reasonable starting point. 

One question that comes up for me is around "banners" and international practices. Are these applicable at all to other countries? I guess that a banner without a background image or just a plain lozenge could be pretty universally applied even if it doesn't reflect actual sign formatting.

style/CONTRIBUTE.md
ZeLonewolf: I didn't fully answer the question - this limit is due to aesthetic concerns, but we can certainly explore what 5-character banners look like to decide if this limit can/should be expanded.

ZeLonewolf:
> Are these applicable at all to other countries? I guess that a banner without a background image or just a plain lozenge could be pretty universally applied even if it doesn't reflect actual sign formatting.

I added a note about the generic shield being displayed.  I don't think we know yet where, when, or if these banners are appropriate in other countries, but I'm hoping we can use these principles to make that assessment when we come to it, and if needed, we'll just amend the list of rationale.

1ec5 (approved):

ZeLonewolf:
#205 explores (among other things) the case of TOLL banners on highway shields for road networks which are not exclusively motorway (and thus would not get the separate coloring for toll motorway roads discussed in #206).  In order to guide our decision-making, we should address this case here.  Possible rationale we might consider regarding banners:

* To indicate toll road networks on non-motorway roads, because such roads are not colored differently from non-toll roads.
* When a short text of up to 4 characters is a significant stylistic element of a shield that can't reasonably be incorporated into the main shield graphic for aesthetic reasons.

1ec5:
> When a short text of up to 4 characters is a significant stylistic element of a shield that can't reasonably be incorporated into the main shield graphic for aesthetic reasons.

A good test case of this criterion is the standard Florida state toll road shield:

Florida Toll

The yellow banner looks similar to the yellow banner placed atop toll Interstates, which would ordinarily call for a “TOLL” modifier without a background. However, it’s integrated into the shield, which might be as recognizable without something yellow atop it.

ZeLonewolf:
I've added the bullets in https://github.com/ZeLonewolf/openstreetmap-americana/pull/186#issuecomment-1053509780 to this PR.

@Mashin6, @zekefarwell do you have input on the guidance in this PR?  If you don't feel strongly about it, that's okay too.

zekefarwell (approved):
This looks good to me.  It may need to be revised as we encounter new situations but I think it lays out the basic guidelines well.

ZeLonewolf:
Given the lack of objections, let's move forward with this.

style/CONTRIBUTE.md
1ec5: > we might run into trouble with “SCENIC”

#545 uses “SCEN”.

PR #187, opened 2022-02-22
Keep offset label style on primary at z14/15

1ec5:
As a followup to #184, keep the offset label style on primary roads at zoom levels 14 and 15 to keep the labels from overlapping with dark-style roads.

before after<br>_Before and after_


adamfranco (approved):
Nice catch!

PR #189, opened 2022-02-23
Consolidate road label layers

1ec5:
Consolidated the road label layers for low and high zoom levels using `step` expressions. Consolidating layers that only vary by paint or layout properties will make it easier to grok the codebase once we add more layers that differ more significantly.

Theoretically, minimizing the number of layers results in better performance, though I suspect the performance hotspots lie elsewhere at this point. A more tangible effect is that labels transform directly from one treatment to another when crossing the threshold between low and high zoom levels. Before, if a label would be placed in the same location on either side of the threshold, it would still appear to fade out and back in with some delay. The new behavior feels a little snappier.

https://user-images.githubusercontent.com/1231218/155242795-3aed653a-3e4c-4d7c-a6e4-54c462c06114.mov

https://user-images.githubusercontent.com/1231218/155242809-7d1dc545-5a68-4bba-9855-6d05aa53b5c4.mov

The samples show a minor difference in label placement depending on whether a layer has a constant or variable `text-size` and `text-anchor`, but I don’t see a corresponding difference on the actual map. The difference was already there, but it’s a little more pronounced in the samples now:

before
after

### DIFF
--- a/style/americana.js
+++ b/style/americana.js
@@ -232,16 +232,11 @@ americanaLayers.push(
   lyrRoadLabel.motorway,
   lyrRoadLabel.trunk,
   lyrRoadLabel.primary,
-  lyrRoadLabel.primaryHZ,
   lyrRoadLabel.secondary,
-  lyrRoadLabel.secondaryHZ,
   lyrRoadLabel.tertiary,
-  lyrRoadLabel.tertiaryHZ,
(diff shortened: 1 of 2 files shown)


ZeLonewolf (approved):
Everything seems in order, and code is much cleaner now!

Issue #141, opened 2022-02-05
Ref values pre-processing


Mashin6:
Some countries have `ref` values that include additional characters such as `A1`. But in real life traffic signs only show the number: `1`.
![image](https://user-images.githubusercontent.com/5265707/152661800-57ca2b39-2619-4806-b7b7-e4e6fa8317fc.png)
vs
![image](https://user-images.githubusercontent.com/5265707/152661870-f4ad0383-ba9d-4de7-8c4e-fc5a04bbdd4c.png)

The biggest problems this is causing is in Romanian trunk roads shields (#99), but this issue is present in other countries as well:
Germany, Switzerland, Italy, Czechia.

A possibility would be to add a function that would strip of letter characters from refs in certain countries.

1ec5:
In #119, I implemented a `refsByWayName` lookup table specified as part of an individual shield definition object. It could be generalized into a function: these European shield definitions would implement the function by deleting the letter, while the Kentucky parkway shield definition could look up the value based on the `ref` (assuming we don’t find a more data-driven way to implement that shield definition).

ZeLonewolf:
I think ref pre-processing is something we should do under only extremely exceptional circumstances, if even at all.  It sounds like there is a relatively small list of countries that have tagged their highway route relations in a way that's different from the rest of the planet.  If that's the case, I'm fairly comfortable rendering them wrongly and suggesting that they might change the tagging to be more consistent with the global trend.

I noticed Italy in that list, and Italy does have the prefix values in their signs so I think in that case the ref prefix should actually be there and be rendered.

1ec5:
I think a transformation mechanism would be useful anyways. Thinking ahead to when we add the full complement of Texas shields, we’ll probably want to display most of the networks’ shields as something like `[FM 1234]` and `[PA 1502]`, which would require prepending something to the bare `ref` that ordinarily shouldn’t be there in the data.

ZeLonewolf:
> Thinking ahead to when we add the full complement of Texas shields, we’ll probably want to display most of the networks’ shields as something like `[FM 1234]` and `[PA 1502]`, which would require prepending something to the bare `ref` that ordinarily shouldn’t be there in the data.

This could also be accomplished by making the prefix part of the shield artwork.

1ec5:
The letters would have to be counterbalanced with whitespace on the other side of the image, which would shift the shield away from the roadway in many cases.

ZeLonewolf:
[...]

PR #190, opened 2022-02-23
Add Vietnam expressway shields

1ec5:
Added a shield for expressways in Vietnam. The tagging scheme for Vietnamese route relations coalesced around the `network=vn:*` prefix too recently to be picked up by the tile server, but here’s how it’ll look:

[CT.01](https://zelonewolf.github.io/openstreetmap-americana/#8.77/10.3671/106.1288)

As in https://github.com/ZeLonewolf/openstreetmap-americana/issues/141#issuecomment-1046409219, I’m not sure putting `CT.` in the `ref` is quite correct, since no one would ever say or write “expressway CT.01”. But it does match what’s on the shield, so this is expedient for now. Even if there’s consensus to drop the `CT.` from `ref`, it might be desirable to drop the “CT.” from the map as a simplification anyways.

### DIFF
--- a/style/js/shield_defs.js
+++ b/style/js/shield_defs.js
@@ -644,6 +644,15 @@ export function loadShields(shieldImages) {
     },
   };
 
+  shields["vn:expressway"] = roundedRectShield(
+    "#ffcd00",
+    "black",
+    "black",
+    2,
+    1,
+    null
+  );
+
   // Europe
   shields["e-road"] = {
     backgroundImage: [


ZeLonewolf:
It appears that the standard we chose, `network=vn:expressway`, is no longer the one most commonly in use:
https://taginfo.openstreetmap.org/search?q=network%3Dvn%3Aexpressway#tags

1ec5:
I discussed this back in February 2022 with [ZioDragon](https://www.openstreetmap.org/user/ZioDragon), who at the time was the only mapper working on Vietnam’s national route and expressway relations. The `network` tags were all over the place, but I recall that they decided on `vn:national` and `vn:expressway`. To my surprise, last December, they unilaterally changed the `vn:expressway` to [`VN:CT`](https://www.openstreetmap.org/changeset/129787222) and then [`VN:Expressway`](https://www.openstreetmap.org/changeset/129787222), but now it’s inconsistent with national routes. I reached out to the [OSM Vietnam Facebook group](https://www.facebook.com/groups/openstreetmapvietnam/posts/1653123425098640/) again for opinions. 🤞

ZeLonewolf:
Thanks for reaching out. If there is a change in local consensus on how to map Vietnam route relations, of course we should change our scheme to match whatever consensus emerges.

Issue #122, opened 2022-01-30
Asian Highway shields


1ec5:
The [Asian Highway Network](https://en.wikipedia.org/wiki/Asian_Highway_Network) is analogous to the E-road network (and overlaps with it in several countries). Like the E-road network, it is designated by a UN agency and largely consists of concurrencies with national routes, making it another showcase for Americana’s concurrency support.

Unfortunately, [the shields](https://commons.wikimedia.org/wiki/Category:Route_shields_of_the_Asian_Highway_Network) aren’t standardized and vary from country to country:

* Blue rectangle, white text 🇦🇲🇰🇭🇲🇾🇲🇲🇵🇭🇷🇺🇹🇭🇻🇳
* Green rectangle, white text 🇦🇿🇨🇳🇮🇷
* White rectangle, black text 🇯🇵🇰🇷
* White rectangle, green border, black text 🇮🇩
* Green variant of Indian National Highway shield, white text 🇮🇳

Most countries use the same prefix on the sign:

* “AH” 🇦🇲🇰🇭🇨🇳🇮🇩🇯🇵🇲🇾🇵🇭🇷🇺🇰🇷🇹🇭🇻🇳
* “AH-” 🇦🇿
* “A” 🇮🇷🇹🇭
* “A-” 🇲🇲
* No prefix 🇮🇳

ZeLonewolf:
Well, I guess the good news is we already support 🇯🇵🇰🇷 😬

I think this is the first example we've encountered where a route network has location-specific shields.

zekefarwell:
It's feeling more and more like we need a `route_shield_variant` tag in addition to `network`

1ec5:
One possible solution would be to convince mappers in each country to create a child relation with a country-specific `network` value like `VN:AH`. (It wouldn’t be uniform: `CN:AH` is already taken for Anhui’s provincial routes.) But that’s a lot of national communities to convince, and for something that in their eyes would be easily solvable by data consumers.

[...]

PR #195, opened 2022-02-24
Philippines route shields

1ec5:
Added route shields for the Philippines.

[Tacloban](https://zelonewolf.github.io/openstreetmap-americana/#12.13/11.22426/125.00547)<br>_Tacloban_

Route relation tagging in Metro Manila is inconsistent. National routes are a mix of the oversimplified `network=nhn` (based on recreational route tagging), `network=PH:nhn`, and the `network=PH:N:*` found elsewhere in the country. National expressways all tagged the same way as national routes despite distinct signage. Radial and circumferential route relations lack a `network` tag. I’ve omitted `network=nhn` because it would make little sense to show these shields in Europe, where this value is endemic. I included both the other styles because it’s unclear which the community stands behind more; `PH:nhn` is far more common, but `PH:N`/`E` is better distributed throughout the country.

[Metro Manila](https://zelonewolf.github.io/openstreetmap-americana/#12.26/14.5242/120.99913)<br>_Metro Manila_

The Philippines handles concurrencies between national routes/expressways and Asian Highways (#122) by combining them onto [a special sign](https://commons.wikimedia.org/wiki/Category:Diagrams_of_Asian_highway_signs_of_the_Philippines). But at the size that we show route shields, it would be better to separate them into two independent shields displayed in series. Unfortunately, these concurrencies are currently tagged like [`ref=E2;AH26`](https://www.openstreetmap.org/relation/3790802), which we can’t do anything reasonable with.

[AH26/E2](https://commons.wikimedia.org/wiki/File:AH26_%28E2%29_sign.svg)


ZeLonewolf:
I'm concerned about:
> National expressways all tagged the same way as national routes despite distinct signage.
 
and:
> I included both the other styles because it’s unclear which the community stands behind more

It sounds like the tagging of highway routes in the Philippines is not mature and/or lacks consensus.  In #180 we removed rendering support for Swiss highway route relations for this reason.  Perhaps we might consider reaching out to the Filipino mapping community to see whether there might be support for standardizing on one of the variants and also separating out relation `network=*` for the cases of distinct signage?

1ec5:
> It sounds like the tagging of highway routes in the Philippines is not mature and/or lacks consensus.

To be clear, I was describing the state of affairs in Metro Manila specifically. The rest of the country seems to be more consistent, and I only included `network` values that are qualified by `PH:`, ignoring the problematic `nhn` value. That said, outreach to the Philippines mapping community is a good idea.

TagaSanPedroAko:
@1ec5 Hi there! From Philippines. but currently in Canada, I'll share what seems to be the current situation with Philippine expressway/numbered route relations.

In its current state, tagging of Philippine expressways is kind of a mess. It was in 2015 when those yellow or white shields began springing up (and have seen one since then), and one user has just began adding relations for those  numbered routes, but it kind of follows an idiosyncratic scheme of network=ph:nhn and nhn:class=expressway/primary/secondary. I have been bringing the network tagging into accordance with the wider network tagging scheme of [country code]:[road type] but there's hundred of these across the thousands of the Philippines' islands, let alone Metro Manila.

style/js/shield_defs.js
```diff
@@ -610,6 +610,32 @@ export function loadShields(shieldImages) {
...
+      };
+  shields["PH:E"] = {
+    backgroundImage: shieldImages.shield40_ph_expressway_2,
+    textColor: "black",
+    padding: {
+      left: 2,
+      right: 2,
+      top: 2,
+      bottom: 6,
+
```
1ec5, line 638: This is a bad merge, causing the build to fail.

TagaSanPedroAko:
@1ec5 Well, relations that use `PH:N:primary` and `PH:N:secondary` should be easily migrated into the simpler `PH:N`, which us justifiable as there aren't much difference between the shields used for a national primary (1-2 digits) and a national secondary (3 digits).

1ec5:
I don’t have a strong opinion on whether to include the primary/secondary/tertiary classification in the `network` value. It’s easy enough to give them all the same shield. However, if the number already reliably indicates the classification and you don’t don’t think it’ll be needed for, say, a Philippines renderer that colors each classification differently, then the plain `PH:N` would be simpler. As I understand it, `designation` is already [widely used](http://taginfo.geofabrik.de/asia/philippines/tags/designation=national_secondary_road#map) for the more granular classifications.

TagaSanPedroAko:
@1ec5 Agreed. Just added various route relations last week and retagged those still using the old scheme. ph: suffix is plain wrong (that would be an ISO language code). Also removed AH26 from the main reference where tagged (those can already be covered by int_ref).

FYI, Philippines uses a road classification scheme based on function and importance, not one based on the number (as with most of Southeast Asia). Expressways, yes, will generally be motorway, but below that, it's much like with the US; 1- to 2-digit routes may not have the importance of a trunk, some 3-digit ones form the "best" routes between major cities, tertiary is at best an administrative classification and roads classified that best tagged by their importance, and there's a lot of bypass construction that would justify moving large classifications away from built-up areas.

style/js/shield_defs.js
```diff
@@ -798,6 +798,33 @@ export function loadShields(shieldImages) {
     },
   };
 
+  shields["PH:N"] =
+    shields["PH:N:primary"] =
+    shields["PH:N:secondary"] =
+      {
```
ZeLonewolf, line 804:
```suggestion
  shields["PH:N"] =
      {
```

Let's ship `PH:N` and `PH:E` as initially-supported tagging that has clear community support.

1ec5:
By the way, I didn’t find an existing tagging scheme for route relations for Metro Manila’s [circumferential and radial roads](https://en.wikipedia.org/wiki/List_of_roads_in_Metro_Manila#Circumferential_and_radial_roads), but those should be easy to implement once they’re tagged.

ZeLonewolf (approved):

TagaSanPedroAko:
@1ec5 I wouldn't think circumferential/radial road numbers should be rendered at all. Well, the circumferential road numbers are in common use save C-1, but the only radial road number commonly used is R-10 (more commonly said of as Road 10). After all, those numbers are mostly for planning purposes, and there isn't a standard shield for them. I'll be in favor of rendering only AH26, national road numbers and expressway numbers; Circumferential/radial road numbers should be migrated to unsigned_ref in the route relations.

Issue #13, opened 2021-05-25
West Virginia "fractional" route symbols


ZeLonewolf:
West Virginia has unusual 2-part "fractional" route designations as indicated here:
https://www.flickr.com/photos/andrew-turnbull/5904974507

How should these be rendered in the style?

~Blocked by #6~

1ec5:
> How should these be rendered in the style?

We could render it either horizontally or vertically in a rounded rectangle. A horizontal layout would use a solidus, like `(35/12)`. Either way, we’d have to stretch it beyond a circle to keep the numerals legible.

PR #199, opened 2022-02-26
Prevent waterway label underlap

### LINKED ISSUE #198: Waterway labels underlap bridges
#197 introduced waterway labels that tend to underlap bridges. While it is an interesting skeuomorphic effect to have the text running beneath bridges, it comes at the cost of readability, since we don’t have a very robust casing around bridges. These conflicts look like a bug, however well intentioned. On the other hand, a waterway label overlapping bridges looks unsophisticated too. We would have to add a robust halo around the waterway label to keep the text from becoming indiscernible above the red motorways and trunk roads.

[Brent Spence Bridge](https://zelonewolf.github.io/openstreetmap-americana/#13/39.0909/-84.52257)

I think we should explore options for keeping the labels away from bridges. The fewer conflicts between layers, the fewer edge case we need to consider for additional contrast.

1ec5:
Added an invisible spacer icon to bridges along a motorway, trunk road, or primary road to push waterway labels away from the bridge.

Before z12 After z12

The spacer layer is placed just before the highway label layers, so it shouldn’t affect highway labels. However, it does prevent some labeling of waterways where there are many crossings. Perhaps we could decrease the waterway label layer’s [`symbol-spacing`](https://maplibre.org/maplibre-gl-js-docs/style-spec/layers/#layout-symbol-symbol-spacing) if that becomes a problem.

The spacer icon technique could also help push shields away from road–road separations where they would be ambiguous.

Fixes #198.

### DIFF
--- a/style/layer/road_label.js
+++ b/style/layer/road_label.js
@@ -130,6 +130,8 @@ export const smallService = {
   "source-layer": "transportation_name",
 };
 
+// A spacer label on each bridge to push any waterway label away from the bridge.
+// https://github.com/ZeLonewolf/openstreetmap-americana/issues/198
 export const bridgeSpacer = {
   id: "bridge_spacer",
   type: "symbol",
--- a/style/layer/road_label.js
+++ b/style/layer/road_label.js
@@ -129,3 +129,24 @@ export const smallService = {
   source: "openmaptiles",
   "source-layer": "transportation_name",
 };
+
+export const bridgeSpacer = {
+  id: "bridge_spacer",
+  type: "symbol",
+  source: "openmaptiles",
+  "source-layer": "transportation",
+  filter: [
+    "all",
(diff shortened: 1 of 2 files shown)


style/layer/road_label.js
```diff
@@ -129,3 +129,24 @@ export const smallService = {
   source: "openmaptiles",
   "source-layer": "transportation_name",
 };
+
+export const bridgeSpacer = {
```
ZeLonewolf, line 133: Please add a comment to explain to a future developer what this sorcery does.

ZeLonewolf (changes requested):

ZeLonewolf:
Underlap still occurs at #13/41.82853/-71.36219

image

1ec5:
Seekonk River

ZeLonewolf (approved):
Looks great, thanks!

PR #200, opened 2022-02-26
Optimize SVGs

1ec5:
I’ve lost count of the times @ZeLonewolf has asked me to optimize the product of my meticulous button-mashing in Inkscape into more compact SVG markup. `npm run code_format` now runs all the SVG icons through [SVGO](https://www.npmjs.com/package/svgo/)’s default preset list of optimization plugins. I haven’t really explored the plugins and their configuration options, so please scrutinize each of the resulting SVGs for perceptible differences.


1ec5:
By the way, I’m not entirely sure why we need this optimization, since the SVGs get compiled down into the spritesheet at build time anyways. But as long as the maintainers expect optimization, it should be applied automatically, so that the contributor doesn’t have to remember to save as Optimized SVG instead of just saving in Inkscape.

ZeLonewolf:
> I’m not entirely sure why we need this optimization

Because extra, unnecessary code is where bugs go to hide :)

ZeLonewolf:
npm warns:

```npm WARN deprecated svgo@1.3.2: This SVGO version is no longer supported. Upgrade to v2.x.x.```

Should we be using a newer one?

1ec5:
I specified `^2.8.0`. Not sure how you’re getting v1.3.2.

ZeLonewolf:
Looks like spritezero is the culprit for that warning message.  I looked in node_modules and it's clearly got the newer version in there so, I think we can ignore that message.

ZeLonewolf (approved):
Shields all look normal to me in the map, thanks for automating this process!

1ec5:
There’s a very subtle difference in style/icons/shield40_us_us_2.svg that I feel uneasy about: the optimized SVG seems to be offset ever so slightly to the left, making the numerals appear uncentered. The effect is present on the map but not when viewing the SVG directly in a browser. I can reproduce this issue on a 2× display; it’s possible that it doesn’t reproduce on a 1× display.

before
after

ZeLonewolf:
Looks completely normal in Inkscape:

image

However, in the sprite sheet generated from this branch, you can see that it's cut off slightly on the left side (where it jams into the KY parkway shield):
image

Seems like spritezero is somehow not properly detecting the edge of the graphic.

ZeLonewolf:
The problematic optimization is `convertPathData`, which I've disabled in 1000314 and re-processed the icons.  Needs a review to confirm we're seeing the expected behavior.

1ec5:
Thanks! I still don’t quite grok all that these plugins are doing, other than being impressed at their ability to resolve transforms. I suspected it might have to do with a lack of precision, since Inkscape’s optimizer defaults to five decimal places while SVGO defaults to three. But I was unsuccessful in configuring the plugin – probably didn’t do it correctly. Maybe at some point we’ll figure it out, but in the meantime, the optimized code looks reasonable compared to Inkscape’s output.

PR #201, opened 2022-02-26
Improve visibility of lake labels

ZeLonewolf:
This PR slightly increases the size of lake labels, and adds rendering of lake labels represented as points in the tiles.  The quality of label placement is driven by the data provided by the tiles, but this should at least be an improvement over the current labelling.

Lake Ontario, before and after:
image
image

Finger Lakes, before and after:
image
image

New Hampshire, before and after:
image
image

### DIFF
--- a/style/americana.js
+++ b/style/americana.js
@@ -50,6 +50,7 @@ americanaLayers.push(
 
   lyrWater.waterwayLabel,
   lyrWater.waterLabel,
+  lyrWater.waterPointLabel,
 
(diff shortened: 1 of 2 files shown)


jleedev:
Yeah this layer leaves a bit to be desired, but it's good to show everything that's there.

1ec5 (approved):
These two symbol layers have a lot in common. I’d be tempted to consolidate them into a single layer, but it’s possible that a future change might cause the point and line labeling to diverge more, so this keeps things simple for now.

PR #205, opened 2022-02-26
Texas shields

amcginni:
This PR distinguishes several networks in the Texas highway system by employing bannered shields in a similar fashion to the interstate guide signs.  It does not attempt to provide a faithful reproduction of the state outline found on the Farm-to-market and Ranch-to-market reassurance shields.  An example of how some networks are displayed on the guide signs, from the TxDOT Freeway Signing Manual, is shown below.
image

The following deviations have been made from the guide signs.
* The "TEXAS" banner was dropped from the `US:TX` network.  For now, the square shield in Texas will imply that the route is a state highway.
* For `US:TX:Beltway`, the "BELTWAY" banner was shortened to "BELT" to improve readability.
* With the exception of toll roads in Harris County, which are not addressed in this PR, all toll road shields are rendered the same, as a blue rectangle with the "TOLL" banner on top.

Some examples are provided below.
TX-Belt
TX-Farm-Ranch
TX-Loop-Spur
TX-NASA
TX-Park
Dallas-Tollways


ZeLonewolf:
Relates to #186

style/js/shield_defs.js
```diff
@@ -559,6 +559,52 @@ export function loadShields(shieldImages) {
...
+    ["BELT"]
+  );
+  shields["US:TX:FM"] = banneredShield(
+    roundedRectShield("white", "black", "black", 1, 1, null),
+    ["F.M."]
+  );
+  shields["US:TX:RM"] = banneredShield(
+    roundedRectShield("white", "black", "black", 1, 1, null),
+    ["R.M."]
+  );
```
1ec5, line 590: It hadn’t occurred to me to reproduce the periods from the signs, but I guess it looks OK either way. This is nicer than “FARM”, since it matches the guide signs, and we wouldn’t be able to fit “RANCH” in four characters anyways.

FM 1220

style/js/shield_defs.js
```diff
@@ -559,6 +559,52 @@ export function loadShields(shieldImages) {
...
+  shields["US:TX:NASA"] = banneredShield(
+    roundedRectShield("white", "black", "black", 1, 1, null),
+    ["NASA"]
+  );
+
+  shields["US:TX:Toll"] = banneredShield(
+    roundedRectShield("#003f87", "white", "white", 1, 1, null),
+    ["TOLL"]
+  );
```
1ec5, line 603: I like how we can distinguish toll express lanes from ordinary lanes:



The “TOLL” banner is technically redundant to the blue background, but it does help with usability, since the toll shield very often appears alongside a non-toll shield with the same number.

Do we intend to also add the “TOLL” or “EXPRESS” banner to tolled Interstate and U.S. highways? The express lanes in the Dallas area appear to be tagged with [`network=US:I:Express:Toll`](https://www.openstreetmap.org/relation/8097271) and so on.

style/js/shield_defs.js
```diff
@@ -559,6 +559,52 @@ export function loadShields(shieldImages) {
...
+    ["F.M."]
+  );
+  shields["US:TX:RM"] = banneredShield(
+    roundedRectShield("white", "black", "black", 1, 1, null),
+    ["R.M."]
+  );
+  shields["US:TX:Recreational"] = banneredShield(
+    roundedRectShield("white", "black", "black", 1, 1, null),
+    ["R"]
+  );
```
1ec5, line 594: I’m unaware of any guide signs ever being posted for Recreational Roads, but how about giving it a brown outline as a nod to the reassurance marker?

R

style/js/shield_defs.js
```diff
@@ -559,6 +559,52 @@ export function loadShields(shieldImages) {
...
+    ["PARK"]
+  );
+  shields["US:TX:Beltway"] = banneredShield(
+    roundedRectShield("white", "black", "black", 1, 1, null),
+    ["BELT"]
+  );
+  shields["US:TX:FM"] = banneredShield(
+    roundedRectShield("white", "black", "black", 1, 1, null),
+    ["F.M."]
+  );
```
1ec5, line 586: There are a few Business Farm-to-Market Roads, which are tagged [`network=US:TX:FM:Business`](https://taginfo.openstreetmap.org/tags/network=US%3ATX%3AFM%3ABusiness). Use `banneredShield()` to give it a “BUS” banner on top of the “F.M.” banner.

style/js/shield_defs.js
```diff
@@ -559,6 +559,52 @@ export function loadShields(shieldImages) {
   shields["US:SC:Business"] = banneredShield(shields["US:SC"], ["BUS"]);
   shields["US:SC:Alternate"] = banneredShield(shields["US:SC"], ["ALT"]);
 
+  shields["US:TX"] = roundedRectShield("white", "black", "black", 1, 1, null);
```
1ec5, line 562: The default generic shield for unsupported networks is also a white rectangle, just like the intended State Highway shield, so this line technically isn’t required.

More importantly, we should make sure as many Texas networks as possible are supported here so that they don’t get confused with State Highways. One that comes to mind is [`network=US:TX:PA`](https://wiki.openstreetmap.org/wiki/Tag%3Anetwork%3DUS%3ATX%3APA) for the one Principal Arterial State System route. Its signs say “PA”, without periods, but periods would be fine too.

1ec5 (commented):
Thanks for messing with Texas! 😉 I agree with your design choices so far. My feedback below is primarily about the long tail of edge cases that Texas is famous for.

style/js/shield_defs.js
ZeLonewolf: I think from the informal discussions we've had on Slack, we intend to render toll roads in a different color entirely on the roadway itself.  So I think we probably don't explicitly need to put a toll banner on the highway shields for those roads.

style/js/shield_defs.js
1ec5: #213 makes it necessary to explicitly declare white rectangular shields in this file, as you’ve done for state highways.

style/js/shield_defs.js
1ec5: The banner guidelines adopted in #186 call for the “TOLL” banner to be kept as in this PR, because toll State Highways aren’t necessarily freeways. For example, [Loop 49](https://en.wikipedia.org/wiki/Texas_State_Highway_Loop_49) is currently an expressway. It’s tagged `US:TX:Loop:Toll`, which is a combination that isn’t handled by this PR yet. On the other hand, `US:I:Express:Toll` may not need both “EXPRESS” and “TOLL” as banners; I’m not sure.

style/js/shield_defs.js
```diff
@@ -718,6 +718,51 @@ export function loadShields(shieldImages) {
...
+    roundedRectShield("white", "black", "black", 1, 1, null),
+    ["NASA"]
+  );
+
+  shields["US:TX:Toll"] = banneredShield(
+    roundedRectShield("#003f87", "white", "white", 1, 1, null),
+    ["TOLL"]
+  );
+  shields["US:TX:CTRMA"] = shields["US:TX:Toll"];
+  shields["US:TX:NTTA"] = shields["US:TX:Toll"];
```
claysmalley, line 764: The Brazoria County Toll Road Authority is the only one in the Houston area with shields conforming to the state standard. `network=US:TX:BCTRA` probably belongs here too.

style/js/shield_defs.js
claysmalley: Existing `network` values in Texas containing `Toll` are:

* `US:I:Express:Toll`
* `US:TX:Express:Toll`
* `US:TX:Loop:Express:Toll`
* `US:TX:Loop:Toll`

Other banners like "EXPRESS" and "LOOP" should definitely be displayed, and anything with `US:TX:…:Toll` should get a blue shield.

I'm not convinced it's necessary to have the "TOLL" banner if toll roads will have a different color. But for now, it's probably fine to keep it in—we can always revisit this once toll roads get their new style.

style/js/shield_defs.js
```diff
@@ -718,6 +718,51 @@ export function loadShields(shieldImages) {
   shields["US:SC:Business"] = banneredShield(shields["US:SC"], ["BUS"]);
   shields["US:SC:Alternate"] = banneredShield(shields["US:SC"], ["ALT"]);
 
+  shields["US:TX"] = roundedRectShield("white", "black", "black", 1, 1, null);
+  shields["US:TX:Loop"] = banneredShield(
+    roundedRectShield("white", "black", "black", 1, 1, null),
+    ["LOOP"]
+  );
```
claysmalley, line 725: Texas has at least one city route, Andrews Loop 1910. [Here's what the shield looks like.](https://www.andrewscountynews.com/sites/andrewscountynews.etypegoogle12.com/files/4100009ffd_Ar00103004.jpg)

For `network=US:TX:Andrews:Andrews:Loop`, I think a shield matching the State Loop shield with dark blue instead of black would work okay. Obviously we can't include the copyrighted artwork, not that it's particularly helpful for navigation.

style/js/shield_defs.js
```diff
@@ -718,6 +718,51 @@ export function loadShields(shieldImages) {
...
+    ["NASA"]
+  );
+
+  shields["US:TX:Toll"] = banneredShield(
+    roundedRectShield("#003f87", "white", "white", 1, 1, null),
+    ["TOLL"]
+  );
+  shields["US:TX:CTRMA"] = shields["US:TX:Toll"];
+  shields["US:TX:NTTA"] = shields["US:TX:Toll"];
+
```
claysmalley, line 765: Existing `network` values for Texas county roads are:

- `US:TX:CR:Caldwell`
- `US:TX:CR:Collin`
- `US:TX:CR:Collingsworth`
- `US:TX:CR:Colorado`
- `US:TX:CR:Cooke`
- `US:TX:CR:Coryell`
- `US:TX:CR:Fannin`
- `US:TX:CR:Gonzales`
- `US:TX:CR:Guadalupe`
- `US:TX:CR:Hays`
- `US:TX:CR:Lamar`
- `US:TX:CR:Runnels`
- `US:TX:CR:San Patricio`
- `US:TX:CR:Smith`
- `US:TX:CR:Van Zandt`
- `US:TX:CR:Wilbarger`
- `US:TX:CR:Wilson`
- `US:TX:CR:Wood`
- `US:TX:CR:Wood:PR` (private roads)

AFAIK, county roads are signposted with either the standard blue pentagon or nothing at all. We haven't done the research yet on which of the 254 counties do and don't give their roads shields, so unless anybody has any insight on these specific counties, we could probably hold off on this.

style/js/shield_defs.js
ZeLonewolf: "EXPR" would be a reasonable abbreviation if we need that in a banner.

style/js/shield_defs.js
ZeLonewolf: I think if you're OK with distorting the state shape a whole bunch, it's achievable with a wide shield.

style/js/shield_defs.js
ZeLonewolf: I'm inclined to say it might be better to drop punctuation in banners, for consistency across the style.  Though I do agree these look pretty decent in the screen caps.

style/js/shield_defs.js
amcginni: Taking some liberties with the state outline, I can get a single digit route to display legibly.

We should also consider whether a banner would be necessary on this shield anyway, since it looks so similar to the Farm and Ranch--to-market reassurance markers.
image

ZeLonewolf: That's barely workable for a 1-character shield, never mind a 5-character shield (including the space).  I think you'll need to go pretty drastic, similar to how Louisiana and Georgia are drawn in rebusurance:

https://github.com/1ec5/rebusurance/blob/main/img/canvas.png

style/js/shield_defs.js
amcginni: I agree with Minh.  County routes can be addressed in a separate PR.

style/js/shield_defs.js
```diff
@@ -718,6 +718,34 @@ export function loadShields(shieldImages) {
...
+  shields["US:TX:Toll"] = banneredShield(
+    roundedRectShield("#003f87", "white", "white", 1, 1, null),
+    ["TOLL"]
+  );
+  shields["US:TX:BCTRA"] = shields["US:TX:Toll"];
+  shields["US:TX:CTRMA"] = shields["US:TX:Toll"];
+  shields["US:TX:NTTA"] = shields["US:TX:Toll"];
+  shields["US:TX:Express:Toll"] = banneredShield(shields["US:TX:Toll"], ["TOLL", "EXPR"]);
+  shields["US:TX:Loop:Toll"] = banneredShield(shields["US:TX:Toll"], ["TOLL", "LOOP"]);
+  shields["US:TX:Loop:Express:Toll"] = banneredShield(shields["US:TX:Toll"], ["TOLL", "EXPR", "LOOP"]);
```
claysmalley, line 747: "TOLL" is usually below the shield. Obviously we don't support modifiers below the shield, and "LOOP" should probably remain last in order to indicate that these are really State Loops with some modifiers. But IMO "EXPRESS" should go before "TOLL".

Is `US:I:Express:Toll` still missing?

1ec5:
Texas road networks are a thorn in the side of every map designer. It’s worth considering how the map industry has handled them over the years. Some online maps, notably Apple Maps, do attempt to cram numbers into a heavily distorted shape of Texas. However, I’d contend that Apple’s approach reflects a possible unfamiliarity with Texas, consistent with how they’ve applied California-style cutout shields to every state, but only for U.S. Routes. Locally published maps show a variety of approaches, but none attempt to reproduce the state shape on F.M. or R.M. roads.

TxDOT and Rand McNally invent some shapes and prefixes that bear no resemblance to any signs, in order to maintain a clear distinction as compactly as possible. This approach relies on the presence of a legend, which we don’t have. Yet F.M.s and R.M.s are still rectangles:

TxDOT 2017 Rand McNally Dallas 1994

Five Star Maps says rectangles are for “county roads”, but as far as I can tell, rectangles actually mark F.M.s, R.M.s, Loops, Spurs, etc.:

Five Star Texas 2009

Mapsco uses rectangles with text at the top for F.M.s, Loops, Spurs, etc., much like the concept proposed in this PR:

Mapsco Fort Worth 1992 Mapsco Fort Worth 1992 Spur 303

style/js/shield_defs.js
claysmalley: The `network` values for those relations were tagged by me, so don't read too much into the ordering. I think it's more important to match the order displayed on the signs. If that gives the shield and modifiers an hourglass figure, so be it.

style/js/shield_defs.js
claysmalley: Even if we're turning this one into a rectangle due to constraints, I still like the idea of matching the brown of the original sign.

amcginni: Commit 8817776acf893549237e2ae7fc465bf086510e4c adds the brown outline on recreation road shields.
image

claysmalley (approved):
Looks good to me!

ZeLonewolf (approved):
Thanks for messing with Texas!

PR #207, opened 2022-02-28
Update maptiler key

ZeLonewolf:
Updates the Maptiler Cloud key.

I cannot explain why it's suddenly different, but this is the key now showing in my MapTiler account.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -19,7 +19,7 @@ jobs:
         run: |
           cd style
           npm install --include=dev
-          sed 's/<your MapTiler key>/53iZvB2drcamS0Ge0xiD/g' configs/config.maptiler.js > config.js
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         run: |
           cd style
           npm install --include=dev
-          sed 's/<your MapTiler key>/53iZvB2drcamS0Ge0xiD/g' configs/config.maptiler.js > config.js
+          sed 's/<your MapTiler key>/hJFMGrd2JszhwfJx5M2k/g' configs/config.maptiler.js > config.js
           npm run build
         # MapTiler key 53iZvB2drcamS0Ge0xiD only allows requests from zelonewolf.github.io
```
1ec5, line 24: Update this comment too?

ZeLonewolf: Fixed

1ec5 (approved):

Issue #118, opened 2022-01-29
Extract loadShields objects into JSON


1ec5:
The `loadShields()` function in shield_defs.js will get unwieldy as we continue to add support for more networks. I think most of the objects in this function could be converted to JSON without losing expressiveness; then, the function could load the data and transform it into the data structure needed by its callers. A JSON file will be less intimidating for non-programmers to contribute to and possibly be easier to refactor in the future.

ZeLonewolf:
Sounds good to me.  We would have to convert the function references into an enum scheme to make that work, I think?

zekefarwell:
I agree the the `loadShields()` function will get unwieldy as it grows (at over 500 lines long it's arguably already there).  Extracting the `shields` data structure out of that function definitely makes sense, but I'm not sure it needs to be JSON.  I think I'd prefer a standard object declaration in JS.  More flexible syntax, possible to use variables, etc.  It seems unlikely to me that non-programmer would really find JSON more approachable than a JS object declaration that looks basically the same.  It's been a while since I was a non-programmer though, so hard to put myself in those shoes.

ZeLonewolf:
The point about using variables is a good one.  There's a bunch of duplication that we're avoiding by re-using common shapes.  The NJ counties alone would be a significant inflation in size if we produce a JSON version of the shield definitions.

1ec5:
A literal translation to JSON would result in redundancy, but I’m pretty confident we could figure out ways around that. In particular, we should rename heavily reused icons like shield40_us_county to be more generic, as suggested in https://github.com/ZeLonewolf/openstreetmap-americana/pull/113#issuecomment-1025002165.

PR #209, opened 2022-03-01
Generate maps of supported states, countries for shields

1ec5:
Similar to #179, this PR adds a Node script that generates choropleth maps of the United States and world that show which states’ and countries’ shields are supported by OpenStreetMap Americana. The SVG maps are placed in the documentation folder, making it convenient to link to the GitHub repository in conversations about shield support. See this PR’s rich diff for the current output. The maps are based on public domain base maps of [U.S. states](https://commons.wikimedia.org/wiki/File:Blank_USA,_w_territories.svg) and [world countries](https://commons.wikimedia.org/wiki/File:World_map_configurable.svg) on Wikimedia Commons.


ZeLonewolf:
Some states (notably, Maine and all of southern New England) have plain black rectangles with a white background for shields.  I consider these to be officially supported by the default symbology.  How should we handle those cases?  Should we explicitly declare those as default rectangles in the shield def?

1ec5:
I think we should declare them somewhere. That would also allow our support for them to show up in taginfo. One idea I had for #118 is to describe standard shields by their shape and color directly in the definition, instead of calling out to any function, then resolve those descriptions later. We could detect descriptions that happen to match the default and avoid drawing anything special for them.

ZeLonewolf:
> describe standard shields by their shape and color directly in the definition, instead of calling out to any function

This is a pretty straightforward refactor.  I initially hadn't done this because we don't have a current use case for marshalling and unmarshalling the shield style.

ZeLonewolf:
Looks good to me and seems to work.  What's the plan for doing something with these generated maps?  It isn't linked to any of the documentation files that I can see.

1ec5:
Before merging:

* [ ] Come up with a stopgap solution for shields that should be white rectangles, in anticipation of #118
* [ ] Embed the maps into a Markdown guide tracking the project’s progress on shields

ZeLonewolf (approved):
Looks essentially correct and javascript previously tested.  I'm not able to confirm that the  links deploy properly since that's done with github actions, so please check that the links are working after the merge.

1ec5:
> Looks essentially correct and javascript previously tested. I'm not able to confirm that the  links deploy properly since that's done with github actions, so please check that the links are working after the merge.

The links in this PR are confined to files that are excluded from the package distribution. They’re only intended to appear here on GitHub. The rich diff of this PR is representative of what visitors to this repository will see.

PR #211, opened 2022-03-03
Filter out disputed boundary lines

### LINKED ISSUE #210: Disputed boundaries are depicted as normal boundaries
For the most part, disputed international boundaries are depicted just like normal boundaries, so the map reflects what OpenStreetMap considers to be the de facto boundary claim according to the “on the ground” rule, with same caveats. But the user would be better served by indicating that these particular boundary claims are disputed.

[Ladakh](https://zelonewolf.github.io/openstreetmap-americana/#6/33.191/78.668)<br>_Boundaries disputed between India and China._

Americana attempts to hide disputed boundaries using OpenMapTiles’ `claimed_by` property:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/affa498126535befbe31740543fd404fe5aed0f0/style/layer/boundary.js#L17

However, [`claimed_by`](https://wiki.openstreetmap.org/wiki/Key:claimed_by) is not very reliably tagged on disputed boundary claims in OpenStreetMap. `disputed=yes` and [`disputed_by`](https://wiki.openstreetmap.org/wiki/Key:disputed_by) are much more common. OpenMapTiles [rolls these properties](https://github.com/openmaptiles/openmaptiles/blob/b4b897999d6951fa29c3389070ee2401d6979e03/layers/boundary/mapping.yaml#L5) and some others into a unified [`disputed`](https://openmaptiles.org/schema/#disputed) property.

We can easily change the filter to check for `disputed` being set to 0 instead of `claimed_by` being absent. However, if the point of this filter was to express a more neutral stance, hiding the disputed boundary might be inappropriate because it could be perceived as taking a side. I think we should render the disputed boundary claim but depict it specially so the user knows that the boundary may not be definitive.

Maps that use line width to distinguish national boundaries from subnational boundaries typically use broken lines to indicate disputed boundaries:

Rand McNally World Atlas

Since we’re already using dashed lines for subnational boundaries, I think a broken line for disputed boundaries would get confusing. But we could hide the line itself and show just the casing or vice versa. Ideally we could label the boundary with the claimant, but I don’t think OpenMapTiles is exposing that information.

1ec5:
Filter out disputed boundary lines based on `disputed` instead of `claimed_by`, but keep showing casings for the boundary lines that OSM includes in boundary relations.

before after<br>_Boundaries disputed [between India and China](https://zelonewolf.github.io/openstreetmap-americana/#6/33.191/78.668)._

Fixes #210.

### DIFF
--- a/style/layer/boundary.js
+++ b/style/layer/boundary.js
@@ -178,6 +178,7 @@ export const countryCasing = {
   filter: ["all", ["in", "admin_level", 2], ["==", "maritime", 0]],
   minzoom: 2,
   layout: {
+    "line-cap": "round",
     "line-join": "round",
     visibility: "visible",
   },
--- a/style/layer/boundary.js
+++ b/style/layer/boundary.js
@@ -14,7 +14,7 @@ export const city = {
   filter: [
     "all",
     ["==", "admin_level", 8],
-    ["!has", "claimed_by"],
+    ["==", "disputed", 0],
     ["==", "maritime", 0],
   ],
   minzoom: 11,
@@ -42,7 +42,7 @@ export const countyCasing = {
   filter: [
     "all",
     ["==", "admin_level", 6],
-    ["!has", "claimed_by"],
+    ["==", "disputed", 0],
     ["==", "maritime", 0],
   ],
   minzoom: 11,
@@ -66,7 +66,7 @@ export const county = {
   filter: [
     "all",
     ["==", "admin_level", 6],
-    ["!has", "claimed_by"],
+    ["==", "disputed", 0],
     ["==", "maritime", 0],
   ],
   minzoom: 9,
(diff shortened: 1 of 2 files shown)


ZeLonewolf (approved):
After extensive spitballing, I'm unable to come up with a better way to represent this, and I think it's important given the current world environment that we render disputed boundaries distinctly ASAP.

style/layer/boundary.js
```diff
@@ -178,6 +178,7 @@ export const countryCasing = {
   filter: ["all", ["in", "admin_level", 2], ["==", "maritime", 0]],
   minzoom: 2,
   layout: {
+    "line-cap": "round",
```
1ec5, line 181: Especially along disputed boundaries, the boundary lines can be split up into multiple LineString features that OpenMapTiles doesn’t union together, due to differing tags like `claimed_by` and `disputed_by`. These splits result in gaps that need to be filled in with line caps. A round line cap won’t look great with a translucent line, but overlaps aren’t noticeable at the zoom levels where the line would be translucent.

line-cap

PR #213, opened 2022-03-06
Explicitly declare plain white shields

ZeLonewolf:
This PR explicitly declares shield definitions for state highway routes that are represented as plain white rectangles with black borders (the default).  This is needed in order to support use cases such as #209.


style/js/shield_defs.js
```diff
@@ -785,6 +823,29 @@ export function loadShields(shieldImages) {
...
+    "GD:Dianbai",
+    "GD:Huazhou",
+    "GD:Maoming",
+    "GD:Maonan",
+    "GD:Xinyi",
+    "HA:Luoyang",
+    "HA:Mengjin",
+    "JS:Liyang",
+    "JS:Xuzhou",
+    "JS:Wuzhong",
```
1ec5, line 838: It would be quite inconvenient to track all the counties, districts, and villages in China. Local route signs in China have a uniform appearance, but I’m unsure if numbers get reused from one locality to another.

ZeLonewolf: We might consider a more flexible map key scheme, such as: https://www.npmjs.com/package/regexp-map

1ec5: If the local route numbering system is also uniform across the country, maybe we could see if mappers in China would be open to more generic `network` values.

1ec5 (approved):
I still have misgivings about having to maintain a practically unbounded set of local `network` values for the PRC for what ends up being a uniform shield and numbering scheme. However, this concern shouldn’t block the other changes. Maybe the refactor in #118 could introduce a regular expression option for certain `network` tagging schemes that are uninteresting to enumerate.

PR #215, opened 2022-03-06
Roads under construction

### LINKED ISSUE #212: Roads under construction
The style should show roads that are under construction and possibly some major proposed roads. Otherwise, there can be prominent, otherwise inexplicable gaps or dangling roads.

[A-35](https://zelonewolf.github.io/openstreetmap-americana/#13/45.11697/-73.12638)<br>_Québec Autoroute 35 is being built in phases._

[Roebling](https://zelonewolf.github.io/openstreetmap-americana/#15/39.09246/-84.50986)<br>_This bridge across the Ohio River has been closed for repairs long enough to be tagged as being under construction too._

## Precedents

The predominant convention among North American maps is to depict a road under construction as a broken line:

OklaDOT

This convention extends to maps published in Canada or for tourists from continental Europe:

Ontario Ministry of Transportation 1994 Hallwag California

The essence of this treatment is that the casing is also broken, not just the fill, and the fill has alternating transparent segments, not white segments. I prefer this treatment to what openstreetmap-carto does, which evokes a barber pole or candy cane:

[candy cane](https://www.openstreetmap.org/#map=16/45.0935/-73.0849)

In many cases, inaccessible or unpaved roads get almost the same treatment. It seems the general idea is for a broken or dashed line to warn of a bad day:

MnDOT 2019–20 Utah DOT 2011

Some maps call them “future” roads and include roads still in aspirational planning stages, while others distinguish between proposed roads and roads under construction:

Rand McNally Dallas Metro 1994 American Map New York City Eastern Long Island INDOT LaDOTD 2014

The Ohio Department of Transportation at one point used a translucent red dotted overlay to indicate future roads, just like openstreetmap-carto’s `access=no` treatment:

ODOT 1993

Most maps in my collection, even city-scale maps only show limited-access highways under construction but omit less important roads under construction:

Best Western Nevada West Virginia DOT AAA Hawaii 2016 Georgia DOT 2005–06 Mississippi DOT 1996–97

This could be because they only depict freeways for which a right of way has already been cleared, leaving a major gap. But it could also be because smaller roads tend to get completed more rapidly than paper maps can reliably keep up with.

Some maps do show multiple levels of roads under construction; these maps generally don’t already use dashes to convey surface conditions:

South Carolina DOT 2005 North Carolina DOT 2005 ODOT Alabama DOT 2007–08 Mobil (Rand McNally) New York Shell (Gousha) East Central States PennDOT 2003

I think it would benefit mappers if Americana shows any road under construction, regardless of classification, but if the complexity of supporting yet another facet of road classification proves too complex, at least there would be precedent for only showing the most major construction projects.

Since not-yet-completed freeways are relatively rare, most maps annotate them with a special label explicitly stating that they’re planned or under construction, even giving an expected date of completion if the DOT is confident enough about it:

KYTC Universal Map Cincinnati Area Mapsco DFW SH 170 Mapsco DFW IH 30/820 Mapsco DFW SH 360

1ec5:
Added path and casing line layers for roads under construction, which are depicted as broken lines. Labels for these roads end with the annotation “(UNDER CONSTRUCTION)” is slightly smaller type.

## Examples

https://user-images.githubusercontent.com/1231218/156917234-2e4ccdb4-8044-4312-a7af-ed3694b47a92.mov

z7 | z8 | z9 | z10 | z11
----|----|----|----|----
sample-z7 | sample-z8 | sample-z9 | sample-z10 | sample-z11

z12 | z13 | z14 | z15 | z16
----|----|----|----|----
sample-z12 | sample-z13 | sample-z14 | sample-z15 | sample-z16

https://user-images.githubusercontent.com/1231218/156917539-2eed3ea2-7d85-4128-b9ac-f906d9303d0e.mov

z7 | z8 | z9 | z10 | z11
----|----|----|----|----
a35-z7 | a35-z8 | a35-z9 | a35-z10 | a35-z11

z12 | z13 | z14 | z15 | z16
----|----|----|----|----
a35-z12 | a35-z13 | a35-z14 | a35-z15 | a35-z16

## Caveats

Aligning the path and casing line layers has been a challenge. The standard approach to casing – putting a wider line layer below the path line layer – breaks down along curved lines, because the [`line-dasharray`](https://maplibre.org/maplibre-gl-js-docs/style-spec/layers/#paint-line-line-dasharray) paint property is scaled by the line width. Instead, both layers have the same line width, but the casing layer has a [`line-gap-width`](https://maplibre.org/maplibre-gl-js-docs/style-spec/layers/#paint-line-line-gap-width) property. Despite the property being measured in pixels, I can’t get the casing to be any thinner than 5&nbsp;pixels wide.

I gave the dashes and gaps an exponential ramp that should keep it discernible as a broken line at any zoom level. Unfortunately, the dash pattern shows seams at tile boundaries or anywhere two LineString features connect, such as at the ends of bridges. (I gave bridges and tunnels under construction exactly the same style, for simplicity.) It’s also very easy for divided highways to wind up with alternating dash patterns that blend into one another at some zoom levels.

I’ve marked this PR as a draft not because these imperfections are fatal to the broken line concept but rather because the code changes are so messy. The road line code is very rigidly structured but shows signs of strain with these changes. The `line-gap-width` approach breaks the assumption that there are distinct line widths for the path versus the casing, and the mere presence of a construction variant of each road layer means lots of conditionals. Since each layer definition is ultimately an object, we probably should be relying on prototypes instead of classes and the Builder pattern (no pun intended).

Fixes #212.


ZeLonewolf:
The fact that we don't have merged lines on dual carriageways is squarely a server-side problem, and I'm comfortable ignoring unwanted display artifacts due to that.  I'm thinking the same about the tile boundary artifacts, and am hoping this is something that can some day be resolved in maplibre.

ZeLonewolf:
The use of a hard-coded English label is new ground for this style.  Is there a path to i18n of this and any other fixed strings in the event someone wants to implement a language switcher?

1ec5:
> The use of a hard-coded English label is new ground for this style. Is there a path to i18n of this and any other fixed strings in the event someone wants to implement a language switcher?

I factored out the suffix into a translation table with code to automatically select the translation that best matches the user’s locale. In time, we can factor it out again into a separate file in a format like [Structured JSON](https://docs.transifex.com/formats/json/structured-json) that a translation management system like Transifex can understand.

matkoniecz:
Is it planned to display tunnel/bridge status?

1ec5:
For visual simplicity, I did not include special treatment of tunnels and bridges under construction. The idea is that a road under construction is only being “sketched in” with a broken line, making it unnecessary to faithfully represent how it will connect to roads around it (or not). This seems to match the approach taken by every print map that I’ve been able to find. However, if you have examples of this extra detail being depicted, I’d be interested to know how they approach the problem while retaining intuitiveness.

adamfranco:
_Noting for reference:_ As mentioned in Slack, one styling option for `highway=track` is the broken double-black line treatment shown as "Uninmproved road" in this example (from #216). This treatment may collide with some treatments of roads under construction in this PR. 
Screen Shot 2022-06-09 at 1 56 00 PM

1ec5:
#502 took a different, more workable approach to rendering roads under construction. #503 tracks reviving the localization system in https://github.com/ZeLonewolf/openstreetmap-americana/pull/215#issuecomment-1060067427 to implement glosses in labels.

PR #217, opened 2022-03-10
South Korea shields

### LINKED ISSUE #138: Dynamically drawn shields should not be drawn without ref
`roundedRectShield()` renders empty shield in some special cases.

Missing ref:
image

When `noref = yes` is set:
image

Too long ref. Though some text scaling could be applied before decision not to draw it.
image

1ec5:
Added shields for national expressways, national highways, and local highways in South Korea.

[Bongdam](https://zelonewolf.github.io/openstreetmap-americana/#13/37.19903/126.97461)

The national expressway shield is a little slice of home away from home. The shape is based on [this sign blank on Wikimedia Commons](https://commons.wikimedia.org/wiki/File:Korean_Highway.svg), which is in the public domain. It’s very similar to the U.S. route shield, but the contours have subtle differences. It doesn’t leave very much room for the numerals, so ironically, local highways are much more prominent than national expressways.

~~The national highway shield is supposed to be an oval, even when a two-digit number would fit inside a circle, so I eschewed the `roundedRectShield()` function in favor of a custom image. Unfortunately, spritezero turns the ellipse into a hendecagon.~~

I’m unable to implement shields for [metropolitan city roads and expressways](https://en.wikipedia.org/wiki/Highway_system_in_South_Korea#Special_Metropolitan_City_roads_and_Metropolitan_City_roads). They are mapped as route relations, but the relations aren’t tagged with `network`. Instead, they’ve been added to `network` relations representing the [highway system](https://www.openstreetmap.org/relation/13157356) or the [municipal government](https://www.openstreetmap.org/relation/10987351). We need them to be tagged with `network=KR:*` like the other route types so we can support them properly. The `ref`s for Seoul’s special metropolitan city roads include both the signposted number and the non-signposted official number, which should go in `unsigned_ref`.

Along the way, this PR fixes an issue that caused routes with matching `network`s but missing `ref`s to appear as blank shields.

Fixes #138.


1ec5:
Somehow I overlooked the `circleShield()` function. I added some options to specify the text color and fix the width, since the shield should always look like an oval:

Bongdam

1ec5:
1× simulation in Firefox:

1ec5:
> In testing, I found some blank shields, at the location `#12/36.87401/127.49871`
> 
> Not sure what's happening there?

It seems to be specific to [this route relation](https://www.openstreetmap.org/relation/8384009) that’s tagged `network=KR:national` but lacks a `ref`. The image ID is `KR:national=`, which `missingIconLoader()` parses into `KR:national` as `network` and the empty string as `ref`. That satisfies the following check, because the programmatically drawn ellipse has `backgroundDraw` defined:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/a54a346044b36f8a6f3a5706d02bcf9952f0a500/style/js/shield.js#L177-L180

The logic seems inverted to me, but it was apparently part of a fix for Allegheny County belt roads in 92d6c5bf965cdf068c73aa069f0c0f4bd41361a0 for #72.

ZeLonewolf (approved):
Looks great!  Thanks for fixing that blank shield bug.  Also confirmed that the Pittsburgh shields still work as expected.

PR #223, opened 2022-03-12
Shields of Bangladesh

1ec5:
Added route shields for national and regional highways in Bangladesh. District roads are not yet mapped as relations.

[Barishal](https://zelonewolf.github.io/openstreetmap-americana/#14/22.67995/90.35098)<br>_National and regional highways in Barishal._


ZeLonewolf (approved):
Looks good.  I only found one regional road mapped, let's hope that having a rendering will help encourage the local community to fill out the network.

PR #224, opened 2022-03-12
Shields of Nepal

1ec5:
Added route shields for national highways and feeder roads (regional roads) in Nepal.

[Kathmandu](https://zelonewolf.github.io/openstreetmap-americana/#13/27.69229/85.2804)<br>_National highways and feeder roads in Kathmandu._


ZeLonewolf (approved):
Looks great, thanks!

Issue #218, opened 2022-03-10
Alberta highway shields


TagaSanPedroAko:
Alberta provincial highways numbers should be rendered with these shields:

- Highways 1 and 16 (network=CA:transcanada): Maple leaf outline shield (would desire a simple maple leaf without green rectangle, if to take in account the slightly different design used by Newfoundland and Labrador, a green shield with a maple leaf, as against most provinces that use a green rectangle with the leaf)
- 1-216 series highways excluding Highways 1 and 16 (network=CA:AB:primary): squared-off white shield. Highways 3 (Crowsnest Highway), 35 (Mackenzie Highway) and 88 (officially Bicentennial Highway) use standard Alberta highway shield, but with extra elements that are under Crown copyright; most maps would defer to using the simple shield
- 500-986 series highway (network=CA:AB:secondary): oblong

TagaSanPedroAko:
Reference for standard highway shield shape: Alberta Highway Guide and Information Sign Manual (https://open.alberta.ca/dataset/d2890998-9341-4c3b-bebb-96e5103eb57a/resource/f823c311-b7ba-40db-9bb2-05339e83c038/download/guide-info.pdf), Section A4.4

TagaSanPedroAko:
@1ec5 @ZeLonewolf Demos with the desired shields?

1ec5:
It’s high time we embark on support for Canada’s route shields, so thanks for providing us with this information!

I think we’d definitely prefer the “guide sign route number shield” designs per [section A4.4.1](https://open.alberta.ca/dataset/d2890998-9341-4c3b-bebb-96e5103eb57a/resource/f823c311-b7ba-40db-9bb2-05339e83c038/download/guide-info.pdf#page=47), as well as just the maple leaf outline (with the banner below it?) for the Trans-Canada Highway. That’s the same approach we’ve been taking in various U.S. states for simplicity.

The [special trailblazers](https://open.alberta.ca/dataset/d2890998-9341-4c3b-bebb-96e5103eb57a/resource/f823c311-b7ba-40db-9bb2-05339e83c038/download/guide-info.pdf#page=51) are intimidating (and copyrighted), but it sounds like they’re redundant to standard numbered shields, correct? Could we gain any usability by showing a small, simplified replica of one of these trailblazers?

[...]

PR #226, opened 2022-03-14
Trans-Canada Highway shield

1ec5:
Added a route shield for the Trans-Canada Highway based on @TagaSanPedroAko’s recommendation in https://github.com/ZeLonewolf/openstreetmap-americana/issues/218#issue-1164762399. The shield background is based very loosely on [this public domain image](https://commons.wikimedia.org/wiki/File:BC-1_%28TCH%29.svg) on Wikimedia Commons, with numerous adjustments to make the shape better accommodate an inscribed number.

[Portage La Prairie](https://zelonewolf.github.io/openstreetmap-americana/#8.54/50.0033/-98.3254)

[Ontario inscribes road names on the TCH shield](https://commons.wikimedia.org/wiki/File:TCH-COR.svg), similar to the Palisades Interstate Parkway shields in New York State. However, this PR keeps the named TCH shields blank because they only supplement the standard provincial route shields. It’s similar to how we’d treat some state scenic route shields like California’s. Unfortunately, OpenMapTiles sorts the TCH shield before the provincial route shields, probably because `CA:transcanada:namedRoute` sorts before `CA:ON:primary`.

[Sault Ste. Marie](https://zelonewolf.github.io/openstreetmap-americana/#8/46.37/-84.324)

[Nova Scotia puts the number on a rectangular plaque](https://commons.wikimedia.org/wiki/File:Nova_Scotia_Highway_104_%28TCH%29.svg) beneath the TCH shield, similar to how U.S. Bicycle Routes are signposted. For concision and consistency and expediency, the number has been inscribed onto the shield proper:

[Amherst](https://zelonewolf.github.io/openstreetmap-americana/#11/45.817/-64.2043)


ZeLonewolf (approved):
Looks good to me.  Note that I am also seeing the blank shield in Quebec, so if that's not expected, we should take a look at that.

TagaSanPedroAko:
That's also what I was looking to post here as well.

1ec5:
> Looks good to me. Note that I am also seeing the blank shield in Quebec, so if that's not expected, we should take a look at that.

Yes, it is expected. The TCH follows Autoroutes that have the standard red, white, and blue shield.

PR #227, opened 2022-03-14
Shields of Alberta

### LINKED ISSUE #218: Alberta highway shields

1ec5:
Added route shields for standard Alberta highway routes in the two- and three-digit series. Factored out the Philippines national highway shield from #195 as a general-purpose shield shaped like an inverted home plate (technically a straightened square escutcheon, for those who keep track of such things).

[Okotoks](https://zelonewolf.github.io/openstreetmap-americana/#13/50.69472/-113.88739)

Fixes #218. Depends on #226.


TagaSanPedroAko:
@1ec5 Well, you might consider renaming the Philippine national road shield if that should be treated as a generic design used across different countries:

- Argentina - national routes (rutas nacionales)
- Australia - national routes (network=NR), used in Queensland, Western Australia, Northern Territories (slowly being replaced by alphanumeric route numbers) as well as on 1 Alt in Melbourne.
- Brazil - federal highways (network=BR). Actual shield in use has some curves to the lower tip
- Mexico - state highways
- Venezuela - trunk highways (troncales)

For Canada, aside from Alberta, the same shield should be used on Manitoba provincial trunk highways as well (see #229). A slightly different version should be used for BC (shields used has rounded bottom)

1ec5:
> Well, you might consider renaming the Philippine national road shield if that should be treated as a generic design used across different countries:

Yes, this PR takes care of renaming it to “home down white”, which is the best, most descriptive name I could think of at a late hour. 😅 (Ironically, a “square” shield has curved lower edges.) You can see the rename if you click on the “Files changed” tab toward the top of this page.

1ec5:
Thanks for providing information about similar shields in other provinces and countries. I’d prefer to keep each PR scoped to an individual jurisdiction or highway system, adding thorough support for a single system at one time, to avoid the scope-creep that caused #139 to get blocked and then stall. But we can track those similar shields in separate issues.

ZeLonewolf (approved):
Looks good both in Alberta and Philippines.  Also +1 for calling it a "home plate" :-D

PR #228, opened 2022-03-14
Add shield for Indiana East-West Toll Road

jleedev:
Applying essentially the same treatment as the NYS Thruway, we remove the text and enlarge the state.

image


1ec5 (approved):
Thanks!

PR #231, opened 2022-03-15
Explicitly declare Montana as a plain shield

Pengor:
Explicitly adds the plain white rounded rectangle shield for Montana state routes.

From the Montana DOT's Sign Catalog:
image

I don't think there's a way to legibly convey the `MONTANA` wording on the shield at the scale used in _Americana_, so I'm proposing we use the plain white rounded rectangle shield for Montana state routes.

### DIFF
--- a/doc-img/shield_map_us.svg
+++ b/doc-img/shield_map_us.svg
@@ -14,7 +14,7 @@
 .ca,.gu {fill:#FF0000}
 In this example, Kansas, Montana and Pennsylvania are colored blue, and California and Guam are colored red.
 Place this code in the empty space below. */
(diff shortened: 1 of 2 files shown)


1ec5 (approved):
Yes, a simple white rectangle is perfectly reasonable and consistent with how we’ve treated similar shields like Indiana’s.

Montana also has a distinctive shield for its secondary highways. Do you intend to introduce it here as well? If not, please open an issue for the secondary shields so we don’t forget. Thanks!

Pengor:
Ah, you beat me to it, thanks! :grin:

PR #234, opened 2022-03-17
Add Montana's secondary highway shield

### LINKED ISSUE #233: Add shield for Montana secondary state highways
As a followup to the Montana primary state highways in #231, we should add a sprite and shield definition for Montana’s secondary state highways, which are tagged [`network=US:MT:secondary`](https://wiki.openstreetmap.org/wiki/Tag:network%3DUS:MT:secondary) and have a distinctive arrowhead shape:

[S-236](https://commons.wikimedia.org/wiki/File:MT-sec-236.svg)

/cc @Pengor

Pengor:
Fixes #233

1x display:
image

2x display:
image


1ec5:
Thanks for contributing this shield! Do you think the bumps along the edges of the arrowhead are significant enough that we should make them more pronounced and legible? Or should we consider smoothing (er, sanding down) those edges?

Pengor:
Good question, I just tried to keep the proportions as close as I could to the physical shield (other than flattening the top, because I figured the bumps would like weird at this scale). I could really go either way, but at this scale it might make more sense to smooth the bumps rather than accentuate them. It's pretty hard to see the bumps at all in 1x display screenshot above. Also, it's been a long time since I've been to Montana so I'm certainly not a local expert (so I don't exactly have any particular attachment to this shield) :thinking:

1ec5:
I don’t feel strongly about how smooth this arrowhead has been cut; how you have it is OK.

The arrowhead’s proportions do look accurate, but it ends up being a suboptimal shape for the numerals inside. Secondary highway numbers are always three digits long, so I think this might end up being the smallest frequently occurring font size of any shield in the style so far. Some ideas for making it better accommodate the number:

* Widen the whole shape by 5&nbsp;pixels.
* Reduce the height of the top portion above the number by a couple pixels.
* Make both sides bow out a couple pixels more.

It’s basically an exercise in seeing how much we can get away with before the arrowhead stops being recognizable – which might not be very much.

Pengor:
Shield graphic has been enlarged:
1x display:
image

2x display:
image

1ec5 (approved):
Much more legible, especially in 1×. Thanks again!

PR #236, opened 2022-03-18
Fixed reference to wrong filename for Montana secondary shield

Pengor:
I was working with too many files and left an intermediate filename in the copy of shield_defs.js that I pushed. This PR puts the proper filename back.


1ec5 (approved):
Oh, good catch. This is probably something we should add automated test for, but I’m not sure how with the current JavaScript-based structure. Another point in favor of #118.

PR #238, opened 2022-03-19
Add shield for Wyoming

Pengor:
Adds a black-outlined rounded yellow square as the highway shield for Wyoming state routes.

[This is the real world appearance](https://commons.wikimedia.org/wiki/File:Wyoming_Route_59.JPG), but for display purposes (and due to the copyright status for the [Bucking Horse and Rider](https://en.wikipedia.org/wiki/Bucking_Horse_and_Rider) emblem) I did not include any additional graphical elements.

It's possible this shield could be generated without the svg if the `roundedRectShield` function was tweaked (I tried it out but when setting a fixed width of 20 there was too much padding added for the ref numbers), but I didn't want to dive into trying to improve upon that because it is used in so many places.

Here's a render on a 1x display:
image

And at 2x:
image


1ec5:
> It's possible this shield could be generated without the svg if the roundedRectShield function was tweaked (I tried it out but when setting a fixed width of 20 there was too much padding added for the ref numbers), but I didn't want to dive into trying to improve upon that because it is used in so many places.

The width doesn’t need to be fixed. A lot of maps depict shields that vary in width (to accommodate numbers cleanly), even if the physical signs are at fixed widths (to save on manufacturing costs). It’s true that `roundedRectShield` currently makes the shields look rather large, but it’d be easier to fix that for all the rectangular shields at the same time than to manually update each SVG to be consistent.

ZeLonewolf:
I would agree with just making this a drawn rounded rectangle with the appropriate yellow/black colors.  We've played fast and loose with shield aspect ratios to prioritize readability on a digital map, rather than trying to strictly conform to on-the-ground shields.

ZeLonewolf:
I propose that we re-open #242 as the implementation for Wyoming shields.

Pengor:
Maybe it's just me but the square shield seems slightly more recognizable as the Wyoming shield than the short/wide rectangle 🤷‍♂️

ZeLonewolf:
Closed in favor of the approach in #242

PR #243, opened 2022-03-19
Fix corrupted shields

ZeLonewolf:
Sometimes this happens:

image

This occurred in cases where there was something wrong with the `network`+`ref` value, and we just returned null to the mapLibre function that requested missing graphics.  It turns out this was a Very Bad Idea™ and caused the bad graphics shown, in addition to filling the console with cryptic errors.  Instead, this PR ensures that the missing graphics function always returns a 1x1 pixel transparent image if the network is unhandled.

This results in no more corruption:

image

Yay!


1ec5 (commented):
The [`styleimagemissing`](https://docs.mapbox.com/mapbox-gl-js/api/map/#map.event:styleimagemissing) event is designed to give the handler the choice of whether to call `addImage` or not. If the glitching happens because the layer refers to a missing image, then a more robust fix would be to modify the `text-field` property to wrap the `image`  expression in a `has` or `coalesce` expression.

ZeLonewolf:
> a more robust fix would be to modify the text-field property to wrap the image expression in a has or coalesce expression

Hmm, I'm gonna need you to dumb this down for me...

1ec5:
This expression takes the image name and converts it to an image to place inside the format expression:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/1e2a661785a33a5047cc26ed8f55be43cb349e3e/style/layer/highway_shield.js#L10

Wrap it in a `coalesce` expression so that, if the style has no image by the given name, it inserts nothing into the format expression.

ZeLonewolf:
How would I express "if the style has no image by the given name" ?  I don't see how you could know that a priori.

1ec5:
The existing [`image`](https://maplibre.org/maplibre-gl-js-docs/style-spec/expressions/#types-image) expression will already evaluate to `null` if the image by the name of its first argument is absent from the style at the time of evaluation.

This is a synchronous process, so it assumes that `addImage` has already been called, if necessary, by the `styleimagemissing` handler if the image wasn’t already part of the style. If the image doesn’t even get added by that event handler, then it resolves to `null`. There’s probably a bug in the `format` expression that a `null` image shows up as glitching instead of being ignored, but we can suppress that glitching by with a `coalesce` expression that replaces the `null` with an empty string.

I haven’t tried this myself, just going off of memory here.

ZeLonewolf:
I seem to be striking out trying to find a combination that works here.  My most recent attempt:

```
function routeConcurrency(num) {
  return [
    "case",
    ["!=", ["get", "route_" + num], null],
    [
      "coalesce",
      [
        "image",
        [
          "concat",
          "shield\n",
          ["get", "route_" + num],
          [
            "match",
            ["get", "route_" + num],
            "US:KY:Parkway=",
            ["concat", "\n", ["get", "name"]],
            "",
          ],
        ],
      ],
      ["literal", ""],
    ],
    ["literal", ""],
  ];
}
```

ZeLonewolf:
I'm no longer seeing the shield corruption on the main branch which has a production version of maplibre running.  Can someone else confirm?  If so, we can abandon this fix.

jleedev:
Negative:
Screenshot_20220415-205655

ZeLonewolf:
Oh wait I'm still seeing it too 😅

1ec5:
> I seem to be striking out trying to find a combination that works here.

I finally got around to trying out my suggestion. Newflash: it doesn’t work. 😅

The `styleimagemissing` event is fired by `ImageManager` long after style evaluation, including after `coalesce` validates its arguments for image availability. `coalesce` [keeps track of the first requested image](https://github.com/maplibre/maplibre-gl-js/blob/9e6d57aaf7f77161c3d2f9dfd330a16227b60f19/src/style-spec/expression/definitions/coalesce.ts#L59-L60) so that it can be passed into the `styleimagemissing` event. If the event handler comes back empty-handed, all that’s left to do is to [issue a warning](https://github.com/maplibre/maplibre-gl-js/blob/224a8c86c8bb3b9abe2a95193b840e710eeba3b6/src/render/image_manager.ts#L201), but there’s no opportunity to go back and change how the style is evaluated.

I think the spacer image approach you’ve taken is the only way to make the fallback work as long as we keep adding static image data to the style on `styleimagemissing`. The downside is that we need to add a spacer image for each individual route relation we have no intention of rendering. If this turns out to be a performance problem, we should consider conditionalizing the `image` expression on whether the `route_*` contains a recognized network and short-circuiting the image stuff entirely if it does not. `match` is very performant for that purpose. We could generate a `match` expression based on the keys in the object returned by `loadShields()`, similar to what we’re already doing for the coverage maps and taginfo.

style/js/shield.js
```diff
@@ -245,8 +247,11 @@ export function missingIconLoader(map, e) {
     ctx = colorCtx;
   }
 
-  var imgData = ctx.getImageData(0, 0, ctx.canvas.width, ctx.canvas.height);
+  uploadShieldToSpriteSheet(map, id, ctx);
```
1ec5, line 250: Let’s refactor this method to avoid early returns. Declare `imgData` at the top. If the conditions aren’t met, generate the blank image here. We’d end up calling `getImageData()` and `addImage()` no matter what.

ZeLonewolf: Thanks!  This refactor is a significant improvement in readability.

style/js/shield.js
```diff
@@ -219,13 +233,11 @@ export function missingIconLoader(map, e) {
...
   var wayName = id.split("\n")[2];
 
-  var colorLighten = ShieldDef.shieldLighten(network, ref);
-
   var ctx = drawShield(network, ref, wayName);
 
   if (ctx == null) {
     //Does not meet the criteria to draw a shield
-    return;
+    return ShieldDraw.blank();
```
1ec5, line 240: This is cleaner, but there are still early returns and an unnecessary helper function that obscures what all these code paths share in common. For future reference, what I had in mind was getting rid of early returns along these lines:

```js
function getShieldCtx(id) {
  var ctx;
  if (id !== "shield_") {
    var network_ref = …
    ctx = drawShield(network, ref, wayName);
  }
  if (ctx) {
    drawBanners(ctx, network);
    var colorLighten = ShieldDef.shieldLighten(network, ref);
    if (colorLighten) {
      let colorCtx = Gfx.getGfxContext(ctx.canvas);
      …
      ctx = colorCtx;
    }
  }
  return ctx || Gfx.getGfxContext({ width: 1, height: 1 });
}
```

This flow reduces repetition for the fallback cases without a trivial helper function. It also makes it clear that the colorLighten stuff is a better candidate for factoring out. However, this is very much in the realm of a nitpick, so unless we anticipate more edge cases and more early returns down the line, this is good enough to merge.

1ec5 (approved):

PR #245, opened 2022-03-20
Add shield for Southern Tier Expressway

jleedev:
For reference, here's the original vector from wikimedia:

Screenshot 2022-03-19 23 15 44
Screenshot 2022-03-19 23 16 25

Here's the simplified version in this branch:

Screenshot 2022-03-20 13 47 09
Screenshot 2022-03-20 13 46 49


ZeLonewolf:
Simplifying the text down to STE is good.  What do you think about also stripping out the header text and bib?

jleedev:
Simplified and cut out the shape:

Note that that's not text, it's a pattern, but replacing it with a stripe seems to work.
Screenshot 2022-03-20 19 44 18
Screenshot 2022-03-20 19 44 37
Screenshot 2022-03-20 19 44 54
Screenshot 2022-03-20 19 45 12

ZeLonewolf (approved):
Looks pretty good to me.  Could use a white stroke around the perimeter like the Interstate shield for visual separation.

jleedev:
Screenshot 2022-03-21 08 42 41
Screenshot 2022-03-21 08 42 27

Issue #248, opened 2022-03-21
Colorado route E-470


Pengor:
Colorado's toll route E-470 has a separate shield than the state DOT's numbered highways. As such, its `network` tag needs to be updated and a separate shield should be added to this project for that network.

See also discussion on OSM US Slack Workspace [here](https://osmus.slack.com/archives/C01V02K52UX/p1647884121104639).

@amcginni has drawn a shield for this.

amcginni:
https://www.openstreetmap.org/changeset/118826107 changes the `network` tag to `US:CO:E470`.  This name was chosen because the name of the toll road operator is "E-470 Public Highway Authority."  The dash in the `ref` tag was also removed to match the signs.

I am going to hold off on submitting a PR for the E470 shields until #249 is finalized.  I think it would look good if the logo in those shields was the same height as the Colorado flag in the state route shields.  For example, the following mock-up uses the same vertical proportions for the E470 shield as the Colorado shield.  I want to replicate this effect.
![image](https://user-images.githubusercontent.com/46615954/159736207-6f9799a3-ba95-44df-9490-78d86b776ef7.png)

ZeLonewolf:
Another option to consider is simply a square shield with just the mountain icon.  That would take some artistic liberty but might be nicer looking than the really wide version.

1ec5:
Would the mountain be recognizable enough on its own? Some people may depend on the number enough that they may mistake SH&nbsp;470 for E-470.

ZeLonewolf:
Yeah I have no idea which portion of the shield is more important.  The shield gives basically equal weight to either part, and the graphic above drops the mountain down to unrecognizable.

![image](https://user-images.githubusercontent.com/3254090/159837187-50f271ee-185e-4bd3-b49c-f5e910c10c0c.png)

PR #249, opened 2022-03-21
Add highway shield for Colorado state routes

Pengor:
Adds a highway shield graphic and definition for Colorado state routes. 

Note that with current tagging Colorado's E-470 shows up with this shield, but it is not part of the same highway network. See #248 (this PR does not close that issue).

1x display:
image

2x display:
image

### DIFF
--- a/doc-img/shield_map_us.svg
+++ b/doc-img/shield_map_us.svg
@@ -17,6 +17,7 @@ Place this code in the empty space below. */
 .us,
 .az,
 .ca,
+.co,
 .ct,
 .de,
 .ga,
--- a/style/icons/shield40_us_co.svg
+++ b/style/icons/shield40_us_co.svg
@@ -5,6 +5,7 @@
  <path fill="#003f87" d="M1 1H19V3.5H1z"/>
  <path d="m0.5 18.5v-17a1 1 135 0 1 1-1h17a1 1 45 0 1 1 1v17a1 1 135 0 1-1 1h-17a1 1 45 0 1-1-1z" fill="none" stroke="#000"/>
  <path d="m1 9.5h18" stroke="#000"/>
+ <path fill="#ffcd00" d="M3 3H6.5V7H3z"/>
  <circle cx="5.25" cy="5" r="1.25" fill="#ffcd00" stroke="#ffcd00"/>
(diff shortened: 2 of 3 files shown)


1ec5 (commented):
Very nice, it looks like the real deal and it’s stylish. Have you considered smooshing the Colorado flag a little, perhaps by allowing the C to touch the top and bottom of the flag? Removing the black line between the flag and the bottom part would also leave a little more precious room for the route number.

Pengor:
I tried to keep things as closely proportional to the real world shield as possible with this attempt. Also I did discover one issue with how this SVG renders at 2x (gap between the red C and the yellow circle from anti-aliasing?) that I will push a fix for shortly (top is before, bottom is after):
image

That said, I did look in to you suggestions. However, it would seem that in order to get all the three digit routes to increase in font size the primary constraint is the width of the shield. I quite like the similarity to the real world shield that this SVG has and it seems readable to me (I have use a 23" 1080p monitor), but I can explore widening and shrinking the flag area if it's deemed necessary.

amcginni:
[Wikipedia](https://en.wikipedia.org/wiki/List_of_state_highways_in_Colorado) uses wider shields for the three-digit Colorado state routes.  I don't know if that's what CODOT does on the ground, but it probably doesn't matter for the purposes of this PR.  We have made more drastic deviations from some of the state shields for readability already.  So, if font size is constrained by shield width, a wider shield for thee-digit routes is an option.

I do not have an HD monitor.  The numbers in the 1x rendering are bordering the "legible but hard to read" territory for me.  If we could smoosh the flag by about four pixels, I think that would help.

Pengor:
On the ground only one shield width is used (and it's square). However, I think I've come up with a good compromise that only slightly elongates the shield and adds an extra pixel of height (by removing the center black line) to the text area of the shield so that we achieve the minimum font height listed in the project's shield guidelines (8 pixels).

Commit forthcoming, but here's a preview at 1x:
image

And 2x:
image

ZeLonewolf:
Haven't had the chance to run the code yet but the screen shots look great.

ZeLonewolf:
Seems like a reasonable approach to get these on the map  Definitely pushing the lower limit of text legibility but it seems workable and definitely captures the essence of the shield.  It might be possible to squeeze a couple more pixels of vertical out of the banner if we play with the graphics but let's leave that for future exploration.  This is a slightly different approach than I took for the South Carolina shield, where I removed the South Carolina state outline from its banner graphics.  At some point someone might examine whether to homogenize that approach but let's get this on the map as a solid starting point.

ZeLonewolf (approved):

1ec5:
This shield wouldn’t’ve been recognizable without something resembling the state flag, anyhow. In the case of South Carolina, I think it’s at least somewhat recognizable without the state shape at the top, though we probably could stick a blue triangle up there without sacrificing legibility.

PR #250, opened 2022-03-21
add shield for US Virgin Islands

### LINKED ISSUE #237: US Virgin Islands highway shields
`network=US:VI`. Shield is the MUTCD standard state route (black text on a white oval). Route numbers can have up to 4 digits.

Identical to `network=US:KY`.

claysmalley:
fixes #237

### DIFF
--- a/style/js/shield_defs.js
+++ b/style/js/shield_defs.js
@@ -821,6 +821,8 @@ export function loadShields(shieldImages) {
     null
   );
 
+  shields["US:VI"] = circleShield("white", "black");
+
   shields["US:VT"] = {
     backgroundImage: shieldImages.shield40_us_vt,
     textColor: "#006b54",
--- a/doc-img/shield_map_us.svg
+++ b/doc-img/shield_map_us.svg
@@ -45,6 +45,7 @@ Place this code in the empty space below. */
 .sc,
 .tx,
 .va,
+.vi,
 .vt,
 .wa,
 .wv { fill: #8250df; }


style/js/shield_defs.js
```diff
@@ -821,6 +821,8 @@ export function loadShields(shieldImages) {
     null
   );
 
+  shields["US:VI"] = circleShield("white", "black");
```
1ec5, line 824: USVI highways are apparently numbered [into the thousands](https://en.wikipedia.org/wiki/List_of_United_States_Virgin_Islands_highways), so we should make sure `circleShield` can accommodate four-digit numbers. If it looks too pointy, `roundedRectShield` would be an alternative, though it would result in a Kentucky-style elongated circle instead of an ellipse for any three- to four-digit shield.

claysmalley: I think it looks alright. Why is Kentucky using `roundedRectShield` instead?

Screenshot from 2022-03-23 22-25-03

ZeLonewolf: We used rounded rects in Kentucky because that's how they're shaped:
image

ZeLonewolf: Meanwhile in Delaware they use ellipses on their wider shields:
image

1ec5: Kentucky’s shields are elongated circles (rounded rectangles) in reality, and four-digit route numbers are very common there, so the rounded rectangles would be familiar. That isn’t the case in every state that uses circular shields, so it’s fine to use the ellipses here.

ZeLonewolf: Since I'm a super math nerd, I implemented drawing options for both form factors :-D

1ec5 (approved):

style/js/shield_defs.js
1ec5: USVI highways apparently use narrower text, keeping a strict circle, as Iowa does. So an ellipse is a reasonable compromise for legibility on this map.

USVI 104

PR #242, opened 2022-03-19
Wyoming shields

amcginni:
This PR adds support for state route shields in Wyoming.  Wyoming uses square shields with a yellow background, like the one shown below.
WY-59
This PR omits the Bucking Horse and 'WYOMING' header text, instead opting for a plain rectangle shield.  Examples are shown below.
image
image

PR #251, opened 2022-03-24
add Wyoming shields

claysmalley:
Adds Wyoming shields, represented as black text on a yellow rectangular shield. The word "WYOMING" and the cowboy illustration are removed.

Screenshot from 2022-03-24 08-08-23


claysmalley:
I'm an idiot and didn't look at the list of open pull requests: #242

PR #253, opened 2022-03-24
added Alaska shields

claysmalley:
Here's a first attempt at an Alaska shield. The word "ALASKA" and the North Star are removed, while the Big Dipper is retained and slightly rotated. Numbers are intentionally off-center, as present on signage.

Spacing may not be optimal—feedback is welcome.

Shield dimensions were copied from the Idaho shield (#129, thanks @Pengor).

Screenshot from 2022-03-24 10-53-14
Screenshot from 2022-03-24 10-53-51


style/js/shield_defs.js
```diff
@@ -206,6 +206,17 @@ export function loadShields(shieldImages) {
     colorLighten: "#613214",
   };
 
+  shields["US:AK"] = {
+    backgroundImage: shieldImages.shield40_us_ak_2,
```
Pengor, line 210: I'd suggest removing the `_2` from the SVG filename to make it consistent with other shields that have only a single shield

Pengor:
Oh one last thing, this PR should also update `/doc-img/shield_map_us.svg` to include Alaska. You can generate the change by running `npm run build`

ZeLonewolf:
Is the route number supposed to be right-justified?  If so, we can probably add an option in the shield definition structure to support that.  Currently all text layout uses centered text but it doesn't seem like it would be that hard to make a right-justified version.

claysmalley:
It's not right-justified. It's centered on a point, but that point isn't in the center of the shield.

ZeLonewolf:
> It's not right-justified. It's centered on a point, but that point isn't in the center of the shield.

Sure, that's how it's implemented here, but is that the intended layout, or is that a hack because we don't have a right-justified text option?

Looking at https://en.wikipedia.org/wiki/List_of_Alaska_Routes it's not totally clear whether it's right-justified or offset-centered.

claysmalley:
It's clear to me—the single-digit routes have significantly more whitespace to their right than the double-digit routes, except for "11" which predictably has as much whitespace to its right as a single digit.

Pengor (approved):
Looks good to me!

ZeLonewolf (approved):
Thanks!

PR #255, opened 2022-03-26
add Northwest Territories shields

claysmalley:
Adds shields for Northwest Territories, Canada. This territory has a polar bear on every highway shield, reminding you of the presence of these adorable, intimidating animals.

Screenshot from 2022-03-25 20-27-40

The official shield looks like this. The tall rectangle has been shortened to a square, the text "NWT" has been removed, and the polar bear has been shrunk to accommodate larger text. The polar bear design is trademarked but not copyrighted, and the design used in this pull request has different geometry.


ZeLonewolf (approved):
Adorable.  Since the routes max out at 10, I guess we don't have to worry about wide shields.

PR #256, opened 2022-03-26
Add state route highway shields for South Dakota

Pengor:
Adds 2-digit and 3-digit highway shields for South Dakota state routes. The 3-digit shield also covers the two four digit South Dakota routes (which both have 1s in them thankfully).

### Previews for 1x display:
image
image

### And 2x display:
image
image


ZeLonewolf:
These look pretty good, but I wonder if it might make sense to square the top and left side of the state shape to get a cleaner line.  I seem to recall @1ec5 describing that this was done for some version of Ohio state routes.  I'm questioning whether we gain much visually from the exact adherence to a slightly off-horizontal or off-vertical line versus bringing it to perfectly horizontal/vertical.

Pengor:
Yeah, that did cross my mind. I generated the outline from data from the South Dakota DOT so this is more of less how the outline came. I did try experimenting with shifting the perspective of the path but I'm still somewhat of an Inkscape novice and wasn't happy with the results, so I left it as-is...

1ec5:
To straighten out the northern edge in Inkscape, you can double-click the path to show the control points and delete any point that’s extraneous. If it results in a wild curve at the corners, select the points at the corners and click the button in the toolbar that makes the path have a corner instead of a curve at that point.

To align the western edge, turn on the grid as mentioned in CONTRIBUTING.md, then turn on snapping to the grid. Then you can drag the northwestern and southwestern points to be vertically aligned. Alternatively, select each point and manually set its Y value to the same value.

Pengor:
Okay, how's this look?

1x:
image
image

2x:
image
image

claysmalley:
Looks great!

I agree with the decision to straighten the top and bottom lines. Every other state-shaped shield design seems to be based off of the Mercator projection, straightening the lines of latitude that many state lines fall on. South Dakota seems to be the odd one out on this. The curved lines may be truer to how the shields are in real life, but we're aiming for consistency and the shield is just as recognizable either way.

ZeLonewolf (approved):
Looks great!  Congrats on your newfound Inkscape skills.

Pengor:
Yeah, that did cross my mind. I generated the outline from data from the
South Dakota DOT so this is more of less how the outline came. I did try
experimenting with shifting the perspective of the path but I'm still
somewhat of an Inkscape novice and wasn't happy with the results, so I left
it as-is...

Issue #239, opened 2022-03-19
Guam and Northern Mariana Islands shields


claysmalley:
Guam and the Northern Mariana Islands each have only one route network and one shield design. In both cases, `ref` values may have up to 3 characters.

Guam shield (`network=US:GU`):

![Guam Route 1 shield](https://upload.wikimedia.org/wikipedia/commons/6/65/Guam_Route_1.svg)

NMI shield (`network=US:MP`):

![NMI Route 30 shield](https://upload.wikimedia.org/wikipedia/commons/d/df/Northern_Mariana_Islands_30.svg)

PR #257, opened 2022-03-26
added Guam and Northern Mariana Islands shields

claysmalley:
Adds shields for Guam and the Northern Mariana Islands. Both territories have quite narrow shields, but route numbers can have up to 3 digits, presenting a design challenge.

Guam's shield has white text on a blue background in the shape of a vertically-oriented canoe (or American football). A few highways have letter suffixes, forcing 3 characters into a narrow space. In the interest of legibility, the canoe shape is widened for 2- and 3-digit highways. The letter-suffixed highways still have quite small text compared to shields elsewhere—this sort of shape may necessitate relaxing the 20-pixel height guidelines.

Screenshot from 2022-03-26 16-23-22
Guam_Route_8
Guam_Route_30A

NMI's shield has black text on a white background in the shape of a [latte stone.](https://en.wikipedia.org/wiki/Latte_stone) Real latte stones are taller and thinner than the ones on the highway shields, but the general shape is still recognizable with a bowl-shaped top and a slightly conical base. NMI has a wider variant for 3-digit highways.

Screenshot from 2022-03-26 16-22-35
Northern_Mariana_Islands_30
Northern_Mariana_Islands_304


claysmalley:
fixes #239

1ec5:
Both these shields look pretty good. The NMI shield works well with this gentler inward curve. The Guamanian canoe doesn’t look as pointy as it could at two and three digits. Maybe the curve on either side could be straighter to keep the tips legible? Since the shield is vertically oriented in real life, we could just make it taller than 20 pixels. There’s already some precedent for that.

claysmalley:
Guam shields enlarged to a height of 24 pixels. I think this makes it visually a similar size to most 20px-height shields.

2-digit routes now have the thinner variant (except for routes like 3A where one character is a letter). This gives the majority of routes on Guam a shield with the original dimensions.

Screenshot from 2022-03-27 11-26-42
Screenshot from 2022-03-27 11-26-55

1ec5 (approved):

PR #258, opened 2022-03-28
added American Samoa shields

### LINKED ISSUE #241: American Samoa shields
American Samoa has one highway network, tagged with `network=US:AS`. Shield is white text on a hut shape with a black outline and green background. All `ref` values have 3 digits, no less. The words "AMERICAN SAMOA" can probably be omitted while staying recognizable.

Looks like there's two slightly different variants on the hut shape. One has a completely circular roof, and the other has a point in the middle of the roof (gothic window shape). My hunch is that the circular roof is the newer design.

claysmalley:
Fixes #241 

This design removes the text "AMERICAN SAMOA" and shortens the bottom line to be flush with the base of the hut.

Screenshot from 2022-03-27 20-29-02


1ec5 (approved):
Thanks! Yes, I agree that hanging horizontal lines at the base would’ve been awkward (and possibly confusable with a road).

PR #259, opened 2022-03-29
add Utah shields

claysmalley:
Beehive shields for the Beehive State!

Screenshot from 2022-03-28 21-15-54
Screenshot from 2022-03-28 21-17-15

This shield design stays mostly true to the dimensions of the 2-digit and 3-digit variants, with some changes:

* the bottom line has been moved up
* the notches on either side have been removed
* the hole has been shortened
* one stroke across the top of the beehive has been removed, and the remaining one has been flattened

Utah_24
Utah_222


ZeLonewolf:
I just want to address that there is no truth to the rumor that the Utah state shield is modeled after 💩.

ZeLonewolf:
I found this online, it looks like maybe there is a newer style of shield that further reduces detail?  If so, perhaps we can dispense with the little knob on the top.
image

mvexel:
> I found this online, it looks like maybe there is a newer style of shield that further reduces detail? If so, perhaps we can dispense with the little knob on the top. image

I think this is actually the older style. If it's important enough I can reach out to UDOT to confirm, or get an 'official' SVG.

mvexel:
Their response:

> Currently UDOT uses two types of beehive shields. The one with the ridges and a more smoothed one.  The one with the ridges is what is out in the field. The smoothed one is used more for graphic/paper maps.

ZeLonewolf:
Well clearly the smoothed on is in the field also, so they're full of 💩 :-D

1ec5:
I’m confused. Which one is smoothed? They both look [`smoothness=very_horrible`](https://wiki.openstreetmap.org/wiki/Tag:smoothness%3Dvery_horrible) to me.

ZeLonewolf (approved):
Just tested this, looks good to me.  Are we ok to merge, or should I leave this open a bit longer so we can further make fun of Utah's funny-looking shields? 😂

PR #261, opened 2022-03-31
added Charlotte shield

claysmalley:
Charlotte has only one numbered route. Its shield has the same dimensions as the MUTCD county shield, with a white-on-green color scheme. Charlotte puts its city logo where the county name typically goes.

Screenshot from 2022-03-30 21-15-43
Charlotte_Route_4_shield


ZeLonewolf (approved):
This is why the Europeans make fun of us.

PR #262, opened 2022-03-31
Add North Dakota state highway shields

Pengor:
Adds the highway shields for North Dakota state routes.

### Design choices:
* Discarded `NORTH DAKOTA` lettering above state outline shape
* Straightened/aligned left (West) edge of state outline to the pixel grid
* Decreased the width of the state shape from the proportions on the shield so that it wasn't overly large in comparison to other shields

### 1x renders:
image
image

### 2x renders:
image
image


1ec5 (approved):
Thanks, looks great!

PR #263, opened 2022-03-31
add Nevada shields

claysmalley:
One more state shape down!

Nevada curiously does not have any single-digit routes. Nearly all routes have 3 digits, except for routes 28 and 88. Due to the challenge of fitting 3 digits into a narrow state shape, I went with a height of 24px and a width of 18px.

The Clark County Road 215 shield is conflated with a standard MUTCD county shield here. This PR does not attempt to reproduce the special Clark County shield.

Screenshot from 2022-03-31 17-58-39
Screenshot from 2022-03-31 17-59-08
Screenshot from 2022-03-31 17-59-58
Screenshot from 2022-03-31 18-00-19

The shields as they are IRL:

Nevada_28
Nevada_318
Clark_County_Route_215_NV

### DIFF
--- a/doc-img/shield_map_us.svg
+++ b/doc-img/shield_map_us.svg
@@ -43,6 +43,7 @@ Place this code in the empty space below. */
 .nh,
 .nj,
 .nm,
+.nv,
 .ny,
 .oh,
 .or,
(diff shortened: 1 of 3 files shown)


ZeLonewolf (approved):
A little shorter in one dimension and a little fatter in another dimension?  What the hell, why not.  Everything else in this country is non-standard which makes this shield standard.  Tested and looks good.

PR #264, opened 2022-04-01
add Nebraska shields

claysmalley:
_"I tried to ford the river and one of my oxen died"_

Remember the good ol' days of the Oregon Trail? Now all you have to do is zoom in on a map of Nebraska! These covered-wagon-inspired shields are sure to make you ooh, and awe, and catch dysentery.

Screenshot from 2022-04-01 08-25-30
Screenshot from 2022-04-01 08-26-02
Screenshot from 2022-04-01 08-26-33

The shields in real life are as follows. Artwork and the state name are removed, and the words "LINK", "SPUR", "RECREATION ROAD" and "BUSINESS" are moved above the shield as banners. The latter two are abbreviated to "REC" and "BUS", respectively.

N-2
N-370
N_LINK_55X
N_REC_88B
N_SPUR_55G

### DIFF
--- a/doc-img/shield_map_us.svg
+++ b/doc-img/shield_map_us.svg
@@ -40,6 +40,7 @@ Place this code in the empty space below. */
 .mt,
 .nc,
 .nd,
+.ne,
 .nh,
 .nj,
 .nm,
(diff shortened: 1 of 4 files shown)


ZeLonewolf (approved):
Congratulations on being the first person to beat Oregon Trail!

PR #265, opened 2022-04-04
add Louisiana shields

claysmalley:
One wildly distorted state and two cups of coffee later, I finally have something working for Louisiana. It's not optimal, but feedback is welcome.

Screenshot from 2022-04-04 10-00-31
Screenshot from 2022-04-04 10-01-59
Screenshot from 2022-04-04 10-01-04
Screenshot from 2022-04-04 10-01-27
Screenshot from 2022-04-04 10-08-18
Screenshot from 2022-04-04 10-07-55

This design removes the text "LA", significantly distorts the state shape and adds a wider variant for designations 3 characters or longer. This is a departure from reassurance shields (first image below), but is more consistent with the way they are displayed on guide signs (second image below).

Louisiana_3134_(2008)
la-hwy-shield


ZeLonewolf:
Thanks, I think the graphics look great!

Looking at the screenshots, the numbers look a little tight on the top with room to spare below then.  Would it look any better to shift the padding south by a pixel or half a pixel?

claysmalley:
If I squish the padding by half a pixel, all shields become the wider variant. So I just scooched part of the outline up half a pixel.

Screenshot from 2022-04-04 10-58-08
Screenshot from 2022-04-04 10-53-43

1ec5:
I wholeheartedly agree with following the guide sign markers instead of the standalone markers for this state. To that end, there are a couple things that can be cleaned up:

* The Mississippi River Delta retains too much of its shape. The guide signs suggest both Lake Pontchartrain and the delta by making the eastern edge of the shape bow slightly inwards and back outwards by a larger distance.
* The upper portion of the boot should be widened to the left and right, particularly for the two-digit variant. This extra width compensates for the decreased height compared to the much taller lower portion.

claysmalley:
Is this good, or should the sides bow in a little more?

Screenshot from 2022-04-04 17-41-40
Screenshot from 2022-04-04 17-41-26

1ec5:
I think that works at 2× resolution. It’s not quite how the guide signs and other stylized depictions have it, but it’s close enough to be suggestive. At 1×, the curves on the two-digit shield are too small in both directions to be discernible.

The upper part of the two-digit shield no longer looks out of scale, though the ankle is normally depicted taking up slightly more than half the width of the state (as you have it in the 3+-digit shield).

PR #203, opened 2022-02-26
Thicken road label halo

1ec5:
Doubled the thickness of the halo around road labels and changed them to an off-white color matching the background. These changes make the halo look more like a knockout effect than an outline around the text.

The previous treatment doesn’t provide enough contrast when the label overlaps a dark element. Since the black text on its own didn’t have enough contrast with the dark element beneath, it almost seemed like the text itself was a translucent white casing with no fill.

The halo blur isn’t strictly necessary anymore, but I left it in because it gives the halo a bit of sophistication versus a solid halo, which can look outdated.

6th before 6th after<br>_The road labels along [this freeway](https://zelonewolf.github.io/openstreetmap-americana/#15/39.1005/-84.54523) are legible against the constant red color of the freeway beneath._

Queensgate before Queensgate after<br>_The road labels at [this freeway junction](https://zelonewolf.github.io/openstreetmap-americana/#15/39.10045/-84.52528) are legible against the alternating red color of freeways crossing perpendicularly beneath the labels._

Mt. Airy before Mt. Airy after<br>_The off-white halo keeps the labels from drawing too much attention in and around [this park](https://zelonewolf.github.io/openstreetmap-americana/#14/39.1674/-84.58289)._

Lake Pontchartrain Causeway before Lake Pontchartrain Causeway after<br>_Labels for [road bridges over water](https://zelonewolf.github.io/openstreetmap-americana/#11/30.078/-90.1115) are more easily readable at a glance._

PR #266, opened 2022-04-04
Render ferry routes

Pengor:
### Adds the rendering of ferry routes starting at zoom level 11 (the lowest available)
The absence of them is particularly noticeable in places that heavily rely on ferries for transportation like Puget Sound. For styling I reused the `waterLabel` color definition to stick with the same palette. I think a navy blue color makes sense as it's a blend of the colors used for roads (black) and water (light blue). Dashed lines seemed like a pretty standard choice in denoting ferry routes and I went with something a bit bolder than the dashes used for boundaries to avoid confusion.

### Before:
image

### After:
image


ZeLonewolf:
Initially I thought the line style was a bit thick, but it works a lot better with the labels present.  I'd like to get some opinions from the other maintainers.

zekefarwell:
I think this looks generally pretty good.  Thanks for working on it!  The lines feel a bit heavy to me, making them stand out from the water color a bit too much.  Maybe a slightly thinner line and/or a lighter shade of blue would help with this.  I also think the labels would look better if they were closer to the styling of the other water labels with dark blue text matching the line color and a much more subtle halo.  The current halos feel like they have an overly high contrast to me.

jleedev:
Maptiler's transportation_name has `subclass=ferry` but Planetiler has `class=ferry`.

Pengor:
@zekefarwell maybe it's my bias from living in the Puget Sound area—where most of these ferry routes are technically state highways that would basically have a classification of `highway=primary` (and in my opinion some should even be connections to some `highway=trunk`)—that made me go for a heavier design here. But as there's no such tagging hierarchy for ferries, I'm open to suggestions that might better suit more regions!

I also did try out both a subtler halo and using the `waterLabel` color for the ferry labels, but it didn't look as good to me in the _Metropolis Light_ font at size 12. But here's a preview of what it would look like to more closely match the water styling, I think it looks good on its own, but I don't find it as legible where there are several routes tightly packed (I think this is where the larger halo helps):
image

I also think we would want to drop one of either the italics or the bold to help distinguish the ferry labels from water labels.

Pengor:
> Maptiler's transportation_name has subclass=ferry but Planetiler has class=ferry.

I didn't even see any documentation for using `ferry` as a value for `class` (in the `transportation` layer) and `subclass` (in the `transportation_name` layer) in the @OpenMapTiles [schema documentation](https://openmaptiles.org/schema) so I had to do some digging to find that these were even usable 🙁

Pengor:
>I also think we would want to drop one of either the italics or the bold to help distinguish the ferry labels from water labels.

Here's a style that might work well for the labels, it's pretty much halfway between the road label style and the water label style. It uses _Metropolis Regular Italic_ for the font, a slightly decreased halo width (1.5), a dark blue (same hue and saturation as `waterLabel` with 8% lower value) for the font color, and a slightly darker halo color (`#eee` instead of `#fff`):
image

I'm going to post this in Slack as well and commit it after hearing feedback ✏️

zekefarwell:
I think the choices you originally made for font, weight (not bold), and style (not italic) were good.  I was only suggesting an adjustment to one or more of the text color, halo color, and halo size.  Using exactly the same text treatment as the water labels as shown in [this comment](https://github.com/ZeLonewolf/openstreetmap-americana/pull/266#issuecomment-1089078955), is definitely not what I'm suggesting.  The last screenshot feels like a definite improvement to me.  To me the text still feels not quite blue enough and the halo feels a bit too strong.  Maybe going further would negatively impact readability though.  

Your perspective on the imprortance of these ferry routes, and why they should be prominent makes sense.  Since there aren't different classifications for ferry routes, though, we have to strike a balance that will work well for both important and unimportant ones.  I'll pull down this branch and look at some different areas tonight to get a better feel for the current line weight.

Pengor:
New York City is another good area to look at, with much higher density of ferry routes. I just noticed a couple of issues from looking there...

Overlapping routes cause what appears to be a solid line rather than a dashed one (but I'm not sure if this is fixable on the style side or if there are features that need to be merged in the data):
image

Body of water labels covered by ferry lines (I'll commit a fix shortly by reordering the layers):
image

Which begs another question: Should body of water labels or ferry route labels take priority? I think I'd lean toward giving priority to the body of water labels, as those serve as larger navigational aids (in my opinion) than an individual ferry route. At high zoom both should coexist fine.

msbarry:
> > Maptiler's transportation_name has `subclass=ferry` but Planetiler has `class=ferry`.
> 
> 
> 
> Pinging @msbarry to see if there might be a planetiler bug exposed here?

On `transportation` layer, class=ferry in both planetiler and maptiler. Shouldn't the class/subclass on `transportation_name` match the class/subclass on corresponding lines in `transportation`?

zekefarwell:
> Shouldn't the class/subclass on transportation_name match the class/subclass on corresponding lines in transportation?

That would be a sensible assumption, but currently in the tiles transportation has class=ferry while transportation_name has subclass=ferry for some reason.  Seems like this may be a mistake, but hard to say since [the schema docs](https://openmaptiles.org/schema) don't say anything about ferries at all.

image

zekefarwell (commented):
I pulled down this branch and messed around with the colors a bit.  I think these slight adjustments are an improvement.  Makes the ferry routes and labels feel a bit more cohesive with the water fill color.  What do you think?  Does it reduce ferry prominence too much? 

A couple of screenshots with the color tweaks:
image
image

And one without for comparison:
image

msbarry:
I could see maybe adding subclass=ferry to planetiler output in transportation and transportation_name layers in addition to class=ferry for compatibility but I wouldn't drop class=ferry since this seems like an openmaptiles bug. I think the intent should be for transportation and transportation_name to be in sync. Is this an issue for aerialways or railways too?

ZeLonewolf:
> I could see maybe adding subclass=ferry to planetiler output in transportation and transportation_name layers in addition to class=ferry for compatibility but I wouldn't drop class=ferry since this seems like an openmaptiles bug. I think the intent should be for transportation and transportation_name to be in sync. Is this an issue for aerialways or railways too?

I tend to agree that this sounds like a bug in OpenMapTiles, and I've opened openmaptiles/openmaptiles#1373 describing this issue.

As a workaround, @Pengor I think we need to modify this PR to match on either `class=ferry` **or** `subclass=ferry` which will ensure that we're compatible with either variant.  Please also add a comment in the code referencing openmaptiles/openmaptiles#1373 so that we can remove the double-match at some future time when this issue is corrected.

style/layer/ferry.js
```diff
@@ -0,0 +1,41 @@
...
+  "source-layer": "transportation",
+};
+
+export const ferryLabel = {
+  id: "ferry_label",
+  type: "symbol",
+  filter: ["all", ["==", "subclass", "ferry"]],
+  paint: {
+    "text-color": "hsl(211, 43%, 20%)",
+    "text-halo-color": "#eee",
```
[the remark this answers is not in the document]
Pengor: Updated in [1c0ad3b](https://github.com/ZeLonewolf/openstreetmap-americana/pull/266/commits/1c0ad3be30ae129d9789d5898b843852462d7a54)

style/layer/ferry.js
```diff
@@ -0,0 +1,41 @@
+"use strict";
+
+import { waterLabel } from "../constants/color";
+
+export const ferry = {
+  id: "ferry",
+  type: "line",
+  paint: {
+    "line-color": waterLabel,
```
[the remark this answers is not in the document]
Pengor: Updated in [1c0ad3b](https://github.com/ZeLonewolf/openstreetmap-americana/pull/266/commits/1c0ad3be30ae129d9789d5898b843852462d7a54)

Pengor:
Okay, after adding a few suggestions I think this is likely in good shape now. Colors now match @zekefarwell's proposal, I think they worked well ⛴️

Pengor:
Preview for Puget Sound:
image

And New York City:
image

style/layer/ferry.js
```diff
@@ -0,0 +1,43 @@
...
+
+export const ferryLabel = {
+  id: "ferry_label",
+  type: "symbol",
+  filter: ["any", ["==", "class", "ferry"], ["==", "subclass", "ferry"]],
+  paint: {
+    "text-color": "hsl(211, 53%, 15%)",
+    "text-halo-color": "hsl(211, 70%, 90%)",
+    "text-halo-blur": 0.5,
+    "text-halo-width": 1.5,
```
1ec5, line 30: If we can be pretty sure these labels will always be over water, do they need halos? The dark blue text probably has enough contrast against the light blue water, just as the ferry lines themselves do.

Pengor: I initially thought the same thing, but without the halo the text becomes hard to read in places where there are several dense ferry routes and some text may be rendered on top of other route lines

zekefarwell: Example with no halo looking janky on top of nearby lines
image

Setting halo color the same as water fill color produces a knockout effect that looks better.  Looks kind of odd to have a halo on the water labels but not to have one on the ferry labels though.
image

Knockouts for both label types could be an option.  Gives a bit of a different look though.
image

zekefarwell: Except the faux knockout effect breaks down when a water label ends up partially over land.  So that's not really an option.
image

1ec5: > Setting halo color the same as water fill color produces a knockout effect that looks better. Looks kind of odd to have a halo on the water labels but not to have one on the ferry labels though.

This knockout effect is much less distracting; thanks for trying it out. I agree that it’s inconsistent to keep the halo on the water body labels, but those labels very often overlap non–water bodies, as you’ve seen. Also, I think the water body would tend to be more notable than the ferry routes that traverse it, so the added prominence is justifiable.

zekefarwell: I guess halos on the water labels and knockouts on the ferry labels isn't as weird as I initally thought and your right that they are less distracting.  I think the halos look good too, though.  Can we be sure these labels will rarely be over land?  I've found a couple of cases where they do go over a bit of land.  For example:
image
It's not terribly noticeable, but might be more of an issue if a ferry route goes close to the shore for longer.  The halos avoid this.

1ec5: What if we have these knockouts fade like the halos? Then it wouldn’t look like the label took a bite out of the peninsula, or at least it would be less decisive in that regard.

zekefarwell (approved):
This is all looking good and ready to merge to me.  Also fine with a bit more discussion if others feel more tweaks are in order.

style/layer/ferry.js
zekefarwell: It's tricky.  Maplibre doesn't seem to leave much space around the letters for the halo to fade.  It quickly starts looking blocky at about 2px.  For the faux knockout I had the halo width set to 3px which looks basically like a rectangle if over a different color but over the same color you don't see it.  reducing the width and increasing the blur makes it look better over a different color, but makes the faux knockout much less effective at de-cluttering the lines behind the text. 🤷‍♂️

Pengor: Thanks for the renders @zekefarwell! I think it's fair to bring up accessibility considerations here too: dark blue on lighter blue (knockout) is less contrasting than dark blue on white-gray (halo)

1ec5: Halos can also impair legibility as much as a lack of contrast, especially when combined with tight letter spacing and italics. The halos pictured above are more legible than what we fixed in #203; however, if contrast is an issue, then it might be more effective to make the text black, as seen on some paper maps that nonetheless have blue ferry lines.

ZeLonewolf: > I guess we should be ok with the opposite (label taking a bite out of land) as well

This approach seems sensible to me!

ZeLonewolf: Clear as mud! 

In general I would avoid translucency as layering can be pretty unexpected.  Are we still playing with the knockout halo width?

Pengor: I'm happy with it as-is, but if other folks want it tweaked a bit feel free to propose new values or maintainers can add a commit 😃

ZeLonewolf:
Ship it!

PR #271, opened 2022-04-07
Add Saskatchewan shields

claysmalley:
Adds shields for Saskatchewan. This introduces two colo(u)r-swapped variants of the "home down" shield, as well as new artwork for rural municipal routes.

Note that the 2-character variant of the "home down blue inverted" shield added here is actually unused, as all northern secondary routes have 3 characters.

Screenshot from 2022-04-07 10-17-31
Screenshot from 2022-04-07 10-13-13
Screenshot from 2022-04-07 10-16-59
Screenshot from 2022-04-07 10-16-31

Saskatchewan sorts its route numbers into hundreds series, as is common in Canada. Routes are numbered as follows:

* 1-99: primary
* 100-199: northern primary
* 200-399: secondary
* 600-799: rural municipal
* 900-999: northern secondary

Please note that these do **not** correspond to values of `network=*`. The current tagging is probably based on a misunderstanding of these series, but it is at least consistent and has remained untouched since 2018.

Primary, northern primary, and secondary routes (`network=CA:SK:primary`) are given a blue fixed-width home-plate-shaped shield with white text and light blue decorations along the top. Northern secondary routes (`network=CA:SK:tertiary`, despite there being no "tertiary" roads in Saskatchewan) have an almost identical shield, except the background is white and the text is blue. This PR removes the decorations on top, changes the blue to match the existing colo(u)r palette, and introduces a wider 3-character variant shield.

Saskatchewan_Highway_2_(jct)
Saskatchewan_Highway_39A_(jct)
Saskatchewan_Highway_962_(jct)

Rural municipal routes (confusingly, `network=CA:SK:secondary`) are signed with two stacked shields. The top shield, with no uniquely identifying information, is much larger and only has green line art of a wheat sheaf on a white background. The bottom shield is modifier-size and only contains green text on a white background. This PR simplifies and significantly shrinks the sheaf, conflates the two shields into one, and changes the green to match the existing colo(u)r palette.

Saskatchewan_Rural_Municipal_Route
Saskatchewan_Route_600


ZeLonewolf:
Looking pretty good.  One thing that might improve is that there doesn't seem to be enough space between the green mushroom thing and the number.  Perhaps the stem can be trimmed upward a bit?

image

1ec5:
Hehe, it’s supposed to be a green sheaf of wheat. It might actually look like a mushroom if we trim it further.

ZeLonewolf:
This is what I see in my head...
image

claysmalley:
How's this?

Screenshot from 2022-04-08 20-41-00
Screenshot from 2022-04-08 20-40-47

ZeLonewolf (approved):
Ohhh yeah that's the ticket.

PR #273, opened 2022-04-09
Add Prince Edward Island shields

claysmalley:
Adds shields for Canada's littlest province, Prince Edward Island. 

Screenshot from 2022-04-09 17-49-34
Screenshot from 2022-04-09 17-49-59

PEI only has one fixed-width shield design (other than the Trans-Canada Highway). The design introduced in this PR removes the tree silhouette artwork, changes the green to match the existing colour palette and shrinks the top border. The top border remains thicker than the rest of the border to allude to the removed artwork.

PEI_Highway_2
PEI_Highway_25A


1ec5 (approved):
Thanks!

claysmalley:
I posted this to talk-ca@ with the intent of getting feedback, but I guess it's okay as is. If any locals think this design isn't faithful enough to the original, I can always make a new PR.

PR #275, opened 2022-04-09
Add New Brunswick shields

claysmalley:
Adds shields for New Brunswick. The work here was mostly already done for me, as this mostly square province was made even more square when it got distorted for the provincial highway shield. I kept the original geometry of New Blobswick almost intact, but removed some islands in the northeast and some inlets in the south.

Screenshot from 2022-04-09 17-47-31
Screenshot from 2022-04-09 17-47-46

New Brunswick's three shields have identical designs with colour palette swaps. One- and two-digit routes are Arterial highways (`CA:NB:primary`), which have green backgrounds. The 100-series is Collector highways (`CA:NB:secondary`), with blue shields. And the remainder of routes, 200-999, are Local highways (`CA:NB:tertiary`), with black shields. The designs introduced here adjust the colours to match the existing palette.

NB_1
NB_100
NB_550


style/js/shield_defs.js
```diff
@@ -168,6 +168,28 @@ export function loadShields(shieldImages) {
...
+    top: 5,
+    bottom: 5,
+  };
+  shields["CA:NB:primary"] = {
+    backgroundImage: shieldImages.shield40_ca_nb_primary,
+    textColor: "black",
+    padding: padding_ca_nb,
+  };
+  shields["CA:NB:secondary"] = {
+    backgroundImage: shieldImages.shield40_ca_nb_secondary,
```
1ec5, line 183: I think the `colorLighten` option would allow us to consolidate these three images into one:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/70db71e788467bb452f4ce68ea3e88e9c3129a6d/style/js/shield_defs.js#L285

claysmalley: Does that apply to the text as well, or only the background image?

1ec5: Not sure; haven’t tried yet. Here’s the documentation for it:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/70db71e788467bb452f4ce68ea3e88e9c3129a6d/style/js/shield_defs.js#L1594-L1597 https://github.com/ZeLonewolf/openstreetmap-americana/blob/70db71e788467bb452f4ce68ea3e88e9c3129a6d/style/js/shield.js#L241-L243

If it doesn’t keep the text black, then the approach you took is perfectly reasonable.

ZeLonewolf: The current implementation applies the recoloration to both the text and shield.  In retrospect this is a bug because text color is separately configured.

ZeLonewolf: On second thought, it's not _necessarily_ a bug because we rely on this behavior for the Georgia special case shields 🤔

claysmalley: I depended on it for the Yukon shields too :grimacing: 

I wish there were an easy way to palette-swap SVGs with arbitrary fill and stroke. That would have made the extra "home down" plates introduced in #271 unnecessary.

1ec5: OK, it’s not a big deal to me if we can’t take advantage of this optimization here.

ZeLonewolf:
Screenshots look good, and I'm fine with the 3 SVGs approach.  We can always go back later and make improvements to the recoloring code to obviate the need for that duplication, but at this point, a too-large sprite sheet is not a significant issue.  I have not yet run this code to verify functionality.

ZeLonewolf (approved):
Thanks, I hate it.  You've accurately captured the essence of these ugly shields.

PR #279, opened 2022-04-15
Add place labels for villages

Pengor:
Adds labels for `place=village` locations from zoom 11 to 14. This adds to the existing rendering for settlements tagged `place=city` and `place=town`. The remaining "stand-alone" place tag, `place=hamlet`, should also be added but that is outside the scope of this PR.

For consistency, styling is similar to `place=city` and `place=town` except with smaller text and icon.

### Before:
image

### After:
image


claysmalley:
I checked it out and looked around a variety of states and metropolitan areas. Looks good to me!

ZeLonewolf (approved):
Looks good, thanks!

zekefarwell (approved):
Looks good to me

PR #283, opened 2022-04-15
Use colorLighten on black shield to draw Saskatchewan 'tertiary' shield

claysmalley:
Removes two of the redundant SVGs I made in #271, one of which is unused. The Saskatchewan Northern Secondary shield (`network=CA:SK:tertiary`) is instead drawn with the black-on-white shield and tinted blue with `colorLighten`.

Shields look just the same as before.

Screenshot from 2022-04-15 16-46-11


Pengor (approved):
Looks good to me, having both `homeDownBlueWhiteShield` and `homeDownWhiteBlueShield` as variable names is a bit confusing but not a deal breaker in my opinion.

Issue #206, opened 2022-02-27
Distinguish toll roads from free roads


1ec5:
Toll freeways should be filled with a different color than free freeways. American road maps that color free freeways red tend to color toll freeways yellow or green. (A red-green distinction may be problematic for some users with color blindness.)

There are also toll roads besides freeways. Most of them are bridges, but the [Chickasaw Turnpike](https://en.wikipedia.org/wiki/Chickasaw_Turnpike) in Oklahoma is a prominent example of a toll surface road. There are also toll two-lane expressways, such as [Toll 49](https://www.openstreetmap.org/relation/365653) in Tyler, Texas. To handle all kinds of toll roads, American road maps seem to rely on appending “(TOLL)” to the road name label. For roads symbolized with a shield instead of a road name label, a “TOLL” banner on the shield may be appropriate: https://github.com/ZeLonewolf/openstreetmap-americana/pull/205#discussion_r815370152.

PR #288, opened 2022-04-18
Add Florida shields

claysmalley:
Adds shields for Florida, including toll and county road shields.

Florida doesn't quite have a state shape conducive to fitting text inside, so they went with an outline of the state and removed a chunk of the Gulf Coast to make it work. The state highway shield (`network=US:FL`) incorporates this broken Florida outline in a black-on-white rectangular shield. The design introduced in this PR mostly retains the dimensions of the state shield.

State toll roads (`network=US:FL:Toll`) shields have a round bottom, a square top and the black text "TOLL" on a yellow background above the black-on-white broken Florida outline. The design introduced in this PR removes the top text but retains the yellow banner, and squashes the shield vertically to maintain a 20px height.

All counties in Florida use the standard MUTCD blue pentagon for their county roads.

Screenshot from 2022-04-17 21-08-16 Screenshot from 2022-04-17 21-07-46
Screenshot from 2022-04-17 21-08-39 Screenshot from 2022-04-17 21-08-57

Florida_93
Florida_9336
Toll_Florida_112


style/js/shield_defs.js
```diff
@@ -439,6 +439,99 @@ export function loadShields(shieldImages) {
...
+    "Lake",
+    "Lee",
+    "Leon",
+    "Levy",
+    "Liberty",
+    "Madison",
+    "Manatee",
+    "Marion",
+    "Martin",
+    "Miami-Dade",
```
1ec5, line 508: [Wikipedia says](https://en.wikipedia.org/wiki/Miami-Dade_County,_Florida#County_roads) none of Miami-Dade County’s county roads have signposted route numbers.

claysmalley:
Added "TOLL" banner above `network=US:FL:Toll` shield, as there is at least one non-motorway toll road in this network.

Screenshot from 2022-04-20 10-18-17 Screenshot from 2022-04-20 10-18-51

1ec5:
That’s fine for now, though I’d imagine we’d end up indicating tolls on non-freeways too for #206.

1ec5 (approved):

PR #292, opened 2022-04-18
Clean up lingering inactive code & optimize NY STE shield

Pengor:
I lumped these two things together because they're minor clean up activities that improve maintainability:
* [LGTM](https://lgtm.com/projects/g/ZeLonewolf/openstreetmap-americana) found some inactive variables and an unused import
* The NY STE shield has long been needing to be run through `npm run code_format` (which calls SVGO) to match other SVG icons


ZeLonewolf (approved):
Everything seems in order.  Tested and no errors observed in the console.

PR #294, opened 2022-04-21
Add Washington, DC shield

claysmalley:
Adds the shield for DC's only signed route.

Washington, DC has a portrait-oriented white shield with a black border, a black outline of the District, and the black text "DC" inside the outline. The DC outline is broken twice in the bottom portion to accommodate the number 295, which would otherwise overlap.

This was a tough one. If I go by the original shield's dimensions and a 20px height, I'm left with only a 16px width, and not enough vertical space to cram the number 295 within the breaks of the outline. So I went with a 22px height and 20px width, compressed the DC outline vertically, and tweaked the ends of the breaks to fit more whitespace.

Screenshot from 2022-04-21 16-34-39
Screenshot from 2022-04-21 16-36-53
DC-295

PR #296, opened 2022-04-22
Add Oklahoma shields

claysmalley:
Adds shields for Oklahoma state highways. (Not implemented in this PR: Oklahoma turnpikes, Tulsa expressways)

A decade or so ago, Oklahoma DOT decided to abandon the nationally popular but boring circle shields and go with a state shape. Oklahoma would have an ideal compact shape to fit text within, if it weren't for all the historical plot twists that gave it its iconic panhandle. With this limitation in mind, ODOT chose to put a broken black outline of the state on a white background, an approach also seen with DC (#294) and Florida (#288) . Where Oklahoma differs, though, is in the placement of the breaks in the outline. DC has only one route number to worry about, so the outline is always broken in the same place. Florida DOT chose a specific chunk of the Gulf Coast to remove in order to accommodate text.

But Oklahoma's outline only breaks when it gets too close to text—effectively, this is a white buffer around the black text. Unfortunately, adding a buffer or halo around the text is not supported on a per-shield basis, so I did the best I could to break the outline of Oklahoma in a way that looked good. For 2- and 3-character routes, I think it looks good, though 1-character routes have a lot of extra whitespace.

Screenshot from 2022-04-22 14-46-33 Screenshot from 2022-04-22 14-47-00
Screenshot from 2022-04-22 14-47-25 Screenshot from 2022-04-22 14-47-56
Screenshot from 2022-04-22 14-48-53 Screenshot from 2022-04-22 14-49-08
Oklahoma_State_Highway_1
Oklahoma_State_Highway_156
Oklahoma_State_Highway_412B


style/icons/shield40_us_ok_2.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="20" height="20" version="1.0" xmlns="http://www.w3.org/2000/svg">
+ <rect style="fill:#fff;fill-rule:evenodd;stroke:#000" width="19" height="19" x=".5" y=".5" rx="1" ry="1"/>
+ <path style="fill:none;stroke-width:.0520836" d="m 178.9297,481.11321 c 0,-0.69037 0.55963,-1.25001 1.25001,-1.25001 h 17.49998 c 0.69038,0 1.25001,0.55964 1.25001,1.25001 v 17.49999 c 0,0.69037 -0.55963,1.25001 -1.25001,1.25001 h -17.49998 c -0.69038,0 -1.25001,-0.55964 -1.25001,-1.25001 z" transform="translate(-178.93 -479.863)"/>
+ <path style="color:#000;clip-rule:evenodd;fill:#000;fill-rule:evenodd;stroke-width:.05208;stroke-miterlimit:3.86369;-inkscape-stroke:none" d="M 1.2070312 2.375 L 1.2070312 4.2089844 L 1.4140625 4.2089844 L 7.3730469 4.2089844 L 7.3730469 7 L 7.7890625 7 L 7.7890625 3.7910156 L 1.6230469 3.7910156 L 1.6230469 2.7910156 L 17.371094 2.7910156 L 17.371094 3.6953125 L 18.289062 6.9453125 L 18.289062 11.121094 L 17.087891 10.697266 L 16.076172 10.974609 L 16 11.025391 L 16 11.525391 L 16.25 11.357422 L 17.072266 11.134766 L 18.705078 11.710938 L 18.705078 6.8867188 L 17.789062 3.6367188 L 17.789062 2.375 L 1.2070312 2.375 z"/>
```
1ec5, line 4: I think the gap in the state shape in one-digit routes looks OK at this small size, even if it’s hard to unsee now that you’ve pointed it out. 😅 If we want to close the gap, I can think of two solutions that might come in handy in the future:

* Allow a shield definition to constrain a particular image to a particular ref length
* Allow a shield definition to apply a casing around the ref, just like we do (unconditionally) for banner text

ZeLonewolf: I agree that the gap is fine.  I'm happy to move forward with these visuals and let's open a new issue to not lose track of it in case someone wants to polish the cannonball later.

claysmalley: New issue for Oklahoma outline gap: #302

ZeLonewolf (approved):
Looks good to me!

PR #260, opened 2022-03-30
add Wisconsin shields

claysmalley:
Adds shields for Wisconsin. This includes State Trunk Highways, County Trunk Highways and State Rustic Roads.

(This PR does not attempt to add support for named routes that may have shields, such as the Kettle Moraine Scenic Drive or the Great Lakes Circle Tours.)

Screenshot from 2022-03-29 18-55-16

When Wisconsin first numbered its State Trunk Highways, an inverted triangle design was chosen. Eventually, they realized how narrow and impractical it was, so they superimposed a rounded rectangle on top of it.

Shields are fixed-width in real life, but this PR introduces a 3-digit variant with different dimensions to keep text roughly the same size.

WIS_35
WIS_133

County Trunk Highways have a simple fixed-width rectangular shield with the word "COUNTY" at the top. This PR eschews the top text and opts for a variable-width rectangle, again to keep text at a fixed size.

WIS_County_E
WIS_County_EEE

[Rustic Roads](https://en.wikipedia.org/wiki/Rustic_Road_\(Wisconsin\)) are a concept unique to Wisconsin, as far as I know. They are state-maintained, though they generally have low speed limits, are meant to be lightly traveled local access, and may even be unpaved. This is more-or-less a historic preservation system for road routes that have certain scenic characteristics.

Each Rustic Road has a yellow-outlined brown shield, with white and yellow line art of a scenic road and the yellow text "RUSTIC ROAD" superimposed on a white outline of Wisconsin. Below this shield, a brown modifier-size shield (not pictured) shows the number of the route in yellow, prefixed by an R. The design in this PR simplifies various characteristics:

* the line art has been removed
* the text "RUSTIC ROAD" has been replaced by the R-prefixed number of the route
* the outline of Wisconsin has been given a fixed stroke width
* the modifier-sized shield below the Rustic Road shield has been removed

Rustic_road

PR #298, opened 2022-04-23
Add shields for remaining New Jersey counties

claysmalley:
Adds support for Bergen, Cape May and Hunterdon Counties. Bergen County is unique in having white-on-black rectangular shields, identical to this project's default. Cape May and Hunterdon Counties use blue pentagons like the rest of New Jersey, but were previously missing from the definitions.


ZeLonewolf:
On the original pass of this, I remember there being some discussion of underscores versus spaces in the network name for multi-word county names, are we satisfied that's resolved?

claysmalley:
Wisconsin #260 and Florida #288 already solidified some inconsistent tagging in this regard. While everything seems consistent within each state we support, I could see the utility of harmonizing county road tagging nationwide. I guess it's now or never if we want county roads to follow a standard scheme.

ZeLonewolf:
Off the top of my head, the more specific problem was inconsistent tagging for the same county network in NJ, and a difference in convention for different counties within NJ.  I think we had both `Cape May` and `Cape_May` when we first looked at it.

claysmalley:
70 relations have `network=US:NJ:Cape_May`, and only one has `network=US:NJ:Cape May`. Looks like the underscore is heavily preferred here.

ZeLonewolf:
Cool, sounds like underscore is the winner then.

ZeLonewolf (approved):
Seems straightforward to me!

Issue #214, opened 2022-03-06
Change demo map tile server to self-hosted planetiler server


ZeLonewolf:
This issue is intended to capture discussion related to switching from MapTiler Cloud to a self-hosted planetiler tile server running on AWS.

The [public demo](https://zelonewolf.github.io/openstreetmap-americana) is currently using MapTiler Cloud as the tile server.  This configuration has the following issues:

1. MapTiler Cloud is updated infrequently.  The current MapTiler Cloud planet has not been updated since Dec 2021, when OpenMapTiles 3.13 was released.
2. MapTiler Cloud is free for limited use, however, as the demo becomes more popular, we may exceed the "free tier" limits.

I've been experimenting with [planetiler](https://github.com/onthegomap/planetiler) as an option for self-hosting our own tiles so that we can have more frequent updates.  I've collected a series of [scripts](https://github.com/ZeLonewolf/planetiler-scripts) for building tiles on AWS.  I'm able to generate an updated tileserver in 1-3 hours.  I'm paying for this out of pocket, but so far costs are low, in the "cups of coffee" range, since the tileserver itself is fairly light-duty, and the renderer only needs to run for the 1-3 hours that are needed to generate a new planet.

The running tileserver is here: https://6ug7hetxl9.execute-api.us-east-2.amazonaws.com/

The AWS architecture for this setup is as shown in the diagram below:

![AWS Architecture](https://user-images.githubusercontent.com/3254090/156907874-466db0be-8897-4536-b9d0-d03117ab7bb6.svg)

### Proposed

1. Replace MapTiler Cloud source with AWS planetiler server, with periodic updates "whenever I get around to it"

### Open issues

1. Planetiler currently does not replicate the OpenMapTiles functionality that abbreviates names, for example, "St" for "Street", "Rd" for "Road", etc.  Thus, upon this change, these names would be rendered in full.
2. Planetiler has an option to enable or disable the zoom 13 building merge.  When enabled, small adjacent buildings are merged into larger polygons at z13.  However, this causes some overlap problems in cities where the merged buildings overlap roads.  When disabled, small buildings at z13 are simply dropped.  We need to decide which configuration we want turned on.  At the time of the writing of this ticket, I have building merge disabled.  It's also a compute-intensive operation that takes up approximately 1/3 of the render time.

1ec5:
[...]

PR #304, opened 2022-04-27
Add shields for New York counties

claysmalley:
Adds support for New York county routes that use the standard blue pentagon shield.

A handful of counties around New York City and Lake Ontario do not signpost shields. Erie County posts shields with a county outline, not included in this PR.


style/js/shield_defs.js
```diff
@@ -1102,16 +1102,60 @@ export function loadShields(shieldImages) {
...
+    "Otsego",
+    "Putnam",
+    "Rensselaer",
+    "Rockland",
+    "Saint Lawrence",
+    "Saratoga",
+    "Schoharie",
+    "Schuyler",
+    "Steuben",
+    "Suffolk",
```
1ec5, line 1150: These routes are currently tagged [`US:NY:CR:Suffolk`](https://taginfo.openstreetmap.org/tags/network=US%3ANY%3ACR%3ASuffolk). Should New York county routes using the blue pentagon include the `:CR:` infix or no?

claysmalley: I don't think so, because these route numbers aren't state-assigned. They should probably be retagged to match county roads in the rest of the state.

1ec5: OK, I thought that was the difference between a `:CR` suffix and a `:CR:` infix, but this non-CR format is indeed very common even in states that standardize the sign designs.

claysmalley: Hm. I was under the impression that the `:CR:` infix meant the numbering is generally state-coordinated but counties can design their own shields. I guess we should clarify this with the community before supporting one tagging scheme over another.

claysmalley: Suffolk County routes have been retagged: https://www.openstreetmap.org/changeset/120479526

claysmalley:
Converted to draft until spaces vs. underscores is resolved.

ZeLonewolf:
This might be the forcing function for finally dealing with #214

claysmalley:
I guess I should leave this as a draft until [spaces vs. underscores is resolved.](https://wiki.openstreetmap.org/wiki/Proposed_features/County,_city_and_local_highway_networks_in_the_United_States)

claysmalley:
Abandoning in favor of #355.

PR #306, opened 2022-05-06
Add styling for (non-protected-area) parks

AnonymousAlligatorOSM:
These [are tagged](https://openmaptiles.org/schema/#fields) in OpenMapTiles in the landcover layer with subclass=park, not in the park layer (which is just for protected areas).

Screenshot from 2022-05-05 22-48-36

Screenshot from 2022-05-05 22-48-55

### DIFF
--- a/style/americana.js
+++ b/style/americana.js
@@ -34,6 +34,7 @@ var americanaLayers = [];
 americanaLayers.push(
   lyrBackground.base,
   lyrPark.fill,
+  lyrPark.parkFill,
 
   lyrBoundary.countyCasing,
   lyrBoundary.stateCasing,
@@ -44,6 +45,7 @@ americanaLayers.push(
   lyrWater.waterwayIntermittent,
 
   lyrPark.outline,
+  lyrPark.parkOutline,
 
   lyrBoundary.city,
   lyrBoundary.county,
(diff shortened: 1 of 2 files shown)


ZeLonewolf:
Also, please do an `npm run code_format` and commit to clear the failed lint check.

claysmalley:
I like the idea of styling parks and protected areas similarly.

That said, how come protected areas have labels and parks don't?

ZeLonewolf:
That's a good point.  We should probably copy whatever we're using for labelling protected areas and do the same with parks.

AnonymousAlligatorOSM:
> That said, how come protected areas have labels and parks don't?

Protected area names are in the `parks` layer, but park names are in the `poi` layer.

ZeLonewolf:
Yup, can confirm:
image

ZeLonewolf (changes requested):
Please add a name label from the poi layer, `class=park`, with the same styling used in protected area labels.

AnonymousAlligatorOSM:
Added labels, although they don't show up until zoom level 14 in the tiles.

ZeLonewolf:
Screen shot of NYC at z14:
image

ZeLonewolf (approved):
The labelling situation is not great, but that's a limit of OpenMapTiles, which labels these features as POIs (point and label) rather than as areas.  Therefore maplibre has no way of knowing that certain parks are big and others are small in order to influence render decisions.  This will be improved when we can implement an improved parks & protected areas layers.

Given the limitation in the underlying data source, I think this is the best we're able to with OpenMapTiles.

ZeLonewolf:
I've documented a ticket to consolidate park features in OpenMapTiles in the next major revision at openmaptiles/openmaptiles#1387

PR #307, opened 2022-05-06
Add latin text highway shields for Japan

ZeLonewolf:
This PR is a subset of #219 which provides comprehensive shield rendering for Japan highways.  Support is added for those highway shields which are consistent with international norms and prefixed with the JP country code.  This allows us to add the shield to the map that we all agree should be rendered, while deferring the routes with more controversial network values.


style/icons/shield40_jp_urban_expressway.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="20" height="20" viewBox="0 0 5.292 5.292" xmlns="http://www.w3.org/2000/svg">
+ <path d="m0.60707 0.13229c-0.26207 0-0.47478 0.21271-0.47478 0.47478v2.0388c0 1.3875 1.1261 2.5135 2.5135 2.5135 1.3875 0 2.5135-1.1261 2.5135-2.5135v-2.0388c0-0.26207-0.2127-0.47478-0.47478-0.47478h-2.0388z" fill="#006747" stroke="#fff" stroke-linejoin="round" stroke-width=".265"/>
```
1ec5, line 2: This SVG is unused for now. Let’s leave it for #219.

1ec5 (approved):
As in https://github.com/ZeLonewolf/openstreetmap-americana/pull/219#issuecomment-1120233443, I agree that we should move forward with these uncontroversial networks and resolve the urban expressways separately.

PR #309, opened 2022-05-07
Delete unused Japanese urban expressway shield

1ec5:
This shield is unused. It was originally added in deaaf8740abcb04d8316de6d6a376d377f1321eb for #219, but its usage was removed in 3def876f49dce6e66b73cd169b1fcfded90b7875 for #307. Since we don’t have a good way of tracking unused images yet, let’s leave this image out of the repository until #219 lands. We can also continue to refine the image there if necessary.


claysmalley (approved):
Fine by me.

ZeLonewolf (approved):
YAGNI

PR #310, opened 2022-05-07
Remove non-namespaced Croatian networks

1ec5:
Removed shield definitions for `network=Autoceste` and `Državne ceste`, which are prevalent in Croatia. Although these `network` values are apparently written in Croatian, nothing in these values explicitly constrains them to Croatia, making them even more ambiguous than the Japanese urban expressway values under discussion in #219 and akin to the German values under discussion in #149. As far as I can tell, `network=Autoceste` is undocumented. `network=Državne ceste` is [documented](https://wiki.openstreetmap.org/wiki/Special:PermanentLink/2320179#Princip_označavanja), but the wiki’s handling of spaces in article titles has resulted in [some confusion](https://taginfo.openstreetmap.org/tags/network=Državne_ceste).

Since this project is pioneering the use of `network` to choose shields, it would be irresponsible for us to “lock in” an atypical, potentially ambiguous value before the community has had a chance to consider less risky approaches. We wouldn’t want the community to be constrained by backwards compatibility concerns just because this project rushed to support the existing tag.

Internally to the project, these definitions also piggyback on unrelated shield definitions from other countries that happen to look similar, resulting in some confusing uses of these country codes. In cases where multiple unrelated networks happen to require similar-looking shields, we should factor out a generically named shield definition that can be reused without this additional confusion.

Rest assured, routes tagged `network=Autoceste` and `network=Državne ceste` will continue to be labeled, just with generic white shields instead of Slovene and Czech shields. The removal of these definitions affects neither the world shield coverage map nor the taginfo project file, because both build steps rely on country code prefixes in `network` values.

/ref #149 https://github.com/ZeLonewolf/openstreetmap-americana/pull/219#discussion_r825241144


claysmalley (approved):
Looks good. I agree with dropping support for these shields until a discussion is had among the Croatian community about tagging guidelines for highways in Croatia.

ZeLonewolf (approved):
I concur with this change for the reason described in https://github.com/osm-americana/openstreetmap-americana/pull/219#issuecomment-1120237287

PR #313, opened 2022-05-08
Add Truck-bannered North Carolina state routes

claysmalley:
One-liner; adds support for Truck-bannered state routes in North Carolina.
Screenshot from 2022-05-08 09-40-40


ZeLonewolf (approved):
Thanks!

PR #314, opened 2022-05-08
Refactor shield drawing code to be more sane and default plain shields to 2px rounded corners

ZeLonewolf:
This is an initial refactor pass to enable fixes for the current shield banner shadow overlap, which is impossible to fix currently due to code complexity.  I'm hoping this edit reduces the amount of complexity in the shield drawing code and makes it easier for contributors to work with.

In addition, I refactored out commonality in the shield definition for "default" white rectangles.  They had a mix of 1px and 2px rounded corners, so I defaulted this to 2px shields, which I think looks a little nicer.

### DIFF
--- a/style/js/shield.js
+++ b/style/js/shield.js
@@ -101,12 +101,6 @@ function drawShield(shieldDef, routeDef) {
   var bannerCount = ShieldDef.getBannerCount(shieldDef);
   var shieldBounds = null;
 
-  var drawnRef = routeDef.ref;
-
-  if (drawnRef === "" && shieldDef.refsByWayName) {
-    drawnRef = shieldDef.refsByWayName[wayName];
-  }
-
   var shieldArtwork = getRasterShieldBlank(shieldDef, routeDef);
   var compoundBounds = null;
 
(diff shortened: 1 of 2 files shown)


style/js/shield_defs.js
```diff
@@ -1431,7 +1414,7 @@ export function loadShields(shieldImages) {
     },
   };
 
-  shields["US:TX"] = roundedRectShield("white", "black", "black", 1, 1, null);
+  shields["US:TX"] = shields["default"];
   shields["US:TX:Andrews:Andrews:Loop"] = banneredShield(
     roundedRectShield("white", "#003f87", "#003f87", 1, 1, null),
```
claysmalley, line 1419: Let's change this one to a 2px radius too. Same for the rest of the rectangles defined for Texas, Wyoming and Nova Scotia.

ZeLonewolf: Fixed in 6f5db25

style/js/shield.js
```diff
@@ -242,9 +237,14 @@ function generateShieldCtx(id) {
     return ShieldDraw.blank();
   }
 
+  // Handle special case for Kentucky
```
claysmalley, line 240: Kentucky is the only place that's taking advantage of this right now, but certainly other route networks will need this special casing too.

ZeLonewolf: That's terrifying.

claysmalley: Off the top of my head, HCTRA and some New York State Parkways are gonna need this treatment.

claysmalley (approved):
Looks good! Thanks for doing this.

Issue #232, opened 2022-03-15
Military base styling


AnonymousAlligatorOSM:
Military bases are missing entirely from the map style. We can't have that, this is America! Compare this area near Salt Lake City:

![Screenshot from 2022-03-15 10-54-07](https://user-images.githubusercontent.com/101657206/158419466-210cd372-891a-42b2-ad00-b3363f473e31.png)
![Screenshot from 2022-03-15 10-54-02](https://user-images.githubusercontent.com/101657206/158419489-d14e5278-8fb0-4228-b78a-5be40ecfbd59.png)

Unfortunately, OpenMapTiles doesn't appear to include names for these areas, but it does include outlines:
![Screenshot from 2022-03-15 11-03-59](https://user-images.githubusercontent.com/101657206/158420444-d9390d34-7e47-488b-b2df-64706e95ca6d.png)

AnonymousAlligatorOSM:
A couple more examples. All the styles I could find that show them at all use a similar striped pattern.

Humanitarian:
![Screenshot from 2022-03-15 11-16-01](https://user-images.githubusercontent.com/101657206/158423299-f2260f07-5ded-4347-aa3d-e7a9ca240342.png)

CyclOSM: 
![Screenshot from 2022-03-15 11-17-39](https://user-images.githubusercontent.com/101657206/158423396-09be6abd-5799-400e-b3f9-4e10db5ab292.png)

1ec5:
Most of the commercial and state DOT road maps in my collection use pastel hues for land use areas and choose a light gray for military bases or other restricted areas, being less interesting from a consumer standpoint:

[...]

Issue #188, opened 2022-02-22
Airports and airfields


1ec5:
Airports and airfields are a prominent feature in the built environment, taking up a lot of space even at middle zoom levels. They would be represented on a conventional map as:

* A label and/or icon for the overall airfield that indicates whether it’s a commercial, general aviation, or military field
* Simple lines for the runways at low zoom levels
* Outlines or fills for runways, taxiways, and possibly aprons at high zoom levels

<img width="392" alt="INDOT" src="https://user-images.githubusercontent.com/1231218/155105333-f3b1c2eb-6fb7-44d0-9f3e-a27c6a902812.png"> <img width="397" alt="San Mateo County 2006, Compass Maps" src="https://user-images.githubusercontent.com/1231218/155105816-3114ae67-6c67-46d0-ae70-4f78f48acc20.png">
<img width="607" alt="Univeral Cincinnati" src="https://user-images.githubusercontent.com/1231218/155106378-71c543eb-5373-434f-8a05-96b4dfc680cc.png">
<img width="272" alt="OklaDOT" src="https://user-images.githubusercontent.com/1231218/155107044-8cb6904d-8ab1-4762-94d3-3b513364de4c.png">

ZeLonewolf:
There is a class attribute in OpenMapTiles that allows us to distinguish between significant and insignificant airports.

1ec5:
OpenMapTiles exposes runways and taxiways only as linear features based on `aeroway=*`, not polygons based on `area:aeroway=*`, so we’ll only be able to reliably achieve the simpler line treatment seen in the map of the Cincinnati/Northern Kentucky airport above, not necessarily the more detailed appearance of the Indianapolis or San Francisco airports.

1ec5:
Fixed in #318. Bon voyage!

PR #318, opened 2022-05-09
Add airport styling

AnonymousAlligatorOSM:
Add styling for airports. See #188.

## Zoomed out
At zoom level <=10 and when the long name doesn't fit, the IATA or ICAO code is shown.

Screenshot from 2022-05-08 22-11-24

## Large airport
Screenshot from 2022-05-08 22-11-44

## Smaller airport
Screenshot from 2022-05-08 22-12-12

## Very small airport
Screenshot from 2022-05-08 22-20-23

## Gate numbers
Bonus points if you recognize this airport

Screenshot from 2022-05-08 22-12-51

# Remaining Issues
- Military air bases are not styled differently--see #232.
- Should international airports stand out more somehow?


claysmalley:
I think the outline without runways looks a bit empty. How about having the runways show up at the same zoom as the outline? They're pretty big features and I think they'd look good at low zoom.

Likewise, taxiways show up at a somewhat high zoom and airports look empty next to residential streets. What if we bump taxiways down to where runways are currently showing up?

Screenshot from 2022-05-09 00-00-39 Screenshot from 2022-05-09 00-00-15 Screenshot from 2022-05-09 00-00-26

claysmalley:
I do like how everything looks overall, though! Fantastic work.

AnonymousAlligatorOSM:
Agreed, I think that looks good.

wmisener:
I really like the use of IATA codes when zoomed out, and I think that combined with the runways will make it clear these represent airports at low zoom, even without an airplane icon. I just want to point out that this is the first non-car transport-related rendering, and also the first use of purple on the map, so this might set a precedent for purple being a transportation-related color (similar to Carto). Not that this is a bad thing (I think it looks good), but something to consider going forward, especially if purple was being considered for any other large features.

claysmalley:
I like the stroke width and fill transparency, but a different color would help distinguish aeroways from boundaries. The shades of purple in this PR are very close to the ones used for administrative boundaries, which can interact with runways and taxiways confusingly:
Screenshot from 2022-05-09 09-20-08 Screenshot from 2022-05-09 09-18-40

AnonymousAlligatorOSM:
Changed to a more pink color to avoid confusion with boundaries:
Screenshot from 2022-05-09 10-36-29

And made airport labels without IATA and ICAO codes not show up until higher zoom levels:
Screenshot from 2022-05-09 10-37-55

OpenMapTiles doesn't expose the runway surface tag.

ZeLonewolf:
The pink doesn't feel quite right to me.  I think the original purple shading was better, but there may be other colors that work as well.

The intersection of boundaries and airports feels like a bit of an esoteric case to me.  Perhaps we can monkey with the draw order and not necessarily show all parts of the boundary drawing.  I'm not sure why a map user would care about the precise position of a town boundary through an airport anyways.

I'm also thinking that when we're down to the IATA codes that it should have an airplane icon, though I'm happy to defer that to a future issue if the author isn't graphically inclined.

adamfranco:
I'd have to agree with Brian that the purple "looked better". This style already uses a lot of red tones and I'm wondering if a red tone would also be appropriate for military areas. I like the idea of using a consistent color-range for public transit broadly (like busways, airports, ferry terminals, train & bus stations). I don't know that choosing a public-transport color needs to block this PR, but maybe we can take an initial stab (purple or otherwise) and then shift it later as more features get added.

adamfranco:
I just noticed that highway tunnels are rendering on top of runways and taxi-ways ([LAV on localhost](http://localhost:1776/#13.79/36.08308/-115.15235), [ORD on localhost](http://localhost:1776/#16.61/41.958981/-87.927631)):
Screen Shot 2022-05-09 at 12 01 15 PM

These lines should be moved after the highway tunnels:
```
  lyrAeroway.runway,
  lyrAeroway.runwayArea,
  lyrAeroway.taxiway,
```

AnonymousAlligatorOSM:
Reverted the color and added icons:
Screenshot from 2022-05-09 11-31-10

ZeLonewolf:
Looks like the only question is boundary collisions.  How would people feel about putting airport areas on top of boundaries?

Also bear in mind that at some point I anticipate putting labels on boundaries to indicate what territory is on either side.  This would be awfully cluttered to have a bunch of boundary labels running through ATL.

claysmalley:
France, fix your administrative boundaries...
Screenshot from 2022-05-09 18-49-40

claysmalley:
In all seriousness, how do people feel about a more skeuomorphic runway style, with a thick black line and a thin dashed white line down the middle?

ZeLonewolf:
>  how do people feel about a more skeuomorphic runway style, with a thick black line and a thin dashed white line down the middle?

If we do that for runways it should be done for roads also, and IMO that would only really make sense at the highest zooms.

zekefarwell:
Although I like the skumorphic idea, I think it would make more sense in a map style going for a realistic look and feel.   The highway map aesthetic we are going for in this style.is more abstract and symbolic.

zekefarwell:
I like the look of this so far.  Definitely also a fan of the cooler purple over the warmer pink.  However, a boundary running through an airport does look rather muddled with the similar shade of purple.  I wonder if desaturating the purple further or even moving entirely to a shade of gray would work.

I noticed that the North Carolina state highway map shows the runway structure of airports in the city insets.  White with a black outline is used.  Some examples below.  If anyone else has maps that show airport details like this, please share.
PXL_20220510_004955769
PXL_20220510_004915950
PXL_20220510_004735699

style/constants/color.js
```diff
@@ -12,3 +12,8 @@ export const borderCasing = `hsl(${hueBorderCasing}, 35%, 86%)`;
 export const parkFill = "hsl(136, 41%, 89%)";
 export const parkOutline = "hsl(136, 41%, 79%)";
 export const parkLabel = "hsl(136, 71%, 29%)";
+
+export const airportFill = "hsla(292, 41%, 89%, .3)";
```
ZeLonewolf, line 16: Alpha blend will be problematic because the color will be different depending on what's behind.  I used the color picker, and the equivalent hsl is `hsl(333, 33%, 93%)`

ZeLonewolf:
Exploring grayscale as suggested:

image

ZeLonewolf:
I played around with changing the layer ordering to see what we could do about airfield/boundary collisions.  The results weren't great.  First, I tried moving the boundary casing below the airfield fill.  This had the effect of keeping the boundaries on top of the airfield, but making them incoherent as you lose the state/county/city distinctions.  I also tried moving the boundary casing and fill below the airfield, and this kept the airfield intact no problem but it creates a love triangle between airfield, water, and boundary.  Airfield wants to be above boundary, boundary wants to be above water, and water wants to be above airfield (for example, small retaining ponds and so forth).  And that's the ball game.

I think the bottom line here is that having a boundary run through an airport is obviously stupid (in reality) and therefore this stupidity is reflected, accurately, on the map.  So we probably have to live with those collisions unless someone has a magic sauce solution here.

I'd propose that we come to a consensus on what color scheme we like and push ahead.

claysmalley:
What about grayscale runways/taxiways and purple airport boundaries? I like the grayscale runways, and maybe that alone is enough to make them conflict with administrative boundaries less.

jleedev:
More distinguishable from boundaries, harder to tell what's a road
Screenshot_20220513-125408
Screenshot_20220513-125447

ZeLonewolf:
Of note, boundaries on top of airports is silly on osm-carto as well :-D
https://www.openstreetmap.org/search?query=CDG%20airport#map=14/49.0135/2.5557

ZeLonewolf:
The main issue with the screenshot in https://github.com/ZeLonewolf/openstreetmap-americana/pull/318#issuecomment-1126258309 is that we chose to put tertiary roads in gray, hence the conflict.  If we could undo that, we might have this nailed.

style/constants/color.js
```diff
@@ -12,3 +12,8 @@ export const borderCasing = `hsl(${hueBorderCasing}, 35%, 86%)`;
...
+export const airportOutline = "hsl(292, 41%, 79%)";
+export const airportRunway = "hsl(270, 41%, 79%)";
+export const airportLabel = "hsl(270, 71%, 29%)";
```
zekefarwell, line 19: I think shifting the color hue more towards more of a bluish-purple makes the warmer purple boundary line more distinct while also leaving a distinction between gray roads and bluish-purple taxiways.

Screen Shot 2022-05-13 at 11 10 38 PM
Screen Shot 2022-05-13 at 11 10 57 PM

Here's the code change:

```suggestion
export const airportFill = "hsl(250, 41%, 95%)";
export const airportOutline = "hsl(250, 41%, 79%)";
export const airportRunway = "hsl(250, 41%, 79%)";
export const airportLabel = "hsl(250, 71%, 29%)";
```

Moving the dotted boundary line (but not the purple casing) above everything else (except labels) also is an improvement I'd say, but this should probably be considered in a different PR to keep this one focused on airports.

Screen Shot 2022-05-13 at 11 09 55 PM

style/layer/aeroway.js
```diff
@@ -0,0 +1,250 @@
...
+      base: 1.7,
+      stops: [
+        [12, 3],
+        [15, 15],
+        [17, 30],
+      ],
+    },
+  },
+  layout: {
+    "line-cap": "round",
```
claysmalley, line 85: Try using `square` instead of `round` on line caps for runways and taxiways. Since these features tend to have square ends in real life, I think they look a little better this way.

ZeLonewolf:
Go team!

1. I really like the color work from @zekefarwell in https://github.com/ZeLonewolf/openstreetmap-americana/pull/318#discussion_r872921754.  I think we have a winner there unless anyone objects.
2. I also like the skeuomorphic runway lines that @claysmalley demonstrated, but only at the highest zoom where it's clear that we're looking at dashed lines.  But, we can do that as a separate PR
3. I agree with the suggestion by @claysmalley for square end caps https://github.com/ZeLonewolf/openstreetmap-americana/pull/318#discussion_r872984695

Thanks for spearheading this @AnonymousAlligatorOSM, this is the first significant small-area feature and it will set the stage for other area features.

ZeLonewolf (approved):
I think this is ready to ship in this state.  We might consider some improvements at zoom 10 as far as which features to show for what types of airports, but I think those are tweaks that can be separately investigated.

claysmalley (approved):
Looks good to me! Thank you for working on this.

adamfranco:
I'm away from a computer and can't check myself, but just be sure that
there runway dashes don't conflict with unpaved roads. I can't quite tell
from the screenshots posted if that is an issue or not.

ZeLonewolf:
@adamfranco the runway dashes are not included in this PR.  @claysmalley only posted samples.

PR #325, opened 2022-05-16
Move aeroway layer above road tunnels

claysmalley:
Reorders layers so that runways and taxiways are above tunnels.

before:
Screenshot from 2022-05-15 21-00-50

after:
Screenshot from 2022-05-15 20-58-22


ZeLonewolf (approved):
Good catch.  Makes sense to me!

Issue #301, opened 2022-04-25
Clipping on right-side of rounded-rect shields


ZeLonewolf:
The same problem is present in Kentucky shields:
![Screenshot from 2022-04-24 19-58-30](https://user-images.githubusercontent.com/1732117/165002165-8de84a15-e866-4013-b2a5-b81add24fd66.png) ![Screenshot from 2022-04-24 19-58-51](https://user-images.githubusercontent.com/1732117/165002166-f897c300-b3d7-49cb-8cb8-07a93c762824.png) ![Screenshot from 2022-04-24 19-59-01](https://user-images.githubusercontent.com/1732117/165002167-aa151455-1416-415b-92c9-49bdda3e9436.png)

I think this is a bug in `roundedRectShield()`.

_Originally posted by @claysmalley in https://github.com/ZeLonewolf/openstreetmap-americana/issues/278#issuecomment-1107944727_

PR #323, opened 2022-05-14
Add Zia sun symbol for New Mexico shield

claysmalley:
Changes the shields of Iowa, Kentucky, New Mexico and Virginia (secondary) to use `circleShield()` instead of `roundedRectShield()`. I intentionally left this out of the previous PR because it's a stylistic choice and it'd need more discussion.  

For shields that are circle-shaped for 1 or 2 digits, behavior is different when expanding to 3 or more digits. Delaware, Mississippi and New Jersey use an oval. Kentucky uses a geometric stadium. Iowa, New Mexico, and the secondaries of Virginia maintain a circle and compress the text, though they have been assigned stadium shapes in this project.

In my opinion, this distinction is not worth keeping. All of these networks are just as recognizable if we stick to one shape for shields with 3 or more digits. I think ovals look better than stadiums for these cases.

(Note that this does not change `network=CA:ON:private_toll`, which would be unrecognizable without the stadium shape, and is probably a good candidate for an SVG shield anyway.)

before:
Screenshot from 2022-05-14 14-33-04 Screenshot from 2022-05-14 14-32-49

after:
Screenshot from 2022-05-14 14-32-14 Screenshot from 2022-05-14 14-32-28

before:
Screenshot from 2022-05-14 14-34-09 Screenshot from 2022-05-14 14-34-21

after:
Screenshot from 2022-05-14 14-34-58 Screenshot from 2022-05-14 14-34-46

PR #327, opened 2022-05-16
Factor out pill shield function

claysmalley:
Just the parts of #323 that had broad consensus.

### DIFF
--- a/style/js/shield_canvas_draw.js
+++ b/style/js/shield_canvas_draw.js
@@ -121,6 +121,7 @@ export function roundedRectangle(
   } else {
     var width = rectWidth * PXR;
   }
+  width = Math.ceil(width);
 
   var ctx = Gfx.getGfxContext({ width: width, height: CS });
 
--- a/style/js/shield_defs.js
+++ b/style/js/shield_defs.js
@@ -23,7 +23,7 @@ const colors = {
   yellow_green: "#c4d600", // Pantone 382
 };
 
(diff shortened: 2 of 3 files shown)


ZeLonewolf:
The code looks essentially correct, and I appreciate the consolidation and reduction in line count!  I just haven't had the opportunity yet to run the code and independently confirm everything is working.

ZeLonewolf:
I've noticed an issue with shrinkage in the pill shield that I can't explain.

Current Americana:
image

PR:
image

claysmalley:
That was a result of me changing the corner radius from 8 to 10 to make it more circley. I haven't taken a look at `roundedRectShield()` to figure out why it does that.

ZeLonewolf:
Oh yeah, that makes sense.  I think the layout math excludes the corner arcs entirely rather than trying to bring the layout area up to the edge of it.  So probably need to add the arc radius times square root of two "somewhere" in that math...

claysmalley:
`roundedRectTextConstraint()` now returns the value of `ellipseTextConstraint()` instead if it is passed a `radius` of `8` or greater:

Screenshot from 2022-05-19 17-33-25
Screenshot from 2022-05-19 17-33-38

claysmalley:
Fixes #301.

Screenshot from 2022-05-19 17-49-45 Screenshot from 2022-05-19 17-52-11

ZeLonewolf:
Oh you fixed the clipping bug!  So awesome.

PR #319, opened 2022-05-09
Add black and white oval shields for town maintained sections of Vermont routes

zekefarwell:
This PR should close #268 by adding standard black and white oval shields for `network=US:VT:Town`.  It looks like this network value is not in the tiles yet (at least for [the example I checked](https://www.openstreetmap.org/relation/13646610)), so an updated tileset will be required before they are visible on the map.
Screen Shot 2022-05-09 at 5 14 49 PM

Issue #268, opened 2022-04-05
Vermont shields and banner modifiers


ahallification:
Vermont is not nearly as complex as other states (Hi Ohio!), but there is some small nuance not currently captured by americana. Vermont has one state highway system and adopted their highway shield in the mid '90s. The default VT shield has already been created for americana. However, when the State of Vermont rolled out their highway shields, they only applied them to state maintained highways. The remaining numbered highways are town maintained, and are represented by the default state route shields, which are the default black and white ellipse. 

To capture this nuance, editing in osm has already been completed as follows:

1. For state maintained highways, no changes were made. The relationships still carry `network=US:VT`
2. For town maintained highways, the relationships were retagged with a new `network=US:VT:Town`. 
3. For highways that are split between state and town maintenance, the original relationship was split into two, with one relationship marked `name=VT xx (state)` and `name=VT xx (town)`. The town relationship was retagged with the new `network=US:VT:Town`.
4. There are also a couple bannered routes for "Alternate". These relationships were newly created and tagged `network=US:VT:Alternate`.

Since these are new edits, they are not represented in the current build (circa December 2021). The build will need to be updated as part of any changes to address this issue.

Example: [Junction of VT 22A and VT 73 shows the town and state maintained shields of VT 73](https://www.mapillary.com/app/?pKey=4361714303880587&mapStyle=esri&lat=43.804411522411&lng=-73.304979958851&z=17&focus=photo)

ZeLonewolf:
#214 will fix the data issue.

ZeLonewolf:
I've resolved #214, please let me know if/when a newer data update is needed to resolve this.

ahallification:
I believe the data has been pulled correctly. This probably needs some magical code written followed by one of those pull request thingamabobbers. Though, with my non-software development background, I’m not entirely certain. I’m happy to advise, discuss, be coached, etc. whatever it takes to make my small contribution live in the demo.

ZeLonewolf:
Step 1 is to head over to the [map](https://zelonewolf.github.io/openstreetmap-americana/#7/43.961/-72.139) and figure out if the shields are displaying as expected.  If they aren't, step 2 is figure out if the roads in question are tagged with the right tagging scheme.  Once that's determined, then we can figure out about thingamabobbers.

ahallification:
All VT highways display the same shield currently. I’m pretty sure the data in the demo is up to date (tagging is correct), as I have recent edits that are now reflected.

[...]

PR #328, opened 2022-05-17
Add bannered shield for VT Alt routes

zekefarwell:
Follow on from #268 that was meant to be included in #319


ZeLonewolf (approved):
Seems pretty straightforward to me.

PR #334, opened 2022-05-20
Add styling for toll roads

### LINKED ISSUE #206: Distinguish toll roads from free roads

claysmalley:
Fixes #206

Adds style definitions with a yellow hue for toll roads of all classes, from motorway to service.

Also removes "TOLL" banners from all route shields, as it would be implied by the distinct toll road styling.

Screenshot from 2022-05-20 10-19-20 Screenshot from 2022-05-20 10-22-35 Screenshot from 2022-05-20 10-24-45 Screenshot from 2022-05-20 10-21-35 Screenshot from 2022-05-20 10-23-50


style/layer/road.js
```diff
@@ -329,12 +371,92 @@ class TrunkExpressway extends Trunk {
...
+
+function expresswayCasingColor(hue, minZoom, transitionZoom) {
+  return [
+    "interpolate",
+    ["exponential", roadExp],
+    ["zoom"],
+    minZoom,
+    `hsl(${this.hue}, 0%, 75%)`,
+    transitionZoom,
+    `hsl(${this.hue}, 0%, 23%)`,
```
jleedev, line 449:
```suggestion
    `hsl(${hue}, 0%, 75%)`,
    transitionZoom,
    `hsl(${ hue}, 0%, 23%)`,
```

jleedev:
Some exotic examples of non-freeway toll roads:

Screen Shot 2022-05-20 at 10 48 05 AM
Screen Shot 2022-05-20 at 10 48 27 AM
Screen Shot 2022-05-20 at 10 48 42 AM

zekefarwell:
Thanks for working on this, Clay, and thanks for the map examples, Adam.  I've got a map by American Map (Arrow Inc) that uses red/green:  
  
I guess the red here is almost more of a magenta. Regardless, probaby not the best choice with accessibility for colorblindness in mind.

Personally I much prefer the blue/green color scheme as it makes controlled access highways really stand out from the red lines used for lesser highways.  A while ago [I posted in slack](https://osmus.slack.com/archives/C01V02K52UX/p1642626472296500?thread_ts=1642570938.266000&cid=C01V02K52UX) about a potential change to blue motorways I was trying out, but haven't had time to take it further yet.  I would still like to explore this if other maintainers are open to the idea.  Two road atlases I have (Rand McNally & Michelin) use thick green casing around a yellow (or maybe bright greenish yellow) center line.  This looks really nice to me and communicates the expressway styling very well.

 

However, if the consensus is to keep motorways red, then I can get behind a golden yellow or orange for toll expressways.  It looks like AAA uses a more of an orange and I think this works well for contrast with lighter background colors. 

I'm not so sure about using a different style for lesser toll roads that aren't expressways.  Perhaps it could work, but I'd like to see some examples of other maps that do this.  It seems to to me that maps highlighting toll roads in a different style generally only do it for expressway level roads.

Pengor:
Looks good to me! I will say that removing the Toll banners from shields makes sense in the long run, but perhaps we should keep them until we actually have a key/legend for the map?

ZeLonewolf:
> Looks good to me! I will say that removing the Toll banners from shields makes sense in the long run, but perhaps we should keep them until we actually have a key/legend for the map?

I'm not sure if the key/legend is a part of this style or an external component that would get built by someone integrating the style into a consumer-facing map.  Or if we need to create some kind of API that could get invoked by a style integrator to display a legend.  So this could be a significant effort to sort out, and may even be outside of our scope, so I wouldn't want to hold up our intended end design for lack of a key/legend.

To my American eyes, the colored toll roads scream "toll road" to me visually and I think I'd lean towards saying this coloring is sufficient to drop the toll banners.

claysmalley:
> However, if the consensus is to keep motorways red, then I can get behind a golden yellow or orange for toll expressways. It looks like AAA uses a more of an orange and I think this works well for contrast with lighter background colors.
> 
> I'm not so sure about using a different style for lesser toll roads that aren't expressways. Perhaps it could work, but I'd like to see some examples of other maps that do this. It seems to to me that maps highlighting toll roads in a different style generally only do it for expressway level roads.

I think it's valuable to show `toll=*` information on all roads at high zoom (>15). Though I can make a demo where non-expressway toll roads are undifferentiated at low zoom, and I'll try the blue and green color scheme with it as well.

ZeLonewolf:
At high zoom, my gut feel is that the mustard yellow (35% lightness) is too dark.  From experimenting with this, I think the floor for toll roads at high zoom should be 45%.  However, both toll and non-toll motorways use a common function `motorwayFillColor`, so some refactoring would be needed to change that zoom stop for toll roads but not for motorways generally.

**Current high-zoom color**:
image

ZeLonewolf:
Fixes #206

claysmalley:
1. Current version
2. Version with toll roads primary and lower undifferentiated at low zoom
3. As 2), but with a blue and green color scheme

Screenshot from 2022-05-21 13-37-10 Screenshot from 2022-05-21 13-36-39 Screenshot from 2022-05-21 13-35-35

ZeLonewolf:
I'm annoyed that the east-west toll section of I-635 is an accurate rendering of reality.

jleedev:
Here's how they look as a separate layer:

Screenshot_20220521-140215

claysmalley:
The toll express lanes on I-635 enter a tunnel underneath the mainlanes in this example, and we would not want to show a tunnel above a surface feature. I think the guidance that toll roads show up above free roads is applicable when the two roads are on the same layer.

In general, toll express lanes like the ones on I-635 are highways where I'd personally rather _suppress_ the toll styling until high zoom, effectively a reversal of the rendering order. Obviously this conflicts with the need we have for highlighting tolls in cases where a road is tolled in one direction, which I think takes precedence.

With that limitation in mind, I think red and yellow blend together better while still highlighting the distinction at low zoom. The green overpowers the blue quite a bit, and it makes it hard to tell this isn't just one road with a weird blue casing:

Screenshot from 2022-05-21 14-46-08 Screenshot from 2022-05-21 14-47-18

claysmalley:
before:
Screenshot from 2022-05-21 17-02-15

after:
Screenshot from 2022-05-21 17-04-41

ZeLonewolf:
This look awesome!

Newport Bridge (RI):
image

Texas:
image
image

Naples Tangenziale
image

Switzerland
image

ZeLonewolf:
There seems to be a lot of other refactor stuff in here, especially with regard to colors, can you confirm that's intended and not bleeding over from other work?

claysmalley:
Ope. I moved the shield colors to the color constants file but that was intended for another branch :facepalm: 

Does it matter a huge deal, or is the PR okay as-is?

ZeLonewolf:
Yep seems ok to me.

ZeLonewolf (approved):

PR #336, opened 2022-05-20
Add pier styling

claysmalley:
Adds styling for piers.

before/after:

Screenshot from 2022-05-20 17-32-50 Screenshot from 2022-05-20 17-32-04
----
Screenshot from 2022-05-20 17-33-00 Screenshot from 2022-05-20 17-32-15
----
Screenshot from 2022-05-20 17-33-17 Screenshot from 2022-05-20 17-32-24

### DIFF
--- a/style/americana.js
+++ b/style/americana.js
@@ -62,6 +62,9 @@ americanaLayers.push(
   lyrWater.waterLabel,
   lyrWater.waterPointLabel,
 
+  lyrBackground.pierArea,
+  lyrBackground.pierLine,
+
   lyrRoad.motorwayTunnel.casing(),
   lyrRoad.trunkExpresswayTunnel.casing(),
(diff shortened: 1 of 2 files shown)


wmisener:
Why isn't it rendering the little offshoots where the ferries launch from? E.g. https://www.openstreetmap.org/way/69707606. They seem to also have `man_made=pier`. It because they're also tagged `public_transport=platform`?

claysmalley:
Those piers in particular are represented in OpenMapTiles as public transit platforms instead, which haven't been implemented yet.

ZeLonewolf (approved):
This looks great!  Thanks for refactoring out the common background color.  I agree that rendering piers with the default background color is a reasonable starting point, and we can discuss in the future whether piers should be rendered differently from the background.  But rendering them as background is clearly better than not rendering them.

PR #337, opened 2022-05-21
Fix banner drop shadow overlapping shield

ZeLonewolf:
This PR fixes the issue we have with banner text drop shadows obscuring the top of a shield.  In addition, this removes the kludge we were using to draw a glow effect and replaces it with an actual glow effect using a text stroke with blur.

**Before**
image

**After**
image

### DIFF
--- a/style/js/shield.js
+++ b/style/js/shield.js
@@ -18,17 +18,15 @@ function loadShield(ctx, shield, bannerCount) {
   ctx.drawImage(drawCtx.canvas, 0, bannerCount * ShieldDef.bannerSizeH);
 }
 
-function drawBanners(ctx, network) {
+function drawBannerPart(ctx, network, drawFunc) {
   var shieldDef = ShieldDef.shields[network];
 
   if (shieldDef == null || typeof shieldDef.modifiers == "undefined") {
     return ctx; //Unadorned shield
   }
 
-  ctx.fillStyle = "black";
-
   for (var i = 0; i < shieldDef.modifiers.length; i++) {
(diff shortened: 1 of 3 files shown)


claysmalley:
Would it be easy to scoot the banners up a pixel or two? Right now they're a little close to the shields:
Screenshot from 2022-05-20 22-51-20

ZeLonewolf:
Sure.  The cheap way is to just increase the banner height:

image

I'll play with a fix to insert an actual gap.

ZeLonewolf:
Why yes, that is satisfying now that you mention it:

image

claysmalley:
I'm not seeing any change when there's a single banner:
Screenshot from 2022-05-21 08-54-29

The presence of banners on one shield noticeably scoots nearby shields down with it. This bug is present in `main`. If it's not fixable in this PR, let's open a new issue for it:
Screenshot from 2022-05-21 08-54-09

claysmalley:
Never mind, I just had to clear my cache. Looks much better!
Screenshot from 2022-05-21 12-01-24

The unbannered shields being too far down is still an issue though.
Screenshot from 2022-05-21 12-03-13

ZeLonewolf:
> The unbannered shields being too far down is still an issue though.

That one isn't easily solvable.  It would require a change to maplibre to adjust the baseline for shields, or some other pixel gymnastics that I haven't thought of yet.

claysmalley (approved):
Looks good.

Banner issue tracked in #340.

PR #320, opened 2022-05-10
Update non-European shields to have 2px corner radius and 1px stroke width

claysmalley:
Fixes #315

Changes various shields in the US and Canada to have standard rounded corners with a 2px radius:

Screenshot from 2022-05-10 18-33-56
Screenshot from 2022-05-10 18-35-14
Screenshot from 2022-05-10 18-38-15

Also puts South Carolina's outline into its shield:

Screenshot from 2022-05-10 19-21-52

And introducing... wide Virginia:

Screenshot from 2022-05-10 18-39-54

PR #348, opened 2022-05-26
Adjust straight-line elements of Alabama shields to pixel boundaries

claysmalley:
A minor change to Alabama shields that slipped my mind on #320. Moves the straight-line segment of the Alabama-Florida border down half a pixel to conform to pixel boundaries, and also adds a little hook in the Mobile Bay.

Screenshot from 2022-05-26 16-45-11 Screenshot from 2022-05-26 16-44-56


Pengor (approved):
Looks good 👍

PR #352, opened 2022-05-30
Add US Indian Route shield

### LINKED ISSUE #351: Indian Routes (United States)
The Bureau of Indian Affairs (BIA) signposts routes in and around Indian Reservations in the United States. Tagging consensus for these routes seems to have formed around `network=US:BIA`.

Route shields may have up to 3 digits, though 4-digit routes might exist. The Eastern Cherokee Reservation certainly has 4-digit BIA routes, though they aren't signposted with the Indian Route shield (instead, they're next to the street name, as North Carolina's secondary state routes are often signed).

Indian Route shield:

claysmalley:
Fixes #351.

Screenshot from 2022-05-30 17-18-04 Screenshot from 2022-05-30 17-18-43
Screenshot from 2022-05-30 17-18-27 Screenshot from 2022-05-30 17-18-52


style/icons/shield40_us_bia_3.svg
```diff
@@ -0,0 +1,4 @@
+<svg height="20" width="26" xmlns="http://www.w3.org/2000/svg">
+ <path style="fill:#000" d="M 4.5,0 C 3.9477381,5.5217905e-5 3.5000552,0.44773812 3.5,1 v 0.6894531 c 0.036268,0.070803 0.080854,0.1370266 0.132813,0.1972656 l -0.466797,-0.1875 C 2.6308551,1.4856636 2.0264032,1.7673238 1.845703,2.3144531 L 0.408203,6.6875 C 0.3747909,6.7883097 0.35764916,6.8937978 0.357422,7 c 4.3e-5,6.112868 5.505377,11.092155 12.378906,12.964844 0.08631,0.02359 0.17499,0.03538 0.263672,0.03538 0.08868,0 0.177364,-0.01179 0.263672,-0.03538 C 20.137201,18.092155 25.642535,13.112868 25.642578,7 25.642351,6.8937979 25.62521,6.7883097 25.591798,6.6875 l -1.4375,-4.3730469 C 23.973598,1.7673233 23.369145,1.4856629 22.833984,1.6992187 l -0.466797,0.1875 C 22.419146,1.8264797 22.463732,1.7602565 22.5,1.6894531 V 1 C 22.499945,0.44773812 22.052262,5.5217905e-5 21.5,0 Z"/>
```
1ec5, line 2: At a glance, this shield looks a little more like a corner-notched projectile point than the polished vessel I came up with for Rebusurance, but it’s still rather blunt. How about allowing the text to shrink a bit more, easing the pressure to widen the shield, as in #234? Angling the notches downward could help too.

claysmalley: Here's the thin variant only with no other changes. It looks fine for up to 3 digits, though 4-digit routes look crowded. However, I'm not convinced that any 4-digit routes are actually posted with these shields, so I guess it's a moot point.
Screenshot from 2022-05-31 08-42-50 Screenshot from 2022-05-31 08-43-00
Screenshot from 2022-05-31 08-43-37 Screenshot from 2022-05-31 08-43-53

claysmalley: And with the notch angled downward a little more:
Screenshot from 2022-05-31 09-06-15 Screenshot from 2022-05-31 09-06-41
Screenshot from 2022-05-31 09-06-28 Screenshot from 2022-05-31 09-06-50

ZeLonewolf:
Does the choice of artwork here have any cultural sensitivities we need to consider, along the lines of the Zia symbol in New Mexico?

claysmalley:
> Does the choice of artwork here have any cultural sensitivities we need to consider, along the lines of the Zia symbol in New Mexico?

Unlikely. Arrowheads like these aren't exclusive to one culture, and the design used in the shield is significantly different from any actual arrowheads.

ZeLonewolf:
What's a good test location where we can find these in the map?

claysmalley:
* San Carlos/Peridot, AZ [(localhost link)](http://localhost:1776/#12.47/33.29872/-110.4547)
* Farmington/Bloomfield, NM [(localhost link)](http://localhost:1776/#12.51/36.61673/-108.04675)

ZeLonewolf:
I showed the Arizona example (route 3) to a non-map nerd, and the shape was not recognizable as an arrowhead.  Since we have flexibility in arrowhead design, I'm thinking it might be clearer to make the tail more of a knob rather than the two linear notches.  Also the current shape seems more like a spade (like in a deck of cards), and I'm wondering if we might get better recognizability for the shape to be a bit pointier at the tip and/or jagged on the edges.

claysmalley:
Shortened the two diagonal notches; added a notch on top.
Screenshot from 2022-06-04 18-24-28 Screenshot from 2022-06-04 18-24-43

Pengor:
> ...and/or jagged on the edges.

When I did the [Montana Secondary shield](https://github.com/ZeLonewolf/openstreetmap-americana/blob/main/style/icons/shield40_us_mt_secondary.svg) I opted to keep the jagged edges and I think it's fairly recognizable. This one's trickier though since it needs to support 4 digits...

claysmalley:
I'm not sure it actually needs to support 4 digits. I haven't found a real-life example of such a shield. Most of the 4-digit routes seem to actually be unsigned inventory numbers. I'm not worrying too much about the appearance of the shield with any more than 3 digits.

claysmalley:
4-digit BIA routes tracked here: #382

PR #356, opened 2022-06-01
Adjust shield for belt routes of Allegheny County, PA to conform to guidelines

claysmalley:
Another one that slipped past me in #320.
Screenshot from 2022-06-01 15-47-41 Screenshot from 2022-06-01 15-47-55

### DIFF
--- a/style/js/shield_canvas_draw.js
+++ b/style/js/shield_canvas_draw.js
@@ -29,21 +29,27 @@ export function paBelt(ref) {
   switch (ref) {
     case "Red Belt":
       ctx.fillStyle = Color.shields.red;
+      ctx.strokeStyle = Color.shields.black;
       break;
     case "Orange Belt":
       ctx.fillStyle = Color.shields.orange;
+      ctx.strokeStyle = Color.shields.black;
       break;
     case "Yellow Belt":
       ctx.fillStyle = Color.shields.yellow;
+      ctx.strokeStyle = Color.shields.black;
       break;
     case "Green Belt":
       ctx.fillStyle = Color.shields.green;
+      ctx.strokeStyle = Color.shields.white;
       break;
     case "Blue Belt":
       ctx.fillStyle = Color.shields.blue;
+      ctx.strokeStyle = Color.shields.white;
       break;
     case "Purple Belt":
       ctx.fillStyle = Color.shields.purple;
+      ctx.strokeStyle = Color.shields.white;
       break;
     default:
       return null;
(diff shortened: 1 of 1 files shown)


jleedev:
Purple should be a lot brighter, really more of a violet. It's not at all that dark in real life: https://commons.wikimedia.org/wiki/File:Pittsburgh,_Pennsylvania_(8482609967).jpg

Sure, it reads as "purple", but it's unclear to me what's the goal of limiting the color palette this much.

Blue and green look reasonable but I wonder if there's too little contrast when they're side by side. Can they be brighter? Compare: https://commons.wikimedia.org/wiki/File:Pittsburgh,_Pennsylvania_(4190168264).jpg

Perhaps reduce the stroke width on the circles, or remove it, it makes the colors more readable side by side.

Screenshot 2022-06-01 16 40 11
Screenshot 2022-06-01 16 39 54


Some signs are in fact being replaced without a stroke on the circle, so far some green and yellow; purple never had a stroke.

claysmalley:
I think the colors pop a little better with a 0.5px black stroke around the circle:
Screenshot from 2022-06-01 16-50-51 Screenshot from 2022-06-01 16-51-03
Screenshot from 2022-06-01 16-52-55 Screenshot from 2022-06-01 16-53-06
Screenshot from 2022-06-01 16-53-58 Screenshot from 2022-06-01 16-54-16
Screenshot from 2022-06-01 16-55-12 Screenshot from 2022-06-01 16-55-27
I agree on changing the shade of purple, though. According to the photo you've linked, the Purple Belt has a different shield altogether from the rest of the belt system. Would it make sense to show this route without the white square background (i.e. just a purple circle with a white stroke)? If not, I can just remove the stroke for the Purple Belt.

Also, how would you feel about these having "BELT" modifier text above them? Right now it's kind of unclear what these routes are if you're unfamiliar with Pittsburgh.

ZeLonewolf:
> Sure, it reads as "purple", but it's unclear to me what's the goal of limiting the color palette this much.

The idea is that if we have a constrained palette there will be less chance of a color clash.  We wouldn't want 5 slightly different shades of blue to match real-world shield colors.  If the current palette is too limiting, by all means add colors to them.  The goal is to eliminate slight variations to reduce color collision issues.

claysmalley:
Here's just a circle with a hue that matches the image. Looks out of place IMO.

Screenshot from 2022-06-01 19-55-41
Just for fun, here's a blue background:

Screenshot from 2022-06-01 20-07-09 Screenshot from 2022-06-01 20-08-05

jleedev:
Yeah let's do the square background on purple.

LGTM

claysmalley:
To be clear, you mean the blue square background, and the lighter purple?

jleedev:
Sorry, I mean the white square. The purple on blue doesn't really translate.

And the purple color from the last image I think.

PR #359, opened 2022-06-02
Fix draw order of shieldLighten

### LINKED ISSUE #358: Halos on text banners should always be the same color
Currently, halos on text banners are a color other than ~white~ background color in cases where colorLighten is applied.

For example:
image

See also comment by @claysmalley in https://github.com/osm-americana/openstreetmap-americana/issues/357#issuecomment-1145299752_

ZeLonewolf:
Fixes #358

This change rearranges the draw ordering of color-composited shields such that the compositing will apply ONLY to the shield itself and not to any banner or banner halo.

It seems to work on a cursory check but I didn't have a handy list of test areas.


claysmalley:
Maryland business route shields are no longer drawing [(localhost link):](http://localhost:1776/#13.65/38.88683/-75.83695)

Screenshot from 2022-06-02 16-50-22

claysmalley (approved):
Looks good to me.

PR #360, opened 2022-06-03
Change colors and stroke widths of European shields

claysmalley:
Changes definitions of currently supported European shields to follow guidelines on color palette, stroke width and corner radius.

Before/after: Czechia, Slovakia, Austria, Hungary, Slovenia
Screenshot from 2022-06-03 10-53-50 Screenshot from 2022-06-03 10-53-10

Before/after: Netherlands
Screenshot from 2022-06-03 10-55-44 Screenshot from 2022-06-03 10-56-22

Before/after: Italy
Screenshot from 2022-06-03 10-59-30 Screenshot from 2022-06-03 10-58-51


ZeLonewolf:
Thanks for consolidating the colors on these shields.  Reducing the number of color shades in use in highway shields is important for avoiding color clashes down the road, even if it makes the colors slightly different from what's on the ground.

The only one that jumps out as substantially different is the yellow Netherlands shield.  I do think it looks a bit nicer with the black stroke border, but I'm hoping someone from .nl can weigh in as to whether it looks okay from a Dutch perspective.

ZeLonewolf:
I found this photo of Netherlands route shields:

image

The yellow shields do not seem to be stroked in on the ground usage.  Do we follow the on-the-ground usage, or should we make a rule that all shields are to be stroked?  I suspect there are other countries we will find where there is no stroked border for some cases.

vincentvd1:
Here you also have a Dutch sign with a green E-route (european highway routes). The proposed green is a bit to dark to match these (or does it otherwise conflict with other green colors in other countries?).

dutch-highway-sign-motorway-e-kilometer-distance-to-city-gouda-utrecht-dutch-highway-sign-motorway-e-123540266

claysmalley:
The proposed green may not match signage exactly, but it follows the color scheme used in this project elsewhere in an attempt to emulate an American road atlas style and reduce the chance of clashing colors. If the green on shields is too dark, we can adjust all shields to have a different shade of green, though that would be out of the scope of this pull request.

Does the black stroke around the yellow shield look good, or is it unnecessary?

vincentvd1:
I prefer to have the black outline removed from the yellow shield. Also see the image below. I think it will look strange on these black roads. Also, the official signs don't have the outline. So if it not much of an issue to remove the black outline then I prefer that. It not, then the outline is also acceptable (or make it white?)

claysmalley:
Black stroke has been removed from `NL:N` route shields:
Screenshot from 2022-06-04 09-09-51

ZeLonewolf (approved):
Looks good to me.

Blijbol:
The general theory in the Netherlands stipulates that red-blue and green-blue transitions require a border, whereas yellow-blue does not.

https://www.wegenwiki.nl/Kleurenregel

The N-shields are therefore usually rendered without border, because the background is usually blue. They do get a black border on a white background, because yellow-white does require a border. Example:

https://media.indebuurt.nl/eindhoven/2018/03/13231644/randweg-eindhoven.jpg

So neither option is wrong. If you do add a border, black is the correct colour.

In the long run I suspect it's better to add borders to all shields for map readability reasons (some countries like Belgium even have white shields without borders).

ZeLonewolf:
I think we could go either way.  The current rendering is more faithful to on the ground of course, but at some point we might have something yellow in the background that clashes.  If we want "all shields have a border" as a rule, then we should discuss that as a general rule in its own issue and add it to our style guidelines - not buried in a PR about European shields.  At this stage in the project it's easy to just push in "good enough" solutions and defer hard questions to when we have more features on the map.

PR #365, opened 2022-06-03
Preliminary support for rendering Swedish shields

riiga:
This is a WIP pending the addition of this kind of tagging to the Swedish route relations.

All shields are simple rounded rectangles with a blue background and white text. European roads are also signed, but support for that already exists.


claysmalley:
Thanks for taking on this work.

Do you have example photos of how these shields look in real life?

Could you describe the tagging scheme you've devised? I see `network=SE:RV`, `network=SE:LV`, and `network=SE:*` for each county's abbreviation. What do `RV` and `LV` mean? And is it necessary to use separate values for each county if they all have identical shields?

style/js/shield_defs.js
```diff
@@ -2158,6 +2158,57 @@ export function loadShields(shieldImages) {
     null
   );
 
+  // Sweden
+  shields["SE:RV"] = roundedRectShield(
+    Color.shields.blue,
+    Color.shields.white,
+    Color.shields.white,
+    4,
```
claysmalley, line 2166: We've standardized on a corner radius of 2 pixels. If these shields are unrecognizable without a 4px radius, that is okay, but otherwise it should be changed to 2.

style/js/shield_defs.js
```diff
@@ -2158,6 +2158,57 @@ export function loadShields(shieldImages) {
...
+    1,
+    null
```
claysmalley, line 2168: Do these shields have a fixed width or variable width? If the width is fixed, I would recommend this value be 30.

style/js/shield_defs.js
riiga: If 2 is the default I can roll back to that, I just thought 4 pixels looked better as 2 pixels makes it almost look like there's no rounded corners at all.

style/js/shield_defs.js
riiga: They are variable in width, the more numbers in the ref, the wider the shield.

riiga:
Here's an example of a county road with a 3-digit ref (the blue sign).

[(Source)](https://commons.wikimedia.org/wiki/File:Eker%C3%B6v%C3%A4gen_2008h.jpg)

The proposed scheme is
`SE:RV` for national roads (*riksväg*, commonly abbreviated RV). Numbered 1-99.
`SE:LV` for county roads (*länsväg*, commonly abbreviated LV). These can span multiple counties, numbered 100-499.
`SE:[county_letter]` for secondary or tertiary county roads. These are as a rule unsigned (there are exceptions), but more often than not shown on online maps and also referred to by their numbers in newspapers, road works, etc. The county letter uniquely identifies which county the county road is in. While the shields themselves look the same, it can be useful for other data consumers to be able to identify in which county the route is located. Numbered 500+ and numbers not unique per county.

ZeLonewolf:
@riiga thanks for taking this on and working to socialize it with the local community in Sweden.  Once you've gotten consensus and have added some of these routes to the map, please let us know so that the tileserver can be updated to test this PR.

ZeLonewolf:
talk-se thread for reference:
https://lists.openstreetmap.org/pipermail/talk-se/2022-June/003993.html

style/js/shield_defs.js
riiga: Changed to default

riiga:
@ZeLonewolf For testing purposes, I've taken the liberty of adding the network tag to one route of each type in my vicinity, so feel free to update the tileserver at your convenience.

claysmalley:
> `SE:[county_letter]` for secondary or tertiary county roads. These are as a rule unsigned (there are exceptions), but more often than not shown on online maps and also referred to by their numbers in newspapers, road works, etc.

This is interesting. In our project, we've been working with the assumption that shields should only show up on the map when they're visible on the ground. On `route=road` relations, we've been using `ref=*` for signed routes and `unsigned_ref=*` for unsigned routes. `unsigned_ref=*` is (intentionally) not supported by any software in order to prevent shields for these routes from showing up.

How are secondary and tertiary county roads in Sweden typically signed? Are their names posted instead? What kind of navigation signs do you find at highway junctions?

ZeLonewolf:
The tileserver has been updated!

riiga:
> How are secondary and tertiary county roads in Sweden typically signed? Are their names posted instead? What kind of navigation signs do you find at highway junctions?

They aren't signed at all, except with navigation signs listing a destination and the distance to it. The roads are named in built-up areas, but not in the countryside.

[Typical rural junction](https://www.google.se/maps/@58.2672316,15.7407194,3a,83y,152h,88.57t/data=!3m6!1e1!3m4!1sLIzBdgjNTVWHMz_YRzO3sg!2e0!7i13312!8i6656), with one or two destinations in each direction and their distance. No shields posted.

[Very rare posting with a shield](https://www.google.se/maps/@58.3866447,15.5920161,3a,39.2y,158.68h,86.49t/data=!3m6!1e1!3m4!1sh78S8OVirMX4ZtDMVtOe_g!2e0!7i16384!8i8192). This is because the road has a thru traffic prohibition for trucks, [signed just beforehand](https://www.google.se/maps/@58.387496,15.5915079,3a,28.5y,183.39h,89.37t/data=!3m6!1e1!3m4!1sDk0wQA11-Nv6HAXMnd8hvA!2e0!7i16384!8i8192) with reference to the road number. Sign reads "Applies to thru traffic on road 708 between Halshöga and road 702 in 2.2 km".

ZeLonewolf:
I think it's not really for us (US mappers) to say when it's appropriate or not to put the ref tag in various in the ground signage situations.  We'll render it where tagged and if that looks wrong the Swedes can sort out how the tagging ought to be adjusted.

style/js/shield_defs.js
```diff
@@ -2160,6 +2160,65 @@ export function loadShields(shieldImages) {
...
+    "O",
+    "S",
+    "T",
+    "U",
+    "W",
+    "X",
+    "Y",
+    "Z",
+    "AC",
+    "BD",
```
claysmalley, line 2201: If `network=SE:RV` and `network=SE:LV` have identical shields to county routes, `RV` and `LV` belong at the top of this list.

Inline comments describing what these two abbreviations mean should be enough to make it clear that they are not counties.

claysmalley (changes requested):
The tagging scheme looks sufficient to describe roads in Sweden. There hasn't been much input from other Swedish mappers, but if it turns out to be controversial, we can always revisit it.

As a matter of style, `network=SE:RV` and `network=SE:LV` don't need separate shield definitions if their shields are identical to secondary and tertiary county routes. Once that is fixed, this will be ready to merge.

style/js/shield_defs.js
riiga: Changes made

claysmalley (approved):
Looks good!

style/js/shield_defs.js
```diff
@@ -2159,6 +2159,52 @@ export function loadShields(shieldImages) {
...
```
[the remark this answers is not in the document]
1ec5: (Or `SE:county`, following [the more common convention](https://wiki.openstreetmap.org/wiki/Key:network#Basic_format) for countries that don’t make a lot of local distinctions.)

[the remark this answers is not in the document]
1ec5: As a point of reference, a lot of U.S. states also have county route networks that can repeat numbers across county lines and use a uniform shield. That shield typically bears the county name, and people even call them “Abc County Route 123” sometimes. But the `network`s got simplified to omit this detail because most renderers don’t need that extra context and a spatial query can already provide it if necessary. Geocoders would also find these county-qualified `network`s redundant to the containing boundary information they already provide. Another kind of `network` consumer is editors, but ideally editors would understand secondary tags like `is_in:county` or (hypothetically) `disambiguator` to label routes less confusingly in a list of relations.

Fortunately, the list of counties in Sweden is pretty manageable and hopefully stable. By contrast, we should probably stop rendering `network`s for China’s county, district, and village routes, because there’s no reasonable way to maintain that list over time as it grows and changes: https://github.com/ZeLonewolf/openstreetmap-americana/pull/213#discussion_r820214498.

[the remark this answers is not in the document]
zekefarwell: Yes I saw that explanation.  If the tagging already had significant use and any other data consumer was already using it I wouldn't question it.  I just want to make sure we as style authors don't end up promoting a needlessly complex tagging scheme that no data consumer actually _does_ find useful.  Without input from any other Swedish mappers it just makes me a bit uncomfortable to be the first to adopt the new scheme.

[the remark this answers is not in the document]
ZeLonewolf: In the US we've gone with nomenclature that matches how the network is abbreviated in customary use.  For example, our Interstate highways are abbreviated as "I" so "Interstate 95" is normally described verbally as "I-95" so hence `network=US:I`.

If you're not getting any pushback and you've done due diligence in notifying the community then I don't see much issue in going along with what you've decided.

ZeLonewolf:
Rendering looks as expected for one sample network:

image

ZeLonewolf:
taginfo is showing a network values of `SE:LS` which is not mentioned in this PR.  Can you clarify what this is or if it should be included here?

image

riiga:
`SE:LS`is part of this change (but were added a bit later), it is for local loops ("lokal slinga"), a rare type of shield. It should show up if you look at the roads in Linköping, there are two ringroads signed with "C" and "Y" respectively. [Street view for reference](https://www.google.se/maps/@58.3927901,15.6429454,3a,24.8y,279.88h,87.97t/data=!3m6!1e1!3m4!1s3TWm8Cj99QQi4OW2uhge3w!2e0!7i13312!8i6656).

ZeLonewolf:
Ah, looks like I didn't scroll far enough down, thanks for clarifying.  Of course, with the standard black-on-white shield, that's the same as the default anyways :)

image

ZeLonewolf:
Thanks @riiga for the contribution and working with the Swedish OSM community!

PR #368, opened 2022-06-04
Change shield banners to use knockout halos

ZeLonewolf:
This PR changes the shield banners to use a 1px "knockout" effect rather than the previous fade halo:

image


style/js/shield_text.js
```diff
@@ -241,15 +242,16 @@ export function drawBannerHaloText(ctx, text, bannerIndex) {
   ctx.font = Gfx.shieldFont(textLayout.fontPx);
   ctx.textBaseline = "top";
   ctx.textAlign = "center";
-  ctx.shadowBlur = 5;
-  ctx.lineWidth = 1.2;
+  ctx.shadowBlur = 0;
+  ctx.lineWidth = 2;
```
claysmalley, line 246: Let's multiply this by the pixel ratio so it doesn't look too thin at 200% zoom:
Screenshot from 2022-06-04 15-14-47
Screenshot from 2022-06-04 15-14-29

ZeLonewolf: Resolved in 277bfb8 but please verify.

claysmalley (approved):
Looks good!

PR #369, opened 2022-06-05
Shields: change default max font size to 14px

Blijbol:
Max font size in shields defaulted to 16px while the documentation
suggested targetting 8-12px. This changes the default to 14px and the
documented target range to 8-14px.

Rationale:

1. 16px is huge, making it very hard to get consistent font sizes across
shields. 16px fits only in the simplest of shield shapes, and even then
still looks large.
2. 12px creates excessive vertical padding in simple rectangular
shields.
3. 14px solves problem 1 without creating problem 2 yet.

Preview with 14px:
Firefox_Screenshot_2022-06-05T13-13-51 473Z


claysmalley:
Seems like a good change to me.

Some other shields have huge text by accident as well and could benefit from this treatment, like `network=US:VT`. This maximum might be worth setting globally instead of on a per-shield basis.

The global maximum is on line 171 of `js/shield_text.js`, currently set to `100`. This should be changed to `14`, and line 178 should be changed to use the lower value between the global maximum and the shield's maximum.

ZeLonewolf (changes requested):
Thanks for opening this PR.  I agree with reducing the max font size for shields.  Please resolve changes requested in https://github.com/osm-americana/openstreetmap-americana/pull/369#issuecomment-1146807685

Thanks!

Blijbol:
Pushed another commit to implement it as a global maximum as suggested.

claysmalley (approved):
Looks good!

ZeLonewolf (approved):

ZeLonewolf:
Thanks @Blijbol for your contribution!

PR #371, opened 2022-06-06
Build improvements

jleedev:
- Don't build sprites three times
- Use [`npm-run-all`] to make scripts that call other scripts more readable.
- Use [`npm ci`] in CI to read from `package-lock.json`.

In a nutshell:

```json
{
  "scripts": {
    "example": "npm run foo && npm run bar && npm run baz"
  },
}
```

Becomes:

```json
{
  "scripts": {
    "example": "run-s foo bar baz"
  },
}
```

And there's also `run-p` to run in parallel.

[`npm-run-all`]: https://www.npmjs.com/package/npm-run-all
[`npm ci`]: https://docs.npmjs.com/cli/v8/commands/npm-ci

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Install and Build 🔧
         run: |
           cd style
-          npm install --include=dev
+          npm ci --include=dev
           cp configs/config.aws.js config.js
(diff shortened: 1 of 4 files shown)


ZeLonewolf (approved):
Works for me here.

Had to re-run `npm install`

PR #373, opened 2022-06-07
Reorder bridge layers to match other layers

claysmalley:
Reorders highway layers so that links render below non-links on bridges.

[Cleveland, TX](http://localhost:1776/#14.59/30.3122/-95.10965)
before:
Screenshot from 2022-06-06 21-37-32

after:
Screenshot from 2022-06-06 21-38-08


ZeLonewolf (approved):
Tested this in Boston, which is my go-to location for complicated highway/ramp stacks.  Everything looks good, thanks!

Issue #142, opened 2022-02-05
Assign highway shield based on ref value


Mashin6:
In some communities they have chosen to group all their roads under a single `network` value with an assumption that certain letter codes in `ref` would be sufficient to split them apart. An option would be to define shield graphics not only by an exact match of `network`, but also by an additional regex match in `ref` values.
This will have to be carefully crafted around ref preprocessing #141 in case it will be implemented.

Countries that are affected:
Hungary [#69](https://github.com/ZeLonewolf/openstreetmap-americana/issues/69#issuecomment-1013750719)
Portugal
Bulgaria
North Macedonia

1ec5:
> In some communities they have chosen to group all their roads under a single network value with an assumption that certain letter codes in ref would be sufficient to split them apart.

As I understand it, the main reason for migrating to route relations with distinct `network`s was to keep data consumers from having to sniff out alphabetic codes the way they had been with `ref`s on ways. It would be helpful to know the reasoning behind replicating the same unstructured format on relation `ref`s. Maybe these communities could be convinced to make their route relation tagging more structured.

Mashin6:
With Hungarian community [that didn't work out.](https://github.com/Mashin6/detect-mutations)

There is also an option of not doing anything about it. It's all just a suggestion.

ZeLonewolf:
My impression is that the countries listed have grouped together multiple de facto highway networks under a common network value, and that this is properly solved with a tagging solution rather than a rendering one.  Thus, I agree with closing this issue upon merge of #355, as the ask here "assign highway shield based on ref value" will be resolved.  I don't think we should make special code for the case of insufficiently structured `network` tagging schemes.

PR #376, opened 2022-06-08
Add support for defining shields by ref

### LINKED ISSUE #282: Queen Elizabeth Way shield (Ontario)
The Queen Elizabeth Way in Ontario has a unique shield that is a derivative of the King's Highway shield, with a blue-on-yellow colour scheme and a serif font. The QEW is currently tagged with `network=CA:ON:primary`, which corresponds to the black-on-white shield used for the rest of the King's Highway. For renderers to support giving the QEW its unique blue-on-yellow shield, it would need a unique `network=*` value. A suggested value that follows guidelines would be `network=CA:ON:primary:QEW`.

Ontario_QEW_crown

### LINKED ISSUE #287: Arkansas Highway 980 (airport) shields
Arkansas Highway 980 is a number used many times around the state on short roads leading to airports. All instances are currently tagged `network=US:AR`. A decision would need to be made on tagging of the `network=*` value, and whether this would add any special cases, before support can be added to this project.

Compared to the standard Arkansas state route shield, the Arkansas shield is color-swapped, the ref of the road is shrunk and moved to the bottom of the state outline, additional text is added along the top, and line art is placed in the middle. Highway 980 has "AIRPORT" as its top text, an airplane as the line art, and a white-on-blue color scheme.

Arkansas_980(Airport)

claysmalley:
Poking the hornet's nest. Fixes #282, fixes #287.

This PR adds support for overriding shield definition values based on the a route's `ref=*` value, enabling the Queen Elizabeth Way and Arkansas Highway 980 to use their unique shields.

The `shieldLighten()` function is removed in favor of defining all override parameters at the end of `loadShields()`.

Screenshot from 2022-06-07 19-33-21
Screenshot from 2022-06-07 19-33-58


claysmalley:
Also changes the method of colorizing Yukon route shields. The previous method was to use `colorLighten` on a black shield, and now colors are defined in the base shield definitions without relying on `colorLighten`.

Colors look sharper - check the edges of the numeral 2. Before/after:
Screenshot from 2022-06-07 21-02-32 Screenshot from 2022-06-07 21-11-36

claysmalley:
I dunno, but I just refactored Allegheny County belt routes that way. There might be an even more succinct way to refactor these, but I don't know enough JS to figure that out. (cc @jleedev)

1ec5:
> wondering if there's a similar improvement we can/should make with the Kentucky parkway code?

Kentucky’s parkways have no `ref`s to speak of. As a last resort, `refsByWayName` maps way names to synthetic “refs” to place inside the shield. If KYTC had done the sensible thing and kept each parkway’s shield a different color, then we could combine `refsByWayName` with `overrideByRef` for a very nice effect. Alas.

zekefarwell (approved):
This change looks good to me.  The code is well structured, readable, and the terminology overrideByRef makes it clear that this is an override of the standard design, to be used for exceptions.  I.e. most routes in a network use a certain shield design, but a few are unique.  

What I don't think we should do is use this same kind of logic to support a single network value for multiple different shield designs, where multiple, heirarchical network values would be entirely appropriate, as described in #142.  To this end, I think it might be better to not handle the PA belt system with this override feature, but it can be argued there is only one route for each shield, so they are all exceptions 😀.  In general I think we should strive to minimize use of this override feature as much as possible in favor of well structured network values.

PR #374, opened 2022-06-07
Replace mapnik with sharp for SVG rendering

jleedev:
Replace spritezero with the (just-released) [sprites], which uses the
same [shelf-pack] logic but uses [sharp] for SVG rendering, which has
the advantages of:

- Provides [binaries] for several platforms (windows, macos, linux,
  linuxmusl) and architectures (arm64 and x64) - Fixes #132.
- Compiles without a fuss on other platforms (e.g. 32-bit linux arm),
  unlike mapnik which is, shall we say, large.
- Seems to render certain SVGs more correctly than mapnik. I haven't
  prepared a gallery, but there were obvious fixes in  the area of
  antialiasing and downsampling. Of note, mapnik was inflating the
  stroke width in star_state_capital.svg and dot_city (if we liked those,
  we should adjust the stroke width in the actual icon). There are minor
  changes to dimensions due to rounding:

```
$ curl -fs https://zelonewolf.github.io/openstreetmap-americana/sprites/sprite.json | jq -c .shield40_us_oh_tus_salem
{"height":19,"pixelRatio":1,"width":25,"x":283,"y":61}
$ <style/dist/sprites/sprite.json jq -c .shield40_us_oh_tus_salem
{"width":25,"height":20,"x":85,"y":221,"pixelRatio":1}
$ head -n1 style/icons/shield40_us_oh_tus_salem.svg
<svg width="25" height="19.999" version="1.0" xmlns="http://www.w3.org/2000/svg">
```

As a result, the entire sprite sheet looks different, so a comparison is
full of noise, but efforts to align SVGs to a pixel grid can be more
useful now.

[binaries]: https://github.com/lovell/sharp/releases
[shelf-pack]: https://www.npmjs.com/package/@mapbox/shelf-pack
[sharp]: https://www.npmjs.com/package/sharp
[sprites]: https://www.npmjs.com/package/@basemaps/sprites

PR #378, opened 2022-06-08
Adjust SVG dimensions and stroke widths

claysmalley:
A bunch of minor changes to shields, mostly just rounding values to whole numbers where they should be.

Changes are minuscule but may reduce unexpected behavior when comparing two sprite packing programs, as in #374.


jleedev (approved):
Here's mapnik vs sharp after this change:
after-mapnik-vs-sharp

Definitely a lot closer, since shelf-pack produces the same output.

Here's sharp before and after this change, with a threshold as low as it will go:

before-vs-after-sharp

Probably some rounding errors in Georgia, since you scaled it horizontally by 0.5%.

Issue #372, opened 2022-06-06
Increase default highway width to better match physical width at high zoom


adamfranco:
In the current implementation, road widths stay small as zoom increases, leading to them being much smaller than physical widths at ~z17 and higher.

[Current Americana](https://americanamap.org/#18/42.19522/-73.36163):
![GB Americana](https://user-images.githubusercontent.com/25242/172229554-d8a4d830-ebbf-4363-8100-840f3d6a694b.png)
For contrast, here is [Carto](https://americanamap.org/#18/42.19522/-73.36163):
![GB Carto](https://user-images.githubusercontent.com/25242/172229571-cb9477cb-7e14-42ca-9d7b-c5d559bd1462.png)

Above a certain zoom we should start increasing default highway line widths to more closely approximate likely physical widths. As a rough starting point for targets we could aim for showing highways at the equivalent of ~8' driveways, 10' service, 14' residential, 18' tertiary, 22' secondary, 30' primary, and 36' trunk & motorway. (More considered width suggestions welcome.)

Note that this would affect _default_ highway widths. Other techniques to show lanes or more accurate physical widths are also welcome.

wmisener:
I said this on Slack, but I'll repeat it here to keep things together: I think increasing the road widths at high zoom is a very good idea. The time is ripe to do it too, before the project moves toward rendering more small-scale features like POIs where high zoom appearance becomes more important. Wider roads will also help make the paved/unpaved distinction (#63) easier to see for residential and service roads (classes which you might expect to be more commonly unpaved, at least in the US), as these roads are so thin that I personally sometimes have trouble seeing the pattern right now, even at high zoom. It would also provide more pixel real estate on these ways to denote access (#216) in the future.

jeisenbe:
Since this style is also intended for Hawaii (and Puerto Rico?) you will also want to consider the apparent width at 18 degrees, where the same pixel represents significantly more width than in the mid-latitudes. And in contrast, in Fairbanks Alaska the rendered road will still be too narrow compared to the actual width, due to the changes in scale with the Mercator projection at higher latitudes.  

It is actually possible to adjust the line rendering based on latitude, but the code is complex. See an example at  https://github.com/gravitystorm/openstreetmap-carto/pull/1853 which also adjusted runways based on width, though I believe you could use a simpler version to just adjust by latitude. 

Also see https://github.com/gravitystorm/openstreetmap-carto/pull/1861

1ec5:
[...]

PR #383, opened 2022-06-09
Clarify and prioritize overall project purpose

ZeLonewolf:
This PR proposes the following:

1. Make "collaboration and common purpose" our top priority.  If we are successful, it's because of our ability to work together.  This map is first and foremost a reflection of our community and we should prioritize our community perspective over any strict adherence to external principles.
2. Explicitly declare that we take our inspiration from paper maps from this region (however, this is secondary to our community opinion).

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,9 @@ _A quintessentially American map style_ [🗺 View the map](https://zelonewolf.g
 
 The purpose of the Americana style is to:
 
-- Express the American experience through cartography
-- Challenge the status quo by showcasing innovation and invention
 - Promote collaboration and common purpose in the American mapping community
+- Express the American experience through cartography, taking inspiration from the familiar features of North American paper maps
+- Challenge the status quo by showcasing innovation and invention
 
 The Americana style is the first digital map to achieve concurrent, state-specific highway shields arranged along the path of road. Representative highway shield rendering is of considerable cartographic importance to the American community. We do this proudly in an open source project using vector tile technology.
 


README.md
```diff
@@ -6,9 +6,9 @@ _A quintessentially American map style_ [🗺 View the map](https://zelonewolf.g
 
 The purpose of the Americana style is to:
 
-- Express the American experience through cartography
-- Challenge the status quo by showcasing innovation and invention
 - Promote collaboration and common purpose in the American mapping community
+- Express the American experience through cartography, taking inspiration from the familiar features of North American paper maps
```
1ec5, line 10: Expressing the American Experience through cartography is a catchy slogan worthy of PBS. ❤️

What would clarify the purpose even better than this section would be a statement affirming the intended audience and geographical scope. https://github.com/ZeLonewolf/openstreetmap-americana/issues/372#issuecomment-1150419745 posed a good question about the degree to which we’re focused on the U.S. versus the rest of the world and how much of the U.S. Additionally, we’re already off the far end of roadgeekery with our shield work, but we also have railfans and trailfans in our midst eager to have a piece of the action. This is FAQ-level material.

Also separately from this mission statement, we should document our sources of inspiration in more detail. Already, we’ve taken inspiration not only from North American road atlases but also from political and topographic maps. Potentially in the near future, trail maps and tourist maps can be added to the mix too. This is important to highlight because our process is fundamentally different than some other OSM projects: many of us don’t put a lot of stock in our own design skills, so we’re actively (re)learning regional cartographic conventions and applying them in every discussion. Familiarity is a shortcut to usable and sensible design.

ZeLonewolf: Do you see an additional statement of scope and/or design/inspiration belonging in the lede section or further down?

bgo-eiu:
I'm overdue to make some trips to the library but perhaps some links to some paper map resources/indexes if anybody knows any would also be a good idea? Could even just be links to github comments where people have shared good pictures.

ZeLonewolf:
Directly including paper map samples in the repo itself would be problematic for a CC0 repository, but we could certainly describe the types of maps or link to my talk at State of the Map US: https://www.youtube.com/watch?v=4T8FxOPlqXA

It sounds like we need to create a section that specifically discusses our cartographic principles and I think that deserves its own discussion issue beyond this high-level purpose PR.

zekefarwell (approved):
Looks good to me

README.md
1ec5: I think we can continue this line of thinking in #385 to clarify what’s being said here. No concerns on my end.

Issue #128, opened 2022-01-31
Icons for points of interest


TheAdventurer64:
Would it be possible to add POI data for the map (restaurants, shops, etc.)?

ZeLonewolf:
Of course.  PRs are welcome for POIs.  Any POIs present in the OpenMapTiles [POI layer](https://openmaptiles.org/schema/#poi) are fair game.

zekefarwell:
We should probably add the rest of the roads too so POIs aren't just hanging out in a random spot on the map 😄

ZeLonewolf:
Yeah, and a half-dozen other incomplete features while we're at it 😂

1ec5:
One possible next step would be to define a color system for categorizing POIs. Lots of print maps, such as those by the National Park Service and Rand McNally, just use black silhouettes to ensure contrast against a variety of backgrounds. On the other hand, many online maps reserve a handful of colors for different kinds of POIs, roughly along the lines of the `amenity`/`healthcare`/`office`/`shop`/`tourism` distinctions OSM makes.

Honorable mention to #17, which adopted the MUTCD symbol signs, which are green for transportation-related POIs, blue for services, and brown for recreational POIs.

jleedev:
From 1974:

![IMG_20220518_131052](https://user-images.githubusercontent.com/23022/169110726-33320b34-f695-4eed-9c3a-4150314546b2.jpg)
![IMG_20220518_130942](https://user-images.githubusercontent.com/23022/169110751-234f71f7-30e2-4827-8180-6870d3667175.jpg)

Main takeaway is you don't have to go back very far at all before the color choices are magenta and cyan.

bgo-eiu:
https://colorbrewer2.org/#type=sequential&scheme=BuGn&n=3

Going to post this color blindness color scheme checker here, partly for myself so I don't suggest something that doesn't work great unintentionally. (There's quite a lot of color blind mappers out there)

bgo-eiu:
[...]

PR #387, opened 2022-06-11
Allow contributors to add points of interest

ZeLonewolf:
This PR sets up the structure for contributors to add POI icons with the following scheme:
* POI icons are black PNGs that are recolored in the style
* All POIs are contained in a single style layer with selectors to set color, icon, etc

As a demonstrar, this PR includes bars/pubs and hospitals to show how the single selector works.  In addition, the source SVGs for the icons are included.

Eventually I'd like to be able to produce the PNGs from SVG at compile time so contributors can deal solely with SVGs.

The initial icons and colors are not great thanks to my limited design skills.  Improvements welcome :)
image


style/layer/poi.js
```diff
@@ -0,0 +1,82 @@
...
+  ["==", ["get", "class"], "hospital"],
+];
+
+const iconColor = [
+  "match",
+  ["get", "class"],
+  "bar",
+  "#0000dd",
+  "beer",
+  "#0000dd",
```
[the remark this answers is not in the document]
1ec5: What belongs in the same category as these two `class`es? Alcohol? Entertainment? Eateries? Food service? Retail?

Blue is an unusual color for any of these criteria. If it’s limited to alcohol, then there would be an unusual emphasis on alcohol, which is a valid editorial choice but one that may limit this style’s relevance. If the scope is broader, then a retail district will have a lot of blue icons. Blue and bold together can also be misinterpreted as a clickable hyperlink. I’d suggest making them all fall back to the generic POI color (black) instead.

style/sdf/health_cross.svg
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="16" height="16" version="1.1" xmlns="http://www.w3.org/2000/svg">
+ <path d="m5 0v5h-5v6h5v5h6v-5h5v-6h-5v-5h-6z" fill-rule="evenodd"/>
```
1ec5, line 3: Some maps use a cross, while others use an “H”. Both symbols would be reasonably intuitive to those familiar with American maps. The former is language-agnostic, while the latter is more specific, ruling out doctors’ offices and urgent care facilities.

style/layer/poi.js
```diff
@@ -0,0 +1,82 @@
...
+    "icon-image": poiIcon,
+    "icon-size": 1,
+    "text-field": label.name_en,
+    "text-anchor": "bottom",
+    "text-variable-anchor": [
+      "bottom",
+      "bottom-right",
+      "bottom-left",
+      "right",
+      "left",
```
1ec5, line 69: Should `left` and `right` come first? Should `bottom-right` and `bottom-left` even be here?

In the screenshot above, “Boom Boom Room” demonstrates how difficult it can be to associate the label with the icon in a dense neighborhood. Since I’m unfamiliar with this neighborhood, I have to do process of elimination to determine that “Boom Boom” isn’t a bar to the northeast of “Room”. A few factors contribute to this difficulty:

* `text-max-width` is wide relative to `text-radial-offset` and the icon size.
* `text-radial-offset` is large compared to `text-height`.
* Line wrapping is enabled.
* `bottom-right` and `bottom-left` are less common for POI icons.

style/layer/poi.js
```diff
@@ -0,0 +1,82 @@
...
+      stops: [
+        [5, 8],
+        [8, 10],
+        [12, 12],
+      ],
+    },
+    "icon-image": poiIcon,
+    "icon-size": 1,
+    "text-field": label.name_en,
+    "text-anchor": "bottom",
```
1ec5, line 63: This property has no effect because of `text-variable-anchor`.

style/layer/poi.js
```diff
@@ -0,0 +1,82 @@
...
+    "text-variable-anchor": [
+      "bottom",
+      "bottom-right",
+      "bottom-left",
+      "right",
+      "left",
+    ],
+    "text-justify": "auto",
+    "text-radial-offset": 0.8,
+    "icon-optional": false,
```
1ec5, line 73: Consider setting `text-optional` to `true` for POIs that have non-generic icons. This will make it easier to increase the icon density without cluttering the map so severely. (Hiding the text but showing a generic icon would be something of a tease.)

style/layer/poi.js
```diff
@@ -0,0 +1,82 @@
...
+
+const iconColor = [
+  "match",
+  ["get", "class"],
+  "bar",
+  "#0000dd",
+  "beer",
+  "#0000dd",
+  "hospital",
+  "#ff0000",
```
1ec5, line 30: The specific color values here could use some refinement to keep the icons from looking gaudy. As a starting point, the red for freeways should be _the_ red for anything other than a fill in this style:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/823df2da905d2fbe26b39273d814277f1043d1cc/style/layer/road.js#L235-L247

style/layer/poi.js
```diff
@@ -0,0 +1,82 @@
...
+  "icon-color": iconColor,
+};
+
+export const poi = {
+  id: "poi",
+  type: "symbol",
+  paint: labelPaint,
+  filter: poiFilter,
+  layout: {
+    "text-font": ["Metropolis Bold"],
```
1ec5, line 51: Can we get away with using a regular typeface for these labels, or would they get lost in a crowd of road labels? It would be nice to reserve boldface labels for things we want to highlight, like places and landmarks that are especially important irrespective of their `class`.

1ec5 (commented):
We should continue the discussion in #128 about whether we should minimize the number of POI icon colors, as we’ve done with roads, or have POIs occupy the full color spectrum.

style/layer/poi.js
ZeLonewolf: Looks like we can use "Metropolis Light" though ultimately I'd like to update the fonts used in this style more broadly to not be handcuffed to the few options we have on the openmaptiles fonts server.

1ec5: #393 will probably force us to migrate to another font.

style/sdf/health_cross.svg
ZeLonewolf: It sounds like maybe hospitals could be an **H** and other health care could be a **+**.

ZeLonewolf:
I've updated this PR based on feedback received.   Some screen shots:

Location: ([localhost link](http://localhost:1776/#17/21.280648/-157.826455))
image

Location: ([localhost link](http://localhost:1776/#17/52.519039/13.423006))
image

Location: ([localhost linkl](http://localhost:1776/#17/40.754149/-73.967023))
image

icons/poi_p.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="32" xmlns="http://www.w3.org/2000/svg">
+ <path d="M 0,0 V 32 H 6 V 22 h 6 A 12,11 0 0 0 24,11 12,11 0 0 0 12,0 H 4 Z m 6,6 h 7 a 5,5 0 0 1 5,5 5,5 0 0 1 -5,5 H 6 Z" style="fill:#003f87;fill-opacity:1;stroke-width:2"/>
```
1ec5, line 2: This path hugs three sides of the viewbox and is relatively thick, causing it to appear much larger than the martini glass icon, even when both are rendered at nominally the same size. For balance, this path should be shrunk down by a few pixels, as should the health cross icon, which has a similarly thick path.

src/layer/poi.js
```diff
@@ -0,0 +1,81 @@
+import * as label from "../constants/label.js";
+import * as Color from "../constants/color.js";
+
+const poiIcon = [
+  "match",
+  ["get", "class"],
+  ["bar", "beer"],
+  "poi_martini_glass",
+  "hospital",
+  "poi_health_cross",
```
1ec5, line 10: Did you want to use an “H” instead of a ➕ for hospitals, per https://github.com/ZeLonewolf/openstreetmap-americana/pull/387#discussion_r917476571?

src/layer/poi.js
```diff
@@ -0,0 +1,81 @@
...
+        [15, 10],
+        [17, 12],
+      ],
+    },
+    "icon-image": poiIcon,
+    "icon-size": 0.5,
+    "text-field": label.name_en,
+    "text-variable-anchor": ["left", "right", "bottom"],
+    "text-justify": "auto",
+    "text-radial-offset": 0.75,
```
1ec5, line 72: This offset should be larger. As it is, the text seems to be flush with the icons. Also try tightening up the `text-height` to keep the icon-label combination looking like one coherent element on the map.

src/layer/poi.js
```diff
@@ -0,0 +1,81 @@
...
+    "parking",
+    17,
+    18,
+  ],
+];
+
+const iconColor = [
+  "match",
+  ["get", "class"],
+  ["bar", "beer"],
```
1ec5, line 35: These classes get repeated several times in the same file. It doesn’t seem likely that these constants will ever get reused. Instead, inline these constants, but pull out different constants like `consumerClasses` that can be reused in the filter and color expressions, if not also the icon expression.

src/layer/poi.js
```diff
@@ -0,0 +1,81 @@
...
+};
+
+export const poi = {
+  id: "poi",
+  type: "symbol",
+  paint: labelPaint,
+  filter: poiFilter,
+  layout: {
+    "text-font": ["Metropolis Light"],
+    "text-optional": true, //TODO: disable this for generic (square dot) POIs
```
1ec5, line 59: Some POIs are sufficiently descriptive by their icons even when there isn’t enough room for the label. This includes community anchor institutions but also more mundane things like parking lots. But I’m not certain that it will always apply to every POI class with an icon. What about explicitly saying which classes can appear without a label?

```suggestion
    "icon-optional": ["match", ["get", "class"], ["hospital", "parking"], true, false],
```

/ref https://github.com/ZeLonewolf/openstreetmap-americana/pull/387#discussion_r895006271

1ec5 (commented):
Do we need anything in the way of documentation? For example, the contributor’s guide could explain the POI color categories and lay out some initial criteria for POI classes at each zoom level.

src/layer/poi.js
ZeLonewolf: I've inlined the constants, however I don't see the class constants being re-usuable.  I think zoom decisions are independent of category.

src/layer/poi.js
ZeLonewolf: Done :+1:

ZeLonewolf:
@1ec5 :
Normalizing icon sizes:

image
image
image

zekefarwell:
This is looking good.  I like the hospital icon, but I find the cross icon for other medical facilities looks inconsistent with it.  I think changing the cross to white inside a blue round rect would unify both icons as the same general category.

CONTRIBUTING.md
```diff
@@ -287,3 +287,31 @@ This style strives to draw representative highway shields wherever they are tagg
 - Shields with a stacked ref configuration, with `/` separating the two lines of text in the `ref` value. Currently, these `ref` values are displayed verbatim on one line, and the code necessary for stacked ref rendering has not been written yet ([#366](https://github.com/ZeLonewolf/openstreetmap-americana/issues/366)). Such cases include:
   - **Italy "Diramazione" (branch) motorways**. Between their main autostrade "A" roads, the Italian motorway network has branch motorways which carry the name of both highways that they connect. For example, the A7 and A26 motorways have a branch motorway named A7/A26, which is correctly tagged `ref=A7/A26` and shown on shields with the two numbers stacked vertically.
   - **West Virginia County Routes**. The West Virginia Department of Transportation posts County Routes, which can have shields with two stacked numbers. For example, in Mercer County, County Route 460/1 is a branch off U.S. Route 460, and County Route 27/6 is a branch off County Route 27. These routes are correctly tagged `ref=460/1` and `ref=27/6` respectively, and shown on shields with the two numbers stacked vertically.
+
```
claysmalley, line 290: I referenced the "bottom section of this page" on line 256, which may have been a bad idea, now that we're adding more sections...

ZeLonewolf: Easily fixed

src/layer/poi.js
```diff
@@ -0,0 +1,74 @@
...
+      ["bar", "beer"],
+      "poi_martini_glass",
+      ["clinic", "doctors"],
+      "poi_health_cross",
+      "hospital",
+      "poi_hospital",
+      "parking",
+      "poi_p",
+      "poi_square_dot", //icon for generic POI, not currently used
+    ],
```
1ec5, line 61: scripts/taginfo_template.json should contain an entry for each of the tags that correspond to the POI types added in this PR.

Ideally, taginfo.json could just run this code and automatically figure out how the `subclass`es map to icon names, but the `subclass`es differ from OSM tags to some extent. It doesn’t look like [this YAML file](https://github.com/openmaptiles/openmaptiles/blob/82b00c13897e46b9d4f39ef0a0ac39b09ebc46ad/layers/poi/poi.yaml#L19-L98) offers a way to map a `subclass` back to the original OSM tag, just the value but not the key.

ZeLonewolf:
Blowing the dust off this PR.  Current state:

image
image
image
image

I see that @zekefarwell made a suggestion to make the cross icon consistent with the hospital "H" icon, so I'm planning to experiment with that next.

CONTRIBUTING.md
```diff
@@ -266,7 +266,7 @@ Additionally, **`refsByWayName`** is an object mapping way names to text that ca
 
 `refsByWayName` only works if there is no `ref` tag and the expression in the `routeConcurrency` function in layer/highway_shield.js includes the `name` property in the image name. The network needs to be listed as an input value that causes the `match` expression to append `name` to the image name.
 
-When using `overrideByRef` or `refsByWayName`, make sure to add a line to the bottom section of this page explaining why it is necessary, as they are only intended for use in special cases.
+When using `overrideByRef` or `refsByWayName`, make sure to add a line to the Special Cases section of this page explaining why it is necessary, as they are only intended for use in special cases.
```
ZeLonewolf, line 269: I added a bunch of text to the contributing guide, so this section is no longer on the "bottom"

src/layer/poi.js
```diff
@@ -0,0 +1,74 @@
...
+export const poi = {
+  id: "poi",
+  type: "symbol",
+  paint: {
+    "text-halo-color": "#ffffff",
+    "icon-halo-color": "#ffffff",
+    "text-halo-width": 1.5,
+    "icon-halo-width": 0.4,
+    "text-halo-blur": 1,
+    "icon-halo-blur": 0.2,
```
1ec5, line 13: I think an opaque, land-colored knockout would feel more consistent than a blurry white halo, which also makes this non-bold, light-colored text pretty difficult to read.

src/layer/poi.js
```diff
@@ -0,0 +1,74 @@
+import * as label from "../constants/label.js";
+import * as Color from "../constants/color.js";
+
+export const poi = {
+  id: "poi",
```
1ec5, line 5: In legend_config.js, add a new “Points of interest” section to `sections` that contains an entry for each of the distinct icons rendered by this layer. Each entry’s filter would likely be based on the `class` feature property.

src/layer/aeroway.js
```diff
@@ -15,8 +15,8 @@ const iconLayout = {
     "match",
     ["get", "class"],
     "military",
-    "military_airport",
-    "airport",
+    "poi_military_plane",
+    "poi_plane",
```
1ec5, line 19: Several entries in the taginfo project file template need to be updated:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/d12a7486607dcf128b6d55d6acc616e6cabecf33/scripts/taginfo_template.json#L132

src/layer/poi.js
1ec5: There may be some overlap between the legend’s needs and the taginfo project file’s needs.

src/layer/poi.js
ZeLonewolf: I came to the same conclusion and implemented this!

src/layer/aeroway.js
ZeLonewolf: Updated 3 entries for military airports.

ZeLonewolf:
Updated screen shot:

image

CONTRIBUTING.md
```diff
@@ -304,6 +304,37 @@ This style strives to draw representative highway shields wherever they are tagg
...
+| Infrastructure         | 294             | <img src="doc-img/pantone_294.svg" height=18 width=50 /> Blue               | 0 63 135    | #003f87     |
+| Consumer               | 717 C           | <img src="doc-img/pantone_717C.svg" height=18 width=50 /> Orange            | 186 82 5    | #d65c05     |
+| Outdoor                |                 | TBD (green?)                                                                |             |             |
+| Attraction             |                 | TBD (brown?)                                                                |             |             |
+| Transportation         | Medium Purple C | <img src="doc-img/pantone_medium_purple_c.svg" height=18 width=50 /> Purple | 78 0 142    | #4e008e     |
+| Knockout               |                 | <img src="doc-img/background.svg" height=18 width=50 /> Lt Grayish Orange   | 249 245 240 | #f9f5f0     |
+
+POIs without a background fill should have a 2px stroke using the "knockout" color above.
+
 ### Shield Test Gallery
```
1ec5, line 338: Should this section come right after the other shield-related information?

src/constants/color.js
```diff
@@ -25,12 +25,14 @@ export const airportOutline = "hsl(250, 41%, 79%)";
 export const airportRunway = "hsl(250, 41%, 79%)";
 export const airportLabel = "hsl(250, 71%, 29%)";
 
+//TODO - rename this variable to "palette"
 export const shields = {
```
1ec5, line 29: Maybe we should keep the shield and POI color constants separate, so we have flexibility to make one less intense than the other, for example. I think it’s fine to have `poi` reference `shield`, but `strong_orange` doesn’t need to be part of the overall palette if it’s specific to POIs.

src/constants/color.js
```diff
@@ -25,12 +25,14 @@ export const airportOutline = "hsl(250, 41%, 79%)";
...
   black: "black",
   blue: "#003f87", // Pantone 294
   brown: "#693f23", // Pantone 469
   green: "#006747", // Pantone 342
   orange: "#f38f00", // Pantone 152
+  strong_orange: "#d45d00", // Pantone 717 C
```
1ec5, line 35: For what it’s worth, #576 uses `#d93f27`, which comes from the California MUTCD, so it’s designed to be used in conjunction with the other colors in this palette.

src/js/legend_config.js
```diff
@@ -35,6 +36,10 @@ export const sections = [
     name: "Route markers",
     source: "Wikidata",
   },
+  {
+    name: "Points of Interest",
```
1ec5, line 40: Should we move this section below Aviation to keep the transportation-related sections together?

1ec5, line 40: The legend uses Sentence case throughout. I realize this is a departure from many print maps, but since the legend is a floating panel separate from the map, I think it’s more relevant that sentence case is more common in UIs these days.

```suggestion
    name: "Points of interest",
```

If we want to use Title Case, then we’d need to install a library to intelligently title-case Wikidata labels, but to my knowledge such libraries only exist for English and would be inappropriate in other languages like Spanish and French anyways.

src/layer/poi.js
```diff
@@ -0,0 +1,103 @@
...
+      15,
+      ["bar", "beer"],
+      16,
+      ["clinic", "doctors", "parking"],
+      17,
+      99,
+    ],
+  ],
+  layout: {
+    "text-font": ["Metropolis Light"],
```
1ec5, line 46: These labels should use the same font as the rest of the map to ensure better language coverage.

```suggestion
    "text-font": ["OpenHistorical"],
```

src/layer/poi.js
```diff
@@ -0,0 +1,103 @@
...
+      Color.poi.infrastructure,
+      Color.poi.infrastructure,
+    ],
+  },
+  filter: [
+    ">=",
+    ["zoom"],
+    [
+      "match",
+      ["get", "subclass"],
```
1ec5, line 35: The `text-color` property matches on the feature’s `class` property, while the filter matches on the feature’s `subclass` property, listing out a similar set of values. Can we stick to one of these properties, or do we need the extra specificity of `subclass`?

src/layer/poi.js
```diff
@@ -0,0 +1,103 @@
...
+    "text-padding": 0,
+    "icon-allow-overlap": false,
+  },
+  source: "openmaptiles",
+  "source-layer": "poi",
+};
+
+export const legendEntries = [
+  {
+    description: "Bar or Pub",
```
1ec5, line 84:
```suggestion
    description: "Bar or pub",
```

src/layer/poi.js
```diff
@@ -0,0 +1,103 @@
...
+    layers: [poi.id],
+    filter: ["in", ["get", "subclass"], ["literal", iconDefs.bar]],
+  },
```
1ec5, line 87: I think this section should be ordered from rarest / largest service area to most common / smallest service area. Community anchor institutions (#435) would end up first, while “Bar or pub” and other storefront POIs would end up last.

src/layer/poi.js
```diff
@@ -0,0 +1,103 @@
...
+  {
+    description: "Hospital",
+    layers: [poi.id],
+    filter: ["==", ["get", "subclass"], iconDefs.hospital],
+  },
+  {
+    description: "Medical facility",
```
1ec5, line 94: Seems odd to have “Medical facility” directly follow “Hospital”. Maybe it could say “Doctor’s office or clinic” more specifically?

src/layer/poi.js
```diff
@@ -0,0 +1,103 @@
...
+      ["bar", "beer"],
+      16,
+      ["clinic", "doctors", "parking"],
+      17,
+      99,
+    ],
+  ],
+  layout: {
+    "text-font": ["Metropolis Light"],
+    "icon-optional": false,
```
1ec5, line 47: I wonder if we should enable `text-optional` at slightly lower zoom levels. Sometimes I think icon-only POI icons can be a bit of a tease, but they might make sense at a zoom level where POI icons are pretty dense, as long as we can ensure they show up consistently without too random collisions. 🤔

src/layer/poi.js
ZeLonewolf: I was worried about keeping the description text short, but your suggestion seems fine.

src/layer/poi.js
ZeLonewolf: For background, when a POI class doesn't have `subclass`, `subclass` just repeats what's in `class`.  I haven't studied the POI structure comprehensively enough to know if it's actually sensible to apply styling at the `class` level versus `subclass`.  We'll have the most flexibility if we go all-in on `subclass` but that might come at the expense of cases where we can group an entire `class` together with common styling.

src/layer/poi.js
ZeLonewolf: Resolved in 107e5be.

src/js/legend_config.js
ZeLonewolf: Resolved in cf9dbf1.

src/constants/color.js
ZeLonewolf: Screen shot with the CA MUTCD vivid orange labels but with the current color icons:

image

I think the California orange doesn't have enough contrast and we want a darker hue?

1ec5: You’re right, though I guess that means this burnt-orange color choice will end up being specific to labels (or icons color-coordinated with them) rather than part of a style-wide color palette that everything adheres to.

src/layer/poi.js
1ec5: If we’re mostly concerned with cases where `subclass` is missing, then a `coalesce` expression could split the difference.

src/constants/color.js
ZeLonewolf: Here's where I am with the burn orange:
image

Should we explore other colors in the palette or try to tweak this or is this good enough?  I also envision that brown will be used for recreational things and so I'm wary about picking a color too close to that.

claysmalley: Burnt orange [as my alma mater defines it](https://brand.utexas.edu/identity/color/) is `#bf5700`. I tried it out, and it's still recognizably orange, but a little darker and arguably more legible:

Screenshot from 2023-01-10 18-21-20

1ec5: You got me craving a Texas navel orange, burnt.

I think any brown color we use as a foreground color would end up considerably darker, so I’m not too concerned with a collision.

ZeLonewolf: For quality assurance, I did an in-person survey and found this flavor of orange to work just fine:
image

src/layer/poi.js
ZeLonewolf: It makes sense, especially for certain ones like hospital.  This might be an optimization that's easier to work with once we have a better spread of icons to examine, though.

CONTRIBUTING.md
ZeLonewolf: Good catch, fixed this.

src/constants/color.js
ZeLonewolf: Resolved

1ec5 (approved):
We’ve come a long way since June. The selection of POI types is rather absurdly minimalistic, but let’s get this layer into users’ hands to give a sense of what’s possible and iterate further in separate PRs.

ZeLonewolf:
image

ZeLonewolf:
image

PR #395, opened 2022-06-12
Adjust pill shield padding

claysmalley:
Removes special casing and defines `pillShield()` as a rounded rectangle with radius 10, with identical text constraints to `ovalShield()`.

before/after:
Screenshot from 2022-06-11 20-40-17 Screenshot from 2022-06-11 20-39-54


ZeLonewolf (approved):
Looks good, tested in Kentucky, no visible degradation, no errors reported

image

PR #398, opened 2022-06-12
Updated CONTRIBUTE.md to reflect that project can run on Windows

bgo-eiu:
From what I can see the previous information that the project cannot run on Windows due to dependency issues was resolved in #374. I cloned and installed the dependencies, and was able to run the project on Windows without issues just now, so this is just an update to the CONTRIBUTE.md to reflect that this is fixed.

I added a note about telling npm where to find `bash.exe` since that can cause issues on `npm start`.


claysmalley:
`npm run code_format` will fix the lint check.

Someone with Windows on their machine will have to review this one :sweat_smile:

jleedev:
I've heard of https://www.npmjs.com/package/shx for writing script commands (such as `rm`).

bgo-eiu:
Ran the `code_format` script, looked like I just left some trailing whitespace in the markdown file

The shx package is good to know about if it makes it any more convenient to run on windows, although I'm not sure how extensively the scripts in the project rely on bash

zekefarwell:
I tested this out on Windows and it works fine, but I figured I'd see how adding shx works in #403 too.  This makes it so the project can be run in PowerShell, the standard Windows CMD prompt, a Windows bash emulator, or WSL.   With that added, I think we can simplify the Windows specific notes to basically just say "it works on Windows", or even just remove them.

zekefarwell:
I think with #374 merged, we can remove the second about Apple Silicon/ARM Macs too.  @adamfranco can you confirm that workarounds are no longer necessary there?

ZeLonewolf (changes requested):
Please update to be consistent with #403.

When merged, it will be possible to build on Windows as long as npm is installed.

bgo-eiu:
Now updated. I opted to just say it now runs on Windows, with parenthetical links to the issues which addressed. (It may still be helpful for people who have seen the previous Contributing page to see exactly what changed to make it work.)

zekefarwell (approved):
Looks good to me.  Thanks!

CONTRIBUTING.md
```diff
@@ -59,8 +59,10 @@ project through Rosetta][63].
 
 ### Windows
 
-Running the project natively on Windows is not currently possible due to [dependency issues][61].
-Instead running on [Ubuntu](#ubuntu-linux) via [Windows Subsystem for Linux][60] is recommended.
+The project can be run from any Windows command line environment provided that Node.js
+and npm are installed. ([Dependency issues][61] which prevented Windows compatibility
+have now been resolved, and the project build scripts have been [updated to no longer
+require a bash analogue][64].)
```
ZeLonewolf, line 65: I think everything after "installed." can be deleted, along with refs 61 and 64.  We don't really need to point out the history of how we got here.

bgo-eiu:
Alright then - removed to keep things simple.

ZeLonewolf (approved):
Looks good to me 🎉

PR #400, opened 2022-06-12
Add AT Schnellstraße shield

gebner:
Adds shields for Austrian Schnellstraßen (S1, etc.); these look the same as the existing shields for the Autobahnen (A1, etc.).  Other road networks (B, L) do not have a `network=` tag.

This is how the shields look like on a street sign:


claysmalley (approved):
Danke schön!

ZeLonewolf:
Looks good here:

image

Not sure what's going on with that toll road tagging though!

ZeLonewolf (approved):
Thanks @gebner for your contribution!

PR #401, opened 2022-06-12
Add Belgium shields

Blijbol:
Probably the most boring country? Only the N-road shields actually differ from the default shields of this project.

I skipped the Bretelles (B-roads) because none have been mapped yet. Of course these too look the same as default shields anyway.

Photos of road signs: https://www.wegenwiki.nl/Bewegwijzering_in_Belgi%C3%AB

Schermafdruk van 2022-06-12 17-02-58

### DIFF
--- a/style/js/shield_defs.js
+++ b/style/js/shield_defs.js
@@ -2548,6 +2548,20 @@ export function loadShields(shieldImages) {
     35
   );
 
+  // Belgium
+  shields["BE:A-road"] = roundedRectShield(
+    Color.shields.white,
+    Color.shields.black
+  );
+  shields["BE:N-road"] = roundedRectShield(
+    Color.shields.blue,
+    Color.shields.white
+  );
+  shields["BE:R-road"] = roundedRectShield(
+    Color.shields.white,
+    Color.shields.black
+  );
+
   // Bulgaria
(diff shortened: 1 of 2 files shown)


style/js/shield_defs.js
```diff
@@ -2529,6 +2529,20 @@ export function loadShields(shieldImages) {
...
+  );
+  shields["BE:N-road"] = roundedRectShield(
+    Color.shields.blue,
+    Color.shields.white
+  );
+  shields["BE:R-road"] = roundedRectShield(
+    Color.shields.white,
+    Color.shields.black
+  );
+
```
claysmalley, line 2545: Let's keep this section sorted by country code. So `be` would go between `ba` and `bg`.

Blijbol: My bad, fixed in force-push.

claysmalley (approved):
Dank u (et merci)!

ZeLonewolf:
Shields that are intentionally black on white should still be explicitly declared, especially now that we are considering #402 which proposes to change the default fallback shield. This can happen in this PR or in a separate one.

Blijbol:
@ZeLonewolf This PR does indeed declare them. (except for the not-yet-mapped B-roads)

Issue #302, opened 2022-04-25
Wide gap in Oklahoma outline for 1-digit routes


claysmalley:
If we want to close the gap, I can think of two solutions that might come in handy in the future:

* Allow a shield definition to constrain a particular image to a particular ref length
* Allow a shield definition to apply a casing around the ref, just like we do (unconditionally) for banner text

_Originally posted by @1ec5 in https://github.com/ZeLonewolf/openstreetmap-americana/pull/296#discussion_r856948851_

PR #402, opened 2022-06-12
Add support for text halo in shields

### LINKED ISSUE #363: Consider changing default shield design
Maybe we should make the default something other than the black and white rounded rectangle so it's clearer when the default fallback is being used vs routes that intentionally use that style.

_Originally posted by @zekefarwell in https://github.com/ZeLonewolf/openstreetmap-americana/issues/362#issuecomment-1146213310_

### LINKED ISSUE #367: Support adding text halos on a per-shield basis
It would be helpful to have the capability to define a halo color and width for shield text. This would fix #302 and enable us to support one popular solution to #363.

claysmalley:
Fixes #367.

Adds support for text halo in shields.

Screenshot from 2022-06-12 15-34-37


ZeLonewolf:
Pros: forces specific designation of black on white shields rather than relying on the default, which ripples out to taginfo to indicate that the tagging is used by a data consumer.  Gives clear mapper feedback that there's a "problem" that needs to be resolved.

Cons: looks like a rendering glitch rather than a deliberate design choice.  Might be too strong feedback in cases where strategic ambiguity might be better, such as cases of non-prefixed route relations in countries that haven't come to a firm consensus on tagging.

Blijbol:
I like this, it's more generic (no assumptions about shape and background colour) and it's consistent with the banner texts.

claysmalley:
Fixes #302.

Screenshot from 2022-06-12 16-03-44

claysmalley:
Text halo color is now identical to background color, for a knockout effect:
Screenshot from 2022-06-13 18-37-23

ZeLonewolf:
The fixes for #302 and #367 should be uncontroversial, and it may be worth separating those changes into its own PR.

claysmalley:
Support for text halos in shields is a prerequisite for this styling, so I've reverted the default shield, and that change will be in a subsequent PR.

1ec5:
> Cons: looks like a rendering glitch rather than a deliberate design choice. Might be too strong feedback in cases where strategic ambiguity might be better, such as cases of non-prefixed route relations in countries that haven't come to a firm consensus on tagging.

Free-floating text can look like a deliberate design choice. In #172, we styled exit numbers as light text surrounded by a thick, dark outline, which ensures contrast and gives the labels enough weight to appear anchored on the roadway. A similar treatment, but larger and white instead of yellow, could look attractive without stealing the show from actual shields.

A distinct treatment for unrecognized networks would also have the benefit of avoiding confusion with overlapping networks that are actually signposted as white rectangles, such as Landesstraßen in Germany and Austria or, occasionally, unclassified roads in the UK.

michaelblyons:
I'm not opposed to the floating number, but I did like the grey text in a white box. (That may be too low-contrast. I'm not an accessibility expert.)

The other thing I have imagined doing is 45&deg; stripes of white and pale grey (or some other colors) as the background of a box. Something light enough that it's not distracting, but noticeable for someone who is looking for shields to add.

ZeLonewolf:
Does `US:FL` also need the same treatment?

src/js/shield_canvas_draw.js
```diff
@@ -86,8 +86,15 @@ export function rectangle(ref) {
   );
 }
 
-export function blank() {
-  return Gfx.getGfxContext({ width: 1, height: 1 });
+export function blank(ref) {
```
ZeLonewolf, line 89: Parameter unused

claysmalley: Are you sure? `ref` is right there on line 91.

ZeLonewolf (approved):
Looks good as far as I can tell and OK shields look right.

claysmalley:
`US:FL` does not need this treatment, because the outline of Florida is always broken in the same place. Oklahoma's outline is broken wherever text happens to collide with it.

PR #403, opened 2022-06-12
Add shx to make npm scripts run on Windows CMD and PowerShell

zekefarwell:
This PR adds [shelljs/shx](https://github.com/shelljs/shx) to make project scripts platform independent instead of requiring a bash shell.  The primary purpose of this is to it easier to run the project on Windows.  The only Windows requirement after this should be to install Node.js (and git obviously).


claysmalley:
I don't have Windows, so unless we need to test that it works on Linux, there's nothing I can do.

zekefarwell:
Yup just wanting to make sure the change doesn't cause any problem for other platforms

claysmalley:
Build is successful on Linux. Everything's working :smile:

zekefarwell:
Oh right, @Pengor is another Windows user who can test this

Pengor (approved):
Yep, builds and runs on Windows (checked with both `pwsh` and `cmd`) 👍

ZeLonewolf (approved):
Works great on Windows CMD and WSL.

ZeLonewolf:
There are conflicts here because some jerks decided to reorganize the repository 😂

zekefarwell:
should have merged this simple change first 😄

jleedev:
Should be good.

PR #406, opened 2022-06-12
Unload the style/ directory into the root

### LINKED ISSUE #386: Collapse style sub-folder into the project root
As noted by @zekefarwell in https://github.com/ZeLonewolf/openstreetmap-americana/issues/344#issuecomment-1137694371 and in other discussions, there is support for removing the style sub-folder and placing its contents into the project root.

The original intent of this structure was to have the style and (future) vector tile schema exist as two separate sub-folders under this repository.  However, there seems to be a consensus that this can be structured as separate projects and managed by versioning rather than in a monorepo.

This will require some care to ensure that the deployment and all the other moving parts still function after the change.

In addition, we should resolve the currently open PRs to the extent possible prior to making this change in order to reduce impact on contributors.

jleedev:
Retain a src/ directory to contain the html and javascript code.
Adjust paths everywhere I could.
Fixes #386


ZeLonewolf:
😱  image

jleedev:
> 😱 image

```
 ~/pkg/openstreetmap-americana ¶ git diff origin/main | grep similarity | huniq -cs 
1 similarity index 85%
1 similarity index 94%
1 similarity index 99%
1 similarity index 89%
1 similarity index 73%
1 similarity index 95%
197 similarity index 100%
```

.github/workflows/deploy.yml
```diff
@@ -17,7 +17,6 @@ jobs:
           node-version: 17.x
       - name: Install and Build 🔧
         run: |
-          cd style
           npm ci --include=dev
           cp configs/config.aws.js config.js
```
ZeLonewolf, line 21: configs is currently in `src/configs`, so either this needs to change or the folder location needs to change

ZeLonewolf: Or is this run from within src?

jleedev: Nope, and this is exactly the sort of change that's impossible to test.

Fixed.

jleedev:
> Nope, and this is exactly the sort of change that's impossible to test.

Well, I think it's possible to write a single action that performs the build on any push or PR, and then conditionally deploys only if the action was triggered by being merged into main.

Would reduce the "write shit that behaves differently in prod" factor.

ZeLonewolf (approved):
Works locally and looks like it "should" deploy.  Might be some minor doc cleanup remaining but nothing that should stop us moving forward.  May as well rip off the bandaid.

Good luck!

image

jleedev:
http://butt.holdings

jleedev:
66ccaf01-5b02-4372-9cdb-a3019c95eeaa_text

PR #414, opened 2022-06-13
More detailed Vermont shields

### LINKED ISSUE #322: Consider changing Vermont shield to better match real world design
When the VT shield graphic was originally added early on in the project, the design goal was for shields to be significantly simplified symbols, representative of the real shield signs.  Since then a lot of shields have been added that are much more detailed and closer to their real world counterparts.  The current VT shield graphic (below) looks a bit oversimplied in comparison now.

Screen Shot 2022-05-14 at 1 36 48 AM

I think it would be appropriate to add some more detail to bring it closer to the real world design (below).

  

  

## Paper highway map examples

Here are some examples of how various paper highway maps symbolize VT route shields.

AAA, 2021

Delorme, 2007

Jimapco, 2014
 

Michelin, 2022
  

Rand McNally, 2022

Vermont Department of Tourism, 2021

Most of these maps just use plain ovals. The VT tourism map does use the chopped oval/bulging rectangle shape, although with red or black instead of green.  Jimapco is the only one with a similar style of real world shield representation to what we are going for in this project.  Here's another Jimapco clip from the VT/NY border showing a very similar NY shield to ours.  

## Ideas for this style

Thanks to @claysmalley for creating these test renders of some different options!  Each option is shown @1x and @2x.

One of our design guidelines is to remove the background when a shield design is non-rectangular shape inscibed within a rectangular sign.  Following this guideline gives us something like this.  It captures the essence, but feels a bit awkward.

Screenshot from 2022-05-13 08-55-30 

Reducing the width of the top bar slightly is an improvement.

Screenshot from 2022-05-13 08-58-03      

Including the full green rectangle behind the white shapes looks very much like the Jimapco style.

Screenshot from 2022-05-13 08-58-03  

An in-between approach.

Screenshot from 2022-05-13 10-05-01  

And for comparison the current shield.

Screen Shot 2022-05-14 at 12 03 16 AM  Screen Shot 2022-05-13 at 11 57 58 PM

There has been some [discussion on Slack](https://osmus.slack.com/archives/C01V02K52UX/p1652446152362849?thread_ts=1652327308.600869&cid=C01V02K52UX) of these options, but a consensus has not yet emerged on whether we should make a change and if so which design to use.

zekefarwell:
This PR implements the Vermont shield design [I proposed here](https://github.com/ZeLonewolf/openstreetmap-americana/issues/322#issuecomment-1132410292), since I didn't hear any opposition.  The buldge on the lower white section is reduced, making it much closer to the Maryland shield MD, while still keeping the suggestion of roundness.  This minimizes the "background", making it more of a variable width border.  Closes #322.

**1x**
Screen Shot 2022-06-12 at 10 03 55 PM

**2x**
Screen Shot 2022-06-12 at 10 03 45 PM


ZeLonewolf:
Vermont shields:
image

ZeLonewolf:
Definitely a more "Vermont shield" feel on the map over the previous icon.  I'm not crazy about the background being included as a _general_ rule, but in this case I don't see any other way to capture the look and feel of the shield without it.

adamfranco (approved):
Thanks, Zeke. Looks good to me!

claysmalley (approved):
Looks good. The Vermont shield is kind of exceptional and it breaks a lot of our assumptions about simplifying shields. I think this is an acceptable compromise.

Issue #396, opened 2022-06-12
Shields of Pakistan


bgo-eiu:
I am creating this issue to start with seeing as there are issues for various country shields, but intend to look into the shields more, as well as how road routes are mapped in Pakistan, then will make a pull request at some point.

Pakistan has a network of National Highway routes designated by the national government which connect the country's major cities, and a Motorway network of routes installed relatively recently with the intent of providing connections between Pakistan's Arabian Sea ports and the Khunjerab Pass at the border with China as part of the "China-Pakistan Belt Road Initiative."

National Highway `ref` designations are prefixed with `N-` and are displayed on a blue hexagonal shield:
![N-10 Pakistan National Highway Shield](https://upload.wikimedia.org/wikipedia/commons/thumb/6/61/Pakistan_N-10.svg/750px-Pakistan_N-10.svg.png)

On guide signage they may look like this:
![N-5 Shield on Guide Signage in Pakistan](https://upload.wikimedia.org/wikipedia/commons/4/4f/Road_sign_in_Islamabad.JPG)

The hexagon outline is not always shown; as you can see on this sign, N-35 (Karakoram Highway) is referenced without the hexagon shape (not specific to the route or carrying any different meaning, just omitted on some signs):
![N-35 Reference Number on Guide Signage in Pakistan](https://upload.wikimedia.org/wikipedia/commons/0/0b/Sign_for_foreigners_Karakoram_Highway_Sep_2004.jpg)

Motorways in Pakistan have `ref` designations prefixed with `M-` and green semi-circular shields:
![M-10 Pakistan Motorway Shield](https://upload.wikimedia.org/wikipedia/commons/thumb/2/29/Pakistan_M-10.svg/550px-Pakistan_M-10.svg.png)

They are often shown on matching green guide signage like this:
![M-2 on Guide Signage in Pakistan](https://upload.wikimedia.org/wikipedia/commons/1/13/M2_Sargodha.jpg)

Pakistan's provinces and territories also each have provincial or territorial highway networks. Their respective subdivisions assign different systems of `ref` designations for each, some of which do have their own shield designs but not all. The available images of signage for these is sporadic compared to the National Highways so it can be harder to discern, but here is an example a shield for the L-20 route, named Lahore Ring Road, of Punjab's provincial highway system:
![L-20 Lahore Ring Road, Punjab, Pakistan Shield](https://upload.wikimedia.org/wikipedia/commons/f/f9/L-20_Logo.png)

[...]

PR #415, opened 2022-06-13
Blue hexagon shields for Pakistan National Highways

bgo-eiu:
As this shape is already included in the repository for Japanese prefectural shields, this PR changes the name of the SVG file for the blue hexagonal shield shape from a country specific description to `hexagon_blue`. The `shield_defs.js` file has then been updated to reflect these changes for Japanese prefectural shields, and to include the Pakistan National Highway shields described in #396 using the same parameters.


bgo-eiu:
Example, N-5 at Torkham 
image

claysmalley (changes requested):
We'll need a better naming scheme to distinguish horizontally-oriented and vertically-oriented hexagon shields. I would recommend adding `horizontal` to the blue hexagon shield's filename. I gave the following shield the name `shield40_hexagon_wide_green.svg`, but `wide` should probably be changed to `vertical`:



Text inside this shield is very small when it has 4 or more characters. The photos in #396 indicate that the hexagon can be horizontally expanded to accommodate longer text, so let's make a variant of this SVG with a width somewhere between 25 and 30 pixels, whatever looks right:

Screenshot from 2022-06-13 08-54-00

ZeLonewolf:
Horizontal means the pointy ends east and west, right?

claysmalley:
That's what I'm thinking. Two of the six sides of the hexagon run horizontally.

claysmalley:
Here's a wider hexagon I think matches the real-life signs a little better:

shield40_hexagon_horizontal_blue_3

Don't worry about renaming the files; I'll refactor it all eventually.

bgo-eiu:
Oh thank you! I've been waiting to have time to make a green semi circle SVG for the motorways and one for Lahore Ring Road. I heard back from the active Pakistani route mapper and they said the tags I suggested for those are fine so long as I don't change any of the tags that relate to them being controlled access highways.

ZeLonewolf:
Setting to draft pending shield update

claysmalley:
Here's a motorway shield:

shield40_pk_motorway

I played around with the Lahore Ring Road shield and I couldn't manage to make the text big enough to be legible. This route may need an `ovalShield` or `pillShield` treatment.

claysmalley:
Shields are up and running:

Screenshot from 2022-07-13 18-07-34

1ec5 (approved):
This looks good to go. We can continue to track provincial and district roads in #396.

claysmalley (approved):

PR #416, opened 2022-06-13
Switch from parcel to esbuild

jleedev:
- Pro: Improved cross platform compatibility - in particular, freebsd. (Also has a wasm package, but you must load it specifically, and it's around 5 times slower)
- Pro: Extremely fast
- Con: Doesn't process html files
  - → css import is moved to the js file and we copy the html file to the output directory on our own
- Con: Esbuild's built-in development server doesn't do hot reloading
  - → wrote something based on https://www.npmjs.com/package/esbuild-serve
- Note: Has similar behavior to parcel that if the page crashes due to compile errors, live reloads are lost and you need to manually reload. (I think? I hadn't played with parcel's server much, preferring my usual http server.)
- Note: esbuild uses polling instead of watching (probably more cross platform)

Loses the parcel feature that shows nice error message overlays - js compile errors are printed to the terminal

Seems to reload multiple times when e.g. vim renames your file to a backup file and writes the file anew, which is less than ideal, although parcel does this too.

Not immediately obvious to me, but it's how they both work, the build target as written is simply "americana.js" but this produces both "americana.js" and "americana.css" by virtue of the codebase importing css somewhere.

It would probably be possible to write a development server that writes a live-reloading index.html but delegates to esbuild's server for everything else. Might be better? We're writing temporary files to dist/ instead of serving them live, e.g.

Verify that the output is comparable in terms of bundle size e.g.

```
  dist/americana.js       801.3kb
  dist/americana.css       68.0kb
  dist/americana.js.map     1.4mb
  dist/americana.css.map   71.7kb

⚡ Done in 1082ms
```

seems like it is.

To verify:

- [x] Developer experience aligns well enough
- [x] Hacked together dev server is functional enough
- [x] Output bundle is comparable in size
- [x] process.argv / import.meta.url hack is cross platform enough
- [ ] Slightly altered build changes are correct wrt css import, html main, etc


ZeLonewolf:
Is the error message shown in some other place, or is it lose entirely?

jleedev:
Compile errors appear in the terminal.

scripts/build.js
```diff
@@ -0,0 +1,35 @@
+import { copyFile, mkdir } from "fs/promises";
+
+import esbuild from "esbuild";
+
+const isMain = import.meta.url === new URL(`file://${process.argv[1]}`).toString();
```
jleedev, line 5: This here is a bit of a hack and I'd like confirmation that it works on windows.

ZeLonewolf:
First thing I noticed is that my web browser didn't auto-launch after an `npm start`, though it did start the server.

jleedev:
Found that https://www.npmjs.com/package/open is popular and straightforward.

ZeLonewolf:
The browser open seems to work!  Though it does this one really _weird_ thing.  Before running it, the console looks like this:
image

After running `npm start` it changes my console font to:
image

Now, I can change the console back to the standard font (Consolas 16) but that's kind of weird that it does that.

jleedev:
🧑‍💻🤸🌋

ZeLonewolf (approved):
Confirmed that it appears to be working on Windows & WSL

PR #417, opened 2022-06-14
Åland shields

riiga:
Support rendering of shields on the Åland Islands (Swedish-speaking autonomous region of Finland). I also plan on getting in contact with the Finnish community to see if it is possible to support more shields in Finland.


ZeLonewolf:
This project is a never-ending geography lesson :-D

riiga:
@claysmalley The difference is that Åland has its own road network, owned and maintained by the government of Åland, while in the rest of Finland they belong to the Finnish Transport Infrastructure Agency. Åland also has their own road legislation separate from mainland Finland. For these reasons `AX` make more sense to me.

There are four main highways on Åland, numbered 1-4.
Red-white

There are also a few additional roads that have higher numbers with a different shield.
Blue-white

ZeLonewolf:
There are six `network=AX:*` routes currently mapped, does this represent the entire route network of Åland?  Looks like the entire network was mapped yesterday :)  I would ask for similar outreach to happen to the national communities involved.  I have no desire to instigate an unexpected international incident!

ZeLonewolf:
I should have brought this up in #365, but now would be a good time to pose the question:

For countries with no existing route networks, what standard of consensus should we hold for establishing support for brand-new network/ref tagging schemes outside the US?

riiga:
@ZeLonewolf Yes, that's the entire network to my knowledge. There are few more roads that don't seem to be signposted but have numbers. 

> Looks like the entire network was mapped yesterday :)
Guilty as charged. I doubt anyone would object, given how little mapping activity Åland gets. I contacted the Finnish community yesterday to ask about the mainland, but I can always ask them about Åland too. Unless they are Swedish-speaking though, I think they know less about Åland than some Swedes do due to the language barrier.

claysmalley:
> For countries with no existing route networks, what standard of consensus should we hold for establishing support for brand-new network/ref tagging schemes outside the US?

I think this depends heavily on how big the country and its mapping community is. For small countries, there may be only one or two regular mappers making decisions on tagging, but numbered road networks for such countries tend to be small and straightforward. If you're the first to map road routes in such a country, and you've notified the local community, it's probably fine. Larger countries may or may not have already reached consensus on tagging, and the resulting consensus is not always something we can support in this project.

Some basic criteria for supporting `route=road` relations:

* Is the `network=*` value the country's [ISO 3166-1 alpha-2 code](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes) alone, or prefixed by the country code and a colon `:`?
  * Or, if this is a route network assigned by an international organization (to which country codes would not apply), is the value of `network=*` unique, avoiding conflict with existing values of `network=*`?
* Can shield appearance be derived from `network=*` (with limited, explicitly documented exceptions)?
* Is the `network=*` tagging scheme free of unnecessary complexities?
* Can `ref=*` be displayed directly as the route number on the shield, without preprocessing (e.g. adding or removing a prefix)?

claysmalley (approved):
This is a country with a small population and a straightforward route network. Seems okay to me.

riiga:
I thought it was by country name and Å comes after Z in the (Swedish) alphabet :roll_eyes: 

I've corrected the order which includes my previous addition of the Swedish shields.

claysmalley (approved):
Tack!

ZeLonewolf:
Code looks good, just need to update planet to test it out.

claysmalley:
Looks good!

Screenshot from 2022-06-18 18-55-12

PR #394, opened 2022-06-11
Complete NL shields

Blijbol:
This completes support for road route shields in the Netherlands:

* Adds support for recreational route shields (e.g. R101).
* Adds support for urban RING roads, which are officially considered special city routes and hence signed identically to routes with S number (e.g. S100). As a consequence, Curacao is now also a supported country.
* Resets padding for city route shields to default padding for rectangular shields, since the custom padding had no visible effect.
* Added a link to https://wiki.openstreetmap.org/wiki/The_Netherlands_road_network in the shield code comments for reference.

Preview with new recreational route shields:

Schermafdruk van 2022-06-11 17-40-42

Note: some network=* values are recently added to the OSM database and require a MapTiler update before showing up.

PR #418, opened 2022-06-14
Documentation: fix Curacao support

Blijbol:
For some reason the CW country code disappeared from the map in #406,
but #394 provides full support for Curacao.

### DIFF
--- a/doc-img/shield_map_world.svg
+++ b/doc-img/shield_map_world.svg
@@ -145,7 +145,7 @@ See the end of this file for a list of available jurisdictions and their codes. 
 .md,
 .me,
 .mk,
-.nl,
+.nl, .cw,
 .pl,
 .pt,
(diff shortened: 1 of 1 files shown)


ZeLonewolf (approved):
Thanks for the update!

1ec5:
This will get overwritten the next time someone runs the build scripts. A more durable fix would be to add `cw` to the array of codes passed into this function:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/2ea6017549a1831035201f66331190c69867ce09/scripts/status_map.js#L44-L47

Or hard-code it in the SVG:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/5e32b51d1525aee6570c8980bf5c2e925057a1f3/scripts/blank_map_world.svg#L117

PR #419, opened 2022-06-14
Remove Tennessee from map of supported state shields

claysmalley:
Reverting the US shield map to its state before #406, which filled in Tennessee on accident. The only supported shields in Tennessee are for a single county, with no routes mapped yet.


1ec5 (commented):
This will get overwritten the next time someone runs `npm run build`. Maybe we could instead add a plain white shield for Tennessee primary state routes for now as a stopgap?

claysmalley:
#418 would be overwritten as well.

Is the tagging scheme for Tennessee settled?

1ec5:
> Is the tagging scheme for Tennessee settled?

Good point. The only outstanding question is whether to write “primary” and “secondary” in mixed case or lowercase. Otherwise, I haven’t gotten any adverse feedback to my changes throughout the state (on admittedly and mercifully short routes).

ZeLonewolf:
I'm confused by this discussion and don't understand if this PR is needed or not needed.

claysmalley:
It's not strictly necessary, though I guess a better approach would be for me to just make the shields...

PR #355, opened 2022-05-31
Change shield definitions for county and city highway networks

claysmalley:
Changes definitions of currently supported county networks, and adds new ones, based on the [local highway network tagging guidelines.](https://wiki.openstreetmap.org/wiki/Proposed_features/County,_city_and_local_highway_networks_in_the_United_States)

Todo:
- [x] wait for voting to end on proposal (2022-06-08)
- [x] add county park access roads for Minnesota
- [x] check Louisiana for parish route networks
- [x] check Tennessee for county route networks
- [x] should anything in Ohio change? @1ec5

PR #427, opened 2022-06-16
Add shields for Kansas counties

claysmalley:
Adds shield definitions for counties of Kansas.

Forgot to include this in #355.


ZeLonewolf (approved):
Based on https://taginfo.openstreetmap.org/search?q=network%3DUS%3AKS this is a bit aspirational, but since we've got an approved proposal in place, no problem to support these routes.

PR #431, opened 2022-06-18
Add shield for Michigan M-185

claysmalley:
Adds shields for Michigan State Trunkline Highway 185, notable for prohibiting automobiles. This shield is a diamond like other Michigan State Trunkline Highways, but has a white-on-brown color scheme and additional text.

Screenshot from 2022-06-18 08-33-36

M-185_(MISPC)


ZeLonewolf:
> Adds shields for Michigan State Trunkline Highway 185, notable for prohibiting automobiles

Uh... is this a road route or trail? 🤔

claysmalley:
Even though general motor vehicle traffic is banned from the island, M-185 still gets heavy horse carriage traffic, and is generally built more like a roadway than a trail. In general, I would consider most of the public thoroughfares on Mackinac Island lined with buildings to be streets and not paths, considering some of them have their own sidewalks. They're built to withstand the occasional motorized emergency vehicle, including one police car, two fire trucks, and [one ambulance](https://emsrig.com/index.php/2021/02/23/mackinac-island-mi-ems-gets-custom-wheeled-coach-ambulance/).

ZeLonewolf (approved):
Weird flex, Michigan, but okay

Issue #366, opened 2022-06-04
Shields with stacked refs


claysmalley:
Some highway networks may contain route shields with two lines of equal-size text. Currently, we display these `ref=*` values verbatim on one very long line. There are two known cases, and in each case the top and bottom text are separated by a slash `/` in the `ref=*` value.

The ability to specify that a network may contain slash-delimited stacked refs, and arrange text for such shields in two lines, would make these shields look more polished.

`network=US:WV:County`, `ref=19/24`:
![Circle_sign_19-24](https://user-images.githubusercontent.com/1732117/171970204-b7bfe506-0f2b-4747-a4db-f11b5b96040c.svg)

`network=IT:A-road`, `ref=A8/A26`:
![Autostrada_A8-A26_Italia](https://user-images.githubusercontent.com/1732117/171970203-54e3679e-66a3-458f-a696-21d9aa1a4911.svg)

1ec5:
The long pills in West Virginia don’t look all that bad to me. If we stack the text, it would be more faithful to the signage, but some numbers may be difficult to squeeze into the circle while keeping it from getting too large.

PR #432, opened 2022-06-18
Clarify documentation about special case shields

claysmalley:
~Removes a line of documentation describing "special code" that does not exist.~

Stacked refs in Italian motorways are a thing, but the current implementation displays the `ref=*` verbatim on one line (#366).


zekefarwell:
I don't think the intent with this chunk of text was to imply that the Italian branch motorway refs were currently supported, but just to document them as a special case that would require special handling.  Perhaps rather than "This requires special code ...", better phrasing would be "These are currently not supported.  Potential future support will require special code ...".  Another alternative to simply removing this from the documenation would be splitting the **Special Cases** list into two lists: **Supported** and **Not Supported**.

claysmalley:
Rearranged the special cases section and added the Italy example back in, with clarification that it is not implemented yet.

CONTRIBUTING.md
```diff
@@ -267,10 +267,15 @@ In all cases, banner text should be no more than **4** characters in length.
...
+- Shields for route networks where each individual route is identified by a color, rather than a number or letter. Such cases include:
+  - **Allegheny County, PA Belt Routes**. Shields for this system use colors, with a colored circle and the words "<COLOR> BELT". These shields are drawn as squares with colored circles, with the `ref` values correctly corresponding to the text on the shield. Because of the common design (white shield with colored circle), these shields are properly part of a common route network. Special code is needed to convert the textual ref values to the colors displayed in the shield.
+- Shields with a stacked ref configuration, with `/` separating the two lines of text in the `ref` value. Currently, these `ref` values are displayed verbatim on one line, and the code necessary for stacked ref rendering has not been written yet (#366). Such cases include:
```
ZeLonewolf, line 279: I don't think #366 will link in this way from a markdown file.

claysmalley: Fixed in merge commit, grumble grumble

ZeLonewolf (approved):
Looks good 👍

PR #441, opened 2022-06-20
Add Houston-area toll road shields

### LINKED ISSUE #244: Houston-area toll road shields
Hot off the heels of #205, a map of Texas shields would be incomplete without the ones for city that started its toll road shield craze: Houston.

Dallas may have built the first few toll roads in Texas, and even gave the Dallas North Tollway a [unique shield](http://www.aaroads.com/shields/img/TX/TX19980021i1.jpg) for a few decades (before changing their minds and adopting the state standard as they added more toll roads to their inventory), but the Houston area really took off with toll roads in 1983. The Harris County Toll Road Authority (`network=US:TX:HCTRA`) chose the standard county pentagon for its new roads, though they opted for purple instead of blue and encircled the Authority's logo with the name of the tollway. At the time, they had only two toll roads in the works, which don't intersect each other (Beltway 8's mainlanes are free at the point where it crosses the Hardy Toll Road, and thus it is not a part of the Sam Houston Tollway), so it wasn't a _huge_ deal that the signs happen to be completely illegible at highway speeds, as long as the shape was recognizable.

Fast forward to today: HCTRA has now crossed the ring-like Sam Houston Tollway with 3 more radial tollways, and 3 suburban counties have formed their own toll road authorities. The newest one, the Brazoria County Toll Road Authority (`network=US:TX:BCTRA`, already implemented in #205), was formed after the standard state toll road shield guidelines had taken ground elsewhere, so they made their own that conformed to it. But the other two, the Fort Bend County Toll Road Authority and the Montgomery County Toll Road Authority (`network=US:TX:FBCTRA` and `network=US:TX:MCTRA`, respectively), felt like one-upping HCTRA with shields even less legible.

HCTRA has 5 toll roads, and the only toll roads signed by FBCTRA are extensions of two of these into Fort Bend County. These counties spell out the full name of the toll road on their shields, though each relation is currently tagged with a `ref` value derived from an initialism of the roadway's name. This is a case similar to Kentucky Parkways, where the best solution would probably be to retag these relations without `ref`, and use hardcoded abbreviations for these routes as a special case.

MCTRA has only one toll road, the 249 Tollway. It may be illegible, but fortunately, no special casing is needed here for the shield text, since the `ref` value is present on the sign.

I hope that in the future, the Houston area toll road authorities can figure something out and redesign their shields to be more legible. But god forbid they ever replace them with the state standard, we'd be in deep SHT.

----

`network=US:TX:HCTRA` on top left, `network=US:TX:MCTRA` on top right, `network=US:TX` on bottom row:

`network=US:TX:FBCTRA`:

claysmalley:
Fixes #244.

Also documents and reorganizes special cases a little.

Screenshot from 2022-06-19 21-36-01 Screenshot from 2022-06-19 21-37-09


src/js/shield_defs.js
```diff
@@ -3038,42 +3025,87 @@ export function loadShields(shieldImages) {
...
+    {
+      refsByWayName: {
+        "East Sam Houston Tollway North": "SHT",
+        "East Sam Houston Tollway South": "SHT",
+        "North Sam Houston Tollway East": "SHT",
+        "North Sam Houston Tollway West": "SHT",
+        "South Sam Houston Tollway East": "SHT",
+        "South Sam Houston Tollway West": "SHT",
+        "West Sam Houston Tollway North": "SHT",
+        "West Sam Houston Tollway South": "SHT",
```
1ec5, line 3097: I look forward to eventually being able to use `short_name` on the relation instead of `name` on the way.

src/layer/highway_shield.js
```diff
@@ -19,6 +19,20 @@ function routeConcurrency(num) {
           ["concat", "\n", ["get", "name"]],
           "",
         ],
+        [
+          "match",
+          ["get", "route_" + num],
```
1ec5, line 24: Combine these match expressions with the Kentucky parkway match expression above. The line that says `US:KY:Parkway=` can contain an array of strings instead of a single string.

1ec5 (commented):
Keep up the good work, Texas!

1ec5 (approved):

PR #444, opened 2022-06-20
Remove unused NCM shield for San Juan County, NM

claysmalley:
Removes the 2-digit variant of the pink-on-white pentagon shield, as all NCM routes in San Juan County, NM have 4 or more digits. I'm not aware of anywhere else that uses this shield.

The "NCM" banner above this shield is also removed, as the shield is still recognizable without it.

Screenshot from 2022-06-20 18-05-08


ZeLonewolf (approved):
Looks okay to me.

PR #447, opened 2022-06-24
Improve support for Finnish shields

riiga:
There are four main classes of signed roads in Finland:

Valtatie (fi)/riksväg (sv)/national road (en)

Kantatie (fi)/stamväg (sv)/trunk road (en)

Seututie (fi)/regionalväg (sv)/regional road (en)

Yhdystie (fi)/förbindelseväg (sv)/[link/connection] road (en)

After consulting the Finnish community (with no objections), I've gone through and cleaned up and/or created route relations, mainly for national and trunk roads.

This pull request adds support for all four kinds.

_(I also removed some unnecessary code for the Swedish shields)_


claysmalley:
I see many route relations in Finland have recently been retagged according to this scheme. We'll need a data update before we can test this.

ZeLonewolf:
I've updated the planet.  I wasn't able to find anything below red just by dragging around, so gonna need to find some test areas.

ZeLonewolf:
Note that there is currently a case-sensitivity issue that needs to get resolved:
image

ZeLonewolf:
image

Trunk 95 isn't showing up even though I think it should 🤔

claysmalley:
The `network=*` values with capitalized `FI` are for train stations. Lowercase `fi` is used on all `route=road` relations.

claysmalley:
Here's Turku:
Screenshot from 2022-06-24 20-30-38

@ZeLonewolf are you on the right branch?

ZeLonewolf:
> @ZeLonewolf are you on the right branch?

Nope, I was browsing production :-D  #fail...

PR #453, opened 2022-06-26
Update world shield coverage map

1ec5:
Added Åland (#417). Added Curaçao more durably than in #418. Running `npm run build` will no longer unhighlight Curaçao.


ZeLonewolf (approved):
Thanks!

1ec5:
#454 not only added Åland but also removed Curaçao, so this PR has now been rebased to ensure that Curaçao remains on the map.

Issue #455, opened 2022-06-26
Shields of Ecuador


1ec5:
[Ecuador](https://es.wikipedia.org/wiki/Carreteras_de_Ecuador) identifica sus carreteras con dos discos que parecen mucho similares con el escudo de autopistas interestatales de los EE.UU. Las carreteras primarias o arteriales tienen un disco azúl y los secundarias o colectoras tienen un disco verde. En OSM, estas rutas tienen las etiquetas `network=EC:national` y `network=EC:provincial`, respectivamente. Los números rutales tienen el prefijo “E”, entonces los números pueden ser más largos que aún el número interestatal más largo.

[Ecuador](https://es.wikipedia.org/wiki/Carreteras_de_Ecuador) marks its roads with two shields that look strikingly similar to the Interstate Highway shield in the U.S. Primary or arterial highways get a blue shield, while secondary or collector highways get a green shield. In OSM, these routes are tagged `network=EC:national` and `network=EC:provincial`, respectively. Route numbers in the shield are prefixed with “E”, so the numbers can get longer than even the longest Interstate route number.

1ec5:
`network=EC:provincial` is incorrect on the collector highways, which are still part of the national route network. In fact, Guayas province has its own _provincial_ primary highways: #457.

1ec5:
He pedido en el chat [OSM Latinoamericana](https://t.me/OSMLatam/35015); un usuario allá he reenviado el mensaje a [Telegram de MappingEcuador](https://t.me/mappingEcuador/7972).

Asked in the [OSM Latin America Telegram](https://t.me/OSMLatam/35015) chat; a user there forwarded the message to [MappingEcuador Telegram](https://t.me/mappingEcuador/7972).

claysmalley:
A [general strike in Ecuador](https://www.rfi.fr/en/international-news/20220629-ecuador-declares-new-state-of-emergency-as-protesters-demand-talks) ([archive](https://archive.ph/7yzOk)) has been going on for the last couple of weeks, so I imagine most Ecuadorians are quite busy at the moment and this issue may be on holld until the political situation is better.

fmvaldezg:
[...]

PR #458, opened 2022-06-26
Add Costa Rica shields

claysmalley:
Initial support for shields of Costa Rica.

Screenshot from 2022-06-26 16-10-49 Screenshot from 2022-06-26 16-10-21


ZeLonewolf:
Route 10111?  Is that a real route?

claysmalley:
Yeah, though it might should be `unsigned_ref=10111`.

icons/shield40_badge_stroke_2.svg
```diff
@@ -0,0 +1,4 @@
...
+ <path style="fill:none;stroke:#000;stroke-width:1px;stroke-linecap:butt;stroke-linejoin:miter;stroke-opacity:1" d="M 1,5.5 H 19"/>
```
1ec5, line 3: Should we include this horizontal bar? It does appear on [guide signs](https://commons.wikimedia.org/wiki/File:P1000465_%284506760395%29.jpg), but if we also add it to the triangular tertiary route shield, it’s going to be even more difficult to fit a number legibly in the shield.

src/js/shield_defs.js
```diff
@@ -551,6 +565,9 @@ export function loadShields(shieldImages) {
   // Yukon
   shields["CA:YT"] = roundedRectShield(Color.shields.white, Color.shields.red);
 
+  // Costa Rica
+  shields["CR:national"] = badgeWithStrokeShield;
```
1ec5, line 569: `network=CR:national` is being tagged on every route in the [National Road Network](https://es.wikipedia.org/wiki/Anexo:Red_Vial_Nacional_de_Costa_Rica), making no distinctions between primary, secondary, and tertiary routes, even though all three classes have different shields. This is similar to the situation in https://github.com/ZeLonewolf/openstreetmap-americana/issues/455#issuecomment-1166417586.

`network=CR:national` está aplicado en cada ruta de la [Red Vial Nacional](https://es.wikipedia.org/wiki/Anexo:Red_Vial_Nacional_de_Costa_Rica), sin distinción entre las rutas primarias, secundarias, y tercerarias, aunque las tres clasificaciones tienen los señales diferentes. Es similar con la situación en https://github.com/ZeLonewolf/openstreetmap-americana/issues/455#issuecomment-1166417586.

[Primaria](https://en.wikipedia.org/wiki/File:CR_RNP_1.svg) [Secundaria](https://en.wikipedia.org/wiki/File:CR_RNS_126.svg) [Terceraria](https://en.wikipedia.org/wiki/File:CR_RNT_606.svg)

icons/shield40_badge_stroke_2.svg
claysmalley: We've kept similar strokes in other shields, like Quebec regional routes: Quebec regional route shield

I anticipated this shield being a commonly used shield in other Latin American countries, so I set it up as a reusable variable. But the shields for those countries may also be recognizable with the stroke removed.

src/js/shield_defs.js
ZeLonewolf: This sounds like a showstopper, right?  Don't we have a similar situation in a country in Europe?

claysmalley: Yeah, I think so. We must be able to determine shield appearance from `network=*`, and three different route shields are all assigned `network=CR:national`. This would need to be fixed before we can support Costa Rican shields.

1ec5: I asked the [talk-cr](https://lists.openstreetmap.org/pipermail/talk-cr/) mailing list for help.

icons/shield40_badge_stroke_2.svg
claysmalley: Horizontal bar has been removed.

ZeLonewolf:
Given the lack of response, I'd say someone could simply define a sensible scheme for Costa Rica's road system.  Note that this PR has to get updated to handle the new shape drawing functions.

PR #460, opened 2022-06-27
Add Ghana shields

claysmalley:
Adds shields for Ghana, all of which are black-on-yellow rectangles.

Techiman, Ghana:
Screenshot from 2022-06-26 22-23-35

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -2397,6 +2397,14 @@ export function loadShields(shieldImages) {
     (county) => (shields[`US:WY:${county}`] = pentagonShieldBlueYellow)
   );
 
+  // AFRICA
+
+  // Ghana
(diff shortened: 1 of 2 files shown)


src/js/shield_defs.js
```diff
@@ -2397,6 +2397,14 @@ export function loadShields(shieldImages) {
     (county) => (shields[`US:WY:${county}`] = pentagonShieldBlueYellow)
   );
 
+  // AFRICA
```
1ec5, line 2400: Ah, I thought we were trying to put the continents in alphabetical order but sorting “North America” as “America, North”. Not a big deal either way, but this reminds me that this method and file need to be split apart at some point.

1ec5 (approved):

Issue #21, opened 2021-07-07
User-switchable languages


ZeLonewolf:
Depends on #20

Functionality description:

A user should be able to open a tray or menu that allows them to set a language and 1 to N fallback languages.  Upon change of this setting, the map should change to use the selected languages.

ZeLonewolf:
An example of language switching can be found here:
https://openmaptiles.org/languages/native/#0.56/0/0

ZeLonewolf:
Perhaps we could use something like Electron to create a tray: https://stackoverflow.com/questions/40485102/rich-html-tray-menu-in-a-desktop-web-application

1ec5:
As of #581, the user can switch the label language by adding `&language=`, followed by an IETF BCP&nbsp;47 language tag, to the hash part of the URL. This issue tracks an in-page control for switching the language.

As far as I know, [this OpenMapTiles YAML file](https://github.com/openmaptiles/openmaptiles/blob/4cb65b0a4f95ba2ad8ea8084003b51f604435ef0/openmaptiles.yaml#L30-L99) enumerates the ISO&nbsp;639 language codes for which Planetiler adds `name:*` fields to the vector tiles. [`Intl.DisplayNames`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames) can turn these language codes into human-readable language names in either the original language or the user’s preferred language.

1ec5:
#609 adds a read-only display of the label language fallback list.

ZeLonewolf:
Here is a proposed workflow for using simplistic text links:

![image](https://user-images.githubusercontent.com/3254090/206789575-787f1d67-2ae5-4a94-962d-ded4ab612946.png)

ZeLonewolf:
Here is a proposed workflow for using simplistic text links:

![image](https://user-images.githubusercontent.com/3254090/206789575-787f1d67-2ae5-4a94-962d-ded4ab612946.png)

Issue #385, opened 2022-06-10
Public statement of Cartographic Design Principles


ZeLonewolf:
Based on discussion in #383, we should be more specific about our design principles for this style.  I'd propose that we add a **Cartography** section to the README.md describing our design principles.

I'll kick this off with some ideas:
* The cartography is inspired by the designs seen in North American paper maps, including atlases, road maps, topographic/hiking maps, city/tourism maps, public transit maps.  Paper maps cover a single fixed zoom; artfully blending single-zoom paper map design principles into a zoomable digital map is a key cartographic challenge.
* Like with paper maps, we aim to render features in a way that prioritizes the conveyance of information to the map reader.  We aim to keep information density within human-digestible ranges, recognizing that this may be a challenge of balance between urban and rural areas.
* We will render a global map according to these principles, while supporting features outside of North America on a "best effort" basis.
* A a community style, we prioritize good mapper feedback over perfection and completionism. We will not fix a mapping or tagging problem with a rendering workaround.

Any others or thoughts/edits?

1ec5:
A statement of design principles will serve this project well. It’ll guide contributions upfront, so that we don’t have to be the bearer of bad news after someone has put too much effort into an ill-fated design. It’ll also help the OSM community know what to expect in the future, even as our technical roadmap remains fluid.

> A a community style, we prioritize good mapper feedback over perfection and completionism. We will not fix a mapping or tagging problem with a rendering workaround.

In the short lifetime of this project, we’ve seen cases where the map’s cartographic needs have driven tagging changes, as well as cases where the tagging has had to catch up to the rendering. Perfectionism aside, we do seem to value inspiring mappers with what’s possible rather than just reflecting what they’ve done. That’s a balance we’ll learn more about over time.

My concrete suggestion for this point is that we prioritize mapper feedback within the design constraints of the kind of map we’re producing. This isn’t quite a [raw data visualization for debugging purposes](https://community.openstreetmap.org/t/get-rid-of-the-standard-map-style/1498/13?u=minh_nguyen). If tomorrow we decide that rendering manhole covers[^manhole] is important for supporting a new water infrastructure mapping initiative, then we’ll need to sketch out enough related functionality that the manhole covers won’t seem like a random addition bolted onto the side.

PR #463, opened 2022-06-28
Add shields for Iran freeways

claysmalley:
Adds shields for national freeways ~and roads~ of Iran.

In real life, highways are posted with both Western Arabic numerals and Persian numerals. Since we're emulating an American road atlas style and using English labels wherever possible, I think it's appropriate to stick with the Western Arabic numerals found in `ref=*`, though the Persian numerals are specified in `ref:fa=*` which may be helpful when implementing language switching in the future (#21).

Screenshot from 2022-06-27 20-04-03

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -2527,10 +2527,6 @@ export function loadShields(shieldImages) {
     Color.shields.blue,
     Color.shields.white
   );
-  shields["ir:national"] = roundedRectShield(
-    Color.shields.green,
-    Color.shields.white
-  );
 
   // Japan
   shields["JP:E"] = roundedRectShield(Color.shields.green, Color.shields.white);
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -2522,6 +2522,16 @@ export function loadShields(shieldImages) {
     },
   };
 
+  // Iran
+  shields["ir:freeways"] = roundedRectShield(
+    Color.shields.blue,
+    Color.shields.white
+  );
+  shields["ir:national"] = roundedRectShield(
+    Color.shields.green,
+    Color.shields.white
+  );
+
   // Japan
   shields["JP:E"] = roundedRectShield(Color.shields.green, Color.shields.white);
   shields["JP:national"] = {
--- a/doc-img/shield_map_world.svg
+++ b/doc-img/shield_map_world.svg
@@ -120,6 +120,7 @@ See the end of this file for a list of available jurisdictions and their codes. 
 .bd,
 .cn,
 .hk,
+.ir,
 .jp,
 .my,
 .np,


src/js/shield_defs.js
```diff
@@ -2522,6 +2522,16 @@ export function loadShields(shieldImages) {
     },
   };
 
+  // Iran
+  shields["ir:freeways"] = roundedRectShield(
+    Color.shields.blue,
+    Color.shields.white
+  );
+  shields["ir:national"] = roundedRectShield(
```
1ec5, line 2530: For some reason, 11 of the 69 occurrences of `network=ir:national` are not in Iran but [in Northern Cyprus](https://overpass-turbo.eu/s/1jHc). It may have just been a mistake, judging from [this relation’s history](https://www.openstreetmap.org/relation/11123980/history). I don’t think it should stop us from supporting Iran’s route shields, but maybe we need to run an analysis of where one country’s ISO&nbsp;3166 code is being used as a namespace in another country, intentionally or otherwise.

1ec5, line 2530: This `network` value conflates first- and second-level national roads. According to the English Wikipedia, only the first-level roads have white on green shields; the [second-level roads](https://en.wikipedia.org/wiki/List_of_roads_and_highways_in_Iran#Second_Level_Roads) have black on white shields. For example, [route 586](https://www.openstreetmap.org/relation/6778799) has been tagged the same way despite its [](https://commons.wikimedia.org/wiki/File:IRN-586.svg)[](https://commons.wikimedia.org/wiki/File:IRN-EN-586.svg) shield.

claysmalley: I hadn't realized there was a distinction. I guess we can leave this one out until we know how the Iranian OSM community wants to handle it.

claysmalley: I think that sort of analysis is a good idea. I also noticed `network=AH` on some Asian Highway relations. `AH` is not an ISO country code, but `network=AsianHighway` is probably the better choice between the two. Two-letter values of `network=*` should be reserved for country codes.

Regardless, I've removed the green shields from this PR, leaving just `network=ir:freeways`.

bgo-eiu:
Would make more sense to include both where signs in real life 

Numerals are closer to symbols than they are to words as far as language is concerned, and it's actually quite rare internationally for highway shields to use the same numerals associated with the local . Treating the Iran shields like this could tentatively mean that one would have to switch languages to see what the shields in Iran look like, then switch back to see what the shields in all the surrounding countries look like, which is counterintuitive.

Also, realistically, anyone doing long distances road travel in this region - which is what route shields are ostensibly for - realistically needs to be able to read both. Unfortunately it's not as common as it used to be for safety reasons, but some of my family used to do road trips to Afghanistan and Iran from Pakistan as tourists, and they would have used paper maps, and understood both numeral systems. (In Pakistan and Afghanistan, the numerals would always be "English" so knowing where they differ is helpful.) So it seems kind of contradictory to the idea of Americana being useful in a similar way to paper maps, as anyone who would be using paper maps in this region would be unlikely to find this interpretation of what numerals are useful for helpful. It's not even that hypothetical that someone doing road trips out of Iran could find Americana useful, things get around quickly online

1ec5:
> If a math textbook didn't print pi because Americans don't speak Greek, that would likely make it less legible

This doesn’t seem to me like an apt analogy. The Greek letter pi has been adapted to other writing systems to the point that Unicode has a different codepoint for pi in a mathematical context. More to the point, the real-world shields feature both numbers, one over the other. ([Asian Highway shields in Iran](https://commons.wikimedia.org/wiki/Category:Diagrams_of_Asian_highway_signs_of_Iran) don’t even have Persian numerals on them.)

Even if we wanted to, we have no technical capability to use `ref:fa` at the moment. The vector tiles don’t include that tag, and none of the currently available fonts contains Persian numeral glyphs.

1ec5 (approved):
Let’s ticket out support for national routes, which are more numerous than the freeways.

bgo-eiu:
The less bad option would be to not include shields in switching languages if that happens, that's really the main thing that would be more confusing than necessary. Looking at pictures though, there do seem to be signs that show one and not the other and I think it would be helpful to reflect that on the map at some point because it's a noticeable distinguishing feature.

That Unicode did that for pi and that more fonts don't support multiple numeral systems is more just an indicator of Eurocentricism on their part, considering there's more people in the world bilingual with both English and a language which uses Perso-Arabic scripts than there are Americans total (and possibly even just more than monolingual English speakers generally). There's also still no way to see a whole range of common Perso-Arabic glyph representations without installing special fonts for it so Unicode isn't exactly logical in what they encode for.

Road signs are different from pi in that you don't actually need to know what values they represent, because you don't need to do any math with them. What's funny is if you look at the Urdu ref tags on route relations where they exist, the translation for "M-25" literally spells out "em" ام like someone sounding out the letter M using Arabic script, then keeps the numerals the same, which would get sounded out by many as well rather than translated. They don't use م-25 even though that's technically the first letter of transliterated motorway, because you don't actually need to know English to know what the letter M is, and treating numbers and letters as symbols rather than abbreviations or representative of words makes more sense in a multilingual setting where 2 or ‌ ‌‌੨ or anything else could be read differently depending on the native spoken language of the reader.

1ec5:
> That Unicode did that for pi and that more fonts don't support multiple numeral systems is more just an indicator of Eurocentricism on their part, considering there's more people in the world bilingual with both English and a language which uses Perso-Arabic scripts than there are Americans total (and possibly even just more than monolingual English speakers generally).

This is quite off-topic, but no, it’s only for representing [certain Greek-derived symbols](https://en.wikipedia.org/wiki/Mathematical_operators_and_symbols_in_Unicode#Mathematical_Alphanumeric_Symbols_block) for certain mathematical purposes, such as symbols for vectors and maps. (Not to be confused with vector maps!) Your elementary school pi is safe; it uses a character from the Greek block.

I get that seeing a transliterated shield in Americana isn’t going to be so useful for roadtripping in certain parts of the world, but keep in mind that this is not a real turn-by-turn navigation map, despite the aesthetic resemblance to some maps that might’ve been used for that purpose in the past. A map that’s actually optimized for turn-by-turn navigation would translate at most place names but not other labels, especially not road names, whereas we’re using all the English translations we can get our hands on.

There may well be other OSM-based projects for which your insights are spot-on and worth raising, but it’s a stretch for this repository as it stands. We aren’t even setting the [`preserveDrawingBuffer`](https://maplibre.org/maplibre-gl-js-docs/api/map/#map-parameters) option that would allow this map to be printed (at a significant performance penalty).

1ec5:
I’m sorry that my response came across as dismissive or condescending. The fact of the matter is that we don’t have the ability to implement the Western Arabic numerals in Iran’s shields at this moment, but for solvable reasons. The font can be changed, the tiles can pull in keys other than `ref`, we could even maybe convince the Iranian mapping community to relegate the Western Arabic numerals to a different key – all that requires work outside of this repository.

Taking a step back, we’re discussing a pull request for an implementation of shields for Iran’s freeways based on the data that’s available to us, so allegations of the Unicode standard being Eurocentric don’t seem relevant to this repository. If due to the numbering situation you believe this pull request should not have been merged, instead leaving these freeway icons as plain black Western Arabic numerals, then please open an issue explaining your reasoning.

You may also want to look at #385, which so far is less ambitious than what you seem to be suggesting as the project’s purpose. It’s not that Americans don’t travel using world maps, but there’s historically been a difference between a local guide map and a world map for studying the world at one’s leisure. There are many examples of both kinds of maps introducing some degree of linguistic variety, such as National Geographic glossing place names in (a Latin transliteration of) the local language, but their ambition does not extend to labeling everything as it is on the ground.

It’s a classic tradeoff: the more this map becomes a lesson in linguistics, the less it can serve as a lesson in geography to a broad audience of English speakers. Both groups of users deserve to be served by a style, but it doesn’t have to be the same style, or there can be a user preference. We will likely make the similar tradeoffs for POI icons, choosing icons that Americans are intimately familiar with at the expense of other users. If we don’t make such tradeoffs, we will undoubtedly end up in the same analysis paralysis that openstreetmap-carto frequently finds itself in, and no one will be well served.

bgo-eiu:
This PR is fine, all I was saying is it's still worth considering showing the full shield at some point given previous discussions on the topic.

It's only a lesson in linguistics to the extent that a "BUS" panel is a lesson in economics. Unlike many (but not all) place names, shields are often understood to be symbolic, especially considering that in most countries that aren't America long-distance travel and cross-country travel are the same thing, and shields are already designed to be understood by foreigners rather than designed in a way that only makes sense to locals. You would be hard-pressed to find a map which displays US interstate shields in another numeral system in world maps for studying at one's leisure produced in any country, and likewise it's impossible to avoid the fact that most contemporary English speakers do not live in America or Europe, and attempts to re-interpret information in multilingual countries as if this isn't the case can result in more confusion than clarity. It would be considered nonstandard to print a map with the name Bombay instead of Mumbai or Nanking instead of Nanjing, or even Ivory Coast instead of Côte d'Ivoire, even though the former in each case is likely more familiar to Americans of not-recent-immigrant-parentage, because it speaks to an antiquated notion of how the English language is used. It's less that it seems like a tradeoff but a rather unorthodox approach which doesn't service geographic or linguistic understanding. (If a non-American map producer were to alter information under the same reasoning, arguably it would come across as uncharitable or even insulting to American English speakers, to suggest that our dialect of English requires a dumbing-down that no other does.)

I was thinking about it, and the true equivalent of "translating" a shield in non-Latin characters would be literally sounding out the name of those characters using Latin characters, like M-5 gets transliterated as ام 5 and not  5-م. That's neither here nor there, but perhaps part of why this seems confusing is that is something which most English-speakers do that Americans oddly don't have much precedent for.

1ec5:
> I was thinking about it, and the true equivalent of "translating" a shield in non-Latin characters would be literally sounding out the name of those characters using Latin characters, like M-5 gets transliterated as ام 5 and not 5-م. That's neither here nor there, but perhaps part of why this seems confusing is that is something which most English-speakers do that Americans oddly don't have much precedent for.

Yes, I was casually suggesting in https://github.com/ZeLonewolf/openstreetmap-americana/issues/465#issuecomment-1168169455 that we find a way to display Taiwan’s provincial routes with the Latin suffixes instead of the signposted Hanzi, e.g. “2a”, “2b”, “2c” instead of “2甲”, “2乙”, “2丙”. There is precedent for this transliteration not only in official contexts but also in foreign-language documentation, such as on the English Wikipedia. However, it ideally would be backed by tags in the data, similar to what the Iranian mappers have done with `ref` and `ref:fa`. Automated transliteration would be far less desirable.

ZeLonewolf:
Thanks for the great discussion here, unfortunately it will be lost buried in a PR for Iranian freeways.  Could we start one or more new issues discussing the specific classes of issue relating to signage in non-latin character sets?  I'm fully in favor of better internationalization but don't want this to get lost here.

PR #469, opened 2022-07-01
Add Texas outline shield

claysmalley:
Adds a state outline shield for FM, RM and Recreational roads of Texas.

The icon added has 24px width and 23px height to account for the limited text space, an exception to the typical 20px height, though nearly identical to the diamond shield (24px square).

This is the most I could distort the state outline while keeping it recognizable :sweat_smile: 

Screenshot from 2022-06-30 20-53-44 Screenshot from 2022-06-30 20-53-35 Screenshot from 2022-06-30 20-52-42 Screenshot from 2022-06-30 20-53-06


claysmalley:
Note: I'm okay if this doesn't get merged. I've pretty much stretched this shield to its limits and the result still has tiny, crowded text. There's a big tradeoff between legibility and recognition here.

zekefarwell:
Maybe instead of "F.M." and "R.M." above the shield there could be a little F or R right inside the northern most part of the outline?  Then the outline could be made a bit bigger to accomodate slightly larger numerals.

I think the Texas-ish shape works decently on Apple maps, though it's a massive departure from the actual shape of Texas.  Obviously Apple includes the black background of the sign and an altered shape like this might not work so well when it's just an outline.  On 1x screens the text is barely readable, but at 2x it's not bad.  
Screen Shot 2022-06-30 at 9 32 34 PM  Screen Shot 2022-06-30 at 9 32 02 PM

1ec5:
Definitely keep experimenting with it in case you arrive at something workable. However, I’d prefer a faithful reproduction of the guide signs than taking as much liberty with the reassurance markers as Apple does. A real Texan would probably feel even more strongly about that. 😄

claysmalley:
> I think the Texas-ish shape works decently on Apple maps, though it's a massive departure from the actual shape of Texas.

I think it looks unrecognizable and overall kind of wonky...

zekefarwell:
Cool.  I don't have strong feelings on the matter.  I guess the questions is if a faithful reproduction at this size, with readable text just isn't possible, then what would be preferred instead?  I had assumed an approximation of the shape would be preferreable over a plain rectangle, but perhaps not.

zekefarwell:
Here's the shield increased to 30px wide give the text some more room.  It looks a little big compared to the interstate shield, but compared to the rectangles the visual weight is still quite balanced.  I also tried stiking the RM up in the corner of the shield graphic to see how that would look.  Of course the spot I chose to screenshot is an FM road.  Oops 😅!  What do you think?

2x Screen Shot 2022-06-30 at 10 48 38 PM   1x Screen Shot 2022-06-30 at 10 48 13 PM

michaelblyons:
I was also interested in having the FM/RM in the top right. This looks nice. Does FM/RM still have a halo when it overlaps a road?

zekefarwell:
> Does FM/RM still have a halo when it overlaps a road?
Yes, the halo is crucial.  For this quick test I just did a white halo, but matching the background color would be better.
Screen Shot 2022-06-30 at 11 14 12 PM  Screen Shot 2022-06-30 at 11 15 27 PM

1ec5:
> I had assumed an approximation of the shape would be preferreable over a plain rectangle, but perhaps not.

For context, the plain rectangle is what TxDOT uses for these roads on guide signs and maps: https://github.com/ZeLonewolf/openstreetmap-americana/pull/177#issuecomment-1046337149 https://github.com/ZeLonewolf/openstreetmap-americana/pull/205#issue-1152483722. So there is precedent for what’s in main. That said, if we can get the state outline to work, the advantage is that it matches the signs for the majority of Farm to Market Roads in rural areas without any guide signs.

It’s entirely reasonable to enlarge a shield to accommodate “tentacles”. The three-digit variant looks very good. The four-digit variant’s text is still quite small. What if we shift El Paso and Beaumont southward by a pixel or two, shortening everything to the south, and widen the shield by a couple pixels for just that variant?

Any thoughts on whether the legend should be hard-coded in the SVG or composited in JavaScript?

zekefarwell:
I'm travelling this weekend and attending a wedding so I probably won't have have time to try any further tweaks until next week.  I pushed up the larger shield test I did on a separate branch ([shields-texas-outline-bigger-test](https://github.com/ZeLonewolf/openstreetmap-americana/tree/shields-texas-outline-bigger-test)) in case anyone else wants to mess around with it and try out more adjustments.

claysmalley:
Okay, I'm back with another Texas. I think this 26x27 SVG satisfies my aesthetic preferences—big enough to fit 4 digits without distorting Texas, but not too big that it looks out-of-place next to other shields.

Also considering removing the "F.M." and "R.M." text altogether. The state outline alone is enough to identify routes as part of these networks.

[Screenshot from 2022-08-03 17-51-45 Screenshot from 2022-08-03 17-52-04](http://localhost:1776/#14.01/30.00112/-95.26863)

[Screenshot from 2022-08-03 17-50-09 Screenshot from 2022-08-03 17-49-47](http://localhost:1776/#13.23/29.93385/-98.22031)

The "R" text is retained on Recreational routes, however. These routes are not very common, and the brown is too similar to black to be considered a distinguishing feature. So it's necessary to keep the text on these, though it looks a bit awkward floating above the state:

[Screenshot from 2022-08-03 17-54-18 Screenshot from 2022-08-03 17-54-06](http://localhost:1776/#13.65/29.60141/-100.86692)

1ec5 (approved):
Dropping the “F.M.” and “R.M.” banner seems to be received well in OSMUS Slack, thanks to the state shape. This will be a significant reduction in numeral size, but I think it’s a fair tradeoff, because areas with many F.M./R.M. shields were pretty loud before; the full-size shields were more prominent than state highways due to the ref lengths.

PR #475, opened 2022-07-04
Optimize Colombia shield SVGs

claysmalley:
@1ec5 forgot `npm run code_format` and so did I when I reviewed it :facepalm:


ZeLonewolf (approved):
LGTM

Issue #25, opened 2021-07-09
Switchable map layers


Mashin6:
The current problem of osm-carto is that it is a map for everything and nothing at the same time. It is supposed to serve mappers to check their mapping progress, but is also supposed to be for general public. It is just filled with too many different objects that make it too dense.

One way to go around is to have several separate tiles styles: 1. for public. 2. for mappers. 3. public transport ,...
-OR- 
I was thinking that since the actual tile rendering is happening on the user side to have some sort of switches that would turn on/off displaying of some objects. So there will be one style for the tiles but the user would choose which parts of that styles would be displayed. For example there is no need to have the default of displaying of manhole covers and utility poles, but with a switch one can display them if needed.

Issues:
I don't know if this is technologically possible
Probably depends on solving of #9

ZeLonewolf:
There are a few distinct issues here that you're raising.

1. **osm-carto is too information dense**.  This is something I wholeheartedly agree with.  Take for example, this screenshot from the Boston area at raster zoom 15.  It does not make sense to draw buildings at this zoom level -- it adds clutter and reduces the readability of the street network.  In the upper right corner there is a [park](https://www.openstreetmap.org/way/29860532), and several athletic fields are rendered.  However, at this zoom level, the more important thing to see would be the park's name.  The individual features within the park should be suppressed until a closer zoom, and this would increase readability.  Similarly, a name on the [school grounds](https://www.openstreetmap.org/way/309668100) would be far more useful to render than the building.  At this zoom, the important features are the road network, road names, and major points of interest.

Ideally I would like Americana to have a "Goldilocks" level of information density: not too dense, but not too sparse, and that will take careful planning.

![image](https://user-images.githubusercontent.com/3254090/125015192-010a2100-e03d-11eb-9fa4-cc67f41ff112.png)

2. **Changing the map style on the fly**.  This is 100% doable with vector tiles and is one of the major selling points for it.  In fact, [here's an example](https://docs.mapbox.com/mapbox-gl-js/example/color-switcher/) of dynamically changing a rendered map style on the fly.  The code for how to do this is not hard -- the hard part is the design work of coming up with what options we should offer to the map user!  I am thinking that we could use the "wiki" section here in GitHub to start working out those design choices.  Design work is hard, and it's not really my skill set.

[...]

PR #476, opened 2022-07-07
Add railroad tracks

### LINKED ISSUE #101: Railroad tracks
Maps published in the U.S. conventionally give railroad tracks a different appearance than European-influenced maps such as openstreetmap-carto. While the railroad ties we’re familiar with may seem kitschy, they are less likely to be confused with roads than the European approach of an alternating fill inside a casing, since the latter treatment is used for secondary roads in American topographic maps and either roads under construction or all-weather roads in general-purpose American road maps.

Here’s a comparison of map legends from [_How to Lie with Maps, Third Edition_](https://books.google.com/books?id=MwdRDwAAQBAJ&pg=PA141):

[Screenshot-2022-1-26 How to Lie with Maps, Third Edition](https://books.google.com/books?id=MwdRDwAAQBAJ&pg=PA141)

Some highlights from maps in my collection that depict railroad tracks:

* Official state maps from Ohio, Oregon, Texas, and Washington only show Amtrak passenger routes, which might be a reasonable option at certain zoom levels to avoid distracting from roads that have a similar line width.
* Pennsylvania makes the line thicker for Amtrak routes.
* North Carolina describes it as “rail corridors”, which probably means spurs are excluded.
* Georgia gives MARTA commuter rail lines a very different appearance: a red line without ties.
* Illinois omits railroad tracks “in congested areas”, relegating them to map insets.
* Pennsylvania and Louisiana color railroad tracks red.
* A map of San Mateo County, California, by Compass Maps distinguishes BART (thick dotted line) from Caltrain (thick solid line).
* Connecticut spaces out the ties and makes them relatively large.

Wyoming DOT 2010 Rand McNally Manhattan 1991 street map
Nevada DOT Oklahoma DOT Kansas DOT 2009–2010 KYTC West Virginia DOT Tennessee DOT 2005 Texas DOT 2016 Ohio DOT North Carolina DOT 2005 Illinois DOT 2005–2006 Seeger Cincinnati 1991 Rand McNally Atlanta 2000
Compass San Mateo County 2004 Pennsylvania DOT 2003 Connecticut DOT 2004–2005

Oklahoma agritourism Louisiana DOTD 2010 Oregon DOT Washington DOT 2008–2009 AHTD 2012 South Carolina DOT 2005 Georgia DOT 2005–2006 Montana DOT 2007–2008

claysmalley:
Closes #101.

At the moment, this is a draft and only implements `railway=rail`. Once finished, this PR will implement all other values of `railway` surfaced by OpenMapTiles:

- [x] `rail`
- [x] `subway`
- [x] `light_rail`
- [x] `tram`
- [x] `monorail`
- [x] `funicular`
- [x] `narrow_gauge`
- [x] `preserved`

Screenshot from 2022-07-06 22-38-40 Screenshot from 2022-07-06 22-26-00 Screenshot from 2022-07-06 22-24-59 Screenshot from 2022-07-06 22-24-16


claysmalley:
Changed to gray:

Screenshot from 2022-07-07 08-51-20 Screenshot from 2022-07-07 08-52-05

src/americana.js
```diff
@@ -370,6 +381,11 @@ var bridgeLayers = [
   lyrRoad.trunkExpresswayTollBridge.surface(),
   lyrRoad.motorwayTollBridge.surface(),
 
+  lyrRail.railBridge.dashes(),
+  lyrRail.railBridge.fill(),
+  lyrRail.railServiceBridge.dashes(),
+  lyrRail.railServiceBridge.fill(),
```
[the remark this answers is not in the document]
michaelblyons: +1 to operator labeling. From my perspective as "not a rail enthusiast," the operator is more interesting than the named section of track.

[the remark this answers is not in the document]
claysmalley: Since name and operator info is not surfaced by the vector tiles, I think it'd be out of scope for this PR, but would be welcome as a future enhancement.

claysmalley:
I'm hesitant to assign purple to railroads because the majority of railroad lines in North America don't carry passenger trains, and there's not much precedent in US cartographic tradition for purple railroads. I'd rather just keep them grayscale.

Screenshot from 2022-07-07 17-17-14

claysmalley:
If busways are rendered in purple in #477, it may set a precedent for public transit railways (subway, tram, funicular, etc.) to be purple as well.

claysmalley:
Added narrow gauge railways,  with slightly thinner lines and dashes a little closer together than `railway=rail`.

Screenshot from 2022-07-15 07-49-32
Screenshot from 2022-07-15 07-50-21

1ec5:
Remember to update scripts/taginfo_template.json with any tags for which this PR would introduce noteworthy treatment.

claysmalley:
Now with bridge casing:
Screenshot from 2022-07-16 16-10-01

src/layer/rail.js
```diff
@@ -0,0 +1,356 @@
...
+      this.service,
+      this.constraints
+    );
+    layer.layout = {
+      "line-cap": "butt",
+      "line-join": "bevel",
+      visibility: "visible",
+    };
+    layer.paint = {
+      "line-color": "hsl(0, 0%, 50%)",
```
1ec5, line 72: At a glance, this casing looked like an abutting street to me.

Consider a knockout effect instead, as we explored in #169: no outer line, just a background that matches the land background. The knockout should be thin to avoid visually intruding on waterways or greenspace. However, it should be more than a hairline: #203.

A more literal representation of a bridge might require work that’d be out of scope for this PR. It seems like you were going for guardrails, but the guardrails would need to flare out at the abutments to avoid looking like road casings. It might be tricky to get those flares to line up with the guardrails.

In other maps, instead of guardrails, there are schematic-like break symbols that don’t necessarily extend the full length of the bridge. For example, the first photo in https://github.com/ZeLonewolf/openstreetmap-americana/issues/169#issuecomment-1043648465 depicts a bridge that actually extends all the way past the river crossing out of view.

claysmalley: I like the knockout casing better:

Screenshot from 2022-07-17 11-57-10

claysmalley:
Experimenting with subways at the same width as conventional rail, but with a lighter fill and no ties:

Screenshot from 2022-07-17 13-10-44

src/layer/rail.js
zekefarwell: that knockout looks great!

claysmalley:
Though it makes it look like a street is missing when there's an elevated `railway=subway` over a street:

Screenshot from 2022-07-17 13-46-43

claysmalley:
More experimentation, with subway darker than conventional rail. Also considering removing the dash pattern when bridge casing is present.
Screenshot from 2022-07-17 14-19-16

claysmalley:
More screenshots of railroad bridges without ties. Looks way cleaner when it's elevated above a street.
Screenshot from 2022-07-17 14-35-37
Screenshot from 2022-07-17 14-36-26

michaelblyons:
I like these. Can you take a shot at low zoom like the first one of Chicago?

I am trying to imagine a nice way to do two things:

1. Prevent people from misinterpreting surface-level or bridge `rail=subway` as "yet another kind of road."
    - Perhaps badges from subway systems like the road badges?
        - This might have copyright implications.
        - You may want to reserve the symbols (if copyright is not a problem) for stations only. 
3. Possibly show subway line color in some subtle way. If feasible, this might take care of 1️⃣.
    - Problems:
        - Might make the map too busy
        - Might not fit in the color scheme
        - How to accommodate co-linear routes of different color?
    - Options:
        - Occasional stripes of the line color(s) 🟫🟫🟫🟫🟫🟫🟫🟫🟫 🟧 🟦 🟫🟫🟫🟫🟫🟫🟫🟫🟫 🟧 🟦 🟫🟫🟫🟫🟫🟫🟫🟫🟫
        - Glow
        - Interior line/dashes

Lastly... There's a playground named "Crack is Wack?" 😳

claysmalley:
> I like these. Can you take a shot at low zoom like the first one of Chicago?

As requested:
Screenshot from 2022-07-17 16-53-04
 
> I am trying to imagine a nice way to do two things:
> 
> 1. Prevent people from misinterpreting surface-level or bridge `rail=subway` as "yet another kind of road."
>    * Perhaps badges from subway systems like the road badges?
>      * This might have copyright implications.
>      * You may want to reserve the symbols (if copyright is not a problem) for stations only.

In the long run, I would love to have station icons based on the services that stop there. But we're trying to figure out version 1 right now, which is probably going to have generic station icons. Right now, this info is not present in the vector tiles, and it's not clear what the path forward is for vector tile support.

> 2. Possibly show subway line color in some subtle way. If feasible, this might take care of one.
>    * Problems:
>      * Might make the map too busy
>      * Might not fit in the color scheme

This treatment would be appropriate for a future switchable layer for public transit (#25). But for a general-purpose map, I think it makes sense to keep railroads grayscale.

claysmalley:
Added lines for light rail, tram, and funicular:

Screenshot from 2022-07-17 17-20-28
Screenshot from 2022-07-17 17-21-07

claysmalley:
Cut railway line widths in half. Looks a lot cleaner, especially on sections of double track.

[Screenshot from 2022-07-19 20-55-44](http://localhost:1776/#15.9/40.431258/-80.002209)
[Screenshot from 2022-07-19 20-56-02](http://localhost:1776/#15.45/39.961779/-75.261038)
[Screenshot from 2022-07-19 20-56-21](http://localhost:1776/#15.16/43.455045/-3.819044)

claysmalley:
Also, `railway=funicular` + `service=*` is removed, since I'm pretty sure it's a tagging error.

jleedev:
Is there a straightforward explanation of the difference between light_rail, subway, and tram?

jleedev:
Heavy rail looks really nice — I've played with this a little bit and it instantly turns to unreadable mush.

Screenshot 2022-07-19 21 04 20
Screenshot 2022-07-19 21 06 34
Screenshot 2022-07-19 21 11 53

This is a light rail yard:

Screenshot 2022-07-19 21 04 53

While this is a subway yard:

Screenshot 2022-07-19 21 15 49

claysmalley:
> Is there a straightforward explanation of the difference between light_rail, subway, and tram?

subway: (almost) completely grade-separated right-of-way
tram: mostly street-running
light_rail: somewhere in between

> e.g. the Cleveland Rapid has parts which are subway and parts which are light_rail, but that's probably not on purpose.

It's on purpose. The subway tracks can support light rail trains, but not vice versa.

1ec5:
> Cut railway line widths in half. Looks a lot cleaner, especially on sections of double track.

It does, though it also makes the bridge knockout more subtle, maybe too subtle. It looks like you tried to compensate in f8e8d76f7204922527c1dc3b03a85b516b39b13a, but I’d keep an eye out for places where the knockout becomes ineffective at indicating a grade separation.

> Heavy rail looks really nice — every time that I've tried writing this style it instantly turns to unreadable mush. I think the very spacious ties make it work.

I’m enamored by the [Moiré](https://en.wikipedia.org/wiki/Moiré_pattern) effect that the ties produce at railyards. I consider it a feature rather than a bug.

> Is there a straightforward explanation of the difference between light_rail, subway, and tram?

It’s all very fuzzy. For example, the [U.S. Capitol subway system](https://www.openstreetmap.org/relation/13896286) is tagged as `light_rail` for some reason. (Part of that system is actually an automated people mover. APMs are tagged as all kinds of things.)

michaelblyons:
If you're looking for monorails, I think there's one in Sydney. Maybe also Las Vegas?

wmisener:
There is a monorail in Vegas. Or, of course, you can find one in Anaheim (https://www.openstreetmap.org/way/134870120) and Orlando (https://www.openstreetmap.org/way/117446424). Good spots to compare a railway=monorail passing over a railway=narrow_gauge, if you'd ever need to.

claysmalley:
What I'm really looking for is examples of paper maps and atlases that include monorails styled differently from other railways.

claysmalley:
The no-tie treatment of subway lines seems to be received well, and monorails are usually completely grade-separated like subways but operate on a smaller scale. So here's monorails that look like subways but thinner:

[Screenshot from 2022-07-21 13-52-27](http://localhost:1776/#15.23/28.419349/-81.579808)
[Screenshot from 2022-07-21 13-53-39](http://localhost:1776/#15.04/47.61661/-122.34376)
[Screenshot from 2022-07-21 13-54-53](http://localhost:1776/#15.33/37.752253/-122.196295)

claysmalley:
Cutting line width in half obviated the need for hiding dashes on bridges, though now bridge dash width is reduced to fit within the casing.

Screenshot from 2022-07-21 19-21-16
Screenshot from 2022-07-21 19-24-39

claysmalley:
Trying a blur for the knockout casing.
Screenshot from 2022-07-22 15-17-34 Screenshot from 2022-07-22 15-17-54

claysmalley:
Updating narrow gauge railroads with a doubled tie pattern so you don't have to squint to distinguish them.

Screenshot from 2022-07-22 19-09-49

claysmalley:
Ties are now disabled on `service=crossover` to clean up ladder tracks a little.

Screenshot from 2022-07-23 15-17-53
Screenshot from 2022-07-23 15-20-37

src/layer/rail.js
```diff
@@ -0,0 +1,672 @@
...
+      ["exponential", railExp],
+      ["zoom"],
+      8,
+      [...serviceSelector, 0.8, 1],
+      12,
+      [...serviceSelector, 2.0, 2.5],
+      20,
+      [...serviceSelector, 6.5, 8.5],
+    ],
+    "line-blur": 0.75,
```
1ec5, line 59: Against water, the blur ends up looking like a halo rather than a knockout. The rail lines seemed to be slightly more legible to me with a solid edge on the casing. On land, road–rail grade separations are quite subtle. I think we got away with a blur in #203 because the halo was sufficiently thick and contrasted, whereas this casing is quite thin.

claysmalley: I think you're right. I went ahead and removed the blur.

claysmalley:
And here's `railway=preserved`.

[Screenshot from 2022-07-29 17-43-54](http://localhost:1776/#16.6/39.485164/-77.35589)

claysmalley:
In all, this adds 214 layers, increasing the layer count from 620 to 834. I could probably reduce the count by using data-driven styling for the fill, but Maplibre does not yet support data-driven styling for dash arrays, so unfortunately the tie pattern is going to give this a high layer count for the time being.

claysmalley:
I managed to combine railway fill into just a few layers, bringing the added layer count to 102. The layer count in `main` has since been reduced to 615, and this PR will make it 717.

claysmalley:
615 → 701, +86

ZeLonewolf:
Are there any lessons learned from this PR that can be applied to the rest of the road layer?

claysmalley:
Yeah, and I plan on working on that next 😅

ZeLonewolf:
Awesome! 

Do we have an open issue for the dashed array expressions in Americana that can be linked to the maplibre issue as a blocker?

claysmalley:
Tracked in #526.

src/layer/rail.js
```diff
@@ -0,0 +1,497 @@
...
+
+var serviceSelector = ["match", ["get", "service"], ["siding", "spur", "yard"]];
+var isService = ["in", "service", "siding", "spur", "yard"];
+var isNotService = ["!in", "service", "siding", "spur", "yard"];
+
+var lineColor = [
+  "match",
+  ["get", "brunnel"],
+  "tunnel",
+  "hsl(0, 0%, 90%)",
```
ZeLonewolf, line 67: I think it would be helpful to pull out these magic colors into the Color class.  As the style grows, it will help us to examine the color palette holistically.

ZeLonewolf:
This looks pretty good to me, visually.  There is maybe a some exploration that can happen to find a way to draw a better visual distinction from the road network, but I think the style choice feels right in terms of the line style with the cross-ties representing rail.

Rail going over/under objects seems to make sense also.  In this case where a rail goes from above ground, to under a building, to underground, the map makes sense:

image

If someone has an example of rail going OVER a building, we should examine that case to see if the ordering still looks right>

Overall a :+1: from me with one minor comment in the code, but like to have more eyes on this.

1ec5:
> If someone has an example of rail going OVER a building, we should examine that case to see if the ordering still looks right

The AirTrain over the terminals at SFO:

[AirTrain](https://zelonewolf.github.io/openstreetmap-americana/#16/37.615309/-122.390071)

src/layer/rail.js
```diff
@@ -0,0 +1,509 @@
...
+    this.minZoom = 14;
+    this.dashWidthFactor = 2.3;
+    this.dashArray = [1, 2];
+  }
+}
+
+/*
+ * TODO:
+ * preserved?
+ */
```
1ec5, line 339: This is done, right?

scripts/taginfo_template.json
```diff
@@ -200,6 +200,90 @@
...
+    {
+      "key": "railway",
+      "value": "narrow_gauge",
+      "object_types": ["way"],
+      "description": "Narrow gauge railroads are represented by a thick light gray line with a sparse doubled tie pattern.",
+      "doc_url": "https://openmaptiles.org/schema/#transportation"
+    },
+    {
+      "key": "railway",
+      "value": "preserved",
```
1ec5, line 220: The [`railway=preserved` documentation](https://wiki.openstreetmap.org/wiki/Tag:railway%3Dpreserved) points out that `railway=preserved` is steadily being [supplanted by `railway=*` `railway:preserved=yes`](https://taghistory.raifer.tech/#***/railway/preserved&***/railway:preserved/yes), which is already twice as popular. Unfortunately, OpenMapTiles doesn’t support `railway:preserved=yes` via property separate from `subclass`. I don’t know the full context around this shift, but it might be an oversight in OpenMapTiles. In the meantime, should we render preserved railways based on this older `railway=preserved` tag?

claysmalley: In retrospect, I should have looked into this before implementing it. I think it's probably appropriate to render these as conventional railways until support can be added for the correct tagging scheme.

claysmalley:
`railway=preserved` is now rendered identically to `railway=rail`, pending OpenMapTiles support of `railway:preserved=yes`.

scripts/taginfo_template.json
1ec5: Sounds good. Let’s ticket out the work to restore this treatment once the upstream issue is resolved.

1ec5 (approved):
I’m on board with these changes! Some of the screenshots of Americana on the wiki and elsewhere are going to look really outdated now that the actual style fills in the gaps with this intricate line work.

Issue #185, opened 2022-02-22
Pedestrian malls


1ec5:
Unlike most walkways, pedestrian malls and bridges are often very prominent features in urban areas. At higher zoom levels, they should be rendered similar to roadways but with different linework or coloring so that they aren’t confused with roads for cars.

For example, Cincinnati’s [Purple People Bridge](https://www.openstreetmap.org/way/23282146) is conspicuously missing on the Ohio River:

[<img src="https://user-images.githubusercontent.com/1231218/155043685-0f33faa4-7e12-42f3-8d70-7fc21a43a167.png" width="400" alt="Purple People carto">](https://zelonewolf.github.io/openstreetmap-americana/#15/39.09788/-84.4978) [<img src="https://user-images.githubusercontent.com/1231218/155043600-5487e341-683b-466d-839e-6cc1e7cf974c.png" width="400" alt="Purple People Americana">](https://www.openstreetmap.org/#map=16/39.0970/-84.4971)

San José State University’s [_paseos_](https://www.openstreetmap.org/way/45409564) are easy to spot as negative space among the buildings:

[<img src="https://user-images.githubusercontent.com/1231218/155044180-99629150-4ca5-45b7-a5a1-fd27f177fd17.png" width="400" alt="SJSU carto">](https://www.openstreetmap.org/#map=16/37.3355/-121.8817) [<img src="https://user-images.githubusercontent.com/1231218/155044038-ed7ae277-e362-4c62-8afe-ac6d347e3840.png" width="400" alt="SJSU Americana">](https://zelonewolf.github.io/openstreetmap-americana/#15/37.33547/-121.8817)

1ec5:
The tiles include pedestrian malls in the `transportation` layer with `class` set to `path` and `subclass` set to `pedestrian`.

PR #477, opened 2022-07-08
Add busways

### LINKED ISSUE #308: Render busways
Busways, often parts of bus rapid transit systems, are roadways dedicated solely to buses and physically separated from other traffic. These are tagged as [highway=busway](https://wiki.openstreetmap.org/wiki/Tag%3Ahighway%3Dbusway), a relatively new tag, approved in April 2021. The tag has quickly gained popularity, rising to [>5000 uses](https://taginfo.openstreetmap.org/tags/highway=busway#chronology), though it has been hampered by [lack of rendering in OSM-Carto](https://github.com/gravitystorm/openstreetmap-carto/issues/4226#issuecomment-1116995998). In the US, such busways have existed for decades (e.g. in [Pittsburgh](https://en.wikipedia.org/wiki/South_Busway)), but are becoming more common, and now exist across the country:
export

An easy initial approach would be to render them identically to service roads. This seems appropriate, as the busways don't form part of the regular road network, but they are important for orientation, as the map appears to have "gaps" without them:
Screen Shot 2022-05-06 at 9 16 51 PM

Due to their importance, it might be good to have them render at an earlier zoom level than other service roads. Once access tagging is developed (#216), rendering similarly to roads with access=private is probably appropriate.

These busways will be more important once more public transport features are rendered. It might be nice in the future to use colors/styles that are in harmony with bus stops, or even with railway features (#101). But I think getting it on the map in some way and refining the rendering later is better than not rendering it at all.

claysmalley:
Fixes #308.

Since it looks like purple is going to be the color for public transit, infrastructure dedicated to public transit might as well be purple too. This may influence design decisions about railways dedicated to public transport (subway, funicular, etc.) as #476 progresses.

Screenshot from 2022-07-07 21-04-50 Screenshot from 2022-07-07 21-05-40


claysmalley:
After some discussion on Slack, I think busways may need lowlighting instead of highlighting. Dull gray works pretty well, and helps indicate that these serve a special purpose but are not accessible for general traffic.

Screenshot from 2022-07-08 09-49-10 Screenshot from 2022-07-08 09-49-30

Bonus screenshots with railroads from #476:

Screenshot from 2022-07-08 09-59-34 Screenshot from 2022-07-08 09-59-11

ZeLonewolf:
At high zoom, the busway is gray cased in black.  Is it our intent that non-drivable roads like this have a common fill color?  Just trying to think through if there's other feature classes that fit the mold here.

claysmalley:
Pedestrian streets come to mind: #185

ZeLonewolf:
Exploring a subtle purple at z15+.  What do we think?

image

1ec5:
Do any busways run near county lines? The faint purple might be too similar to the border highlight for `admin_level=6`.

claysmalley:
Here's a busway in the Philadelphia suburbs that runs close to a county line: [(localhost link)](http://localhost:1776/#14.31/39.99527/-75.30187)

ZeLonewolf:
That's a weird example because the road sits on top of the county line, but this is what it looks like zoomed way in:

image

ZeLonewolf:
Another example, localhost [link](http://localhost:1776/#17/52.230682/4.670594)

image

ZeLonewolf:
Relaying @claysmalley comments in #476:

> If busways are rendered in purple in https://github.com/ZeLonewolf/openstreetmap-americana/pull/477, it may set a precedent for public transit railways (subway, tram, funicular, etc.) to be purple as well.

I agree that what we're searching for here is the general case for how we want that collection of features to render.

claysmalley:
I'm considering a more lavender hue to distinguish from taxiways better. I'm still indecisive on whether busways should be highlighted with a particular color on the default map, or lowlighted with a dull gray.

Amsterdam Schiphol Airport:
Screenshot from 2022-07-10 19-12-43
Screenshot from 2022-07-10 19-13-04

Almere, Netherlands:
Screenshot from 2022-07-10 19-13-25
Screenshot from 2022-07-10 19-14-07

1ec5:
The “highlighting” in https://github.com/ZeLonewolf/openstreetmap-americana/pull/477#issuecomment-1179820878 reminds me of how city street maps highlight road routes in highlighter/pastel colors while keeping the width realistic. This will be an important tool at high zoom levels once we implement realistic road widths (#372 #486), so I’m a bit hesitant about introducing road fill colors for other purposes such as distinguishing modes of transportation. What if we go back to the grayed-out fill but change the casing color?

ZeLonewolf:
Thinking about this more, should we treat a busway like a no-access service road?

ZeLonewolf:
Would it make sense to come up with a design scheme for roads tagged `access=private` and similar, and then unify that with busways?

ZeLonewolf:
Also note that a legend entry is now needed.

claysmalley:
I don't think it's clear enough from casing color alone that this is not a legally driveable road. We'll need to decide how to display private-access roads before we can fit busways into the picture.

Screenshot from 2023-01-11 12-37-24

1ec5:
> I agree that what we're searching for here is the general case for how we want that collection of features to render.

Frankly, I don’t think it’s necessary to make all public transportation infrastructure the same color, as if all public transportation is a thematic overlay apart from the basemap. Public transportation is such a diverse range of features. Two users may not even agree whether air or rail transportation counts as public transportation, based on what’s available in their respective areas, so I don’t think we can count on them to intuit the relationship based on a color alone.

> I don't think it's clear enough from casing color alone that this is not a legally driveable road. We'll need to decide how to display private-access roads before we can fit busways into the picture.

I think it should be the other way around: we’re already showing plenty of restricted-access roads undifferentiated from public-access roads, so adding a few that happen to be busways doesn’t exacerbate the problem very much.

Let’s figure out a general restricted-access treatment separately in #679. In the meantime, a busway shares more in common with a public-access service road than it does with vacant land, which is how we’re depicting busways currently. If we’re concerned about confusion with drivable roads, we could have the label simply say “Busway” instead of the road name, though many will already have “Busway” in the name.

claysmalley:
Bringing back colorful busways, this time in a desaturated mauve to match the public transit icons added in #845.

[Screenshot from 2023-04-06 10-55-49](http://localhost:1776/#map=15.55/52.378615/4.899571)

[Screenshot from 2023-04-06 10-54-19](http://localhost:1776/#map=15.55/52.378615/4.899571)

[Screenshot from 2023-04-06 10-53-01](http://localhost:1776/#map=11.95/52.3611/5.21281)

claysmalley:
Guided busways are included now, styled identically.

[Screenshot from 2023-04-06 11-54-12](http://localhost:1776/#map=13.3/52.22844/0.1346)

ZeLonewolf (approved):
I think this is good enough to add to the map at this point.

We could probably argue about the exact width and level of prominence but scrolling around, it doesn't seem like it's overpowering, nor does it crowd out other features.

image

PR #480, opened 2022-07-09
Refactor shields with common shapes

claysmalley:
A much-needed refactor to give common shield shapes a consistent naming scheme.

A new shield definition parameter is introduced, `verticalReflect`, which draws a shield upside-down when set to `true`.

This was sort of a prerequisite to #458, since it introduces another convex triangle shield that's just a vertically flipped version of an existing shield. In the future, we may be able to further reduce the size of the sprite sheet by drawing some of these shapes directly with code.

Basic shapes include:

* `tri_rounded`
* `tri_convex`
* `diamond`
* `trapezoid`
* `trapezoid_rounded`
* `pent`
* `home`
* `hex_horz`
* `hex_vert`
* `oct`
* `badge`

In each case, a white shield with a black border is treated as the default color scheme. This black border can be recolored with `colorLighten`. Other color variants are specified by appending the background color first, then the border color. If no second color is specified for the border, white is assumed, except when the background is yellow, in which case a black border is assumed.

Additionally, the dimensions of the Malaysian route shield are adjusted to enlarge text and match other shields with similar shapes:

Screenshot from 2022-07-09 10-08-48


ZeLonewolf:
I would suggest dropping the 40 from the filenames also.

ZeLonewolf:
Seems to be working for me.  What are examples of shields using the new verticalReflect parameter?

claysmalley:
* Ontario secondary routes [(localhost link)](http://localhost:1776/#12.25/49.78238/-92.8281)
* Hawaii state routes [(localhost link)](http://localhost:1776/#13.8/21.3797/-158.03063)

claysmalley:
I forgot to multiply the pixel ratio by 2 - now it should look the same as before.

ZeLonewolf (approved):
Seems to be working here.  Ubuntu 22.04/Chrome

Issue #162, opened 2022-02-15
SADC trunk road shields


1ec5:
Routes belonging to the [SADC Regional Trunk Road Network](https://en.wikipedia.org/wiki/SADC_Regional_Trunk_Road_Network) in Southern Africa are tagged [`network=sadc`](https://taginfo.openstreetmap.org/tags/network=sadc) and should be marked by [blue-on-orange hexagons](https://commons.wikimedia.org/wiki/Category:Diagrams_of_route_markers_of_Southern_African_Development_Community_routes).

<img src="https://upload.wikimedia.org/wikipedia/commons/7/7a/SADC_road_sign_GE19-423.svg" height="54" alt="SADC 423">

ZeLonewolf:
Seems decently mapped.  This is equivalent to an e-road or the Asian network, so we should render this.  To be consistent with our style guideline and other shields, the orange backdrop should get removed in favor of a stroked orange line.

![image](https://user-images.githubusercontent.com/3254090/153993919-8cc98c0b-2f1f-45a7-b4c5-0c252bac53c2.png)

1ec5:
We’re unable to render these routes correctly because of what seems to be incorrect tagging:

* Some of these relations are tagged `int_ref=*` instead of `ref=*`. `int_ref` is only necessary for distinguishing the common and international reference numbers of a feature, typically on a way that would be numbered concurrently by multiple authorities. But these route relations are already international by definition, and a single route relation cannot represent a concurrency by itself.
* Other relations in this network are tagged `ref=SADC *`, which is similar to the problem described in https://github.com/ZeLonewolf/openstreetmap-americana/issues/141#issue-1125053298.

Once the tagging issues are fixed, it’ll be trivial to support `network=sadc`. Here’s a shield we can use:

![shield40_un_sadc](https://user-images.githubusercontent.com/1231218/155257013-81d00e0e-6d38-4682-8032-6eb850d509d0.svg)

1ec5:
[...]

Issue #85, opened 2022-01-21
Branch Motorways in Italy


ZeLonewolf:
Branch motorways in Italy are numbered based on the two motorways that they connect:

![image](https://user-images.githubusercontent.com/3254090/150456656-b24bc9e8-e8f2-4ebd-a7cb-d099edc56510.png)

The shields on these roads use a stacked configuration as follows:

![image](https://user-images.githubusercontent.com/3254090/150456767-6a16f4f8-71f1-4eaf-a150-3086ce34bbfc.png)

In the database, these route relations use the form `ref=A7/A26`.  These will need to be parsed and split by the slash in a vertical configuration.

Mashin6:
Would '/' --> '\n' replacement work or are special characters not supported for text rendering?

ZeLonewolf:
Javascript canvas doesn't have multi line support, so that wouldn't work out of the box.

claysmalley:
Stacked refs tracked in #366.

PR #483, opened 2022-07-10
Add South Africa national route shields

claysmalley:
Adds shield support for national routes of South Africa. National routes have a blue pentagonal shield with a white border and yellow text. All routes in South Africa have a letter prefix, which is shown on shields in small font above the route number, though for the purposes of this project, the prefix is displayed inline with the route number.

South Africa has additional route networks on a more local level (provincial, regional, metropolitan), though tagging would need to be more complete and consistent before these routes can have shield support: #484 

Screenshot from 2022-07-09 17-49-55
SA_road_N1


1ec5:
For context, South Africa’s directions signs follow a UK design: no shield, just a number in plain text in gold, with the alphabetic designation inline.



However, the blue pentagon remains [in the SADC-RTSM](https://www.transport.gov.za/documents/11623/21913/SADCRoadTrafficSignsManualVolume4pages487to547.pdf/d96f986b-7699-4c66-b3db-406616e7f22c#page=11) and reportedly on maps. On the pentagon, the “N” appears above the route number, and a cardinal direction can optionally appear below the route number. I think it would be suboptimal to include inline prefixes on these shields or the ones in #484, just as it would be suboptimal to include “US” on a California-style U.S. Route shield.

ZeLonewolf:
This seems like a reasonable graphical representation to me based on the limited photos shared.

1ec5:
Between this and #162, I suspect the prefixes are just there for consistency with way `ref`s. The photos above aren’t representative of how routes are marked on maps. Normally the pentagon contains just the route number, or it has “N” above the route number, just like on what would be the trailblazer sign.

[](https://zoom-maps.com/southafrica) [Freytag & Berndt](https://maps2anywhere.com/south-africa-road-and-shaded-relief-tourist-map.html)
[MapStudio](https://www.mapstudio.co.za/product/businessmans-wall-map-south-africa/)

I think we should at least give the South African community a heads-up about this issue or address it on our end via #141. If we land national route shields with the prefixes, they’ll look out of place next to provincial and metropolitan routes (#484), which likely won’t have enough horizontal space for an inline prefix.

claysmalley:
I guess a potential solution would be to treat the prefix as banner text floating above the shield, which would involve removing the prefixes from `ref` values in `route=road` relations. I had been considering something similar with Brazil (#425).

claysmalley:
Shield icon scrapped in favor of draw function:

Screenshot from 2023-01-16 12-43-01
Screenshot from 2023-01-16 12-42-40

ZeLonewolf:
This is an older PR, but I'd be in favor of merging this as is and opening a separate issue and/or adding to #85 to detail the desire for stacked text in the shield.

Edit: Stacked refs tracked in #366

claysmalley:
Text is really butting up against the border here. But that can be a separate issue—I've been meaning to give the text padding a haircut on pentagon shields.

Example: https://preview.americanamap.org/pr/483/#map=9.02/-25.7743/28.208

[image]()

image

Pengor (approved):

PR #490, opened 2022-07-13
Add shields for national routes of France

claysmalley:
Adds support for French national routes, which have identical signage to French autoroutes.


ZeLonewolf:
Looks like there are also C- and D- routes, but they appear to be mapped in a funny way.

claysmalley:
Yes, see here: #491

PR #492, opened 2022-07-13
Add Turkish motorway shields

claysmalley:
Adds shield support for motorways of Turkey.

Screenshot from 2022-07-13 18-59-44
O3-TR_sign


src/js/shield_defs.js
```diff
@@ -2716,6 +2704,26 @@ export function loadShields(shieldImages) {
...
+    Color.shields.white,
+    Color.shields.black
+  );
+
+  // Philippines
+  shields["PH:N"] = homeDownShield;
+  shields["PH:E"] = homeDownShieldYellow;
+
+  // Turkey
+  shields["TR:motorway"] = hexagonVerticalShieldOrange;
```
1ec5, line 2725: Nit: “Taiwan” sorts before “Turkey” (or “Türkiye”).

1ec5 (approved):
Interesting that the dash looks so close to an underscore or period on the actual signs, but I suppose that’s just a typographical flourish.

Let’s ticket out shields for Turkey’s state roads (D) and, if signposted, the provincial roads too.

src/js/shield_defs.js
claysmalley: That's intentional; it's sorted by ISO country code.

claysmalley:
State roads tracked here: #494

Issue #465, opened 2022-06-28
Refs with other numeral systems


claysmalley:
Some countries have route shields posted with local numeral systems alongside Western Arabic numerals. These could potentially be displayed side-by-side, or one at a time with language switching (#21).

For example, Iran (#463, #464) posts Persian numerals:

![Iran_Freeway_7](https://user-images.githubusercontent.com/1732117/176074500-4b8e1fbe-ba00-4fe6-ae60-a9ba1711e6ba.png) ![Iran_First_Level_Road_44](https://user-images.githubusercontent.com/1732117/176074596-2c758fee-ae5d-480a-8d35-49c2f98a6391.png) ![Iran_Second_Level_Road_122](https://user-images.githubusercontent.com/1732117/176074571-f0a54df1-acad-4827-8df4-65126f3dca66.png)

In the case of Iran, the route relations are tagged with Western Arabic numerals in `ref=*` and Persian numeral in `ref:fa=*`. Only `ref=*` is surfaced in OpenMapTiles, and none of the available fonts have glyphs for Persian numerals. These issues would need to be resolved before support can be added for Persian numerals on Iranian highways.

1ec5:
Saudi Arabia posts its route numbers only in Eastern Arabic numerals, but currently they’re tagged only in Western Arabic numerals: #467.

1ec5:
For some ROC shields, we’re showing Chinese characters instead of the corresponding alphabetic suffixes because that’s what’s in the vector tiles: https://github.com/ZeLonewolf/openstreetmap-americana/issues/141#issuecomment-1157248693.

1ec5:
In addition to numeral systems, some `ref`s contain homoglyphs. For example, in Greece, we display shields for [E roads](https://www.openstreetmap.org/relation/5953973) using the Latin “E” but [motorways](https://www.openstreetmap.org/relation/362575) using the Greek “Α” and [national roads](https://www.openstreetmap.org/relation/7972880) using the Greek “ΕΟ”.

[<img src="https://user-images.githubusercontent.com/1231218/177015045-7bbea58b-9bec-43a5-b6bd-f5538afffcce.png" width="400" alt="Greece">](https://zelonewolf.github.io/openstreetmap-americana/#17.05/38.782723/22.706757)

[...]

PR #493, opened 2022-07-13
Add Armenia shields

claysmalley:
Adds shield support for highways of Armenia.

Screenshot from 2022-07-13 19-13-37
H84_road_(Armenia) M2_Road_signs_of_Armenia


1ec5 (approved):
The Armenian characters may look out of place on some systems where the default fonts have very different metrics. In Latin script, Մ and Հ roads are considered to be prefixed “M” and “H”, respectively. But there’s little we can do about that unless we attempt transliteration (#141 #465). There are also Տ (T) local roads, but as far as I can tell, they aren’t signposted with shields.

1ec5:
By the way, the hyphens in the `ref`s are probably a tagging issue we could raise with the Armenian community. There doesn’t seem to be a basis for the hyphens, other than consistency with other countries’ `ref` formats.

PR #496, opened 2022-07-14
Add U.S. insular areas to shield coverage world map

1ec5:
Added United States insular areas other than Puerto Rico (#240) to the shield coverage world map for consistency with the shield coverage U.S. map.


ZeLonewolf (approved):
Thanks!

PR #501, opened 2022-07-18
Add US:I:Future

jleedev:
Relations with `network=US:I:Future` and `ref=*` appear to represent "Future" interstate routes that are signed in the field with real Interstate shields that say "Future" in the red part of the shield, or as a plaque. This is distinct from e.g. "Future I-nn Corridor" that are not signed as such, and are tagged with `fut_ref=*`.

Add a FUT banner to render these.

Future I-86 and Future I-99 _might_ be mistagged — will render here but might be signed in the field as "Future Corridor" — but I'm not sure. The other ones are all in NC and look correct if Wikipedia can be trusted.

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -673,6 +673,7 @@ export function loadShields(shieldImages) {
   };
 
   shields["US:I:Alternate"] = banneredShield(shields["US:I"], ["ALT"]);
+  shields["US:I:Future"] = banneredShield(shields["US:I"], ["FUT"]);
   shields["US:I:Truck"] = banneredShield(shields["US:I"], ["TRK"]);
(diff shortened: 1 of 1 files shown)


jleedev:
Example from North Carolina of Interstate 26 suspiciously lacking the word "INTERSTATE" in the shield and wearing a FUTURE plaque:
https://www.mapillary.com/app/?pKey=982222942314543

Example of I-86 and I-99 being "Future (Corridor)" and should be fixed in the map:
https://commons.wikimedia.org/wiki/File:Future_I-99_Corridor.jpg
https://en.wikipedia.org/wiki/File:NY17_Future_I-86.jpg

jleedev:
image

PR #502, opened 2022-07-19
Add highways under construction

### LINKED ISSUE #212: Roads under construction

jleedev:
Fixes #212

Treatment is minimalist. Main purpose served is to identify what a gap in the road network might represent, particularly for a long-term road closure. Only attributes included are whether it is trunk/motorway (red) or other roads (blue). Includes the name, in that same color, which additionally contrasts with regular roads.

Omits path, track, and raceway.

image

Doesn't use the words "under construction"; the dotted line should imply the noncorporeal nature of the feature, and the color (and the geometry these features tend to have) should clue the viewer that this is related to the highways.

Includes the name property only. While some roads under construction might only have a ref and some route relations, it generally looks weird to have shields floating above roads that aren't there yet, and it looks perfectly fine for a highway under construction to not have any label.

Layer count: 2.
Avoids attempting to render "the entire road style, but dashed", as that proved to be impracticable in #215, in taming the style, the number of layers, and visually. Ignoring bridge/tunnel is probably the right choice.

This should be sufficiently distinct from railways, unpaved roads, and admin_8 borders, but maybe not.

image
image

### DIFF
--- a/src/layer/construction.js
+++ b/src/layer/construction.js
@@ -60,7 +60,7 @@ export const road = {
       "interpolate",
       ["exponential", 2],
       ["zoom"],
-      12,
+      11,
       0,
       20,
       2,
--- a/src/layer/construction.js
+++ b/src/layer/construction.js
@@ -1,7 +1,7 @@
 const majorConstruction = [
-  "any",
-  ["in", "motorway", ["get", "class"]],
-  ["in", "trunk", ["get", "class"]],
+  "match",
+  ["get", "class"],
+  ["motorway_construction", "trunk_construction"],
 ];
 
 const constructionColor = [
@@ -9,11 +9,11 @@ const constructionColor = [
   ["exponential", 2],
   ["zoom"],
   10,
-  ["case", majorConstruction, "lightcoral", "lightslategray"],
+  [...majorConstruction, "lightcoral", "lightslategray"],
   13,
-  ["case", majorConstruction, "maroon", "lightslategray"],
+  [...majorConstruction, "maroon", "lightslategray"],
   15,
-  ["case", majorConstruction, "maroon", "slategray"],
+  [...majorConstruction, "maroon", "slategray"],
 ];
 
 const constructionFilter = [
--- a/src/layer/construction.js
+++ b/src/layer/construction.js
@@ -43,12 +43,13 @@ export const road = {
   paint: {
     "line-color": constructionColor,
     "line-opacity": ["interpolate", ["exponential", 2], ["zoom"], 10, 0, 11, 1],
+    "line-blur": 0.75,
     "line-width": 1,
-    "line-dasharray": [2, 1],
+    "line-dasharray": [2.5, 1.25],
     "line-offset": 0,
     "line-gap-width": [
       "interpolate",
-      ["exponential", 2],
+      ["exponential", 1.2],
       ["zoom"],
       11,
       0,
--- a/src/layer/construction.js
+++ b/src/layer/construction.js
@@ -1,0 +1,91 @@
+const majorConstruction = [
+  "any",
+  ["in", "motorway", ["get", "class"]],
+  ["in", "trunk", ["get", "class"]],
+];
+
+const constructionColor = [
+  "interpolate-lab",
+  ["exponential", 2],
+  ["zoom"],
+  10,
+  ["case", majorConstruction, "lightcoral", "lightslategray"],
+  13,
+  ["case", majorConstruction, "maroon", "lightslategray"],
+  15,
+  ["case", majorConstruction, "maroon", "slategray"],
+];
+
+const constructionFilter = [
+  "in",
+  ["get", "class"],
+  [
+    "literal",
+    [
+      "motorway_construction",
+      "trunk_construction",
+      "primary_construction",
+      "secondary_construction",
+      "tertiary_construction",
+      "minor_construction",
+      "service_construction",
+    ],
+  ],
+];
+
+export const road = {
+  id: "highway-construction",
+  type: "line",
+  source: "openmaptiles",
+  "source-layer": "transportation",
+  filter: constructionFilter,
+  minzoom: 9,
+  paint: {
+    "line-color": constructionColor,
+    "line-opacity": ["interpolate", ["exponential", 2], ["zoom"], 10, 0, 11, 1],
+    "line-width": [
+      "interpolate",
+      ["exponential", 2],
+      ["zoom"],
+      9,
+      0,
+      12,
+      ["case", majorConstruction, 1, 0.5],
+      15,
+      1,
+    ],
+    "line-dasharray": [2, 1],
+    "line-offset": 0,
+    "line-gap-width": [
+      "interpolate",
+      ["exponential", 2],
+      ["zoom"],
+      12,
+      0,
+      20,
+      2,
+    ],
+  },
+};
+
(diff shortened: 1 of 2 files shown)


ZeLonewolf:
Surprisingly workable based on the screen shots.

jleedev:
Ah, meant to include some over water:

Screen Shot 2022-07-19 at 4 26 18 PM
Screen Shot 2022-07-19 at 4 26 09 PM
Screen Shot 2022-07-19 at 4 25 50 PM

ZeLonewolf:
So the main issue with the prior approach was with dual carriageways merging dash patterns at lower zooms.  Do these look okay at the lowest zoom they appear?

adamfranco:
This looks reasonable to me. The color makes it distinguishable from future (likely brown) track rendering.

jleedev:
Slightly chunky at z12:

image

Slightly mushy at z11.9:

image

This is somewhat unavoidable when using dashes, in part due to them jumping at integer zoom levels. But I think it looks ok.

ZeLonewolf:
Yeah that seems again surprisingly workable.  Like to get a review from @1ec5 as the author of the previous attempt but otherwise looks good to me.

src/layer/construction.js
```diff
@@ -0,0 +1,91 @@
...
+    "line-offset": 0,
+    "line-gap-width": [
+      "interpolate",
+      ["exponential", 2],
+      ["zoom"],
+      11,
+      0,
+      20,
+      2,
+    ],
```
1ec5, line 67: Good idea to make everything thin, including the dashes. That neatly gets around the issues with overlapping, mismatched dashes that I had run into in #215. However, the dashing can be a very subtle effect at some zoom levels. In the screenshot below, the U-shaped road is also under construction:

[Vallée-des-Forts](https://zelonewolf.github.io/openstreetmap-americana/#14/45.09346/-73.08425)

Can we widen the gaps a bit at the lower zoom levels to keep the roads from looking like a different classification of corporeal road? It seems to b more of an issue with non-motorways; the blue contrasts less well with the land background color.

We should also double-check the output on 1× displays.

src/layer/construction.js
```diff
@@ -0,0 +1,91 @@
...
+  type: "symbol",
+  source: "openmaptiles",
+  "source-layer": "transportation_name",
+  filter: constructionFilter,
+  minzoom: 15,
+  layout: {
+    "symbol-placement": "line",
+    "text-font": ["Metropolis Light"],
+    "text-size": 12,
+    "text-field": "{name}",
```
1ec5, line 82: I’d still be interested in revisiting the label glosses from #215 at some point. Without a legend that’s in the user’s face, any relatively rare line treatment will be subject to misinterpretation. This also goes for busways (#308), for example. But we can tackle that in a separate issue. The differently colored labels already do a great job of signaling to the user that there’s something altogether different about these roads.

1ec5 (commented):
Thanks for taking a fresh look at this issue. This approach seems intuitive enough and avoids piling onto the layer situation (which is actually the main thing that caused #215 to stall in my opinion). Even a major freeway under construction doesn’t _need_ to be in the user’s face, so long as it fills a gap.

src/layer/construction.js
jleedev: Changed line-width to 1, always, causes the dashes and gaps to be uniform across both stylings:

image

jleedev:
Doesn't read well at low resolution, scaled up ever so slightly.

Before:
Screenshot 2022-07-19 18 44 20

After:
Screenshot 2022-07-19 18 44 27

src/layer/construction.js
```diff
@@ -0,0 +1,82 @@
...
+  "interpolate-lab",
+  ["exponential", 2],
+  ["zoom"],
+  10,
+  ["case", majorConstruction, "lightcoral", "lightslategray"],
+  13,
+  ["case", majorConstruction, "maroon", "lightslategray"],
+  15,
+  ["case", majorConstruction, "maroon", "slategray"],
+];
```
1ec5, line 17: More concise:

```suggestion
const majorConstruction = [
  "match",
  ["get", "class"],
  ["motorway_construction", "trunk_construction"],
];

const constructionColor = [
  "interpolate-lab",
  ["exponential", 2],
  ["zoom"],
  10,
  [...majorConstruction, "lightcoral", "lightslategray"],
  13,
  [...majorConstruction, "maroon", "lightslategray"],
  15,
  [...majorConstruction, "maroon", "slategray"],
];
```

1ec5 (approved):
My taxpayer dollars at work!

src/layer/construction.js
jleedev: Ah, clever.

PR #504, opened 2022-07-20
Add airport, road construction to taginfo project file

### LINKED ISSUE #354: Add airport-related tags to taginfo project file
The [taginfo project file](https://github.com/ZeLonewolf/openstreetmap-americana/blob/main/style/scripts/taginfo_template.json) should describe this project’s support for the airport-related keys and tags added in #318. For example, it could describe how `iata` or `icao` is used to label airports at lower zoom levels. We probably only need to document the tags that we handle differently than a more established style like openstreetmap-carto.

1ec5:
The taginfo project file now documents tag support introduced in #318 (airports) and #502 (roads under construction).

Fixes #354.

/cc @AnonymousAlligatorOSM

### DIFF
--- a/scripts/taginfo_template.json
+++ b/scripts/taginfo_template.json
@@ -154,6 +154,33 @@
       "object_types": ["area"],
       "description": "Runway areas are filled in with the same color as runway lines.",
       "doc_url": "https://openmaptiles.org/schema/#aeroway"
+    },
+    {
+      "key": "highway",
+      "value": "construction",
+      "object_types": ["way"],
+      "description": "Roads under construction have a dotted line pattern.",
+      "doc_url": "https://openmaptiles.org/schema/#transportation"
+    },
+    {
+      "key": "construction",
+      "object_types": ["way"],
+      "description": "Roads under construction have a dotted line pattern.",
+      "doc_url": "https://openmaptiles.org/schema/#transportation"
+    },
+    {
+      "key": "construction",
+      "value": "motorway",
+      "object_types": ["way"],
+      "description": "Major roads under construction have a dotted line pattern and a more prominent color.",
+      "doc_url": "https://openmaptiles.org/schema/#transportation"
+    },
+    {
+      "key": "construction",
+      "value": "trunk",
+      "object_types": ["way"],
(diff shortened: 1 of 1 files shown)


jleedev:
Taxiway areas?

1ec5:
> Taxiway areas?

Thanks, added.

> I noticed one of the tags is `ref`, which is listed as being used for airport gates. But we use `ref` for other things (highway shields, exit numbers, possibly others I forgot), should that be more generic?

The taginfo project file can contain multiple entries for the same key or tag. There’s already an entry about `ref` on route relations. I added one about `ref` on freeway exit nodes.

scripts/taginfo_template.json
```diff
@@ -65,11 +65,141 @@
       "description": "Populates shields that mark roads carrying the route relation.",
       "doc_url": "https://openmaptiles.org/schema/#route_1"
     },
+    {
+      "key": "unsigned_ref",
+      "object_types": ["relation"],
+      "description": "Suppresses any shields that would mark the roads carrying the route relation, unless the network=* consists of unnumbered routes.",
```
1ec5, line 71: `unsigned_ref` is interesting because we “support” it by omission. There’s no mention of the key in either this repository or OpenMapTiles; rather, it’s a tagging convention we lend our support to by the fact we render `ref` unconditionally.

ZeLonewolf: Pwrhaps unsigned_ref should state something like that it holds a value that would otherwise be in ref if it were signed

1ec5: Taginfo already incorporates the wiki’s definitions of these tags. The purpose of this file is to supplement that definition with how _we_ use the data.

ZeLonewolf: Yes, but our description is wrong.  Unsigned_ref suppresses nothing if ref is also set.

1ec5: OK, I see what you mean. I added a clause explaining that `ref` takes precedence over `unsigned_ref`.

ZeLonewolf (approved):
Seems ok by me.  Please sanity check when taginfo updates.

PR #505, opened 2022-07-21
Refactor some toll roads to generate styling with expressions

claysmalley:
Removes 58 layers by generating toll styling with Maplibre expressions instead of separate layers. Non-integer zoom levels are not handled quite like expected in expressions, so not all highway classes made it into this PR.

Highway classes affected:
* motorway
* motorway_link
* trunk
* trunk expressway
* trunk_link
* primary expressway
* secondary expressway
* tertiary expressway


ZeLonewolf:
Just tested and everything seems to be working as far as I can tell.  Like to have a review from @jleedev who was working on reducing the layer count in the roads layer.

claysmalley:
Forgot `layer-sort-key` to ensure toll roads render above free roads.

jleedev (approved):
Appears to also fix missing surface on certain classes too.

PR #509, opened 2022-07-23
Re-order water layer

### LINKED ISSUE #507: Waterbody labels underlap buildings
Labels of bodies of water are layered beneath translucent buildings. When the labels underlap the faint buildings, the labels look splotchy, as if parts of letters are missing. In general, symbol layers should be above fill and line layers to avoid underlap, but they should especially be above translucent fill layers to avoid splotchiness.

[Screenshot from 2022-07-21 13-52-27](https://zelonewolf.github.io/openstreetmap-americana/#15.23/28.419349/-81.579808)

[Guayama](https://zelonewolf.github.io/openstreetmap-americana/#14/17.97534/-66.11473)

/ref https://github.com/ZeLonewolf/openstreetmap-americana/pull/476#issuecomment-1191779541 https://github.com/ZeLonewolf/openstreetmap-americana/pull/497#issuecomment-1190973068

ZeLonewolf:
Fixes #507

This PR re-orders the water layer labels to be on top of other feature types.  The original thought of putting the water labels lower was to have it go "under" things like bridges, but I think that approach is no longer in vogue and a bit to avant garde for what we're trying to do here.

The screen shot below still feels a little clash-y but it solves the specific issue described in the bug report.

image

### DIFF
--- a/src/americana.js
+++ b/src/americana.js
@@ -59,10 +59,6 @@ americanaLayers.push(
   lyrFerry.ferry,
   lyrFerry.ferryLabel,
 
-  lyrWater.waterwayLabel,
-  lyrWater.waterLabel,
-  lyrWater.waterPointLabel,
-
   lyrBackground.pierArea,
   lyrBackground.pierLine,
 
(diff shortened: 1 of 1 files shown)


1ec5 (approved):
Thanks!

PR #510, opened 2022-07-23
Remove Interstate layer from high zoom

claysmalley:
The `InterstateMotorway` class has instance variables `maxZoomCasing` and `maxZoomFill` that were unused. This PR uses them to block the layer from showing up at higher zoom and covering motorway brunnels.

before:
Screenshot from 2022-07-23 12-03-46

after
Screenshot from 2022-07-23 12-02-58


ZeLonewolf (approved):
Looks good!

PR #511, opened 2022-07-23
Change label halo color for roads under construction

claysmalley:
Changes the label halo added in #502 from white to the background fill color to match other road labels.

Screenshot from 2022-07-23 15-41-27


jleedev (approved):

jleedev:
Well, that appears to be a "before" screenshot.

Screenshot_20220723-174748

Issue #486, opened 2022-07-11
Roadway footprints


1ec5:
The current transition to cased roads as you zoom in points strongly in the direction of giving roads a realistic width at higher zoom levels. #372 tracks improving the default; maybe down the line we can even consider tags such as `lanes`, `width`, and `placement`. But these tags can only describe the roadway’s shape so well; the style specification lacks affordances for sophisticated transitions between roadway segments of differing widths, or for tapered segments.

Where [`area:highway`](https://wiki.openstreetmap.org/wiki/Key:area:highway) areas have been mapped, we should use them to flesh out the roadway’s footprint at high zoom levels. For example, a roadway area fill layer could sit immediately beneath the road layer, and a roadway area casing layer (also a fill layer, slightly buffered) could sit immediately beneath the road casing layer. Labels, shields, and one-way arrows would continue to be placed as they are today. For visual and code simplicity, we should aim to give most road classifications at these zoom levels the exact same treatment, although some sort of highlighter overlay effect might be useful for certain important classifications or routes. Pedestrian malls would also require some visual distinction: #185.

@BjornRasmussen has helpfully mapped [a dense concentration of `area:highway` areas](http://overpass-turbo.eu/s/Xi7) in the Research Triangle of North Carolina that we can use as a rendering testbed.

ZeLonewolf:
Areas tagged `area:highway` are able to be rendered today in planetiler with the new configurable schema option.  We could potentially generate a custom layer which could be hosted in order to experiment with this capability.  I would be skeptical that we could get highway areas added to openmaptiles on any reasonable timeline.

1ec5:
There’s a lot of underutilized space at high zoom levels that we could use to make the map [more pedestrian-oriented](https://community.openstreetmap.org/t/pedestrian-centric-maps/97629) in urban areas, taking inspiration from city guide maps like the ones above. `area:highway` areas are the main missing piece to get to a minimally viable city guide aesthetic.

PR #512, opened 2022-07-24
Add oneway arrows to remaining highway classes

claysmalley:
Adds oneway arrows to non-motorway highway classes, with half the layers.

Screenshot from 2022-07-24 19-39-28


ZeLonewolf:
I honestly didn't even realize this was missing.

ZeLonewolf:
It appears that this change also unifies the link and regular versions of each road class, is that also a deliberate consolidation?

ZeLonewolf:
#513 would be helpful in assessing changes in layer count.

claysmalley:
> It appears that this change also unifies the link and regular versions of each road class, is that also a deliberate consolidation?

Yes - were they separate layers intentionally?

ZeLonewolf:
> Yes - were they separate layers intentionally?

Only in the sense that I (a) didn't really know what I was doing and (b) didn't realize layer count had a performance implication.

ZeLonewolf:
The changes appear to be working from visual inspection.

The arrows on motorway_link are hard to see and we may want to open a separate issue to track it, as reducing layer count is a clear improvement by itself.

image

claysmalley:
Adding oneway arrows seems to cause a significant slowdown at very high zoom, so it's hidden at ≥z20.

ZeLonewolf:
> Adding oneway arrows seems to cause a significant slowdown at very high zoom

This sounds dodgy.

jleedev:
_Something_ is extremely slow at high zooms already.

claysmalley:
How about white arrows for motorways and trunks?

Screenshot from 2022-07-24 22-41-24

claysmalley:
On second thought, white is even less legible against the red on trunk_link. Changing trunk oneway arrows back to black.

Screenshot from 2022-07-24 22-45-36

ZeLonewolf:
There is an artifact in cases of a link road:

image
Location ([localhost link](http://localhost:1776/#17/41.513485/-71.367262))

Is this resolvable with a change in ordering or unavoidable?  Is there a way to prevent the arrows from being drawn too close to the ends of a road?

claysmalley:
We would have to sandwich oneway arrows in between two road fill layers, which would hinder our ability to refactor roads to use more expressions and fewer layers. I don't think there's an easy solution to this.

1ec5:
> Is there a way to prevent the arrows from being drawn too close to the ends of a road?

It is possible to use invisible spacer symbols to collide out any arrows that would be drawn at the endpoints, similar to how we’re pushing waterway labels away from bridges. In this case, the symbol’s [`symbol-placement`](https://maplibre.org/maplibre-gl-js-docs/style-spec/layers/#layout-symbol-symbol-placement) property would be set to `point` so that it gets placed at one of the endpoints instead of repeated along the line. We’d need to pay attention to the spacing settings so that shields and road names don’t become less dense because of collisions suddenly occurring at junctions – or junctions on nearby roadways. Also, openmaptiles/planetiler-openmaptiles#10 proposes to fix wrong-way one-way arrows by disabling line merging of one-way lines; with that stopgap fix in place, the spacers would frequently prevent any one-way arrows or shields from showing up at all on one-way streets.

/ref mapbox/mapbox-gl-js#4003

1ec5:
One-way arrows only come in at z15 for freeways. At this zoom level, we’re already starting to show surface streets in the “cased” style. I think this gives us some flexibility to deemphasize color distinctions based on road classification. It’s also at a high enough zoom level that there’s a less ambiguity as to which ramp goes in which direction, so maybe the ramp arrows aren’t essential.

/ref #486

claysmalley:
Oneway arrows removed from link roads:

Screenshot from 2022-08-30 20-54-39

adamfranco:
Hmm... Without arrows on link roads, figuring out whether a ramp is an on or off requires the eye tracing it up to the lanes it diverges from. I imagine that there are a few cases of stacked interchanges where that becomes really difficult to tease out visually. (for example, the [GW bridge](https://zelonewolf.github.io/openstreetmap-americana/#17.49/40.849807/-73.941963))

ZeLonewolf:
These looks pretty good for all but the lowest highway classes.

For the tiniest roads (alleys and so forth), I'm not sure if I'm seeing arrows or if there's a speck on my laptop screen that needs to be cleaned.  Not sure if there's a good solution to this.  Perhaps a hard minimum size to the arrow and let it overflow a bit?

[Location](https://preview.ourmap.us/pr/512/#map=17/41.490536/-71.31307):
image

Wondering if we might

claysmalley:
This better?

image

ZeLonewolf:
This looks good!

I note that bridges are not included ([example](https://preview.ourmap.us/pr/512/#map=15/41.52993/-71.40433)), however, bridges appear to be not supported presently. Is that intended to be part of this PR or left as future work?

claysmalley:
> I note that bridges are not included ([example](https://preview.ourmap.us/pr/512/#map=15/41.52993/-71.40433)), however, bridges appear to be not supported presently. Is that intended to be part of this PR or left as future work?

That's a mistake. They're in the list of layers, so I don't know why they aren't showing up.

claysmalley:
Figured it out. The culprit was the invisible spacer label that prevents waterway labels from colliding with bridges. This spacer is now disabled if the bridge is tagged `oneway=yes`.

[image](https://preview.ourmap.us/pr/512/#map=15.01/42.36907/-71.06575)

ZeLonewolf (approved):
Good work.  Thanks for sticking it out!
This would be a great use case for #881 if we're not in a rush to push the change.

jleedev:
The arrows seem to serve the spacers' purpose except at _really_ high zooms.

Screenshot_20230621-123530

claysmalley:
> The arrows seem to serve the spacers' purpose except at really high zooms.

Before removing the spacer, oneway arrows would _only_ show up at very high zooms. Something weird is going on and it's cooking my CPU.

jleedev:
Ah, it might be that some combination of spacers + oneway arrows + lettering causes this.

I still also hypothesize that it's overzooming the z14 tile to z22 and then stuffing it full of symbols, i.e. if higher zoom tiles existed that were simply slices of the full ones then it would attempt to do less work.

Might also be that the waterway lettering is attempting to overlap itself severely, could see if the slowdown is localized depending on the presence of waterways or oneways.

https://osmus.slack.com/archives/C01V02K52UX/p1658777024337899

PR #515, opened 2022-07-26
Add shields for New York parkways

claysmalley:
Adds shields for `US:NY:Parkway` and `US:NY:Parkway:NYC` networks.

Still missing are shields for parkways of Long Island and the Palisades Interstate Parkway Commission, because consistent tagging for these routes doesn't seem to have solidified yet.

Screenshot from 2022-07-26 18-24-51
Screenshot from 2022-07-26 18-25-10
jackie-robinson-parkway
Hutchinson_River_Pkwy_Shield


src/js/shield_defs.js
```diff
@@ -3188,35 +3216,44 @@ export function loadShields(shieldImages) {
...
+    "Bronx River Parkway": "BRP",
+    "Cross County Parkway": "CCP",
+    "Hutchinson River Parkway": "HRP",
+    "Korean War Veterans Parkway": "KWVP",
+    "Mosholu Parkway": "MP",
+    "Pelham Parkway": "PP",
+    "Saw Mill River Parkway": "SMP",
+    "Sprain Brook Parkway": "SBP",
+    "Taconic State Parkway": "TSP",
+  };
```
1ec5, line 3255: Just noting here that, aside from the Long Island parkways, these other NYSDOT parkways have been omitted for unavoidable reasons:

* The [Niagara Scenic Parkway](https://www.openstreetmap.org/relation/418069) is currently tagged `network=US:NY`. It has a distinct shield without drop caps, though it would likely simplify down to the same style being used for the other NYSDOT parkways if folks are OK with it saying “NSP” inside anyways.
* The [Lake Ontario State Parkway](https://www.openstreetmap.org/relation/408496) is currently tagged `network=US:NY` `ref=LOSP`. It has a unique shield, quite unlike any of the others.

Let’s ticket out the ones not included here so we don’t lose track of them.

src/js/shield_defs.js
```diff
@@ -3188,35 +3216,44 @@ export function loadShields(shieldImages) {
...
-        "Mountain Pkwy": "MP",
-        "Purchase Parkway": "JC",
-        "Purchase Pkwy": "JC",
-        "Western Kentucky Parkway": "WK",
-        "Western Kentucky Pkwy": "WK",
-      },
-    },
-    shields["US:KY:AA"]
-  );
+  shields["US:KY:Parkway"].refsByWayName = {
```
1ec5, line 3219: Thanks, this reorganization makes a lot of sense.

1ec5 (approved):
The changes look good to me on a technical level and as someone who [likes initialisms a little too much](https://github.com/ZeLonewolf/openstreetmap-americana/pull/119).

It would be great if we could get some quick feedback from a local. I think there were some strong feelings about how to tag these routes uniquely. Fortunately, this PR doesn’t actually rely on any keys other than `network` and `name`, which I’ve been viewing as a stopgap until we can get access to `short_name` or similar.

src/js/shield_defs.js
claysmalley: Tracked in #516, #517.

PR #518, opened 2022-07-28
Data-driven styling of road and ferry labels

claysmalley:
This PR changes labels for roads and ferries to data-driven expressions. This reduces the layer count by about 1% :grimacing:  

Also moves ferry lines above tunnels:

Screenshot from 2022-07-27 20-43-04


src/layer/transportation_label.js
```diff
@@ -0,0 +1,141 @@
...
+      5,
+    ],
+  },
+  source: "openmaptiles",
+  "source-layer": "transportation_name",
+};
+
+// A spacer label on each bridge to push any waterway label away from the bridge.
+// https://github.com/ZeLonewolf/openstreetmap-americana/issues/198
+export const bridgeSpacer = {
```
1ec5, line 125: I forgot about this – we might want to keep waterway labels away from railroad bridges as well in #476.

1ec5 (approved):

src/layer/transportation_label.js
jleedev: I suspect we already are:

image

jleedev: Or at least, this layer is not filtering any class out of `transportation`. Unclear if there's a bug:

image

PR #525, opened 2022-07-30
Combine shield layers and sort rendering

### LINKED ISSUE #524: Prioritize shields by road classification
The north–south freeway seen here in Northern Kentucky is Interstate 75, but you’d be forgiven for thinking it’s Kentucky Route 1292 or 2366. Both those routes run very close to I-75 along frontage roads, their shields colliding out the I-75 shields. It would be nice if there were some way of prioritizing the shields of higher-classed roads over those of lower-classed roads in the event of a symbol collision. This is not to say that the state routes should’ve necessarily gone unlabeled at this zoom level, just that that shields should’ve avoided the higher-classed road. If that would mean hiding all the shields for these lower-classed roads until a higher zoom level, so be it.

[Walton](https://zelonewolf.github.io/openstreetmap-americana/#12/38.85049/-84.61686)

claysmalley:
Fixes #524.

Before/after at http://localhost:1776/#13.01/29.78071/-95.56293

Screenshot from 2022-07-30 15-39-42
Screenshot from 2022-07-30 15-39-11

### DIFF
--- a/src/layer/highway_shield.js
+++ b/src/layer/highway_shield.js
@@ -1,7 +1,5 @@
 "use strict";
 
-import * as Util from "../js/util.js";
-
 function routeConcurrency(num) {
   return [
     "case",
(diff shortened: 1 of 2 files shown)


src/layer/highway_shield.js
```diff
@@ -44,38 +42,65 @@ let shieldLayout = {
...
+export const shield = {
   type: "symbol",
-  layout: shieldLayout,
   source: "openmaptiles",
-  metadata: {},
   "source-layer": "transportation_name",
-};
-
-function shieldLayer(hwyClass, minzoom) {
-  var layer = Util.cp(baseShield);
```
ZeLonewolf, line 58: Oops, looks like I wasn't using minzoom at all.  Which means it's not necessary :+1:

claysmalley: Oop. The `text-opacity` value implements what I think was intended here.

ZeLonewolf: Well, nobody was missing it.

src/layer/highway_shield.js
```diff
@@ -44,38 +42,65 @@ let shieldLayout = {
   "symbol-placement": "line",
   "text-max-angle": 180,
   "text-pitch-alignment": "viewport",
+  "symbol-sort-key": [
```
1ec5, line 45: Ah, so the fix was to combine the layers, the exact opposite of what I had supposed in #524. :sweat_smile: I think `symbol-sort-key` will also end up being handy for minimizing the number of layers needed for POI icons in the future.

src/layer/highway_shield.js
```diff
@@ -44,38 +42,65 @@ let shieldLayout = {
...
-export const minor = shieldLayer("minor", 14);
+  },
+  filter: [
+    "any",
+    ["has", "route_1"],
+    ["has", "route_2"],
+    ["has", "route_3"],
+    ["has", "route_4"],
+    ["has", "route_5"],
+    ["has", "route_6"],
```
1ec5, line 104: If a feature has `route_6`, would it be guaranteed to have `route_1` through `route_5`, making it unnecessary to check for them? (This is probably a minor point performance-wise, but it piqued my curiosity.)

1ec5 (approved):

PR #527, opened 2022-08-01
Upgrade legacy filters

jleedev:


src/layer/boundary.js
```diff
@@ -175,7 +183,11 @@ export const countryCasing = {
       ],
     },
   },
-  filter: ["all", ["in", "admin_level", 2], ["==", "maritime", 0]],
+  filter: [
+    "all",
+    ["in", ["get", "admin_level"], ["literal", [2]]],
```
1ec5, line 188:
```suggestion
    ["==", ["get", "admin_level"], 2],
```

src/layer/boundary.js
```diff
@@ -219,9 +231,9 @@ export const country = {
   },
   filter: [
     "all",
-    ["in", "admin_level", 2],
-    ["==", "disputed", 0],
-    ["==", "maritime", 0],
+    ["in", ["get", "admin_level"], ["literal", [2]]],
```
1ec5, line 234:
```suggestion
    ["==", ["get", "admin_level"], 2],
```

src/layer/place.js
```diff
@@ -373,7 +377,7 @@ export const continent = {
     "text-halo-color": "rgba(255,255,255,0.7)",
     "text-halo-width": 1,
   },
-  filter: ["all", ["==", "class", "continent"]],
+  filter: ["all", ["==", ["get", "class"], "continent"]],
```
1ec5, line 380:
```suggestion
  filter: ["==", ["get", "class"], "continent"],
```

src/layer/place.js
```diff
@@ -211,7 +211,7 @@ export const state = {
     "text-halo-width": 2,
     "text-halo-blur": 0.5,
   },
-  filter: ["all", ["==", "class", "state"]],
+  filter: ["all", ["==", ["get", "class"], "state"]],
```
1ec5, line 214:
```suggestion
  filter: ["==", ["get", "class"], "state"],
```

src/layer/rail.js
```diff
@@ -57,8 +57,12 @@ var defRail = {
 };
 
 var serviceSelector = ["match", ["get", "service"], ["siding", "spur", "yard"]];
-var isService = ["in", "service", "siding", "spur", "yard"];
-var isNotService = ["!in", "service", "siding", "spur", "yard"];
+var isService = [
+  "in",
+  ["get", "service"],
+  ["literal", ["siding", "spur", "yard"]],
```
1ec5, line 63: Use a `match` expression instead of searching for the value inside a literal array.

src/layer/road.js
```diff
@@ -709,8 +733,11 @@ class Service extends Road {
 
     this.constraints = [
       "all",
-      ["!=", "toll", 1],
-      ["!in", "service", "parking_aisle", "driveway"],
+      ["!=", ["get", "toll"], 1],
+      [
+        "!",
+        ["in", ["get", "service"], ["literal", ["parking_aisle", "driveway"]]],
```
1ec5, line 739: Use a `match` expression that evaluates to 0 if it’s one of the cases and 1 otherwise.

src/layer/transportation_label.js
```diff
@@ -127,7 +127,11 @@ export const bridgeSpacer = {
   type: "symbol",
```
1ec5, line 127: There’s also a `text-field` value above that relies on string interpolation instead of `concat` and `get` expressions. Converting to the latter will facilitate automated localization.

1ec5 (commented):
Thanks! All my feedback below is nitpicky, but hopefully it makes the code more readable with less nesting.

jleedev:
Is producing "Expected value to be of type number, but found null instead." Unclear if there's a straightforward way to debug that.

jleedev:
Doesn't like this one:

> `["all",["==",["get","brunnel"],"bridge"],["==",["get","class"],"minor"],["!=",["get","ramp"],1],["!=",["get","toll"],1],[">=",["get","layer"],5]]`

claysmalley:
`["get","toll"]` can be `null`. Try coalescing it to `0`.

jleedev:
Ah, https://github.com/maplibre/maplibre-gl-js/blob/main/src/style-spec/feature_filter/convert.ts

They also evaluate to `false` if an operand is of the wrong type, but it's reasonable to only handle null.

jleedev:
Actually, `aerodrome_label` doesn't seem to have `rank`, will remove.

jleedev:
Resolved all of these.

Test: Zoom all the way out and all the way in and watch for type errors.

1ec5 (approved):

Issue #451, opened 2022-06-25
New Zealand touring routes


1ec5:
New Zealand’s [_Traffic Control Devices Manual_](https://www.nzta.govt.nz/assets/resources/traffic-control-devices-manual/docs/part-2-draft-direction-service-and-general.pdf#page=21) provides for [named touring routes](https://www.nzta.govt.nz/resources/traffic-control-devices-manual/sign-specifications/view/964) marked by one-off shields.

[![named](https://user-images.githubusercontent.com/1231218/173923902-6cf29757-5e4d-4479-b17a-165690b38624.png)](https://www.nzta.govt.nz/resources/traffic-control-devices-manual/sign-specifications/view/964)

So far, two touring routes have been tagged [`network=NZ:Touring`](https://www.openstreetmap.org/relation/14143895) and [`network=NZ:Tourist`](https://www.openstreetmap.org/relation/14143156).

_Originally posted by @1ec5 in https://github.com/ZeLonewolf/openstreetmap-americana/issues/379#issuecomment-1156915732_

1ec5:
> two touring routes have been tagged [`network=NZ:Touring`](https://www.openstreetmap.org/relation/14143895) and [`network=NZ:Tourist`](https://www.openstreetmap.org/relation/14143156)

At least on directional signs, both of these routes are marked by the white-on-brown Twin Coast Discovery shield shown above. One of the relations is tagged `ref=24`, but I can’t find any evidence of “24” being posted in a manner useful for wayfinding.

[![TCD](https://user-images.githubusercontent.com/1231218/175793836-7dc3aaa5-56fc-48af-84db-ed432ddc73a0.png)](https://www.mapillary.com/app/?pKey=637446193914912)

[![24](https://user-images.githubusercontent.com/1231218/175793845-78f28aae-7f34-43b0-978e-66cc79241679.png)](https://www.mapillary.com/app/?pKey=178758087459194)

The Pacific Coast Highway is signposted as described in the TCD, but [its route relation](https://www.openstreetmap.org/relation/6014214) has no tags other than `name`.

[![PCH](https://user-images.githubusercontent.com/1231218/175793892-d8b740c5-6623-47f9-8303-3a2a0b66b3f5.png)](https://www.mapillary.com/app/?pKey=470936030828000)

[...]

PR #529, opened 2022-08-01
Add shield for Western Ring Route in New Zealand

k-yle:
This PR adds a shield for `network=NZ:WRR`, which is a one-of-a-kind route ([example 1](https://www.nzta.govt.nz/media-releases/new-logo-to-guide-motorists-on-western-ring-route/), [example 2](https://goo.gl/maps/4Thd4qVTCpHrtQmo6))

Related to #451


ZeLonewolf:
Thanks @k-yle for opening a PR.

Looking at this image:
image

I have questions.

Is the WRR route relation separate from the route 18 relation?  I would think that we would want a custom graphic for the route, much like we do for special turnpikes in the United States.  This PR declares a pill shield, which draws a circle or pill around the value in `ref`, and I'm not sure that's really the behavior we're looking for here?

k-yle:
yes, it's separate from SH 18, ([this is the OSM relation](https://osm.org/relation/13730708)). I saw that `US:IA` uses the same style, and [renders similarly](https://zelonewolf.github.io/openstreetmap-americana/#13.81/42.22584/-91.86094) to how the WR shields are signposted. 

I wasn't sure if a custom graphic was appropriate in this case (?), since it's a pretty much just a circle with text inside.

ZeLonewolf:
Oh, I see.  Black on white WR in a pill does seem like a workable approximation in that case.

src/js/shield_defs.js
```diff
@@ -3180,6 +3180,7 @@ export function loadShields(shieldImages) {
     },
   };
   shields["NZ:UR"] = homeDownShield;
+  shields["NZ:WRR"] = pillShield(Color.shields.white, Color.shields.black);
```
claysmalley, line 3183:
```suggestion
  shields["NZ:WRR"] = ovalShield(Color.shields.white, Color.shields.black, Color.shields.black, 20);
```

ZeLonewolf:
Code looks essentially correct, needs a tile update to test.

claysmalley (changes requested):
I think the plain text approach is reasonable, though it at least seems important that the shield is a circle rather than a pill or oval. If we use `ovalShield()` and specify a width of 20 it should make the shield circular.

claysmalley (approved):
Looks good!

Screenshot from 2022-08-16 17-58-41

ZeLonewolf:
Thanks @k-yle for your contribution!

PR #531, opened 2022-08-02
Crossover is a rail service

jleedev:
Fixes:

Screenshot_20220802-192413

### DIFF
--- a/src/layer/rail.js
+++ b/src/layer/rail.js
@@ -56,7 +56,11 @@ var defRail = {
   "source-layer": "transportation",
 };
 
-var serviceSelector = ["match", ["get", "service"], ["siding", "spur", "yard"]];
+var serviceSelector = [
+  "match",
+  ["get", "service"],
+  ["siding", "spur", "yard", "crossover"],
+];
 var isService = [...serviceSelector, true, false];
 var isNotService = [...serviceSelector, false, true];
 


claysmalley:
This rendering is intentional. `service=crossover` is not rendered as service track on other renderers, nor should it be.

The styling introduced in #476 does apply special styling to `service=crossover`, but only to disable the dash pattern, not to make the line thinner.

claysmalley:
To clarify, `service=crossover` should only connect a main track to a main track. If one of the connecting tracks is a service track, that is a tagging error—even though it is functionally a crossover, the value of `service=*` should match the adjacent tracks.

jleedev:
Well, it looks a little strange for heavy rail to not all be the same line width. What's the motivation for emphasizing crossover in this way?

Perhaps adding `line-cap: round` would smooth it a little bit.

jleedev:
Got it, so it should always — in principle — be between two parallel non-service rails that extend in both directions.

claysmalley:
I introduced a lot of these tagging errors myself a couple years ago. I went through and undid hundreds of them last week, but I didn't change anything that had been edited by someone else in the meantime. So if you come across any of these and you find me in the edit history, you know why 😬

jleedev:
I do see that osm-carto also renders crossover with this use case apparently in mind:

Screen Shot 2022-08-03 at 08 39 39

PR #532, opened 2022-08-03
Convert some road styling to expressions

claysmalley:
Before: 680 layers; after: 333 layers.

Road styling is still responsible for the majority of layers. But hopefully, in the future, we can further reduce this with more Maplibre expressions as support improves.


jleedev:
Appears to alter casing rendering:

Before: before
After: after
Diff: diff

claysmalley:
Casing width expression updated. Hope this fixes it.

Side note, how are you generating the diff images?

jleedev:
https://www.npmjs.com/package/pixelmatch

https://github.com/maplibre/maplibre-gl-js/ uses it for render tests

jleedev:
Casing is now consistent but the roads are still thicker:

Before: before
After: after
Diff: diff

ZeLonewolf:
Can someone assess whether there is a noticeable performance difference associated with the drop in layer count?

jleedev:
It's difficult to measure any time or memory difference on a high-end macbook. Will try the chromebook tonight.

Hypotheses:

- Consolidating the layers still produces the same amount of rendering pressure (adjusting uniform values and making draw calls), just the output is determined by conditionals rather than discrete layers.
- There is some GC pressure but it's hard to see on a machine with ample memory.

claysmalley:
Anecdotally, it takes about 30 seconds longer for my laptop fan to turn on. Unfortunately I don't have hard data on why.

As for the discrepancy in casing width, I had to insert an interpolation stop at z12 for motorway styling in order to fit it into the expression. I've updated it to be a little closer to the original metrics.

jleedev (approved):
Width change seems within reason to me.

Code is too long for github's little brain.

 2 files changed, 693 insertions(+), 1171 deletions(-)

Good enough.

PR #534, opened 2022-08-04
Add shields for Mendocino County, California

claysmalley:
Roads signposted by Mendocino County, California have a white-on-green square shield: https://wiki.openstreetmap.org/wiki/Talk:California/State_Highway_Relations#Mendocino_County

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -925,6 +925,10 @@ export function loadShields(shieldImages) {
   };
   shields["US:CA:Business"] = banneredShield(shields["US:CA"], ["BUS"]);
   shields["US:CA:CR"] = pentagonShieldBlueYellow;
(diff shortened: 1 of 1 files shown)


src/js/shield_defs.js
```diff
@@ -925,6 +925,10 @@ export function loadShields(shieldImages) {
   };
   shields["US:CA:Business"] = banneredShield(shields["US:CA"], ["BUS"]);
   shields["US:CA:CR"] = pentagonShieldBlueYellow;
+  shields["US:CA:Mendocino"] = roundedRectShield(
```
1ec5, line 928: [Changeset 124,460,057](https://www.openstreetmap.org/changeset/124460057) tags two route relations with this tag. I’m OK with going ahead and merging this PR, since we know what the shield will look like, and the tagging is already settled.

1ec5 (approved):

PR #535, opened 2022-08-05
Add shields for Branson, Missouri

### LINKED ISSUE #285: Branson, Missouri color-coded route shields
The city of Branson, Missouri has a color-coded route signage system similar to the belt system in Allegheny County, Pennsylvania. Only blue, red and yellow are used. All are tagged with `network=US:MO:Taney:Branson`, and the Blue Route for example is tagged with `ref=Blue Route`, `colour=blue`.

claysmalley:
Fixes #285.

[Screenshot from 2022-08-04 20-04-50](http://localhost:1776/#13.32/36.64567/-93.28604)


CONTRIBUTING.md
```diff
@@ -285,6 +285,7 @@ This style strives to draw representative highway shields wherever they are tagg
...
 - Shields for route networks where each individual route is identified by a color, rather than a number or letter. Such cases include:
   - **Allegheny County, PA Belt Routes**. Shields for this system use colors, with a colored circle and the words "<COLOR> BELT". These shields are drawn as squares with colored circles, with the `ref` values correctly corresponding to the text on the shield. Because of the common design (white shield with colored circle), these shields are properly part of a common route network. Special code is needed to convert the textual ref values to the colors displayed in the shield.
+  - **Branson, MO color-coded routes**. Shields for this system use colors, with a colored rectangle and the words "<COLOR> ROUTE". These shields are drawn as squares with colored rectangles, with the `ref` values correctly corresponding to the text on the shield. Because of the common design (green shield with colored rectangle), these shields are properly part of a common route network. Special code is needed to convert the textual ref values to the colors displayed in the shield.
```
1ec5, line 288: It sure would be nice if the vector tiles could pass along the `colour` tag of a route relation. This will become much more important when we tackle recreational routes (#216) and public transportation routes. In the meantime, it is indeed common for color-coded routes to put this information in `ref` in quasi-human-readable format.

src/js/shield_canvas_draw.js
```diff
@@ -35,6 +35,35 @@ export function paBelt(fillColor, strokeColor) {
   return ctx;
 }
 
+// Special case for Branson color-coded routes, documented in CONTRIBUTE.md
+export function bransonRoute(fillColor, strokeColor) {
+  var ctx = roundedRectangle(
+    Color.shields.green,
```
1ec5, line 41: I suppose the shield is analogous to the Allegheny County belts or the Chicago Skyway. The green part of the sign may not add much to the shield’s recognizability, but it keeps the shield anchored, so the color swatch doesn’t feel like it’s floating by itself. Have you tried out how the shield would look with balanced top and bottom padding?

claysmalley: It's a judgment call, though I prefer the bottom-aligned version.

Screenshot from 2022-08-05 12-25-32

ZeLonewolf:
I prefer the variant that looks like the real sign's alignment.

1ec5 (approved):
Thanks for humoring me; I agree that the original alignment looks better.

PR #536, opened 2022-08-11
Add Kansas Turnpike shield

### LINKED ISSUE #290: Kansas Turnpike shield
The Kansas Turnpike (currently tagged `network=US:KS:Turnpike`, no `ref=*`) has a similar shield to Kansas state highways. Differences include:

* The flower has 20 petals instead of 16
* The background is blue instead of black
* In place of the route number is a white-on-blue logo of the Kansas Turnpike Authority, with blue text "KANSAS TURNPIKE" encircling it in the yellow part

The shield is probably recognizable without the text "KANSAS TURNPIKE". Since the extra 4 petals are not a noticeable detail, the geometry can be copied from the existing Kansas shield. If necessary, the KTA logo could be enlarged slightly, or the K and A could be removed to leave the T.

Kansas_Turnpike

K-4 (`network=US:KS`, `ref=4`) for reference:
K-4

claysmalley:
Fixes #290.

Screenshot from 2022-08-11 09-06-49 Screenshot from 2022-08-11 09-06-36


michaelblyons:
What do you think about making the center circle bigger?

claysmalley:
It's already 1.25x the size of the the original. I tried a few different sizes and this is what I landed on - any larger and it looks awkward.

jleedev:
I agree; it works well enough on New York Thruway and Indiana Toll Road as those are largely one color and it just expands the detail area. This PR retains enough of the blue/yellow contrast, see https://upload.wikimedia.org/wikipedia/commons/4/41/Kta_182.jpg

I'd almost suggest that the Garden State Parkway have the yellow word PARKWAY restored for the same reason. https://upload.wikimedia.org/wikipedia/commons/d/de/2020-07-16_13_57_23_View_south_along_New_Jersey_State_Route_444_%28Garden_State_Parkway%29_at_Exit_38A_%28Atlantic_City_Expressway_EAST%2C_Atlantic_City%29_in_Egg_Harbor_Township%2C_Atlantic_County%2C_New_Jersey.jpg

1ec5 (approved):
Balancing the colors is pretty important for recognition. The downside is that the “KTA” is less readable at this size. The only other option I can think of would be to nix the blue disc in favor of a blue outline around the letters “KTA”, but I’m not sure that would actually be an improvement. This shield is sufficiently different than any other Kansas highway shield that I think legibility won’t be an issue in practice.

PR #537, opened 2022-08-13
Add missing dependency

jleedev:


package.json
```diff
@@ -33,6 +33,7 @@
     "esbuild": "^0.14.43",
     "maplibre-gl": "^2.1.9",
     "openmapsamples-maplibre": "github:adamfranco/OpenMapSamples-MapLibre",
+    "openmapsamples": "github:adamfranco/OpenMapSamples",
```
1ec5, line 36: Does NPM not see [this dependency](https://github.com/adamfranco/OpenMapSamples-MapLibre/blob/d0bb7b897c95bd0ef6adcd4e3a2ad409b9266c27/package.json#L27) because OpenMapSamples-MapLibre is only on GitHub as opposed to npm.org?

jleedev: Nope,  yarn and npm don't see this dependency because they're trying to move away from a world where you throw all your transitive dependencies into node_modules and see what happens.

1ec5: Makes sense, thanks.

1ec5 (approved):

PR #542, opened 2022-08-22
Add shield for Lane County, Oregon

arch0345:
Lane County has at least 2 signed county routes which use the generic blue/yellow pentagon shield. I've recently added relations for these:
https://www.openstreetmap.org/relation/14484195
https://www.openstreetmap.org/relation/14484209


claysmalley (approved):
Looks good. Thank you!

PR #543, opened 2022-08-23
Shield test gallery

### LINKED ISSUE #86: Wide shields showing up for 1 & 2 digit shields at DPR 1
DPR2

DPR1

zekefarwell:
In order to more effeciently test out changes to shield rendering, this PR adds a bulk test gallery for shields at http://localhost:1776/shieldtest.html.  I extracted out the Overpass font and shield width logic adjustments to a separate PR (#544).  This way we can easily compare a shield test gallery with current text rendering vs any potential changes.

#### Shield test gallery in Mac Chrome, Safari, and Firefox
sans-serif-mac-chrome-safari-firefox


src/shieldtest.js
```diff
@@ -0,0 +1,233 @@
...
+
+const once = (emitter, name, { signal } = {}) =>
+  new Promise((resolve, reject) => {
+    emitter.once(name, resolve);
+    signal?.addEventListener("abort", reject);
+  });
+
+await once(map, "load");
+
+let networks = [
```
ZeLonewolf, line 16: Would it make sense to load these programmatically from the array that we generate in shield_defs.js?

src/shieldtest.js
```diff
@@ -0,0 +1,233 @@
...
+  "US:WI:Spur",
+  "CA:ON:Waterloo:Wilmot",
+  "CA:ON:Brant:Highway",
+  "CA:ON:Muskoka:West",
+];
+
+// Uncomment for a list of all supported networks.  This makes for a very long page.
+//networks = Object.keys(shields);
+
+let refs = [
```
ZeLonewolf, line 191: Should we choose ref values that represent the narrowest and widest possible characters?  So all `1`s for narrowest.  5 is a good choice for widest because it hits the corners.  Though refs with a W are probably the worst case.

src/shieldtest.js
zekefarwell: That was my initial plan and is what this line does:
https://github.com/ZeLonewolf/openstreetmap-americana/blob/903323c53e44796421fdf602d20d18dde05192d0/src/shieldtest.js#L188-L189

I replaced it with a custom list because it ends up including a lot of duplicate graphic since many networks values use the same shield:
Screen Shot 2022-08-22 at 11 15 05 PM

I left the commented out line there so we could toggle it back on when you want to see _all_ the networks, not just all the shields.

src/shieldtest.js
zekefarwell: That's what I attempted to do, but I may not have hit the mark exactly.  I'm imagining this purely as a developer tool for testing changes so this array could be modified at any time to test with different numbers.

src/shieldtest.js
zekefarwell: The full list also can take quite a long time to load and it crashed my browser tab at times.

ZeLonewolf: Ah yes, counties are dumb.

src/shieldtest.js
```diff
@@ -0,0 +1,233 @@
...
+  "US:GA:Connector",
+  "US:LA:Business",
+  "US:MD:Business",
+  "US:VT:Alternate",
+  "US:MO:Spur",
+  "US:NC:Bypass",
+  "US:ND:Alternate",
+  "US:NE:Link",
+  "US:NH:Bypass",
+  "US:OH:Business",
```
claysmalley, line 175: A bannered Ashland County shape.
```suggestion
  "US:OH:Business",
  "US:OH:ASD:TWP",
```

src/shieldtest.js
```diff
@@ -0,0 +1,233 @@
...
+  "US:OK:Spur",
+  "US:OR:Business",
+  "US:PA:Truck",
+  "US:SC:Business",
+  "US:TX:Park",
```
claysmalley, line 180: A bannered Texas shape.
```suggestion
  "US:TX:Park",
  "US:TX:FM:Business",
```

src/shieldtest.js
```diff
@@ -0,0 +1,233 @@
...
+  "US:WA:Spur",
+  "US:WI:Spur",
```
claysmalley, line 182: A bannered pill.
```suggestion
  "US:WI:Spur",
  "CA:ON:private_toll",
```

src/shieldtest.js
```diff
@@ -0,0 +1,233 @@
...
+  "US:MN:Aitkin:CSAH",
+  "US:TX:Harris:HCTRA",
+
+  "JP:prefectural",
+
+  "my:federal",
+  "TR:motorway",
+  "US:NY",
+  "US:NY:Parkway",
+  "NL:S:Amsterdam",
```
claysmalley, line 62: Non dimenticare gli ottagoni!
```suggestion
  "NL:S:Amsterdam",
  "IT:A-road",
```

src/shieldtest.js
```diff
@@ -0,0 +1,233 @@
...
+  "11",
+  "81",
+  "69",
+  "95",
+  "111",
+  "281",
+  "980",
+  "H201",
+  "480N",
+  "A 562",
```
claysmalley, line 203: Here are some extant 6-character refs, just for fun.
```suggestion
  "A 562",
  "1138-2",
  "A26/A7",
```

claysmalley (commented):
I have a few more examples to add, just to be thorough.

src/shieldtest.js
```diff
@@ -0,0 +1,233 @@
...
+  "CA:ON:secondary",
+  "US:NE",
+  "US:MN:Aitkin:Park_Access",
+  "US:PA",
+  "US:PA:Turnpike",
+
+  "CA:AB:primary",
+  "CA:SK:tertiary",
+  "CA:SK:primary",
+  "PH:N",
```
claysmalley, line 44: Identical to `CA:AB:primary`.
```suggestion
```

src/shieldtest.js
```diff
@@ -0,0 +1,233 @@
...
+
+let networks = [
+  "default",
+
+  // Circles, pills, ovals
+  "US:DE",
+  "KR:national",
+  "US:IA",
+  "US:NM",
+  "US:KS",
```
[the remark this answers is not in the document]
zekefarwell: Sounds good.  I'm tried to get all the unique shape/color variants in the list, but I'm not surprised I missed some.

zekefarwell:
This PR is now only the addition of the shield test gallery with the font and text rendering logic changes extracted to #544.  I think it should be good to merge this piece in unless anyone else requests changes.

claysmalley (approved):

ZeLonewolf (approved):
I just tested and it looks good to me.  As in #549, we should add documentation of this very cool debug feature, either in the contributor's guide or (better) integrate it into the existing samples button.

However, neither of these suggestions should hold up adding this capability.

zekefarwell:
> we should add documentation of this very cool debug feature, either in the contributor's guide or (better) integrate it into the existing samples button.

I agree.  I've just added some documentation to the end of CONTRIBUTING.md.  

I looked into adding a link into the samples control, but it seemed like I'd need to do something hacky to make that work since the control is meant for OpenMapSamples and this is a separate thing.  I originally considered making this gallery as an OpenMapSample, but opted to go with this simpler option since the shields look exactly the same at all zoom levels and (at least for now) the shield rendering is very specific to this project only.

PR #544, opened 2022-08-27
Shield text rendering: width adjustments and Overpass font

### LINKED ISSUE #86: Wide shields showing up for 1 & 2 digit shields at DPR 1

zekefarwell:
The purpose of this draft PR is to try out some changes to shield rendering aimed at improving consistency in text rendering across shield shapes and across browsers.

The first change is to select the shield blank (when there are multiple widths) by the number of characters in the ref.  The first (narrowest) shield blank is for up to 2 characters and each size larger accomodates an additional character.  Currently the shield blank is chosen based on a measurement of the text width.  This leads to two digit refs ending up on wide shields and three digit refs ending up on narrow shields in a variety of situations.  Changing this makes the shield width selection behavior more predictable regardless of font or browser text rendering variations.  This should close #86.

The second change is to set the `maxwidth` parameter on the [canvas fillText() method](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/fillText) that draws the text on the shield blank.  This squeezes any text that is too wide for the space horizontally, without shrinking the font size (height).  This simulates the effect of a condensed font variant even when one is not available.  Currently text that is too wide has its font size shrunk so that it will fit in the space.  This leads to significant variation in character size across different shield shapes and widths.  Changing this behavior to just condense the text horizontally makes the text appear more consistent between short and long refs.

The third change is to add [Overpass](https://fonts.google.com/specimen/Overpass) as a custom font for those sweet Highway Gothic vibes.  I pulled in medium weight (500) as bold (700) felt too heavy, but there's also semi-bold (600) if medium feels too light.  Overpass doesn't have a condensed variant, but I'm hoping the simulated condensed effect described above will do the job nearly as well.  Let me know what you think.  

#### Caveats
A bunch of padding adjustments would definitely be needed with these changes.  I've only done that for the Interstate and US shields so far.  Wider banner text is also cut off in this draft and that would need to be fixt as well.  I'd like to see what folks think of this general approach before going further, though.

#### Current - Windows Chrome, Firefox
sans-serif-windows-chrome-firefox

#### Current - Mac Chrome, Safari, Firefox
sans-serif-mac-chrome-safari-firefox

#### This PR - Windows Chrome, Firefox
overpass-windows-chrome-firefox
#### This PR - Mac Chrome, Safari, Firefox
overpass-mac-chrome-safari-firefox


zekefarwell:
I guess I'm assuming that most 3 digit real world signs are using the wider variant even in the case of narrow numerals like 111.  That probably not true in reality as highway departments aren't terribly consistent.  At any rate, the extra padding looks fine to my eyes and the very simple logic of counting characters makes the code easy to reason about.

If we were to change this so 111 gets a narrower shield I'd suggest we make it an exception.  I.e. If ref length <= 2 or ref = 111.  Measuring text width seems very hard to get right when dealing with cross browser compatibility and font fallbacks.

1ec5:
> I pulled in medium weight (500) as bold (700) felt too heavy, but there's also semi-bold (600) if medium feels too light.

At first I thought it would be nice to ease up on the font weight as in this PR. It’s more faithful to the signs. But something about the new appearance didn’t quite sit well with me, and I think it comes down to the weight. Our shields have bold shapes, strokes, and colors, so at this small size, the medium weight seems too understated by comparison. The condensed effect, while necessary for other reasons, contributes to this understatedness. By contrast, the previous font was halfway to Impact in terms of heft.

> Overpass doesn't have a condensed variant, but I'm hoping the simulated condensed effect described above will do the job nearly as well.

So far it passes for the same thing. However, I’d like to see what this would do to a slightly heavier weight. It might leave some horizontal strokes heavier than vertical strokes, particularly if the browser needs to substitute characters from other fonts in non-Latin regions.

zekefarwell:
> At first I thought it would be nice to ease up on the font weight as in this PR. It’s more faithful to the signs. But something about the new appearance didn’t quite sit well with me, and I think it comes down to the weight. Our shields have bold shapes, strokes, and colors, so at this small size, the medium weight seems too understated by comparison. The condensed effect, while necessary for other reasons, contributes to this understatedness. By contrast, the previous font was halfway to Impact in terms of heft.

I find the current bold font weight too heavy for the line weight used in our shield graphics, but it's a good point that perhaps Overpass Medium 500 goes a bit too far in the other direction when combined with the condensed effect.  Regular 400 (in our current font stack or Overpass) would certainly come out looking too thin in comparison to shield outlines, but the slightly bolder Medium 500 weight seems to be a very close match.  You are right that as the characters are artificially condensed the line weight becomes thinner though.  SemiBold 600 weight is also available and may work better.  If the full bold (halfway to Impact) look is preferred, then I would suggest we make the shield line work thicker to match.  I think this would end up looking too cramped at a such a small icons size though.

ZeLonewolf:
With the merger of #699, these changes will need to get worked into the shield library.

ZeLonewolf:
With #965 and #1010, I believe this PR is no longer needed.

zekefarwell:
This exploratory branch is no longer needed with the merger of the above linked PRs

Issue #540, opened 2022-08-18
Map routes of Tennessee


claysmalley:
Relation coverage of Tennessee state routes is largely complete, but the majority of routes follow the outdated tagging scheme `network=US:TN`. These route relations need to be split and retagged to `US:TN:primary` or `US:TN:secondary`, according to their signage.

This [Overpass query](https://overpass-turbo.eu/s/1lqF) highlights `route=road` relations in Tennessee, color-coded by the value of `network=*`:
* gray: missing `network=*`
* red: `network=US:TN`
* blue: `network=US:TN:primary` or subnetwork
* green: `network=US:TN:secondary` or subnetwork
* purple: `network=US:TN:McMinn` (county roads)

This [wiki page](https://wiki.openstreetmap.org/wiki/Tennessee/State_Route_Relations) tracks the mapping of the individual relations.

michaelblyons:
> highlights `route=road` relations in Puerto Rico, 

You mean Tennessee?

claysmalley:
fixed

1ec5:
By the way, I’m still using `network=US:TN` for route superrelations that incorporate a mix of primary and secondary route segments. Would that be a problem? It should be possible to exclude superrelations from the query by recursing down and subtracting the properly tagged members from the original set.

claysmalley:
> It should be possible to exclude superrelations from the query by recursing down and subtracting the properly tagged members from the original set.

I'm gonna need you to translate that to Overpass QL for me...

jleedev:
Looks like you're already ignoring superrelations here:

```
{{geocodeArea:Tennessee}}->.searchArea;

(
  relation[route=road][ref][network~"^US:TN"](area.searchArea);
  relation[route=road][ref][!network](area.searchArea);
);

out body;
>;
out skel qt;
```

1ec5:
Oh, I just misunderstood the purpose of the query, without being able to really tell on my phone at the time. It highlights all the routes, so including the `route=road` `network=US:TN` superrelations isn’t a problem because they get covered up by the blue and green `route=road` `network=US:TN:primary`/`secondary` relations anyways.

claysmalley:
Mapping is finished, until TNDOT opens a new highway.

PR #545, opened 2022-08-28
Add bannered Tennessee route shields

claysmalley:
Adds two bannered route shields found during mapping of Tennessee routes (#540).

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -2162,10 +2162,17 @@ export function loadShields(shieldImages) {
   shields["US:TN:primary:Bypass"] = banneredShield(shields["US:TN:primary"], [
     "BYP",
   ]);
+  shields["US:TN:primary:Truck"] = banneredShield(shields["US:TN:primary"], [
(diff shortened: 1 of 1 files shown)


src/js/shield_defs.js
```diff
@@ -2162,11 +2162,18 @@ export function loadShields(shieldImages) {
...
+    "TRK",
+  ]);
   shields["US:TN:secondary"] = triangleRoundedDownShield;
   shields["US:TN:secondary:Alternate"] = banneredShield(
     shields["US:TN:secondary"],
     ["ALT"]
   );
+  shields["US:TN:secondary:Scenic"] = banneredShield(
+    shields["US:TN:secondary"],
+    ["SCEN"]
```
1ec5, line 2175: Hopefully this is intuitive enough. Tennessee secondary state routes aren’t the only kinds of routes that can be bannered “SCENIC”, but it is a fairly rare banner.

1ec5 (approved):

PR #548, opened 2022-09-01
Add Truck Bypass US Route shields

claysmalley:
Adds a bannered route network unique to Jamestown, North Dakota.
* [US 52 Truck Bypass](https://www.openstreetmap.org/relation/1944296)
* [US 281 Truck Bypass](https://www.openstreetmap.org/relation/1944294)

Screenshots unavailable because I just retagged these routes.


ZeLonewolf (approved):
Not sure why we need all the extra whitespace but seems fine to me.

PR #549, opened 2022-09-03
Add debug variable to specify shield text halo color

claysmalley:
Adds a debug variable that overrides the text halo color on shields. This can be helpful to ensure text doesn't collide with other design features when determining padding values for shield definitions.

```
const SHIELD_TEXT_HALO_COLOR = null;
```
Screenshot from 2022-09-03 18-11-58
```
const SHIELD_TEXT_HALO_COLOR = "magenta";
```
Screenshot from 2022-09-03 18-12-14

### DIFF
--- a/src/configs/config.aws.js
+++ b/src/configs/config.aws.js
@@ -5,9 +5,14 @@ Planetiler tile server, hosted at AWS
 */
 const OPENMAPTILES_URL =
   "https://6ug7hetxl9.execute-api.us-east-2.amazonaws.com/data/v3.json";
-const SHIELD_TEXT_HALO_COLOR = null;
+
+/*
+Uncomment this variable to override the shield text halo color.  Useful while testing shield design changes.
+Accepts an HTML color name, hex code, or other CSS color value.
+*/
+//const SHIELD_TEXT_HALO_COLOR_OVERRIDE = "magenta";
 
 export default {
   OPENMAPTILES_URL,
-  SHIELD_TEXT_HALO_COLOR,
+  //SHIELD_TEXT_HALO_COLOR_OVERRIDE,
 };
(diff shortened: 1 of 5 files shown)


ZeLonewolf (changes requested):
Looks good, and tested satisfactorily.

Can we add something in the shield contributor's guide that describes the use of this variable?

claysmalley:
Added a blurb in the documentation.

src/configs/config.aws.js
```diff
@@ -5,7 +5,9 @@ Planetiler tile server, hosted at AWS
 */
 const OPENMAPTILES_URL =
   "https://6ug7hetxl9.execute-api.us-east-2.amazonaws.com/data/v3.json";
+const SHIELD_TEXT_HALO_COLOR = null;
```
zekefarwell, line 8: I'd suggest we briefly document the purpose and usage of the variable with a comment.  Prefilling with a useful value and then commenting it out also might be nice.  Then you just have to un-comment to turn this debug feature on.

```suggestion

// Uncomment this variable to override the shield text halo color.  Useful while testing shield design changes.
// Accepts an HTML color name, hex code, or other CSS color value.
//const SHIELD_TEXT_HALO_COLOR = "magenta";
```

I'd also suggest the variable be named a bit more descriptively so it's clear this is a debug/override/test feature, not something that gets used in production.  `SHIELD_TEXT_HALO_COLOR_DEBUG_OVERRIDE` is rather long long.  Maybe just `SHIELD_TEXT_HALO_COLOR_OVERRIDE`?

zekefarwell (approved):
Oh right, the variable has to be commented/uncommented in the export block as well.  
https://github.com/ZeLonewolf/openstreetmap-americana/blob/239d9f0003752e645c8208f4156e2657a579dc3f/src/configs/config.aws.js#L9-L18
I guesshaving it commented out isn't as clean as I was imagining.  Maybe it was better having it just default to null.   Either way , this all looks good to me.

ZeLonewolf (approved):
Looks good, thanks for the writeup!

PR #550, opened 2022-09-05
Minor adjustments to various shields for design consistency

claysmalley:
The shield test gallery added in #543 exposed a lot of nitpicks with certain shields. Changes in this PR include:

* a fixed width assigned to the oval shields of `CA:AB:secondary` and `CA:MB:PR`
* a new `circleShield()` function as a special case of `ovalShield()`, to which `VE:L:*` and `NZ:WRR` are assigned
* widths of fixed-width rectangular shields adjusted to match the max width otherwise defined (`34`)
* reduced horizontal padding for variable-width rectangular and pill shields
* an unused argument (`outlineWidth`) removed from shield definition functions
* another unused argument (`radius`) moved to the end of the `roundedRectShield()` function arguments to keep undefined as a default, but reserve future use
* documentation for the shield definition functions
* a replaced 3-digit octagon shield (`IT:A-road`), enlarged to split the difference between the 2- and 4-digit shields

before:
Screenshot 2022-09-05 at 14-55-42 Shield Test

after:
Screenshot 2022-09-05 at 14-56-22 Shield Test

diff:
out

before:
Screenshot from 2022-09-05 15-06-42

after:
Screenshot from 2022-09-05 15-06-07

diff:
out2


zekefarwell (approved):
Looks good to me.  I definitely appreciate added docblocks explaining what the various shield function parameters do.  That's very helpful!

PR #553, opened 2022-09-24
Add Australia shields

### LINKED ISSUE #397: Shields of Australia
[Routes in Australia](https://en.wikipedia.org/wiki/Highways_in_Australia#Route_numbering_systems) were originally given a hierarchical numbering scheme with different shields for National Highways, National Routes and State Routes, though there has been an organized effort to nationally standardize on an alphanumeric scheme and replace signage with rectangular shields. Notably, Western Australia (the largest state by area, taking up one third of the continent) has not adopted this scheme and retains the old hierarchical system. There are a few odds and ends, like an Alternate National Route and a color-swapped alphanumeric rectangle for toll roads, but the following shields should be enough for basic support.

Most numbered routes in Australia are covered by `route=road` relations, but `network=*` tagging is not country-prefixed. There are a number of `network=*` values in Australia, including `MR`, `NH`, `NR`, `S` and `T`. Documentation is limited on this tagging scheme. A decision would need to be made on whether this tagging scheme is sufficient, and whether it is appropriate mapper feedback to support it.

New alphanumeric shield (nationwide except for Western Australia):
AUS_Alphanumeric_Route_B83

National Highway shield (Western Australia, Tasmania):
Australian_national_highway_1

National Route shield (Western Australia, Queensland):
Australian_national_route_13

State Route shield (Western Australia, Queensland, Victoria):
Australian_state_route_66

Metroad shield (Queensland):
AUS_Metroad_5

Tourist Drive shield (most states):
Australian_Tourist_Drive_357

claysmalley:
Fixes #397.

The shape used by Aus/NZ State Routes is renamed to "fishhead".

Tagging transition work is currently under way. We'll need to wait for that to be finished before a tile update.


src/js/shield_defs.js
```diff
@@ -3272,18 +3314,35 @@ export function loadShields(shieldImages) {
...
-    },
+  // Australia
+  ["ACT", "NSW", "NT", "QLD", "SA", "TAS", "VIC", "WA"].forEach(
+    (state_or_territory) =>
+      ([
+        shields[`AU:${state_or_territory}`],
+        shields[`AU:${state_or_territory}:NH`],
+        shields[`AU:${state_or_territory}:NR`],
+        shields[`AU:${state_or_territory}:S`],
+        shields[`AU:${state_or_territory}:T`],
```
1ec5, line 3325: Some tourist drives in Queensland are marked by [glyphs](https://www.tmr.qld.gov.au/-/media/busind/techstdpubs/Traffic-management/TC-signs/zipped/q-series.pdf#page=62) rather than numbers:

[](https://commons.wikimedia.org/wiki/File:Great_Barrier_Reef_Tourist_Drive_road_sign.jpg)

Some of these relations have been tagged with keyword-like `ref` values, such as [`ref=dugong`](https://www.openstreetmap.org/relation/13493955), based on [a table](https://wiki.openstreetmap.org/wiki/Custom_Highway_Shields#Australia) that’s been on the wiki for a long time. I guess we should exclude them for now, and when we’re able to design copyright-unencumbered artwork for them, we can add a manual mapping from `ref`s or `name`s to images.

claysmalley: Seems like a similar situation to the [Touring Routes of New Zealand,](https://github.com/ZeLonewolf/openstreetmap-americana/issues/451#issuecomment-1169355237) each of which has unique artwork. The solution in that case was to adopt guidelines that tag each route with a unique value of `network` matching `NZ:Touring:*` so that each one can be assigned a different icon. Emojis were chosen as `ref` values as a fallback for less sophisticated renderers, though this wasn't strictly necessary.

@DianAgesson would Queensland mappers be opposed to assigning unique `network` values for their non-numeric Tourist Drives? Even though they may all belong to one "network" in reality, the shield appearance is different enough that using a generic `network` value would produce misleading renderings.

DianAgesson: Oh no, there are always edge cases that haven’t been accounted for aren’t there! What’s wrong with using numbers?! :)

I’ll reach out to the active mappers on discord/mailing list. I should be able to update the tagging guidelines for these shields accordingly.

DianAgesson: Hello,

Just updating you with our progress.

The tagging guidelines have been adjusted for these edge cases. I can see you’ve already updated for the State Strategic Touring Routes, and if there are other cases that come up I’ll pass it on.

Bulk editing has also kicked off in earnest. Lots of contributors have been chipping away in places as well; WA is completely converted, and NT and TAS have transitioned their relations.

Dian

DianAgesson: A large majority of the route relations have now been converted. Two previously undocumented route types where identified.

Overdimensional routes (eg AU:VIC:OD), and freight routes (eg AU:QLD:Freight). Neither type of route requires a shield to be rendered.

It'll take more time to progressively update the highway road segments, but my understanding is that the shield rendering uses the relation anyway?

1ec5: > It'll take more time to progressively update the highway road segments, but my understanding is that the shield rendering uses the relation anyway?

Correct, this style doesn’t consider way refs at all. This means most of the road network is ready for the vector tiles to be regenerated.

ZeLonewolf: Sounds good, we're overdue for a tile update anyways.

1ec5:
[Perth, Western Australia](https://zelonewolf.github.io/openstreetmap-americana/#12/-31.9672/115.9021)
_Perth, Western Australia_

[Humpty Doo, Northern Territory](https://zelonewolf.github.io/openstreetmap-americana/#12/-12.57249/131.08376)
_Humpty Doo, Northern Territory_

[Adelaide, South Australia](https://zelonewolf.github.io/openstreetmap-americana/#12/-34.86488/138.5648)
_Adelaide, South Australia_

[Brisbane, Queensland](https://zelonewolf.github.io/openstreetmap-americana/#14/-27.47306/152.97747)
_Brisbane, Queensland_

[Warwick, Queensland](https://zelonewolf.github.io/openstreetmap-americana/#13/-28.20962/152.03073)
_Warwick, Queensland_

[Charleville, Queensland](https://zelonewolf.github.io/openstreetmap-americana/#8/-26.347/146.507)

[Newcastle, New South Wales](https://zelonewolf.github.io/openstreetmap-americana/#12/-32.90107/151.6024)
_Newcastle, New South Wales_

[Canberra, Australian Capital Territory](https://zelonewolf.github.io/openstreetmap-americana/#14/-35.3066/149.12572)
_Canberra, Australian Capital Territory_

[Melbourne, Victoria](https://zelonewolf.github.io/openstreetmap-americana/#13/-37.81925/144.89609)
_Melbourne, Victoria_

[Hobart, Tasmania](https://zelonewolf.github.io/openstreetmap-americana/#12/-42.89951/147.29409)
_Hobart, Tasmania_

1ec5 (approved):
It’s been a long time coming, but here we are!

PR #557, opened 2022-10-09
Iraq highway shields

1ec5:
Added a shield definition for Iraq (national) highways. Only highways have been mapped as relations so far; numbered freeways and roads don’t have relations yet. As in Iran (#463), route shields in Iraq bear both writing systems on separate panels, but this style is only showing the Western Arabic numerals that have been tagged in `ref`.

[highways](https://zelonewolf.github.io/openstreetmap-americana/#8/32.706/44.497)

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -2853,7 +2853,9 @@ export function loadShields(shieldImages) {
   // Iraq
   shields["IQ:national"] = roundedRectShield(
     Color.shields.green,
-    Color.shields.white
+    Color.shields.white,
+    Color.shields.white,
+    34
   );
 
   // Japan
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -2850,6 +2850,12 @@ export function loadShields(shieldImages) {
     Color.shields.white
   );
 
+  // Iraq
+  shields["IQ:national"] = roundedRectShield(
+    Color.shields.green,
+    Color.shields.white
+  );
+
   // Japan
   shields["JP:E"] = roundedRectShield(Color.shields.green, Color.shields.white);
   shields["JP:national"] = triangleConvexDownShieldBlue;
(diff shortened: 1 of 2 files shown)


src/js/shield_defs.js
```diff
@@ -2845,6 +2845,12 @@ export function loadShields(shieldImages) {
     Color.shields.white
   );
 
+  // Iraq
+  shields["IQ:national"] = roundedRectShield(
+    Color.shields.green,
+    Color.shields.white
```
claysmalley, line 2851:
```suggestion
  shields["IQ:national"] = roundedRectShield(
    Color.shields.green,
    Color.shields.white,
    Color.shields.white,
    34
```

The shield images on [Wikimedia Commons](https://commons.wikimedia.org/wiki/Category:Diagrams_of_route_signs_of_Iraq) appear to be fixed-width.

ZeLonewolf:
Are the wide rectangles a deliberate design decision based on OTG signage?

image

Also, I notice the "5" route in the northeast of the clip above, is that an artifact awaiting a tile update?

1ec5:
> Are the wide rectangles a deliberate design decision based on OTG signage?

This was @claysmalley’s suggestion in https://github.com/ZeLonewolf/openstreetmap-americana/pull/557#discussion_r991330355. You can just barely make out the green highway shields on the left side of this photo:

[Falluja](https://commons.wikimedia.org/wiki/File:الفلوجة.png)

> Also, I notice the "5" route in the northeast of the clip above, is that an artifact awaiting a tile update?

That’s Road 5. So far, [two roads](https://overpass-turbo.eu/s/1odJ) have been tagged `network=IQ-road`. Should we add support for this tag even though it uses a hyphen instead of a colon? As far as I know, roads have the same green shield as highways.

claysmalley:
In general, I'm not in favor of supporting network values like `XX-roads` where `XX` is a country code. I think these network values sprung from a misunderstanding of European road networks (`e-road`, `FR:A-road`, etc.) and I think there's a better country-and-colon-prefixed value to use in every case.

1ec5:
We can take care of roads and motorways in separate PRs once the tagging for those catches up to the tagging for highways.

claysmalley (approved):
Looks good to me.

PR #558, opened 2022-10-09
Cyprus motorway shields

1ec5:
Added a shield definition for A roads (motorways) in Cyprus, as well as Akrotiri and Dhekelia, a British Overseas Territory. So far only A roads have been mapped; B roads (main roads) and E roads (secondary roads) don’t have relations yet.

[larnaca](https://americanamap.org/#11/34.9456/33.6107)

[dhekalia](https://americanamap.org/#13/35.00977/33.71338)

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -4102,6 +4102,13 @@ export function loadShields() {
   );
 
   // Cyprus
+  shields["CY:motorway"] = hexagonVerticalShield(
+    2,
+    Color.shields.green,
+    Color.shields.yellow,
+    Color.shields.yellow,
+    0
+  );
   shields["CY:B"] =
     shields["CY:E"] =
     shields["CY:F"] =


src/js/shield_defs.js
```diff
@@ -3062,6 +3062,16 @@ export function loadShields(shieldImages) {
     Color.shields.white
   );
 
+  // Cyprus
+  shields["cy:national"] = {
```
claysmalley, line 3069: `national` is an unusual name for what seems to be a network of motorways. `cy:motorway` would be more appropriate, as would `cy:A-road` considering the `A` prefix.

It's unclear whether this tagging was invented by a Cypriot. Is there a Cyprus OSM community that we can contact about this?

1ec5: [According to Wikipedia](https://en.wikipedia.org/wiki/Roads_and_motorways_in_Cyprus), “motorway” fits the A-road network pretty well, but it does sound like B-roads and E-roads could also be part of a national numbering scheme. Maybe this was an attempt to conform to what was perceived as international terminology (not the first time I’ve seen it). The tag was introduced in [changeset 44730739](https://www.openstreetmap.org/changeset/44730739).

The community index doesn’t list any local community for Cyprus, and [the national tagging page](https://wiki.openstreetmap.org/wiki/Cyprus#Highways_in_Cyprus) doesn’t provide any guidance on route relations.

claysmalley:
Don't forget to add a line to `src/shieldtest.js`, since this introduces new shield background icons.

src/js/shield_defs.js
claysmalley: I'm hesitant to solidify a tagging scheme that may complicate tagging guidelines for the remaining route networks in Cyprus. It's unclear how other networks in Cyprus should be tagged if we decide to support `cy:national`.

Cyprus is a small country, and the mapping community may only be a few people. Perhaps we could ask the Greek community for their perspective—certainly some of them would be familiar with Cyprus, and if there are any Cypriot mappers, there's a chance we may find them there.

ZeLonewolf:
How should we proceed on this?  Is it worth proposing a route network scheme for Cyprus and advertise it in the appropriate channels, and implement it if there's no objections?

claysmalley:
> How should we proceed on this? Is it worth proposing a route network scheme for Cyprus and advertise it in the appropriate channels, and implement it if there's no objections?

Sounds like a plan.

@1ec5 By the way, the hexagonal icons can be replaced with the new draw functions here.

ZeLonewolf:
What would be a sane scheme for A, B, and E roads?

mlundblad:
It looks like the networks in Cyprus have maybe been updated since this was last discussed

A1:
https://www.openstreetmap.org/relation/359862#map=11/34.9194/33.2148

Using network=CY:motorway

B3:
https://www.openstreetmap.org/relation/17608887#map=13/34.95129/33.75083
Using network=CY:B

The E roads seems to still be missing.

1ec5:
`network=CY:national` still has 14 occurrences while `network=CY:motorway` has nine. Some `CY:national` were changed to `CY:motorway` [a couple years ago](https://www.openstreetmap.org/changeset/155167580). We can support the less frequent tag, but only with some evidence that the local community accepts it.

#1421 tracks B/E/F roads.

1ec5:
Oh, it looks like `network=CY:national` is only used in Northern Cyprus, with the exception of [A9](https://www.openstreetmap.org/relation/359516), which is in Cyprus.

1ec5:
I’ve rewritten this PR to use the drawing functions and `network=CY:motorway` instead of `network=cy:national` (which is no longer in use when considering case). The rewritten shields are less steep than before. I think I was following the diagrams on Commons, but it turns out the actual shields on guide signs are boxier.

[A1 shields](https://commons.wikimedia.org/wiki/File:Latsia,_Cyprus_-_panoramio.jpg)

[A2, A3, A5 in Cyprus](https://preview.americanamap.org/pr/558/#map=11/34.9456/33.6107)

[A3 in Dhekelia](https://preview.americanamap.org/pr/558/#map=13/35.00432/33.72388)

claysmalley (approved):
Looks good!

PR #559, opened 2022-10-09
Kosovo motorway shields

1ec5:
Added a shield definition for the two completed motorways in Kosovo. Once [R 7.1](https://en.wikipedia.org/wiki/R_7.1_%28Kosovo%29) is completed, we’ll need a wider shield image to accommodate it.

[motorways](https://zelonewolf.github.io/openstreetmap-americana/#9/42.3553/20.9496)


claysmalley (approved):
Looks good to me.

PR #560, opened 2022-10-10
Require shield icon PRs to add lines to test gallery

claysmalley:
Since the shield test gallery's introduction (#543), it's grown a little as more shield icons have been added. This PR makes it explicit that new shield icons meant for text overlay should be added to the gallery when submitting a PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -163,6 +163,8 @@ boilerplate in `scripts/taginfo_template.json`.
 2. If you are introducing a novel approach to depicting a layer or feature
    property from the OpenMapTiles schema, document how the corresponding
    OpenStreetMap key or tag is used in `scripts/taginfo_template.json`.
+3. If any shield background icons are introduced, add lines to `src/shieldtest.js`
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -163,6 +163,8 @@ boilerplate in `scripts/taginfo_template.json`.
 2. If you are introducing a novel approach to depicting a layer or feature
    property from the OpenMapTiles schema, document how the corresponding
    OpenStreetMap key or tag is used in `scripts/taginfo_template.json`.
+3. If any shield background icons are introduced, add lines to `src/shieldtest.js`
+   to demonstrate overlaid text on each of them.
```
1ec5, line 167: Would it be possible to generate this list automatically, as with the status map?

1ec5 (approved):

ZeLonewolf (approved):
Sounds good to me.

CONTRIBUTING.md
claysmalley: Maybe, though it's more work than I'm willing to commit to right now.

PR #561, opened 2022-10-10
Add rendering for capital=3, same as capital=4

Blijbol:
Cities with capital=3 are state capitals that are at least as important as state capitals with capital=4, so these cities should receive at least equal rendering. [Based on this OpenMapTiles comment.](https://github.com/openmaptiles/openmaptiles/pull/1401#issuecomment-1155742393).

Example: island of Sint Maarten / Saint-Martin, with a capital=3 city on each side.

| Before | After |
|---|---|
| Firefox_Screenshot_2022-10-10T21-08-51 519Z | Firefox_Screenshot_2022-10-10T21-08-37 203Z |

### DIFF
--- a/scripts/taginfo_template.json
+++ b/scripts/taginfo_template.json
@@ -20,6 +20,14 @@
     },
     {
       "key": "capital",
+      "value": "3",
+      "object_types": ["node"],
+      "description": "Marks the city with a state capital icon.",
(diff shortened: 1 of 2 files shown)


1ec5 (approved):
Makes sense, and thank you for remembering to update the taginfo project file!

PR #562, opened 2022-10-11
Add Blue Ridge Parkway shield

claysmalley:
Adds a shield for the Blue Ridge Parkway.

The illustration is simplified and moved upward, truncating the tree. The text "BLUE RIDGE" is removed, and "PARKWAY" is replaced with "BRP".

Screenshot from 2022-10-10 19-49-57 Screenshot from 2022-10-10 19-50-11

Blue_Ridge_Parkway_shield


icons/shield_us_nps_brp.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="20" height="20" viewBox="0 0 5.292 5.292" xmlns="http://www.w3.org/2000/svg">
+ <rect style="fill:#fff;fill-rule:evenodd;stroke:#693f23;stroke-width:.26458299;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:none" width="5.027" height="5.027" x=".132" y=".132" rx=".397" ry=".397"/>
```
1ec5, line 2: We could introduce a tan/cream color to the standard color palette to avoid simplifying it to white. Caltrans uses a “cream” color `#ddcba4` on some standard signs. This color was chosen to fit in with the standard colors in the national MUTCD, which are the basis for this style’s color palette. It complements the brown color `#693f23` well, but it would reduce the “BRP” legend’s contrast.

ZeLonewolf: The cream color is awfully subtle.  I think the current PR offers a nice balance between minimizing the palette and still visually representing the original in a 100% understandable way.  I'm not convinced we need the cream in this use case, but perhaps a deeper dive of the mentioned Caltrans examples would reveal a need.

1ec5: I’m OK with the white in this case. Just bookmarking this cream option in case we come across other rustic-looking shields in the future and want to make them consistent with each other.

1ec5 (approved):
We haven’t gotten any negative feedback from anyone local to Virginia or North Carolina, so let’s go ahead and merge this design and revisit if there are any concerns. Looking forward to more tourist and scenic route shields!

PR #563, opened 2022-10-11
TagInfo: capital=yes is equivalent to capital=2

Blijbol:
See:
https://github.com/openmaptiles/openmaptiles/blob/master/layers/place/capital.sql

### DIFF
--- a/scripts/taginfo_template.json
+++ b/scripts/taginfo_template.json
@@ -12,6 +12,14 @@
   "tags": [
     {
       "key": "capital",
+      "value": "yes",
+      "object_types": ["node"],
+      "description": "Marks the city with a national capital icon.",
+      "doc_url": "https://openmaptiles.org/schema/#capital",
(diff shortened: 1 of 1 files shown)


ZeLonewolf:
Huh.  Curious.  Looks like `capital=2` is barely used in fact.

Is it permissible to set an array in the "values" block? @1ec5 ?

Blijbol:
> Is it permissible to set an array in the "values" block?

The [documentation](https://wiki.openstreetmap.org/wiki/Taginfo/Projects) does not mention such a possibility.

ZeLonewolf (approved):
Looks good to me in concept.  Another maintainer should sanity check before merging.

1ec5 (approved):
Unfortunately, [`value` is required to be a `string`](https://github.com/taginfo/taginfo-projects/blob/e577219b5a74b570146af0e84b9b9b74ddf58a21/taginfo-project-schema.json#L93). This looks good to me.

PR #576, opened 2022-11-26
49-Mile Scenic Drive shield

### LINKED ISSUE #570: 49-Mile Scenic Drive
The [49-Mile Scenic Drive](https://en.wikipedia.org/wiki/49-Mile_Scenic_Drive) is a slightly less than 49-mile-long circuit along local roads in San Francisco. Its main claim to fame is its iconic shield, which this style should render.

[49-Mile Scenic Drive](https://commons.wikimedia.org/wiki/File:San_Francisco_CA_49-Mile_Scenic_Drive.svg)

The route is tagged `network=US:CA:San_Francisco:49_Mile_Scenic_Drive` with no `ref`.

1ec5:
Added a shield for the 49-Mile Scenic Drive in San Francisco. This route includes segments of I-280 and U.S. 101, so it’s actually a bit more prominent on the map than its name would imply.

[Dogpatch](https://zelonewolf.github.io/openstreetmap-americana/#14/37.75954/-122.39202) [Golden Gate Park](https://zelonewolf.github.io/openstreetmap-americana/#15/37.76889/-122.4753)

This shield image is based on [a public domain graphic](https://commons.wikimedia.org/wiki/File:San_Francisco_CA_49-Mile_Scenic_Drive.svg) on Wikimedia Commons. It departs from this style’s standard blue color because the sky has to be a light blue to be recognizable. To ensure compatibility with the rest of the MUTCD-inspired color scheme, I reused the color of the blue jeans in [this standard California MUTCD sign](https://commons.wikimedia.org/wiki/File:MUTCD-CA_S32-1.svg). I think this color will work well in other shields that depict the sky or anything else that’s much lighter than the standard blue. At high zoom levels, it has just enough contrast from water to avoid confusion with nearby ponds. I also took the custom orange from the same sign, which is intended to complement the light blue, but if desired, I can switch it to the standard orange, which is slightly lighter.

The original shield features a the seagull and the number 49 very prominently. I chose to omit the 49 because it would force us to either reduce the seagull’s head to the point of being illegible or expand the shield horizontally, giving it an unwieldy shape. A white 49 would also contrast poorly with the light blue I’ve chosen, which is slightly lighter than what’s on the actual signs.

Fixes #570.


claysmalley:
I tried the orange from the existing color palette, and I agree that the original orange is better in this case.

I was worried about the light blue contrasting against the white, though maybe a black border looks too jarring:

Screenshot from 2022-11-26 22-27-52 Screenshot from 2022-11-26 22-27-27

1ec5:
The standard blue seems less jarring, and somehow less constraining too:

Dogpatch Golden Gate Park

claysmalley (approved):
That's a good compromise. Looks good to me.

PR #577, opened 2022-11-26
Convert shield definitions to JSON

### LINKED ISSUE #118: Extract loadShields objects into JSON

1ec5:
Added a build script that extracts the JavaScript shield definitions into a JSON file. This will pave the way for validating the definitions against a schema and filling gaps in alternative representations of the shield repertoire, such as this project’s taginfo project file. Eventually, the JSON data could be extracted into a separate package for easier reuse in other projects, regardless of their programming languages. In the meantime:

* [x] Add JSON-compiling script similar to taginfo.js
* [x] Convert color constants back to keywords
* [x] Convert shape functions to keywords
* [x] Convert common shape objects to keywords
* [ ] Convert text layouts to keywords
* [x] Sort definitions by `network` value
* [ ] Allow definitions to inherit properties from other definitions
* [ ] Split JSON file by country for easier maintenance
* [ ] Combine per-country JSON files into a single minified file for distribution
* [ ] Add schema and validate shield definitions against it; requires #594
* [ ] Remove shield_defs.js and compile_shield_defs.js
* [ ] Generate placeholder images for shape keywords for taginfo
* [ ] Generate shieldtest.js networks based on a property in individual shield definitions
* [ ] Update CONTRIBUTING.md

Fixes #118.

### DIFF
--- a/data/shield_defs.json
+++ b/data/shield_defs.json
@@ -1,0 +1,18805 @@
+{
+  "am:national": {
+    "shape": "rectangle",
+    "fillColor": "blue",
+    "strokeColor": "white",
+    "padding": {
+      "left": 3,
+      "right": 3,
+      "top": 3,
+      "bottom": 3
+    }
+  },
+  "AT:A-road": {
+    "shape": "rectangle",
+    "fillColor": "blue",
+    "strokeColor": "white",
+    "padding": {
+      "left": 3,
+      "right": 3,
+      "top": 3,
+      "bottom": 3
+    }
+  },
+  "AT:S-road": {
+    "shape": "rectangle",
+    "fillColor": "blue",
+    "strokeColor": "white",
+    "padding": {
+      "left": 3,
+      "right": 3,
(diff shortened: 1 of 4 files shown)


1ec5:
The unit test harness in #594 would enable us to automatically validate that the shield data conforms to a schema. In the future, we could even install additional tooling to enable pixel diffs of the generated shields, similar to GL JS rendering tests, which will obviate the need to examine shieldtest.js for regressions.

1ec5:
Thanks for running with this idea. Some of the unchecked boxes in the to-do list above would be logical next steps now that #714 has landed.

Issue #20, opened 2021-07-07
Implement language switching in URL querystring


ZeLonewolf:
Depends on #19

A key feature of vector tiles is the ability to switch languages for name labels.  OpenMapTiles provides `name:xx` tags wherever a `name` tag is available.  Proposed is that the following parameters can be parsed from the querystring:

*`lang0=xx` - use `name:xx` tag for name features
*`langN=xx` - use `name:xx` as a fallback language if the `lang0` language is not available.  N is numbered from 1 and counting up

If none of the specified languages or fallback languages are set, this will default to one of the standard name values (one of `name`, `name_int`, or `name_latin`).

In javascript, a JSON fragment will need to be generated which creates a case statement to switch the specified language names, which is then inserted into the appropriate locations in the style JSON variables.

ZeLonewolf:
The following library may be of use:
https://github.com/klokantech/openmaptiles-language

1ec5:
> The following library may be of use: https://github.com/klokantech/openmaptiles-language

This script predates expressions and uses the deprecated string interpolation syntax, so it’s incompatible with this project. It also isn’t distributed as an NPM package. There is [a compatible Mapbox GL JS library](https://github.com/mapbox/mapbox-gl-language/) for localizing labels. However, I don’t think a library is necessary at all. Unlike most styles that are distributed in static JSON format, Americana is built at runtime in JavaScript, giving us the ability to inject environment variables like the language as we build each layer. #578 implements this approach without needing to hard-code a single `name:*` field.

#578 relies on user preferences, but I agree that it would be handy to be able to set the language inside the URL. There are two typical approaches to implementing a URL-based language switch:

* A `?language=` query parameter would be ideal. I think GitHub Pages supports query parameters just fine, but when running `npm start`, appending any query parameter currently results in a 404. We would need nativew/serve#8 to land before taking this approach.
[...]

PR #578, opened 2022-11-27
Use user-preferred language in labels

1ec5:
All map labels now use the `name:*` field corresponding to the user’s preferred language if available, falling back to the user’s second preferred language and so on. If none of the preferred languages is available, labels fall back to the local language. English is no longer hard-coded.

Spanish and Traditional Chinese, falling back to the local language:

[Spanish](https://zelonewolf.github.io/openstreetmap-americana/#6/34.18/-118.798) [Chinese](https://zelonewolf.github.io/openstreetmap-americana/#6/34.18/-118.798)

Working towards #20.

### DIFF
--- a/src/constants/label.js
+++ b/src/constants/label.js
@@ -2,8 +2,7 @@
 
 // Name fields in order of preference
 export const localizedName = (function () {
-  let userLocales =
-    "languages" in navigator ? [...navigator.languages] : [navigator.language];
+  let userLocales = navigator.languages ?? [navigator.language];
   let locales = [];
   let localeSet = new Set(); // avoid duplicates
   for (let locale of userLocales) {
--- a/src/constants/label.js
+++ b/src/constants/label.js
@@ -1,9 +1,24 @@
 "use strict";
 
 // Name fields in order of preference
-export const localizedName = [
-  "coalesce",
-  ["get", "name:en"],
-  ["get", "name:latin"],
-  ["get", "name"],
-];
+export const localizedName = (function () {
+  let userLocales =
+    "languages" in navigator ? [...navigator.languages] : [navigator.language];
+  let locales = [];
+  let localeSet = new Set(); // avoid duplicates
+  for (let locale of userLocales) {
+    // Add progressively less specific variants of each user-specified locale.
+    let components = locale.split("-");
+    while (components.length > 0) {
+      let parent = components.join("-");
+      if (!localeSet.has(parent)) locales.push(parent);
+      localeSet.add(parent);
+      components.pop();
+    }
+  }
+  if (locales.at(-1) === "en") {
+    locales.push("latin");
+  }
+  let nameFields = [...locales.map((l) => `name:${l}`), "name"];
+  return ["coalesce", ...nameFields.map((f) => ["get", f])];
+})();
--- a/src/constants/label.js
+++ b/src/constants/label.js
@@ -1,7 +1,14 @@
 "use strict";
 
-// Name fields in order of preference
-export const localizedName = (function () {
+/**
+ * Returns a `coalesce` expression that resolves to the feature's name in a
+ * language that the user prefers.
+ *
+ * @param {boolean} includesLegacyFields - Whether to include the older fields
+ *  that include underscores, for layers that have not transitioned to the
+ *  colon syntax.
+ */
+function getLocalizedNameExpression(includesLegacyFields) {
   let userLocales = navigator.languages ?? [navigator.language];
   let locales = [];
   let localeSet = new Set(); // avoid duplicates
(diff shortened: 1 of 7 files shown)


src/constants/label.js
```diff
@@ -1,9 +1,24 @@
 "use strict";
 
 // Name fields in order of preference
-export const name_en = [
-  "coalesce",
-  ["get", "name:en"],
-  ["get", "name:latin"],
-  ["get", "name"],
-];
+export const localizedName = (function () {
```
1ec5, line 4: The `text-field` expressions continue to be evaluated statically as the page loads. To dynamically switch languages after the page loads, we could listen for the [`languagechange`](https://developer.mozilla.org/en-US/docs/Web/API/Window/languagechange_event) and tear down and rebuild all the layers. This would require refactoring americana.js.

src/constants/label.js
```diff
@@ -1,9 +1,24 @@
...
+    let components = locale.split("-");
+    while (components.length > 0) {
+      let parent = components.join("-");
+      if (!localeSet.has(parent)) locales.push(parent);
+      localeSet.add(parent);
+      components.pop();
+    }
+  }
+  if (locales.at(-1) === "en") {
+    locales.push("latin");
```
1ec5, line 20: This Latin fallback is vestigial at this point, but there remains another Latin fallback for roads and such:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/830d43690aa2a8e2c2e14a63e79deba259ffd859/src/layer/transportation_label.js#L82-L87

I’d like to replace this with `Label.localizedLabel`, but I’m unsure what the expected behavior would be for street names. Do we really need to split apart the Latin and non-Latin characters in a name? Wouldn’t they already be separate in a typical bilingual `name`?

src/constants/label.js
```diff
@@ -1,9 +1,24 @@
...
 // Name fields in order of preference
-export const name_en = [
-  "coalesce",
-  ["get", "name:en"],
-  ["get", "name:latin"],
-  ["get", "name"],
-];
+export const localizedName = (function () {
+  let userLocales =
+    "languages" in navigator ? [...navigator.languages] : [navigator.language];
```
1ec5, line 6: The user-preferred language could be Arabic or Hebrew, which are both written right to left. As tail work, we should install the [mapbox-gl-rtl-text](https://github.com/mapbox/mapbox-gl-rtl-text/) plugin to keep the map readable for users of these languages.

ZeLonewolf:
How does this work? Is it based on a user's web browser settings?

ZeLonewolf:
Awesome...

image

image

ZeLonewolf (approved):
I'm not going to pretend that I fully understand how this works, but the result is awesome!

src/constants/label.js
jleedev, line 6:
```suggestion
  let userLocales = navigator.languages ?? [navigator.language];
```

jleedev:
I seem to recall that it was once possible for navigator.languages to be empty, but both chrome and firefox appear to default to en-us when I set their respective preference to empty string, so defensive coding there isn't needed.

src/constants/label.js
1ec5: Thanks, this didn’t work when I first tried it because I hadn’t split out a separate `locales` array yet. 👍

1ec5:
> I seem to recall that it was once possible for navigator.languages to be empty, but both chrome and firefox appear to default to en-us when I set their respective preference to empty string, so defensive coding there isn't needed.

This is actually for Node, which doesn’t support `navigator.languages` yet.

jleedev:
> > I seem to recall that it was once possible for navigator.languages to be empty, but both chrome and firefox appear to default to en-us when I set their respective preference to empty string, so defensive coding there isn't needed.
> 
> This is actually for Node, which doesn’t support `navigator.languages` yet.

Node.js doesn't have `navigator` at all, not sure what you mean here.

1ec5:
Oh, right…

jleedev (approved):

1ec5:
I brought the `transportation_label` layers under the new localized label system as well. However, due to openmaptiles/openmaptiles#769, roads can only be in English, German, or the local language.

[Hoover Dam Access Road](https://zelonewolf.github.io/openstreetmap-americana/#15/36.01521/-114.73852) [Hoover-Talsperre](https://zelonewolf.github.io/openstreetmap-americana/#15/36.01521/-114.73852)

src/constants/label.js
1ec5: #580

src/constants/label.js
1ec5: #579

answerquest:
How do we make this change happen? I'm not seeing any place in the main map https://zelonewolf.github.io/openstreetmap-americana to change language

1ec5:
[See the instructions on the readme](https://github.com/ZeLonewolf/openstreetmap-americana#how-to-use).

1ec5:
#622 implemented a control in the lower-right corner to set the map’s language independently of your browser settings, so you no longer need to know the ISO&nbsp;639 code of the language you want to see.

Find out more about Americana’s label localization and its use cases in [this OSMUS blog post](https://www.openstreetmap.us/2022/12/americana-language/).

PR #581, opened 2022-11-27
Localize labels dynamically

### LINKED ISSUE #20: Implement language switching in URL querystring

### LINKED ISSUE #579: Update label language when user language preference changes
The `text-field` expressions for label layers are only evaluated statically as the page loads. To dynamically switch languages after the page loads, we could listen for the [`languagechange`](https://developer.mozilla.org/en-US/docs/Web/API/Window/languagechange_event) and tear down and rebuild all the layers. This would require refactoring americana.js.

_Originally posted by @1ec5 in https://github.com/ZeLonewolf/openstreetmap-americana/pull/578#discussion_r1032848256_

1ec5:
The format of the hash at the end of the application’s URL has changed from `#zoom/latitude/longitude` to `#map=zoom/latitude/longitude`, which is consistent with both openstreetmap.org and iD. For easier map sharing and debugging, the application now accepts a `language=` parameter as part of the hash. For example, `#map=12/29.95/-90.08` zooms to New Orleans, but `#language=fr&map=12/29.95/-90.08` zooms to _La Nouvelle Orléans_, and `#language=vi,es&map=12/29.95/-90.08` zooms to _Nueva Orleans_.

The application is now capable of rebuilding the entire style at an arbitrary time after the page loads. It rebuilds the style any time the user’s language preference changes or the URL hash’s `language` parameter changes. The only dynamic part of the style is label localization. Rebuilding the layers is challenging because of the codebase’s extensive use of global variables, but this PR uses a `metadata` field to persist an indication of the intention to localize a given layer’s `text-field` property, which is resolved just before setting the style.

Fixes #20 and fixes #579.

### DIFF
--- a/src/americana.js
+++ b/src/americana.js
@@ -2,6 +2,8 @@
 
 import config from "./config.js";
 
+import * as Label from "./constants/label.js";
+
 import * as Util from "./js/util.js";
 import * as Shield from "./js/shield.js";
 import * as ShieldDef from "./js/shield_defs.js";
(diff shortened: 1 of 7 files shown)


claysmalley (approved):
Fantastic work!

PR #585, opened 2022-11-27
Draw trapezoid shields mathematically without SVGs

claysmalley:
Adds a function to draw trapezoid shield backgrounds with arbitrary widths, colors, corner angles and corner radii. This function replaces the trapezoid SVG icons.

Screenshot from 2022-11-27 18-34-32
Screenshot from 2022-11-27 18-34-09


ZeLonewolf:
Nebraska:
image

ZeLonewolf:
Ontario secondary:
image

claysmalley:
Those are actually Ontario county roads, same as Nebraska state highways.

Here are some Ontario secondary routes:

Screenshot from 2022-11-27 16-39-55

claysmalley:
That actually exposed a math error on my part... I forgot to assign the sine a sign.

Screenshot from 2022-11-27 18-31-34

src/js/shield_defs.js
```diff
@@ -651,20 +638,37 @@ export function loadShields(shieldImages) {
...
   );
-  shields["CA:ON:Halton"] = trapezoidUpShieldGreenYellow;
-  shields["CA:ON:Peel"] = trapezoidUpShieldBlackYellow;
-  shields["CA:ON:Simcoe"] = {
-    ...trapezoidUpShield,
-    textColor: Color.shields.blue,
-    colorLighten: Color.shields.blue,
-  };
+  shields["CA:ON:Halton"] = trapezoidShield(
+    10,
```
1ec5, line 651: Should the angle default to 10°? It seems to be a pretty common design.

1ec5 (approved):
I agree with generating shapes programmatically when possible. It’ll be a lot less annoying to add new width variants of a shield.

This will remove some icons from this project’s taginfo listing for now, until we finish implementing #577.

src/js/shield_defs.js
claysmalley: I chose 10° for most of these because it was visually similar to the SVG icon, but these shields use different angles in real life. I don't want to set a precedent for all trapezoid shields to look the same just because they're trapezoids—if Nebraska state highways and Ontario county roads have different dimensions, we can try to match those dimensions if we want.

Also, negative numbers indicate a trapezoid with a long bottom (like `CA:ON:secondary`). The top-heavy trapezoid may be more common, and I did assign it a positive number accordingly, though it would feel weird to default to one or the other.

ZeLonewolf:
Looks good:

image

PR #587, opened 2022-11-27
Global Find/replace to convert CR/LF -> LF

ZeLonewolf:
This only impacted one file.  Converts CR/LF endings to LF endings.


1ec5 (approved):
Not sure how this one slipped through the cracks. #585 also runs this file through svgo.

PR #588, opened 2022-11-27
Auto-configure git to use LF endings

ZeLonewolf:
According to Professor Google:

https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/#a-simple-gitattributes-config

...adding this file will make git enforce LF endings when committing to the repository.


jleedev:
Will this avoid corrupting binary files? (Probably, but worth asking).

ZeLonewolf:
Unclear from the documentation I've seen, but when I ran `git add --renormalize .` it didn't change any of the few binary files we have in the repo.

.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto
```
Pengor, line 1: This ought to be fine, just wanted to share one other option for consideration:
```
*.js text
*.md text
etc for each file extension
```
Tradeoff is harder to maintain but less likely to mangle any binary files

Pengor (approved):

.gitattributes
Pengor: After reading more on this I see that explicit exclusion of paths as `text` is also supported and that leads me to support `* text=auto` rather than what I've mentioned above

PR #591, opened 2022-11-29
Draw home plate and vertical hexagon shields with math

claysmalley:
Adds two more shield-drawing functions, one for home plate shapes and the other for vertical hexagons.

Screenshot from 2022-11-28 20-11-51
Screenshot from 2022-11-28 20-12-08
Screenshot from 2022-11-28 20-02-33
Screenshot from 2022-11-28 20-02-57


ZeLonewolf (approved):
Everything seems in order.  The only possible optimization might be to extract out the green/white/white variants of these two shapes, as we seem to have a lot of them, rather than re-constructing them each time.

PR #594, opened 2022-12-01
Add unit tests

1ec5:
Installed Mocha and Chai as development dependencies so that we can write unit tests for parts of this application that can run independently of a browser. Unit tests run as a continuous integration check, or you can run them manually using `npm test`. As a starting point, I added tests of a trivial URL parsing function from #581, but more will come as part of #592.

Updated various documentation to reflect the current state of the deployed map.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
     "start": "run-s clean sprites serve",
     "serve": "node scripts/serve.js",
     "build:code": "node scripts/build.js",
-    "build": "run-s clean sprites build:code taginfo status_map"
+    "build": "run-s clean sprites build:code taginfo status_map",
+    "test": "mocha test/spec/**/*.js"
   },
   "dependencies": {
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -20,7 +20,8 @@
     "start": "run-s clean sprites serve",
     "serve": "node scripts/serve.js",
     "build:code": "node scripts/build.js",
-    "build": "run-s clean sprites build:code taginfo status_map"
+    "build": "run-s clean sprites build:code taginfo status_map",
+    "test": "mocha test/spec/**/*.js"
```
jleedev, line 24: Glob needs to be quoted for portability. https://mochajs.org/#the-test-directory

But I'd just as soon do:

```sh
echo 'recursive: true' > .mocharc.yml
```

and then `mocha` can be invoked with no arguments.

jleedev (approved):

PR #603, opened 2022-12-05
Indicate minimum nodejs version

jleedev:


ZeLonewolf (approved):
Now gives a warning when running `npm install` which is at least an improvement over current behavior!

PR #604, opened 2022-12-05
Deprecation warnings in workflow actions

jleedev:
Boring details:

- Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout@v2.3.1, actions/setup-node@v1
- The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


ZeLonewolf (approved):
Please let me know if any repository settings need to change to support this.

PR #605, opened 2022-12-05
Nix `name:latin` fallback from English

1ec5:
As currently implemented, `name:latin` is generally unsuitable for display: https://github.com/onthegomap/planetiler/issues/86#issuecomment-1050039099 https://github.com/onthegomap/planetiler/issues/86#issuecomment-1153166781. In some languages, it’s sort of like asking Cicero to sound out Shakespearean sonnets or the _Monday Night Football_ teleprompter. Removing this fallback also improves consistency between user-preferred languages.

Also updated the taginfo project file to note the new city label glosses.

### DIFF
--- a/scripts/taginfo_template.json
+++ b/scripts/taginfo_template.json
@@ -13,7 +13,7 @@
     {
       "key": "name",
       "object_types": ["node", "way", "relation", "area"],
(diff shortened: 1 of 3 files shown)


ZeLonewolf (approved):
Looks good!

PR #606, opened 2022-12-06
Change place label halos from pure white, to background fill color

zekefarwell:
Ever since @1ec5 improved the road labels halos in #203, I've found the pure white place label halos a bit distracting by comparison.  This PR tones down the place label halos just a bit by giving them a similar knockout style treatment as the road labels.  It's a fairly subtle difference, but I think it makes the labels feel more grounded and less like they are jumping out of the map, especially at lower zoom levels.

**Northeast Before:**
northeast-before

**Northeast After:**
northeast-after

**Europe Before:**
europe-before

**Europe After:**
europe-after

### DIFF
--- a/src/layer/park.js
+++ b/src/layer/park.js
@@ -37,8 +37,8 @@ export const label = {
   filter: ["has", "rank"],
   paint: {
     "text-color": Color.parkLabel,
-    "text-halo-blur": 1,
-    "text-halo-color": "rgba(255, 255, 255, 1)",
+    "text-halo-blur": 0.5,
+    "text-halo-color": Color.backgroundFill,
     "text-halo-width": 1,
   },
   layout: {
@@ -88,8 +88,8 @@ export const parkLabel = {
   filter: ["==", ["get", "class"], "park"],
   paint: {
     "text-color": Color.parkLabel,
-    "text-halo-blur": 1,
-    "text-halo-color": "rgba(255, 255, 255, 1)",
+    "text-halo-blur": 0.5,
+    "text-halo-color": Color.backgroundFill,
     "text-halo-width": 1,
   },
   layout: {
--- a/src/layer/park.js
+++ b/src/layer/park.js
@@ -37,8 +37,8 @@ export const label = {
   filter: ["has", "rank"],
   paint: {
     "text-color": Color.parkLabel,
-    "text-halo-blur": 0.5,
-    "text-halo-color": Color.backgroundFill,
+    "text-halo-blur": 1,
+    "text-halo-color": Color.parkLabelHalo,
     "text-halo-width": 1,
   },
   layout: {
@@ -88,8 +88,8 @@ export const parkLabel = {
   filter: ["==", ["get", "class"], "park"],
   paint: {
     "text-color": Color.parkLabel,
-    "text-halo-blur": 0.5,
-    "text-halo-color": Color.backgroundFill,
+    "text-halo-blur": 1,
+    "text-halo-color": Color.parkLabelHalo,
     "text-halo-width": 1,
   },
   layout: {
--- a/src/layer/place.js
+++ b/src/layer/place.js
@@ -1,8 +1,9 @@
 import * as Label from "../constants/label.js";
+import * as Color from "../constants/color.js";
 
 const cityLabelPaint = {
   "text-color": "#444",
-  "text-halo-color": "rgb(255,255,255)",
+  "text-halo-color": Color.backgroundFill,
   "text-halo-width": 2,
   "text-halo-blur": 0.5,
 };
@@ -210,7 +211,7 @@ export const state = {
   type: "symbol",
   paint: {
     "text-color": "hsl(45, 6%, 10%)",
-    "text-halo-color": "rgb(255,255,255)",
+    "text-halo-color": Color.backgroundFill,
     "text-halo-width": 2,
     "text-halo-blur": 0.5,
   },
@@ -250,8 +251,8 @@ export const countryOther = {
   type: "symbol",
   paint: {
     "text-color": "#334",
-    "text-halo-blur": 1,
-    "text-halo-color": "rgba(255,255,255,0.8)",
+    "text-halo-blur": 0.5,
+    "text-halo-color": Color.backgroundFill,
     "text-halo-width": 2.0,
   },
   filter: [
@@ -279,9 +280,9 @@ export const country3 = {
   type: "symbol",
   paint: {
     "text-color": "#334",
-    "text-halo-blur": 1,
-    "text-halo-color": "rgba(255,255,255,0.8)",
-    "text-halo-width": 2.0,
+    "text-halo-blur": 0.5,
+    "text-halo-color": Color.backgroundFill,
+    "text-halo-width": 2,
   },
   filter: [
     "all",
@@ -309,9 +310,9 @@ export const country2 = {
   type: "symbol",
   paint: {
     "text-color": "#334",
-    "text-halo-blur": 1,
-    "text-halo-color": "rgba(255,255,255,1)",
-    "text-halo-width": 3.0,
+    "text-halo-blur": 0.5,
+    "text-halo-color": Color.backgroundFill,
+    "text-halo-width": 2,
   },
   filter: [
     "all",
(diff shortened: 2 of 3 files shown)


src/layer/park.js
```diff
@@ -88,8 +88,8 @@ export const parkLabel = {
   filter: ["==", ["get", "class"], "park"],
   paint: {
     "text-color": Color.parkLabel,
-    "text-halo-blur": 1,
-    "text-halo-color": "rgba(255, 255, 255, 1)",
+    "text-halo-blur": 0.5,
+    "text-halo-color": Color.backgroundFill,
```
1ec5, line 92: The neutral background color works better than white against a park’s mint green fill, but I wonder if something closer to green would achieve an even more subtle knockout effect.

claysmalley:
Can we try it with minimal or no blur? I'm seeing rectangular artefacts in some scripts.


Martinique in Korean ([localhost](http://localhost:1776/#language=ko,en&map=3.56/14.63/-60.98)):
Screenshot from 2022-12-05 22-23-16

src/layer/place.js
```diff
@@ -339,9 +340,9 @@ export const country1 = {
   type: "symbol",
   paint: {
     "text-color": "#334",
-    "text-halo-blur": 1,
-    "text-halo-color": "rgba(255,255,255,1)",
-    "text-halo-width": 3.0,
+    "text-halo-blur": 0.5,
+    "text-halo-color": Color.backgroundFill,
+    "text-halo-width": 2,
```
1ec5, line 345: Did you find that the neutral color was still too loud at a width of 3&nbsp;pixels? It is reasonable to use a wide knockout for emphasis while staying grounded, so to speak. Maybe the blur is actually drawing some attention to these labels.

1ec5 (commented):
I agree with trying to achieve a knockout effect instead of a halo effect. Let’s try a couple variations, but this is already better than the status quo.

zekefarwell:
> Can we try it with minimal or no blur? I'm seeing rectangular artefacts in some scripts.

Yes I'm seeing the same, but not as a result of these changes.  The same issue is present on main as well.  I did reduce several of the `text-halo-blur` properties from `1` to `0.5` already, but I'll try some more variations.  It may have more to do with too high of a `text-halo-width`.

src/layer/park.js
zekefarwell: Good idea.  I'll try it out.

src/layer/place.js
zekefarwell: 3 pixels just seemed very thick looking when these large country labels stick out over the water.  Over the land 3 pixels does look good though. 

2 pixels:
Screen Shot 2022-12-05 at 10 54 42 PM

3 pixels:
Screen Shot 2022-12-05 at 10 54 08 PM

Hmm, yeah maybe the 3 pixels is better than 2 even when it appears to create a peninsulas 😀

1ec5: How about without a blur? I know it sounds counterintuitive, but maybe the blur is drawing extra attention to the labels that wouldn’t be as pronounced if the halo were simply wide.

zekefarwell: No blur looks pretty good.  Definitely a much sharper look (on a 2x display anyway), though.  

Screen Shot 2022-12-05 at 11 39 29 PM

To make the halos as thick as possible, no blur may be necessary.  It seems that once the halo or it's blur hits the edge of the letter box, we start getting blocky artifacts.
https://osmus.slack.com/archives/C01V02K52UX/p1670300775628049?thread_ts=1670296592.027569&cid=C01V02K52UX

1ec5: I like this look without the blur. But if you think it’s too eye-catching, we could make the halo color slightly translucent, with an alpha of, say, 90%. It wouldn’t quite be a knockout anymore, but any water or roads that shine through would be less apparent.

src/layer/park.js
zekefarwell: **Background fill halo:**
background-fill-halo-green-text

**Park fill green halo:**
park-fill-halo-green-text
Ok here are some park label halo variations.

**Park fill green halo, black text:**
park-fill-halo-black-text

**lighter greenish/offwhite (F0F4EC) halo:**
F0F4EC-halo-green-text

Simply using the park fill green doesn't work so well in my opinion.  The green text needs more contrast with the green fill and the lighter halo helped with that.  With a darker text color that provides enough contrast the park fill green does work well.  However, I don't want to get into label text color changes in this PR.   A lighter greenish/offwhite halo is an improvement over the park fill green.  I'm not sure if it's better or worse than the background fill though.

src/layer/place.js
```diff
@@ -309,9 +325,14 @@ export const country2 = {
...
-    "text-halo-color": "rgba(255,255,255,1)",
-    "text-halo-width": 3.0,
+    "text-halo-blur": 0,
+    "text-halo-color": Color.backgroundFill,
+    "text-halo-width": [
+      'interpolate', ['linear'],
+      ['zoom'],
+      2, 2,
+      5, 3,
+    ],
```
zekefarwell, line 335: I noticed that the blocky artifacts in label halos were showing up more at lower zooms than higher ones.  This makes sense because the halo width was set to a fixed pixel value while the text size increased with the zoom level.  So 3 pixlels could be a perfect halo width at zoom 5 and higher where the font size is 17, but as you zoom out an the font size drops to 11, available space in the letter box also shinks and a 3 pixel halo starts to hit the edge and cause artifacts.  Replacing the fixed value with an interpolate expression seems to (mostly?) resolve the artifact issue.

I've also set the halo blur to zero as @1ec5 suggested so we can have a look at that as a possibility.  I think it may work nicely.  I'm sure bringing back a small blur would be viable if deemed necessary. It would probably need to be interpolated as well and the halo width would need to shrink a bit to avoid hitting the letter box edges and causing artifacts.

  Using an interpolate expression here instead of a fixed

src/layer/park.js
zekefarwell: I've changed the park label halo color to the lighter greenish/offwhite

zekefarwell:
I've adjusted the place label halos at low zoom (< z4) to have partial transparency, a slight blur, and be a bit thinner.  I think this works better because there are a lot of labels spanning both land and water at low zoom and the transparency makes the labels less disruptive to the coastline.  Lines on the map are fainter and fewer at low zoom, so they don't show through the transparency as much as the would at higher zooms. 

Screen Shot 2022-12-08 at 1 08 40 PM Screen Shot 2022-12-08 at 1 08 16 PM

z4 to z5 is a transition zone where the halo transparency and blur is reduced while the width is increased.  So by z5 we have a fully opaque, faux-knockout effect.  Since there are more darker lines on the map at this zoom level, and fewer labels spanning water and land, full opacity works better here.

Screen Shot 2022-12-08 at 1 16 56 PM

adamfranco (approved):
The translucent low-zoom option looks like a nice compromise.

ZeLonewolf (approved):
I think we can live with the coastline overlaps at higher zooms, at least until someone comes up with a better answer.  Thanks for all the hard work on this!

src/layer/place.js
```diff
@@ -1,10 +1,23 @@
...
+import * as Color from "../constants/color.js";
+
+const labelHaloColor = [
+  "interpolate",
+  ["linear"],
+  ["zoom"],
+  4,
+  `hsla(${Color.backgroundHsl}, 0.8)`,
+  5,
+  `hsl(${Color.backgroundHsl})`,
```
1ec5, line 11: This is identical to the `backgroundFill` constant. Anyways, I think it would be fine to define two different color constants that share the same first three components. Deconstructing the CSS color syntax gets into diminishing returns.

1ec5 (approved):
I think this is an improvement on multiple fronts. We can revisit the translucent halos over water at low zoom levels as part of #608, but I suspect that a lighter tint of blue for water areas will render that issue largely moot. There might be a potential for the darker waterline to visibly cut through the translucent halo, but by then we’ll be able to reduce the transparency without making the coastline so ambiguous. There’s also a halo around water labels that we can revisit as part of that issue.

src/layer/place.js
zekefarwell: Good point.  I've [simplified it.](https://github.com/ZeLonewolf/openstreetmap-americana/pull/606/commits/df867e24e6aada7249a67e8e2ee128806486b709)  I guess I was thinking that the label halo color should be tied to the background color in case it changes, but that's not really necessary.

PR #609, opened 2022-12-09
Display which language codes are currently in use

ZeLonewolf:
This PR adds a small label in the bottom right to display which languages are currently used by the style.

image


src/js/language_label.js
```diff
@@ -0,0 +1,35 @@
...
+    this._map = undefined;
+  }
+}
+
+export var label = new LanguageControl();
+
+export function displayLocales(localeSet) {
+  document.getElementById("language-field").textContent =
+    Array.from(localeSet).join(", ");
+}
```
1ec5, line 35: `Label.getLocales()` returns an array, not a set. Use [`Intl.DisplayNames`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames) to format these IETF language tags as human-readable strings in either the user’s preferred language:

```suggestion
export function displayLocales(locales) {
  let languageNames = new Intl.DisplayNames(locales, {type: "language"});
  let listFormat = new Intl.ListFormat(locales, {type: "disjunction"});
  document.getElementById("language-field").textContent =
    listFormat.format(locales.map(locale => languageNames.of(locale)));
}
```

ZeLonewolf:
Thanks for the human-readable code @1ec5.  Looks awesome!

image

1ec5 (approved):

src/js/language_label.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+}
+
+export var label = new LanguageControl();
+
+export function displayLocales(locales) {
+  let languageNames = new Intl.DisplayNames(locales, { type: "language" });
+  let listFormat = new Intl.ListFormat(locales, { type: "disjunction" });
+  document.getElementById("language-field").textContent = listFormat.format(
+    locales.map((locale) => languageNames.of(locale))
```
1ec5, line 36: If you want to link each language to its Wikidata item:

```js
`https://www.wikidata.org/wiki/Special:ItemByTitle/enwiki/ISO%20639:${languageCode}`
```

or to its English Wikipedia article:

```js
`https://en.wikipedia.org/wiki/Special:Search/ISO%20639:${languageCode}`
```

or to its article in the user’s preferred language:

```js
`https://hub.toolforge.org/P305:${locale}`
```

ZeLonewolf: Tracked in #611, because I'm not sure which would be the best reference.

PR #610, opened 2022-12-09
Waterlines

### LINKED ISSUE #608: Outline waterbodies
This style currently gives bodies of water a relatively dark fill. On land, the water layer overpowers the land in areas with many lakes and streams, undercutting the illusion that the land is a layer atop the water (which is really just an illusion). At sea, the dark color contrasts too much with the boundaries of marine parks.

[Panama](https://zelonewolf.github.io/openstreetmap-americana/#map=5/7.7/-81.53)

[New Orleans](https://zelonewolf.github.io/openstreetmap-americana/#map=8/29.742/-89.715)

[Fairfield](https://zelonewolf.github.io/openstreetmap-americana/#map=13/39.33227/-84.60103)

Print maps use a variety of blues for water. Maps aiming for a more rustic look tend to choose darker shades, but most choose lighter shades, probably to cut down on the cost of ink. A light fill almost always has a hairline stroke around the coastline or shoreline to give it more definition.

Rand McNally Hallwag California San Mateo County 2006 by Compass Maps

I think we should lighten the water fill and add a darker outline that matches the waterway line color. To keep seams in the water layer from appearing at the edges of harbors and along rivers, the line layer should be underneath the fill layer. I think a lighter color will make the marine parks look much less awkward and also dull the effect of faux knockout halos around place labels over sea (#606).

/ref gravitystorm/openstreetmap-carto#3895 gravitystorm/openstreetmap-carto#3896 gravitystorm/openstreetmap-carto#3926

1ec5:
Bodies of water are filled in a pastel blue and are outlined with a medium blue hairline. The outline is hidden at low zoom levels except along coastlines. Intermittent lakes have a dashed outline. Sea and ocean labels have a halo that matches the fill color for bodies of water, which will result in a knockout effect more often than a halo. Park boundaries along coastlines are blended with the waterline; park boundaries over water are much more subtle, no longer appearing to glow.

Before | After
----|----
[Gulf of Mexico before](https://zelonewolf.github.io/openstreetmap-americana/#map=4/28.68/-92.42) | Gulf of Mexico after
[Southeast Louisiana before](https://zelonewolf.github.io/openstreetmap-americana/#map=7/29.623/-89.629) | Southeast Louisiana after
[Mississippi Delta before](https://zelonewolf.github.io/openstreetmap-americana/#map=8/29.184/-89.459) | Mississippi Delta after
[Venice before](https://zelonewolf.github.io/openstreetmap-americana/#map=12/29.27786/-89.34858) | Venice after
[Mobile Bay before](https://zelonewolf.github.io/openstreetmap-americana/#map=7/30.374/-87.621) | Mobile Bay after
[Mobile before](https://zelonewolf.github.io/openstreetmap-americana/#map=8/30.45/-87.893) | Mobile after
[Pensacola before](https://zelonewolf.github.io/openstreetmap-americana/#map=10/30.3448/-87.3757) | Pensacola after
[Florida Keys before](https://zelonewolf.github.io/openstreetmap-americana/#map=5/25.17/-79.2) | Florida Keys after
[Gulf of California before](https://zelonewolf.github.io/openstreetmap-americana/#map=4/29.09/-113.27) | Gulf of California after
[New York Harbor before](https://zelonewolf.github.io/openstreetmap-americana/#map=12/40.69992/-74.01363) | New York Harbor after
[Grand Lake St. Marys before](https://zelonewolf.github.io/openstreetmap-americana/#map=12/40.51248/-84.44934) | Grand Lake St. Marys after
[Fairfield before](https://zelonewolf.github.io/openstreetmap-americana/#map=13/39.33883/-84.59982) | Fairfield after
[Kennecott before](https://zelonewolf.github.io/openstreetmap-americana/#map=11/40.7521/-112.1435) | Kennecott after
[Alviso before](https://zelonewolf.github.io/openstreetmap-americana/#map=13/37.4356/-121.98692) | Alviso after

Fixes #608.

/ref gravitystorm/openstreetmap-carto#3468


src/americana.js
```diff
@@ -45,9 +45,11 @@ function buildLayers() {
     lyrBoundary.stateCasing,
     lyrBoundary.countryCasing,
 
-    lyrWater.water,
     lyrWater.waterway,
     lyrWater.waterwayIntermittent,
+    lyrWater.waterLine,
+    lyrWater.waterLineIntermittent,
+    lyrWater.water,
```
1ec5, line 52: The water fill layer needs to draw over the waterway line layer; otherwise, there will be a prominent line down the middle of the river that looks like a seam.

src/americana.js
```diff
@@ -45,9 +45,11 @@ function buildLayers() {
     lyrBoundary.stateCasing,
     lyrBoundary.countryCasing,
 
-    lyrWater.water,
     lyrWater.waterway,
     lyrWater.waterwayIntermittent,
+    lyrWater.waterLine,
```
1ec5, line 50: The waterline layer needs to draw under the water fill layer; otherwise, there would be visible seams at tile boundaries over water and at the mouths of rivers and harbors.

src/layer/water.js
```diff
@@ -64,13 +64,66 @@ export const water = {
       0.3,
       1,
     ],
+    "fill-outline-color": Color.waterFillTranslucent,
```
1ec5, line 67: Because the waterline is under the water fill layer, only the half of the line that lies outside the waterbody would normally be visible. This outline doesn’t apply the waterline color. By applying some extra blue, it seems to help the waterline overcome the fill’s antialiasing and show through more prominently than it would on its own. The waterline relies on color instead of width because any additional width makes some waterways look cartoonish.

src/layer/water.js
```diff
@@ -64,13 +64,66 @@ export const water = {
...
+  type: "line",
+  filter: [
+    "match",
+    ["get", "class"],
+    bigRivers,
+    [">=", ["zoom"], 8],
+    mediumRivers,
+    [">=", ["zoom"], 16],
+    "lake",
+    ["all", ["!=", ["get", "intermittent"], 1], [">=", ["zoom"], 8]],
```
1ec5, line 84: Unfortunately, there’s no way to distinguish between a small lake and a large one. The Great Lakes conspicuously lack waterlines until zoom level 8. But for smaller lakes near larger bodies of water, such as Lake Okeechobee, this lack of a waterline looks more intentional.

1ec5:
The screenshots above are from a 2× display. Since the waterline is such a delicate hairline, we should make sure it’s still legible on a 1× display.

src/layer/water.js
jleedev: Try experimenting on waterLine with (say) `{"line-width":1.25,"line-blur":0.25}`, especially at 1x vs 2x. Could smooth out jaggies and make it more subtle, or just look terrible.

src/layer/water.js
jleedev: It might just be from years of staring at a lack of Great Lakes in the osm coastline, but it looks intentional to me.

The eastern seaboard looks a little worse.

coastline1
coastline2
coastline3

jleedev:
The main issue this seems to highlight is not mapping but data processing, short segments of river and riverbank disappearing at low zooms and creating gaps.

data1
data2

jleedev:
> The screenshots above are from a 2× display. Since the waterline is such a delicate hairline, we should make sure it’s still legible on a 1× display.

This spot gave me pause:

Screenshot 2022-12-09 07 10 42

Perhaps: `"line-opacity": ["interpolate",["exponential",1.25],["zoom"],7,0,10,1]`.

src/layer/water.js
1ec5: Are you referring to the North Carolina coastline? I’m unsure why none of the waters behind the barrier islands are `ocean` in the vector tiles; it’s inconsistent with similar waters elsewhere along the Eastern Seaboard and Gulf Coast.

1ec5:
> The main issue this seems to highlight is not mapping but data processing, short segments of river and riverbank disappearing at low zooms and creating gaps.

These gaps already exist on the main branch, but the waterline accentuates them as if they’re enormous oxbow lakes. For what it’s worth, I’ve been mulling a change to hide all `river` water areas at mid zoom levels, like most maps do, but it we’d need to experiment with various cutoffs. I was thinking that could be tail work.

zekefarwell:
> The screenshots above are from a 2× display. Since the waterline is such a delicate hairline, we should make sure it’s still legible on a 1× display.

Just tested on a 1x display.  The line ends up blurring, making it a bit thicker and a bit ligher than on a 2x display, but it looks fine.
cape-code-water-outline-1x png

From a 2x display for comparison.
cape-code-water-outline-2x

src/layer/water.js
ZeLonewolf: It's no mystery at all: https://www.openstreetmap.org/way/815122837

Someone pulled a Chesapeake Bay and put the coastline outside.

ZeLonewolf: Also, Virginia's maritime boundary is missing there?

1ec5: In #35, we attempted to hide maritime boundaries except for boundaries between two administrative areas. The vector tiles have North Carolina’s boundary over the Atlantic as `maritime: 0`, whereas it’s `maritime: 1` for the other surrounding states.

ZeLonewolf: Ah, I see.  Thanks @Mashin6 for fixing this a couple days ago :-D

src/layer/water.js
1ec5: The `line-width` makes the line darker (not wider), but the addition of a `line-blur` seems to cancel out the change on a 2× display. I don’t see any difference from the original state of this branch. Is there a particular area I should be testing?

src/layer/water.js
1ec5: > It might just be from years of staring at a lack of Great Lakes in the osm coastline, but it looks intentional to me.

Yes, at zoom level 3, there’s nothing unseemly about the shoreline being absent from the Great Lakes as with any other lake. However, it does start to feel like something is missing at z6 or so, when the coastline isn’t likely to be within view anymore. I’m of two minds about this discrepancy: either we’re OK with it on the basis that the waterline at low zoom levels only reflects `natural=coastline` in OSM, or it’s annoying and we should come up with a solution at the data level for inland seas.

src/americana.js
1ec5: Although there isn’t a seam at mouths of rivers, @zekefarwell pointed out that the waterline ends abruptly at the point where the `water=river` begins, seen here at zoom level 7.99 (but not at z7 or z8):

[St. Lawrence](https://zelonewolf.github.io/openstreetmap-americana/#map=7.99/46.999/-70.696)

My initial inclination was not to worry much about it. Paper maps are unapologetic about beginning and ending strokes where they stop being applicable.

I tried giving lakes and rivers a fainter waterline at low zoom levels, but any color that’s discernible from the water fill ends up overemphasizing the water feature:

[Dallas no waterline](https://zelonewolf.github.io/openstreetmap-americana/#map=6/32.777/-96.655) Dallas waterline

src/layer/water.js
jleedev: Not sure, just an idea. The lakes in eastern Ohio are pretty noisy though.

src/layer/water.js
ZeLonewolf: I've been slowly fixing the NC Outer Banks' coastline to be where it's supposed to be.  This style change does a great job of highlighting the coastline is in the wrong place, and I consider that a feature and not a bug.

ZeLonewolf:
Overall I'm pretty happy with this and I'd be good with merging in its current state as an improvement over what we have presently.

I'd like to discuss adding the line treatment to inland lakes also but that can be a separate discussion.

Also the line treatment looks a little funny when the coastline is very complex, but I don't think that's a showstopper either:

image

Are there any open issues that we think need to be hashed out?

src/americana.js
zekefarwell: I tried out some variations for inland water as well.  It's definitely a challenge.  Between z6 and z8, there are many visible water bodies with quite highly detailed geometry.  The waterline can look ok on smoother water bodies, but on complex shorelines (reservoirs for example) the line ends zig-zagging back and for so much that it ends up looking thicker and/or darker.  I found it can look quite good below z6 as the water bodies become much more generalized at that point and many smaller ones are no longer visible.  I'm optimistic we can find a way to make a water line of some sort work on inland water bodies, but I'm happy to leave that for a future PR.

zekefarwell (approved):
This is looking really good to me.  At first I wasn't quite sure about the water fill.  It seemed maybe a bit too light.  After looking at it for a while now, I think it's just about perfect.  The hairline technique you've come up with here certainly works much better than what I was trying out last year.  

I also would like to see some sort of waterline on inland water bodies at all zoom levels, but it's definitely a challenge and can be looked into later.  Some other things to note for potential future improvement:
- Some admin boundaries seems a bit too dark now that the water is much lighter.  
- Rivers and streams mapped only as lines use the water outline color now and this looks good when they are thin.  As the zoom increases and they get thicker, it starts to look off though.  Maybe river lines need casing like roads at a high enough zoom?
Screen Shot 2022-12-11 at 1 37 36 AM
Screen Shot 2022-12-11 at 1 37 50 AM

src/layer/water.js
1ec5: Sorry, I’m not seeing the `line-width` and `line-blur` make any difference there either. If you mean that the waterlines make the reservoirs in eastern Ohio look much more prominent, I’m seeing the same thing in North Texas, especially when introducing waterlines below z8: https://github.com/ZeLonewolf/openstreetmap-americana/pull/610#discussion_r1045143790.

1ec5:
> This is looking really good to me. At first I wasn't quite sure about the water fill. It seemed maybe a bit too light. After looking at it for a while now, I think it's just about perfect. The hairline technique you've come up with here certainly works much better than what I was trying out last year.

Glad to hear it. Changing the water color really affects the style’s overall mood, so I don’t take this choice lightly. I went back and forth between this and a darker shade for a while, but this lighter shade has grown on me over the past few days, probably because of the additional contrast with the waterline and labels. Plus, we save a little on ink.

> Rivers and streams mapped only as lines use the water outline color now and this looks good when they are thin. As the zoom increases and they get thicker, it starts to look off though. Maybe river lines need casing like roads at a high enough zoom?

Looks like this will happen at z14–15 and especially z16+ any time a `waterway=river` that hasn’t been mapped as an area meets one that has, as with the [Browns River](https://www.openstreetmap.org/way/539727654) meeting the [Lamoille River](https://www.openstreetmap.org/way/366824665). For a river as narrow as the Browns River, it might be misleading to give the line a casing as though it has been mapped as an area and only takes up that small amount of space. I’m also unsure how to case the dashed line we’re using for intermittent waterways.

Instead, I’ve changed the waterway layer to fade to the water fill color by z14 and moved the water line layer under the waterway layer:

[z13](https://zelonewolf.github.io/openstreetmap-americana/#map=13/44.65965/-73.01807) [z14](https://zelonewolf.github.io/openstreetmap-americana/#map=14/44.66087/-73.0185) [z15](https://zelonewolf.github.io/openstreetmap-americana/#map=15/44.66087/-73.0185) [z16](https://zelonewolf.github.io/openstreetmap-americana/#map=16/44.660472/-73.018908)

I think this is an improvement because there’s less of a need for the waterline to provide definition at this zoom level. It also avoids a mixed message to mappers that the waterline sometimes represents the edge of an area they should map but sometimes represents a line casing.

With this latest change, the waterway layer will continue to show through an intermittent water area, such as where the banks of a river fluctuate seasonally. However, the overall appearance does feel different than before this PR. Before, the surrounding area looked like residual moisture from a previous rainy season, whereas now it looks much less skeuomorphic and more like something you’d see on a traditional paper map:

[Colorado before](https://zelonewolf.github.io/openstreetmap-americana/#map=10/36.1231/-114.0735) Colorado after

Unfortunately, the lighter color makes intermittent waterways quite subtle within parks. I’m not so concerned about the main channel of the Colorado here, because the whole thing is the Colorado, but the side streams are barely perceptible even without color blindness:

[Colorado intermittent before](https://zelonewolf.github.io/openstreetmap-americana/#map=14/36.12649/-114.09572) Colorado intermittent after

In principle, we could put a land-colored knockout behind the waterway but in front of the park fill to provide the necessary contrast. Some print maps take a similar coloring book approach to land use fills. But a dashed water fill color cased in near-white surrounded by mint green comes out as a light gray.

[gray](https://zelonewolf.github.io/openstreetmap-americana/#map=14/36.13308/-114.08804)

For now, I’m keeping intermittent waterways in the darker blue, matching waterlines. It would be counterintuitive for a legend, but at least the dashing reduces the user’s expectation that the lines will join up perfectly:

darker

1ec5:
> Some admin boundaries seems a bit too dark now that the water is much lighter.

The boundary line might be difficult to lighten because it also needs to maintain contrast against the boundary casing. However, I lightened ferry lines and labels and turn the halo into a knockout to mitigate the density in some areas, now that it’s easier to maintain contrast against the water fill. Boundaries are slightly more discernible amid a sea of ferry lines:

[New York Harbor before](https://zelonewolf.github.io/openstreetmap-americana/#map=12/40.70042/-74.0194) New York Harbor after

1ec5:
Should we revisit the translucent place label halos in https://github.com/ZeLonewolf/openstreetmap-americana/pull/606#pullrequestreview-1211089504, now that it’s easier to achieve a knockout effect without implying a deformed coastline?

scripts/taginfo_template.json
```diff
@@ -298,6 +298,20 @@
...
       "description": "Displays railroads with a thinner line width but normal tie spacing.",
       "doc_url": "https://openmaptiles.org/schema/#transportation"
+    },
+    {
+      "key": "natural",
+      "value": "coastline",
+      "object_types": ["way"],
+      "description": "Coastlines are represented by a hairline at all zoom levels, except across the mouth of a river.",
+      "doc_url": "https://openmaptiles.org/schema/#water"
+    },
```
ZeLonewolf, line 308: I don't believe that we actually consume OSM coastline tagging, but I might be wrong on that.

jleedev: I believe it's natural earth <z6 and water-polygons ≥z6, the latter being an indirect use of coastline.

ZeLonewolf: :+1: that works for me.

ZeLonewolf (approved):

PR #615, opened 2022-12-12
Make language label clickable

jleedev:


ZeLonewolf (approved):
Indistinguishable from magic...

PR #616, opened 2022-12-13
Enlarge ocean, sea labels

1ec5:
Ocean and sea labels are a lot larger at low zoom levels and grow as you zoom in instead of shrinking.

[Pacific Ocean](https://zelonewolf.github.io/openstreetmap-americana/#map=1/37.1/-140.5) [Caribbean Sea](https://zelonewolf.github.io/openstreetmap-americana/#map=3/19.6/-69.8) [Gulf of California](https://zelonewolf.github.io/openstreetmap-americana/#map=4/27.66/-110.78) [Java Sea](https://zelonewolf.github.io/openstreetmap-americana/#map=4/-7.17/111.98)


ZeLonewolf (approved):
Looks decent to me.  The lack of "Atlantic Ocean" and "Gulf of Mexico" are interesting.

jleedev:
> Looks decent to me. The lack of "Atlantic Ocean" and "Gulf of Mexico" are interesting.

The three Atlantic Ocean nodes (North, South, and regular) lack a `name=*` tag.

https://www.openstreetmap.org/node/305640306
https://www.openstreetmap.org/node/305640288
https://www.openstreetmap.org/node/305639726

The Gulf of Mexico is not a node:

https://www.openstreetmap.org/way/808726001/history

1ec5:
These features unfortunately don’t end up in the vector tiles, but if that gets fixed, then they’d show up automatically. There’s probably a broader discussion to have about why the Atlantic ocean should be tagged so differently than all the others.

PR #617, opened 2022-12-13
Highlight urbanized areas at low zoom levels

1ec5:
Added a goldenrod-colored fill layer for Natural Earth settled area polygons from zoom level 4 up to but not including zoom level 6. At zoom level 6 and above, the same source layer contains only a generalization of some OSM land use areas, which isn’t useful for settled area highlighting.

[Chicago–Washington–Philadelphia–New York](https://zelonewolf.github.io/openstreetmap-americana/#map=4/40.02/-79.15) [Chicago–Detroit](https://zelonewolf.github.io/openstreetmap-americana/#map=5/41.86/-85.18) [Bos-Wash](https://zelonewolf.github.io/openstreetmap-americana/#map=5/40.67/-74.27) [England–Benelux](https://zelonewolf.github.io/openstreetmap-americana/#map=4/51.89/0.66) [England](https://zelonewolf.github.io/openstreetmap-americana/#map=5/52.13/-1.69) [Korea–Japan](https://zelonewolf.github.io/openstreetmap-americana/#map=4/35.96/132.94) [Honshu](https://zelonewolf.github.io/openstreetmap-americana/#map=5/35.78/137.33) [South Korea](https://zelonewolf.github.io/openstreetmap-americana/#map=5/36.58/127.18)


ZeLonewolf:
OMT is shoving NaturalEarth urbanized areas into residential landuse tagging in the tiles?  How on earth did you even discover that?

adamfranco (approved):
This looks great! At some point in the future we may want to revisit the tone used to disambiguate from desert types of land over, tribal areas, or other landuse and landcover that might fall in the yellow/orange/brown pallet. That said, we don't have conflicts now and this looks good and is in keeping with the paper examples, so I support going with this as is for now.

We probably wouldn't want to show precise landuse until higher zooms anyway.

1ec5:
> OMT is shoving NaturalEarth urbanized areas into residential landuse tagging in the tiles? How on earth did you even discover that?

It’s in the [OpenMapTiles schema documentation](https://openmaptiles.org/schema/#landuse). I guess the assumption is that styles would render both low-zoom urbanized areas and high-zoom residential landuse areas in the same gray, as in openstreetmap-carto. We’re bucking that assumption.

> This looks great! At some point in the future we may want to revisit the tone used to disambiguate from desert types of land over, tribal areas, or other landuse and landcover that might fall in the yellow/orange/brown pallet. That said, we don't have conflicts now and this looks good and is in keeping with the paper examples, so I support going with this as is for now.

Landcover would conflict with the overall design direction of this style, which is heavily geared towards the built environment per political and road atlases. Folks here would probably be very interested in developing a style more focused on physical geography at some point, taking the best bits of American cartographic practice for a topographic or outdoors map. I’m not sure if that would be a series of layers in this style or an entirely different style.

I forgot to mention that I think the yellow or gold color for urbanized areas in print maps is supposed to evoke city lights. Hopefully the color I chose communicates that intention reasonably well. I think it works best when some highways are visible. Otherwise, the goldenrod settled areas and green park areas can conspire to look like landcover, as in the screenshot of Western Europe above. But improvements to the vector tiles are gradually introducing highways at z4 in more countries, so I think this will be only a temporary problem.

zekefarwell:
Screenshots look good, but I'm not seeing the goldenrod polygons when testing locally.  Seems like maybe the new layer hasn't actually been added to americana.js?

ZeLonewolf (changes requested):
Please also update americana.js to add the new layer!

1ec5:
Sorry about that – must’ve did a `git checkout` when I meant `git add`.

src/layer/landuse.js
```diff
@@ -0,0 +1,28 @@
+"use strict";
+
+import * as Color from "../constants/color.js";
+
+export const urbanizedArea = {
+  id: "urbanized-area",
+  type: "fill",
+  minzoom: 4,
+  maxzoom: 6,
```
1ec5, line 9: We aren’t using these min/max zoom level properties as much as we could. They could improve the framerate at some zoom levels.

ZeLonewolf (approved):

zekefarwell (approved):
Looks good to me.  It would be nice to keep showing urbanized areas and/or more detailed landuse at medium to high zoom levels as well, but with the polygons in the source layer changing drastically after z6, that doesn't seem viable currently.

**Zoom 5.99:**
image 

Zoom 6.00 (if the maxzoom were removed).  Hartford disappears, Boston shrinks, and a few random blotches show up in Maine...
image

ZeLonewolf:
I think OMT should take these polygons to higher zoom and give them a different name.

Issue #618, opened 2022-12-14
Show highway names in users' language with gloss


adamfranco:
Around the world there are many highways that have translated names tagged in OSM, but highway names are shown in Americana as the local name, which might not be understandable by the viewer, especially in cases where the script of the local language is not familiar to the viewer. As someone who doesn't read Hebrew, Arabic, or other non-latin scripts I'd expect to see my local language or at least a Latin scripts on road names.

Here's an example of the [Armenian Patriarchate (openstreetmap.org)](https://www.openstreetmap.org/way/265414845) in Jerusalem, which has Hebrew script as the local name, but has many translations.
[Armenian Patriarchate](https://www.openstreetmap.org/way/265414845) in Jerusalem.
![Screen Shot 2022-12-14 at 11 36 28 AM](https://user-images.githubusercontent.com/25242/207654168-888fe9dd-a9cf-4664-a589-67e0aa1e257f.png)

[Location in Americana](https://zelonewolf.github.io/openstreetmap-americana/#map=17/31.77716/35.22786)
![Screen Shot 2022-12-14 at 11 36 50 AM](https://user-images.githubusercontent.com/25242/207654240-33059cc9-f6f2-4041-ae43-a5f2cad0cae6.png)

As with place labels, glossing with the local name strings would make the rendering useful for travelers so that can compare street signs while walking about while still relating the translated name to guidebooks or other materials translated into their language.

1ec5:
https://github.com/ZeLonewolf/openstreetmap-americana/pull/578#issuecomment-1328173700 implemented language switching for road names, but OpenMapTiles only contains names in English, German, and the local language: openmaptiles/openmaptiles#769.

1ec5:
Oh, looks like the localization functionality broke for streets has broken since that comment.

ZeLonewolf:
That cited openmaptiles issue (769) is about removing the legacy `name_en`, `name_de`, etc., in favor of colon-separated name tags e.g. `name:en`, `name:de`, etc.  If language name support is missing on roads but should be there, that needs to be a separate OMT ticket.

1ec5:
#619 fixes the regression introduced in #471 so that road name labels are once again localized in the user’s preferred language (as long as the user prefers English or German).

1ec5:
> As with place labels, glossing with the local name strings would make the rendering useful for travelers so that can compare street signs while walking about while still relating the translated name to guidebooks or other materials translated into their language.

[...]

PR #619, opened 2022-12-14
Fix street name localization

1ec5:
Decoupled the `text-field` expression of `transportation_name` layers from other layers, since the available fields differ. The broader problem is that we’re overrelying on constants everywhere, but a deeper copy avoids the immediate issue that changing other layers’ localization to use, e.g., `name:de` overwrites any localization to `name_de` that has already taken place on this layer.

/ref #618

### DIFF
--- a/src/layer/transportation_label.js
+++ b/src/layer/transportation_label.js
@@ -80,7 +80,7 @@ export const label = {
       ["literal", ["OpenHistorical Italic"]],
       ["literal", ["OpenHistorical"]],
     ],
-    "text-field": Label.localizedName,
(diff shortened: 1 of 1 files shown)


ZeLonewolf:
Is there a street somewhere localized in English and German we can use as a test case? :joy:

1ec5:
Yes, see https://github.com/ZeLonewolf/openstreetmap-americana/pull/578#issuecomment-1328173700.

ZeLonewolf (approved):
Looks good to me.

Let's leave #618 open to track that we need to get proper language support added to OMT for other languages.

PR #622, opened 2022-12-19
Control for users to switch languages

### LINKED ISSUE #21: User-switchable languages

ZeLonewolf:
Fixes #21 

This PR adds a widget that allows a user to select the languages for display:

Initial view:
image

When the control is opened
image

If the user clears the language list, it will revert to the browser settings.

We may want to move this over to the buttons, like the samples and the coming legend PR, but I assume that can be tweaked later.

The list of languages is also hard-coded, and the UX is entirely in English.  However, I think this is a reasonable starting point from which we can welcome enhancements.


src/js/language_label.js
```diff
@@ -1,15 +1,176 @@
...
+  { id: "sr", name: "Serbian" },
+  { id: "sr-Latn", name: "Serbian Romanized" },
+  { id: "sv", name: "Swedish" },
+  { id: "ta", name: "Tamil" },
+  { id: "te", name: "Telugu" },
+  { id: "th", name: "Thai" },
+  { id: "tr", name: "Turkish" },
+  { id: "uk", name: "Ukrainian" },
+  { id: "zh", name: "Chinese" },
+];
```
1ec5, line 108: As with https://github.com/ZeLonewolf/openstreetmap-americana/pull/609#discussion_r1044150817, use `Intl.DisplayNames` to format these locale codes. Most of the codes conform to the IETF’s BCP&nbsp;47 standard, so they’ll localize correctly with the following code:

```suggestion
let languageNames = new Intl.DisplayNames(locales, { type: "language" });
let langCodes = [
  "am",
  "ar",
  "az",
  "be",
  "bg",
  "br",
  "bs",
  "ca",
  "co",
  "cs",
  "cy",
  "da",
  "de",
  "el",
  "en",
  "eo",
  "es",
  "et",
  "eu",
  "fi",
  "fr",
  "fy",
  "ga",
  "gd",
  "he",
  "hi",
  "hr",
  "hu",
  "hy",
  "id",
  "is",
  "it",
  "ja",
  // "ja_kana",
  "ja-Latn",
  "ja-Hira",
  "ka",
  "kk",
  "kn",
  "ko",
  "ko-Latn",
  "ku",
  "la",
  "lb",
  "lt",
  "lv",
  "mk",
  "mt",
  "ml",
  "nl",
  "no",
  "oc",
  "pl",
  "pt",
  "rm",
  "ro",
  "ru",
  "sk",
  "sl",
  "sq",
  "sr",
  "sr-Latn",
  "sv",
  "ta",
  "te",
  "th",
  "tr",
  "uk",
  "zh",
].map((locale) => languageNames.of(locale));
```

1ec5: The only nonstandard code is `ja_kana`, but `name:ja_kana` has long been deprecated in favor of `ja-Hira` (which is already in the list). If we want to use `name:ja_kana` when the user requests `ja-Hrkt` (hiragana+katakana), then the translation would happen in `getLocalizedNameExpression()` in label.js.

Pengor (approved):
Tested and reviewed code changes, looks great!

src/js/language_label.js
```diff
@@ -1,15 +1,176 @@
...
+      let langCodes = [];
+      items.forEach((element) => langCodes.push(element.id));
+      let langQuery = langCodes.join(",");
+      let rawHash = window.location.hash.split("&")[0];
+      let langQueryString =
+        langCodes.length > 0 ? `&language=${langQuery}` : "";
+      var currentURL = `${window.location.protocol}//${window.location.host}${window.location.pathname}${rawHash}${langQueryString}`;
+      window.history.pushState({ path: currentURL }, "", currentURL);
+      updateLanguageLabel();
+      hotReloadMap();
```
1ec5, line 152: Similar to `upgradeLegacyHash()`, we should just change the URL hash and the rest of the application will update automatically, adding to the browser history:

```suggestion
      let hash = window.location.hash.substr(1); // omit #
      let searchParams = new URLSearchParams(hash);
      url.searchParams.set("language", langCodes.join(","));
      window.location.hash = `#${searchParams}`;
```

src/americana.js
```diff
@@ -326,10 +326,18 @@ map.on("styleimagemissing", function (e) {
   Shield.missingIconHandler(map, e);
 });
 
-window.addEventListener("languagechange", (event) => {
-  console.log(`Changed to ${navigator.languages}`);
+export function hotReloadMap() {
```
1ec5, line 329: This function doesn’t have to be exported anymore.

src/js/language_label.js
```diff
@@ -1,15 +1,179 @@
...
+      newItems: false,
+    });
+    document.querySelectorAll(".tokenfield").forEach((e) => {
+      Object.assign(e.style, {
+        height: "5em",
+        width: "20em",
+        "margin-bottom": "4em",
+        "margin-top": "0.3em",
+      });
+      e.setAttribute("placeholder", "hi");
```
1ec5, line 145: Does this field need a placeholder? It already has some explanatory text above plus a list of suggestions.

src/js/language_label.js
```diff
@@ -1,15 +1,179 @@
...
+      el: document.querySelector("#language-picker"), // Attach Tokenfield to the input element with class "text-input"
+      items: langCodes,
+      newItems: false,
+    });
+    document.querySelectorAll(".tokenfield").forEach((e) => {
+      Object.assign(e.style, {
+        height: "5em",
+        width: "20em",
+        "margin-bottom": "4em",
+        "margin-top": "0.3em",
```
1ec5, line 143: I only see a couple of the suggestions at a time, because the suggestion menu only appears below the textbox that’s already pretty far down the page. If Tokenfield doesn’t support automatically flipping the menu above the textbox, we should find a different place for the setting as tail work.

src/js/language_label.js
```diff
@@ -1,15 +1,179 @@
...
+  return { id, name: languageNames.of(id) };
+});
+
+function labelControlElement(tag, id) {
+  var element = document.createElement(tag);
+  element.id = id;
+  Object.assign(element.style, {
+    margin: "0 2.5px",
+    color: "#444",
+  });
```
1ec5, line 122: Eventually we should move non-dynamic CSS styling to a stylesheet or at least the `<style>` element in index.html. But this is OK for now if we’re going to redesign the feature soon anyways.

1ec5 (approved):

src/js/language_label.js
ZeLonewolf: I wanted it to be a placeholder, but I couldn't get it to work.  That's why I went with the text above.

src/js/language_label.js
ZeLonewolf: Does not appear to have that option unfortunately:
https://github.com/KaneCohen/tokenfield

PR #627, opened 2022-12-21
Provide a definition of the word "Americana"

ZeLonewolf:
I thought it would be a nice touch to add a definition of "Americana" as explained by an American author, for the benefit of those not familiar with American culture and zeitgeist.

I'll start the bidding with this one but I invite consideration of other quotes or ideas.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ _A quintessentially American map style_ [🗺 View the map](https://zelonewolf.g
 
 <img src="doc-img/osm-americana-logo.png" alt="Americana map style logo" width="200"/>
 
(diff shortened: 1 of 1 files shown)


zekefarwell (approved):
Looks great to me.  This quote is much more interesting than a plain definition which would be something boring like:
> materials concerning or characteristic of America, its civilization, or its culture

michaelblyons:
I like the quote. "Crazy quilt of competing themes" certainly pertains to the variety of route signs, although not as much to the base layer.

I don't think the formatting for GitHub Markdown rendering is quite right. See [preview](https://github.com/ZeLonewolf/openstreetmap-americana/blob/zlw-americana-definition/README.md) with floating `>`. I also do not think the `_`-bracketing to represent an underline is necessary when the title is linked. That already sets it apart.

> _The United States of America is such a glorious mess of contradiction, such a crazy quilt of competing themes, such a fecund mishmash of people and ideas, that defining us is pretty much pointless. There is, of course, a kind of faded notion of "Americana", one that concerns Route 66, diners, freak rock formations, and the like—but even in its halcyon days this "roadside attraction" version of America was never an accurate or nuanced distillation of our massively complicated culture._
> <cite>Hampton Sides, [Americana: Dispatches from the New Frontier](<https://en.wikipedia.org/wiki/Americana_(book)>) (2004)</cite>

Sorry for excessive bike-shedding.

ZeLonewolf:
Thanks for that catch.  I spent 5% of my time on this PR finding a suitable quote, and 95% of my time trying to format it in a way that isn't terrible.

Issue #404, opened 2022-06-12
Shields of Mexico


claysmalley:
**(EN)**

Federal Highways of Mexico have a unique black-on-white shield design. Almost all states of Mexico use a similar design with a square top for their State Highways.

Many numbered routes in Mexico are covered by `route=road` relations, but `network=*` tagging is limited. Values of `network=*` in Mexico are country-prefixed, but it is unclear which values correspond to Federal Highways, highways of Mexico City, and highways of the State of Mexico. A decision would need to be made on whether this tagging scheme is sufficient, and whether it is appropriate mapper feedback to support it.

**(ES)**

Las Carreteras Federales de México tienen un diseño único de escudo en blanco y negro. Casi todos los estados de México utilizan un diseño similar con una parte superior cuadrada para sus Carreteras Estatales.

Muchas rutas numeradas en México están cubiertas por las relaciones `route=road`, pero el etiquetado `network=*` es limitado. Los valores de `network=*` en México tienen el prefijo del país, pero no está claro qué valores corresponden a Carreteras Federales, carreteras de la Ciudad de México y carreteras del Estado de México. Sería necesario tomar una decisión sobre si este esquema de etiquetado es suficiente y si este soporte brindará una buena retroalimentación a los mapeadores.

----

Federal Highway shield:
Rótulo de las carreteras federales:
<img src="https://user-images.githubusercontent.com/1732117/173255177-bd4317cc-f0ff-49c1-b64d-2c093dad6e5f.svg" alt="Carretera_federal_45" width="200" /> <img src="https://user-images.githubusercontent.com/1732117/173255178-2c2d26d4-5292-4ddf-8186-4738d8b2fc0d.svg" alt="Carretera_federal_2D" width="200" />

State Highway shield:
Rótulo de las carreteras estatales:
<img src="https://user-images.githubusercontent.com/1732117/173255176-a17c8141-720c-4d89-acb5-e66d92c82948.svg" alt="Carretera_estatal_176_(Sonora)" width="200" />

1ec5:
#411 tracks Jalisco’s state and rural highway shields.

1ec5:
> it is unclear which values correspond to Federal Highways, highways of Mexico City, and highways of the State of Mexico

[...]

PR #628, opened 2022-12-22
Add Mexico Federal Route Shields

ZeLonewolf:
This PR adds shields for Mexico's federal highways (`network=MX:MX`)

image

Real world example:
image

Mexco federal highways have been mapped for routes 1-69 so far.


ZeLonewolf:
Partial implementation of #404.  Autopistas still need to be mapped and defined.

icons/shield_mx_mx.svg
```diff
@@ -0,0 +1,5 @@
...
...758 7.0305 11.864 12.127 3.1636 5.0972 5.4484 10.897 6.8546 17.4 1.5817 6.3275 2.6363 12.831 3.1637 19.509.70294 6.679 1.0545 13.27 1.0546 19.773-.0001 6.3274-.0001 11.952 0 16.873m-23.464 0c-.00007-2.4605-.0879-5.8-.26364-10.018-.00007-4.2181-.17583-8.7879-.52727-13.709-.35159-5.0969-.87887-10.194-1.5818-15.291-.7031-5.2726-1.7577-10.018-3.1637-14.236-1.2304-4.2181-2.8122-7.6454-4.7455-10.282-1.9334-2.6362-4.3062-3.9544-7.1182-3.9546-3.3395.00016-6.1516 1.4062-8.4364 4.2182-2.1092 2.6365-3.8668 6.1517-5.2728 10.546-1.2304 4.2184-2.197 8.9639-2.9 14.236-.52731 5.2729-.96671 10.458-1.3182 15.555-.1758 4.9214-.26368 9.4911-.26364 13.709.17572 4.0426.2636 7.1184.26364 9.2274-.00004 1.9334-.0879 5.0092-.26364 9.2274-.00004 4.0425.0878 8.6123.26364 13.709.35148 5.0971.79088 10.37 1.3182 15.818.703 5.4486 1.6697 10.37 2.9 14.764 1.406 4.394 3.1636 8.085 5.2728 11.073 2.2848 2.8122 5.097 4.2182 8.4364 4.2182 3.3394.00002 6.0636-1.406 8.1728-4.2182 2.109-2.9879 3.7788-6.6788 5.0091-11.073 1.406-4.394 2.3727-9.3152 2.9-14.764.5272-5.4485.87872-10.721 1.0546-15.818.35145-5.097.43933-9.6667.26364-13.709-.00007-4.2181-.00007-7.2939 0-9.2274" transform="translate(-3.493 -6.358) scale(.02753)"/>
```
1ec5, line 4: Can we get away with omitting or abbreviating the “MEXICO” legend? It becomes a blur at this size at 1× resolution.

ZeLonewolf: Here is a version without the "MEXICO":

image

Would it be better to shrink the crown perhaps?

ZeLonewolf: Version with shrunken crown:

image

claysmalley: Shrunken crown looks significantly better, and it's still recognizable.

1ec5: That looks much cleaner. If any other network in Mexico has a similar-looking shield with a different legend, we could add a banner on top as with the Texas networks.

claysmalley:
[Federal Highway GUA 10D](https://en.wikipedia.org/wiki/Macrolibramiento_Sur_de_Guadalajara) is currently [mistagged](https://www.openstreetmap.org/relation/14467970) with `network=MX:MEX`, otherwise we'd be able to test a 7-character `ref` in an unusually narrow shield. We'll probably need to fudge the design and add a wider variant icon.

ZeLonewolf:
That 10D road sounds like a special case anyways.  I've noticed that the 3-digit Mexico shields are the same width as the 2 digit versions in real world shields.

icons/shield_mx_mx.svg
ZeLonewolf: Shrunken heads it is.

image

Also scooted the padding up a bit since the crown is smaller now.

ZeLonewolf:
Updated GUA 10D in https://www.openstreetmap.org/relation/14467969.
We might need to do a PA Turnpike / Kentucky job on this one.  I'm thinking the Mexico shield with 10D and "GUA" in a banner might look nice.

src/js/shield_defs.js
```diff
@@ -883,6 +883,19 @@ export function loadShields(shieldImages) {
     Color.shields.white
   );
 
+  // Mexico
+
+  // Carreteras Federales
+  shields["MX:MX"] = {
+    backgroundImage: shieldImages.shield_mx_mx,
```
1ec5, line 890: >  I've noticed that the 3-digit Mexico shields are the same width as the 2 digit versions in real world shields.

I think it’s entirely reasonable to depart from the literal sign metrics to make numbers a little more readable when the shape can accommodate that kind of stretching. But I think in this case it’ll work out OK for the normal three-digit routes. GUA 10D would be a different story, but that sounds like a special case.

src/js/shield_defs.js
```diff
@@ -883,6 +883,19 @@ export function loadShields(shieldImages) {
     Color.shields.white
   );
 
+  // Mexico
+
+  // Carreteras Federales
+  shields["MX:MX"] = {
```
1ec5, line 889: I remain a little concerned that `MX:MX` is too easy to confuse with state highways of the State of México (`MX:MEX`). Presumably `MX:MX` was inspired by `US:US`, but we arrived at that value because we literally call them U.S. Routes.

We can land this support for `MX:MX` for now, but stay in touch with the Mexican community in case they flag any concerns as time goes on. If needed, I would suggest two alternatives: either `MX:national`, following the [basic format](https://wiki.openstreetmap.org/wiki/Key:network#Basic_format), or `MX:Fed`, which would be a better analogy to what the other [hierarchical-format countries](https://wiki.openstreetmap.org/wiki/Key:network#Hierarchical_format) came up with. (“Fed.” is a common abbreviation for this network in both English and Spanish.)

1ec5 (approved):

src/js/shield_defs.js
ZeLonewolf: I note that "Fed" isn't used in way refs but maybe it should be. In any case, with the route relations coming online, it'll at least be easy to change later in a bulk edit if a different community consensus emerges.

1ec5: Yes, just like in the U.S., way `ref` formats have historically been heavily influenced by the needs of pre-Americana renderers, which parse out country and state codes from that value.

PR #630, opened 2022-12-22
Refactor layer generation into a separate file and generate style.json

### LINKED ISSUE #68: Pre-generate style.json
Thanks for your great work creating this style. I was wondering if it would be possible to create a style.json from the style? As I understand there are currently multiple .js files that can be used in combination with Maplibre-GL to render a style.

I would like to use the style in combination with for example https://github.com/maptiler/tileserver-gl 

This is (as far as I am aware) only possible with a style.json.

So I was thinking, since all the rules are available as js files, wouldn't it be possible to generate a style.json from this?

ZeLonewolf:
Fixes #68

This PR produces a static style JSON which will be accessible via GitHub pages.  However, it does not (yet) use the pre-generated style JSON when setting up the map object.

This PR also separates the instantiation and ordering of layers into a separate file to improve modularity and separation of concerns.  This will simplify the workflow for anyone wishing to generate a standalone style.json.


src/americana.js
```diff
@@ -33,249 +33,21 @@ import * as search from "./search.js";
...
-    lyrPlace.country2,
-    lyrPlace.country1,
-    lyrPlace.continent
-  );
-
-  Label.localizeLayers(layers, Label.getLocales());
-
-  return layers;
-}
+import { buildLayers } from "./layer/layers.js";
```
1ec5, line 36: I think by convention this would be called index.js, but I’m not sure if that results in anything else happening differently.

1ec5 (commented):
> This will simplify the workflow for anyone wishing to generate a standalone style.json.

Might as well make a new script to facilitate that workflow. Take a look at how taginfo.js is put together and invoked by package.json.

1ec5:
`buildLayers()` returns layers containing labels that are blank until `Label.localizeLayers()` is called. To set a more reasonable initial default of the local-language name(s), replace the empty string in the following lines:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/27a07671a318c114f1dfca7191febbb9543d3789/src/constants/label.js#L125 https://github.com/ZeLonewolf/openstreetmap-americana/blob/27a07671a318c114f1dfca7191febbb9543d3789/src/constants/label.js#L207

with `["get", "name"]`.

src/americana.js
ZeLonewolf: Done

ZeLonewolf:
> `buildLayers()` returns layers containing labels that are blank until `Label.localizeLayers()` is called. To set a more reasonable initial default of the local-language name(s), replace the empty string in the following lines:

This fix seems to break our unit tests.  Perhaps it would make more sense to localize the static file to English (and accordingly, we can generate "style-en.json")?

1ec5:
> This fix seems to break our unit tests. Perhaps it would make more sense to localize the static file to English (and accordingly, we can generate "style-en.json")?

The script could take language codes as an option. Either way, we should update the unit tests whenever there’s new functionality.

scripts/style_json.js
```diff
@@ -0,0 +1,7 @@
+import { copyFile, writeFile, mkdir } from "fs/promises";
+
+import { buildLayers } from "../src/layer/index.js";
+
+export function generateStyleJSON() {
+  writeFile("dist/style.json", JSON.stringify(buildLayers()));
```
1ec5, line 6: This method could call `Label.localizeLayers()`, passing in the contents of [a command line argument](https://nodejs.org/en/knowledge/command-line/how-to-parse-command-line-arguments/) as the locales.

1ec5, line 6: `buildLayers()` only generates the list of layers, not the full style. Any real use case for this JSON would require the full style. Call `buildStyle()` instead, but have that method take a base URL as a parameter instead of hardcoding `window.location`. The base URL can also come from a command line argument.

src/constants/label.js
```diff
@@ -122,7 +122,7 @@ export function localizeLayers(layers, locales) {
 export const localizedName = [
   "let",
   "localizedName",
-  "",
+  ["get", "name"],
```
1ec5, line 125: Change this test to expect “Null Island” instead of the empty string:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/0eb949eaa1618b756ce80a04e8e5ac27c3122492/test/spec/label.js#L243-L252

ZeLonewolf:
Converting to draft until the legends PR is resolved

ZeLonewolf:
I'm going to start over on this project in order to work this in smaller, more manageable chunks.

Issue #569, opened 2022-11-25
Great Lakes Circle Tour


1ec5:
The [Great Lakes Circle Tour](https://en.wikipedia.org/wiki/Great_Lakes_Circle_Tour) is a network of four signposted routes in Illinois, Indiana, Michigan, Minnesota, Ohio, Ontario, Pennsylvania, and Wisconsin, with additional unsignposted route segments in New York State. It is coordinated by the Great Lakes Commission, a U.S. interstate agency. (Ontario is an associate member.)

Each of the four routes has a similar shield consisting of a representation of one of the Great Lakes or its surrounding states encircled by a vaguely circular band in perspective.

[<img src="https://upload.wikimedia.org/wikipedia/commons/2/22/Lake_Erie_Circle_Tour.svg" width="100" alt="Lake Erie Circle Tour">](https://commons.wikimedia.org/wiki/File:Lake_Erie_Circle_Tour.svg) [<img src="https://upload.wikimedia.org/wikipedia/commons/7/77/Lake_Huron_Circle_Tour.svg" width="100" alt="Lake Huron Circle Tour">](https://commons.wikimedia.org/wiki/File:Lake_Huron_Circle_Tour.svg) [<img src="https://upload.wikimedia.org/wikipedia/commons/7/7c/Lake_Michigan_Circle_Tour.svg" width="100" alt="Lake Michigan Circle Tour">](https://commons.wikimedia.org/wiki/File:Lake_Michigan_Circle_Tour.svg) [<img src="https://upload.wikimedia.org/wikipedia/commons/a/a7/Lake_Superior_Circle_Tour.svg" width="100" alt="Lake Superior Circle Tour">](https://commons.wikimedia.org/wiki/File:Lake_Superior_Circle_Tour.svg)

At this style’s usual shield size, these details would easily turn into mush. We’ll need to keep the circular band, since it’s the most recognizable element in the shield, but this will leave very little room for the lake/states if we want to keep them legible.

[...]

Issue #106, opened 2022-01-27
Shields in UK


SK53:
I'm not sure that UK roads truly have shields, but the basic minimum for a reasonably accurate representation is:

* European routes (E-*): these are not signed in the UK and shields are no use for motorists. I'm not sure if the relevant relations are tagged unsigned=yes, and anyway unsigned is probably not supported in OMT
* Motorways should have white lettering on a blue background to the shield. These are pretty common on signposts.
* Trunk roads should have  a yellow lettering on a green background to the shield. Note this conflicts with current E-route rendering.
* Other roads (primary & secondary) should be black letters on a white background. Note virtually all ref on other road classes is a tagging error (these are extremely rarely signed & when they have been the highway authority is not following the guidance) & are best not shown.

A reasonable selection of examples on the [Sabre wiki](https://www.sabre-roads.org.uk/wiki/index.php?title=Direction_Sign).

Note this issue is raised for information, rather than with any expectation that it might be implemented. The only thing which I think is important is the suppression of shields for E-routes, as showing them may affect users' expectations.

ZeLonewolf:
Regarding:
> European routes (E-*): these are not signed in the UK and shields are no use for motorists. I'm not sure if the relevant relations are tagged unsigned=yes, and anyway unsigned is probably not supported in OMT

We have both `ref` and `unsigned_ref` (rather than the barely used `unsigned=ref`) to indicate whether or not a numbered route is signed.  This is how we deal with the situation in the US, where we have both signed and unsigned routes.  This style suppresses shields whenever a `ref` is not present, unless the route is of the type that doesn't have a ref but still has a shield (there are a number of these special cases in the US).

The E route relations that run through the UK are tagged with `ref` and therefore this style displays them as signed.  The EU/UK community should consider this an opportunity to improve the tagging such that unsigned portions of E-road routes are members of relations tagged with `unsigned_ref`.  If this is able to change in the tagging, the E-road shields will disappear from the map on the next update of the underlying vector tiles.

For awareness, the shields displayed by this style are generated solely through route relations and ignore `ref` tagging on ways entirely.  We have been slowly building out shields for mainland Europe as they are displayed on signage and I would consider it in scope to do the same for the UK.

Tasks:

- [ ] Update UK route relations so that unsigned E-roads are members of relations with `unsigned_ref` tagging
- [ ] Add shields for UK roads that are mapped with route relations

- [ ]

ZeLonewolf:
[...]

PR #632, opened 2022-12-22
Generate legend on demand

### LINKED ISSUE #75: Create key/legend for Highway Shields

### LINKED ISSUE #332: Create key/legend for POIs
Similar to #75, it would be useful to be able to show a legend for any POI icons that are currently displayed on the screen.  This is normally a feature of paper maps, so I would consider demonstrating how to do this in the demo map as part of the cartography.  I welcome ideas on how we could do this in our current tech stack.  This has been recently [noted](https://www.openstreetmap.org/user/Cristoffs/diary/399189#comment52494) as a capability gap in the standard tile layer, and is a strong opportunity to demonstrate how this technology could be built.

1ec5:
This PR adds a legend (map key) to the Web application. Click the Legend button in the lower-left corner to reveal a popup illustrating each element of the map style with a description in plain (American) English. The legend’s contents are generated dynamically on demand based on the features visible in the current viewport.

United States
_The legend for the default view shows a bit of everything._

## Motivation

<div align="center"><h3><i>A legend can come from anywhere.</i></h3><i>— Campy coming-of-age film</i></div><br><br>

This project intentionally departs from established conventions among OSM-based maps in favor of established conventions among U.S. map publishers. While this hopefully makes the map more intuitive and approachable to Americans who grew up refolding print maps, the [choices in line style and symbology](https://github.com/ZeLonewolf/openstreetmap-americana/issues/101#issue-1115515924) can be bewildering to those who grew up refolding European maps, or those who have become accustomed to commercial online maps that split the difference between the two cartographic traditions in search of greater marketshare or [platform-level versatility](https://www.justinobeirne.com/cartography-comparison#block-yui_3_17_2_16_1480548256155_17312).

The need for a legend has grown as OSM Americana has become more well-known among the global OSM community. Most mappers are familiar with openstreetmap-carto from using it for both mapping and trip planning. Over the years, openstreetmap-carto has made many design decisions that reflect the European cartographic tradition and the [unique constraints](http://blog.imagico.de/openstreetmap-carto-an-update-on-the-project/) of being a default, well-rounded style for OSM globally.

Coming from that perspective to Americana, mappers may rightly wonder how to tell which road is classified higher than another, or why there appear to be restricted-access roads in unexpected places. Without firsthand experience on U.S. roads, someone from [a country with a more basic route marking system](https://github.com/ZeLonewolf/openstreetmap-americana/issues/106#issuecomment-1024132836) would [look at our menagerie of route shields](https://lists.openstreetmap.org/pipermail/tagging/2022-January/063544.html) and [never be able to figure it out on their own](https://github.com/gravitystorm/openstreetmap-carto/issues/508#issuecomment-45408961).

This project reintroduces American cartographic practices to the OSM community, ensuring that the OSM database accommodates the technical needs of anyone who wishes to reproduce this approach in their own projects. The community is well aware of the need to [avoid overreliance](https://wiki.openstreetmap.org/wiki/Tagging_for_the_renderer) on a particular renderer’s approach, but it’s [easy to develop bad habits](https://community.openstreetmap.org/t/multiple-delimited-names-in-the-name-tag/6803) when the results of those habits aren’t apparent. Just as importantly, the nomenclature associated with various tags also needs to be translated for a lay audience so that the community can connect tagging discussions to a realistic end user experience. A legend can be a powerful tool for reframing discourse around the question, “Which distinctions really matter?”

The Web application already integrates [OpenMapSamples](https://github.com/adamfranco/OpenMapSamples/), but its output is more appropriate for [proofing](https://en.wikipedia.org/wiki/Prepress_proofing) a change during the development process. Educating the end user about the style requires a UI that’s more compact and focused so that they can compare the illustrations to the actual map.

## Architecture and implementation

As a form of user documentation, legends can be challenging to maintain. You need to carefully consider what to illustrate and how much detail to provide, and of course the illustrations need to visually match the map. In the past, various OSM-based renderers have attempted to somehow automate the creation of a legend: https://github.com/ZeLonewolf/openstreetmap-americana/issues/332#issuecomment-1362466431. However, these tools still require a nontrivial amount of curation, leading some legends to become outdated easily: openstreetmap/openstreetmap-website#1307.

The Americana style is still in its infancy and subject to continual change, so any legend needs to be as data-driven as possible to remain relevant and accurate. Therefore, the legend’s configuration is written in JavaScript and runs as part of building the style. The configuration is located alongside the layer definitions and directly references the various layers that constitute the style. If you change a layer and neglect to update the configuration, the application will most likely crash on launch so you won’t overlook the problem.

The configuration isn’t derived directly from the layer definitions, because there isn’t a straightforward one-to-one mapping between layers and the concepts that a user needs to know. For example, rendering each kind of railroad track requires two layers, one of which is common to all the kinds of railroad tracks. Many layers implement data-driven styling for performance and maintainability, so multiple legend entries may draw from the same set of layers. A few layers, such as one-way arrows, have been omitted from the legend, because the meaning should be eminently obvious to most users. Most bridge and tunnel variations have been omitted for brevity, except on a few features like subways and monorails that routinely traverse one of these structures.

For each feature type in the configuration, the legend picks out a representative feature that’s currently visible on the map and generates an illustration based on the associated layers’ evaluated paint and layout properties. GL JS also has the capability to pick out features from the vector tiles that have loaded but aren’t visible. However, it doesn’t evaluate the style properties for these features, so we wouldn’t be able to automatically derive an illustration. Moreover, showing all feature types from all zoom levels would result in a lengthy, chaotic legend with many contradictions because of how layers can vary by zoom level. The popup can take a second or two to appear because it synchronously queries the map for features once per entry. Initially, I combined all the entries into a single query, but there was too much crosstalk among entries that happen to share layers and it was challenging to filter the results effectively.

z0 | z1 | z2 | z3
----|----|----|----
z0 | z1 | z2 | z3

z4 | z5 | z6 | z7
----|----|----|----
z4 | z5 | z6 | z7

z8 | z9 | z10 | z11
----|----|----|----
z8 | z9 | z10 | z11

z12 | z13 | z14 | z15
----|----|----|----
z12 | z13 | z14 | z16

Most illustrations are implemented as styled HTML elements. However, line features are implemented as small embedded SVG documents because the CSS profile for HTML lacks a straightforward way to customize dashed lines. So far, I’ve only implemented the paint and layout properties that this style is using. If a future PR starts using another property, it’ll need to be implemented as part of the legend control.

In most sections, the legend descriptions are hard-coded as part of the application. Every style makes different distinctions requiring different descriptions, so we can’t source the descriptions from, say, id-tagging-schema or the OSM Wiki’s data items. The hard-coded descriptions follow American English to a fault. I tried to capture the essence of how a print map would describe a particular entry if a cartographer had arrived at a definition equivalent to the OSM tag’s actual definition by accident. The highway classifications posed a particular challenge. Print maps are all over the place, but I think the typical functional classification terminology is common enough and close enough to OSM tagging to work here.

The “Route markers” section’s descriptions are fetched dynamically from Wikidata. The first time the user opens the legend after loading the page or changing the preferred language, the legend queries Wikidata for all the items [associated with](https://www.wikidata.org/wiki/Property:P1282) OSM’s `network=*` tags and populates the section with these items’ labels in the user’s preferred language. The shields are grouped by country, with the country names coming from the browser.

QEW<br><i>The “Route markers” section includes any visible variants of a standard shield.</i>

LECT<br><i>Any `network` value outside of a country namespace, such as `GLCT` for the Great Lakes Circle Tour (#569), goes into an “International” section.</i>

MX:MX<br><i>Country subsections and route network descriptions are translated into the user’s preferred language to the extent possible. If no description is available, the legend falls back to the raw `network` value.</i>

The legend is contained inside a standard GL JS popup. Normally a popup is tied to a map marker, but in this case, the Legend button’s location on screen is converted to a geographic location and the popup is anchored to that location. To prevent the popup from becoming visually detached from the button, changing the map camera automatically closes the popup. If we want a more persistent popup, we’d need to implement our own popup UI and style it consistently with the map’s other controls.

## Alternatives considered

Many GL JS–based maps integrate the [mapbox-gl-inspect](https://github.com/lukasmartinelli/mapbox-gl-inspect/) or [maplibre-gl-inspect](https://github.com/acalcutt/maplibre-gl-inspect/) plugin, which shows an individual feature’s raw properties on hover. This plugin is very useful for style development but not as meaningful to end users, who wouldn’t know what to make of raw OpenMapTiles keys and values. (They don’t straightforwardly align with OSM tags either.) Moreover, describing a particular feature doesn’t necessarily help the user figure out the distinctions that the style is making between different features. A legend gives a better lay of the land, directing the user’s attention to the elements we consider most important to the style.

GL JS is capable of taking a screenshot of the feature itself, analogous to how [mapnik-legendary](https://github.com/gravitystorm/mapnik-legendary/) works. However, this results in legend entries that are too large for an in-map popup and not as easy to comprehend at a glance. We don’t really need to reproduce the representative feature’s geometry, only the styling.

This project has [a nice writeup on the OSM Wiki](https://wiki.openstreetmap.org/wiki/OpenStreetMap_Americana) (if I may say so myself). We could double down on documenting the style on the wiki, putting the project out there where mappers cannot but consider it in tagging discussions. We could even [maintain our own wiki](https://imagico.de/blog/en/introducing-styleinfo/) to explain OSM data from the perspective of this style. However, I don’t think anything so voluminous would satisfy the basic user requirement of a visual guide to the map in front of them.

## Future improvements

In the future, we could generalize the legend’s support code a bit and extract it out as a plugin for Mapbox/MapLibre GL JS, as with OpenMapSamples. However, I think that should be a later step once the code stabilizes and we see how it holds up to new additions to the style.

We should make legend descriptions localizable, just like in any other user-facing application. I prototyped a rudimentary solution for localization in https://github.com/ZeLonewolf/openstreetmap-americana/pull/215#issuecomment-1060067427, but this PR introduces enough new UI strings that we’d want to shunt the translation table into a separate JSON file.

Once enough `network=*` tags are tagged in Wikidata to bog down the SPARQL query, we should move the query to a build step and check in the results. Running the query on demand gives mappers instant gratification as they add labels to Wikidata, but it’ll be increasingly wasteful for every user to run this query once per session.

This legend unlocks some possibilities for interactivity, such as [highlighting features on the map](https://osmus.slack.com/archives/C01V02K52UX/p1670622529555479?thread_ts=1670622077.199789&cid=C01V02K52UX) when you hover over a legend entry. Perhaps a checkbox beside each legend entry could toggle the relevant layers on and off: #25. #76 tracks making an individual map feature clickable to reveal more information about the feature, though that would generally require more information in the vector tiles.

----

Fixes #75 and fixes #332 to the extent that we have any POIs to define so far.


ZeLonewolf:
It looks awesome!  I have no idea how any of this work but I tested it and it seems to be functioning exactly as advertised

ZeLonewolf:
I noticed that countries with lower-case versions of the country code don't get converted to names.  I'm thinking we should do a case conversion.

image

1ec5:
> I noticed that countries with lower-case versions of the country code don't get converted to names. I'm thinking we should do a case conversion.

Fixed:

[Slovakia](https://zelonewolf.github.io/openstreetmap-americana/#map=6.86/48.574/19.007&language=en)

ZeLonewolf (approved):

src/js/legend_control.js
```diff
@@ -0,0 +1,593 @@
...
+  }
+
+  /**
+   * Returns the Wikidata Query Service SPARQL query for network metadata.
+   */
+  getNetworkMetadataQuery() {
+    let locales = Label.getLocales().join(",");
+    return `
+SELECT ?value ?network ?networkLabel WHERE {
+  ?network wdt:P1282 ?tag.
```
1ec5, line 585: I considered limiting this query to instances of [highway system (Q25631158)](https://www.wikidata.org/wiki/Q25631158) or [road (Q34442)](https://www.wikidata.org/wiki/Q34442), to avoid confusion with list articles about these networks and also filter out all the `network=*` values about railway and public transportation networks. However, I suspect that would make the query take longer and it’d be wasteful. Maybe once we move this query to a build step, we can make it more sophisticated and targeted.

src/js/legend_control.js
```diff
@@ -0,0 +1,651 @@
...
+   * @param region ISO 3166-1 alpha-2 country code.
+   */
+  getNetworkMetadataQuery(region) {
+    let locales = Label.getLocales().join(",");
+    let triple,
+      filter = "",
+      bind = "";
+    if (region === "GB") {
+      triple =
+        "?network wdt:P361 wd:Q115856945; p:P528 [ ps:P528 ?value; pq:P972 wd:Q110613756 ]";
```
1ec5, line 635: In support of #633, this is an extra query for [parts of](https://www.wikidata.org/wiki/Property:P361) the [UK road coloring scheme](https://www.wikidata.org/wiki/Q115856945) that’s requested whenever the synthesized UK “networks” are listed in the table.

michaelblyons:
This is spectacular!

If may, I'd like to lobby mildly for left-alignment of (the rest of) the legend items. Populated places, Aviation markers, and the country headings when multiple routes are visible.

I recognize that many print-map legends have the neat row of icons down the center with a sample label to the left and the description to the right, but I respectfully assert that it's harder to read than a table with left-alignment. Eyes are trained to return to a left margin when reading or scanning and mismatched left edges detract from that.

I do genuinely think the dynamic legend is incredible. Please pardon my nitpicking.

1ec5:
> If may, I'd like to lobby mildly for left-alignment of (the rest of) the legend items. Populated places, Aviation markers, and the country headings when multiple routes are visible.
> 
> I recognize that many print-map legends have the neat row of icons down the center with a sample label to the left and the description to the right, but I respectfully assert that it's harder to read than a table with left-alignment. Eyes are trained to return to a left margin when reading or scanning and mismatched left edges detract from that.

Thanks for pointing this out. The current placement does result in some ragged left edges. In earlier iterations of this design, I experimented with a left-aligned column, but it often made it more difficult to associate a given icon and label with a description, especially in the “Populated places” section where some labels would be a lot shorter or taller than others. (The screenshots below are from before I implemented support for the `text-max-width` property, which currently serves as a lower bound on the width as well, due to the vagaries of CSS table layout.)

 Arlington Ratisbona

Some legends get around this problem by adding a horizontal rule between each row or a leader line from the icon or label to the description. But those legends tend to have a lot more text or right-align the description column.

Most symbol layers’ labels aren’t that interesting for a legend. It’s only really necessary to include labels that stand on their own, independent of icons, lines, or fills – for example, countries and highway exits. However, for populated places, we’re reusing the same dot icon in multiple entries, varying only the size. The difference in label font size makes this distinction a lot clearer. The capital city entries are mainly about the icons rather than the labels, but if your viewport happens to contain only a country and a capital city, the resulting table looks really imbalanced. Once we start adding layers for POIs (#387), which tend to have much longer labels, we should probably hide nonessential labels from some the POI entries and the ones under “Aviation”.

zekefarwell (approved):
I can't say I fully understand how this all works either, but it seems good to me.  Fantastic work, Minh!  Are we waiting for any more tweaks before merging?  There are definitely plenty of things we can iteratively improve moving forward, but I see no problem with those happening in future PRs.

ZeLonewolf:
Are there other open source maps with dynamic legends like this?  If this is a "first" let's make sure it's well advertised. I teased it in my year in review but it's worth special promotion every time we push the envelope on what open source digital maps can do.

1ec5:
> Are we waiting for any more tweaks before merging?

No blockers, but I’ve been taking some time to clean up the code a little. That’s an open-ended task. I’ve also been trying to add some Wikidata entries for networks in different regions so that folks don’t have to start from scratch when they just want to translate the legend entries.

I had hoped to include some automated tests as part of this PR, but I’m not finding many opportunities because so much of it is tied to map data (mockable but difficult to maintain) or the DOM (not clearly worth mocking).

ZeLonewolf:
Do we need a guide on wikidata editing for route relation networks?  Either here or elsewhere?  Or do we see that as more of a one time task?

1ec5:
This project’s wiki would be a decent place for a guide. We could link to it from the [FAQ](https://github.com/ZeLonewolf/openstreetmap-americana/wiki/Frequently-asked-questions), which is still incomplete. There’s a long tail of networks. I could automate the creation of stub Wikidata items for the more obscure networks’ items based on the shield definitions, but that can be tail work.

ZeLonewolf:
I'm just thinking we want to head off the kind of thing that I did, which was to immediately edit wikidata wrongly to add missing network information.

1ec5:
> Are there other open source maps with dynamic legends like this?

I haven’t looked thoroughly enough to say for sure that this legend is the first of its kind. In the “Alternatives considered” section above, I pointed out some prior art that’s less interactive and some prior art that’s even more interactive (but less usefully so). This PR takes an approach befitting the maps we’re using as design inspiration, so I wouldn’t consider it a baseline expectation for every map as with route shields. Still, I think it is a usability improvement over these other approaches.

src/js/legend_control.js
```diff
@@ -0,0 +1,651 @@
...
+   * Returns a table row representing a route shield.
+   *
+   * @param network The `network=*` value associated with the style images.
+   * @param names An array of style image names.
+   * @returns An HTML table row representing the route shield, or nothing if the style does not render the given network.
+   */
+  getShieldRow(network, names) {
+    // On recreational route relations, network=* indicates the network's scope, not the network itself.
+    // https://github.com/ZeLonewolf/openstreetmap-americana/issues/94
+    if (/^[lrni][chimpw]n$/.test(network)) return;
```
1ec5, line 483: This will be unnecessary once #635 lands.

1ec5:
> I'm just thinking we want to head off the kind of thing that I did, which was to immediately edit wikidata wrongly to add missing network information.

Once we have some documentation on the project wiki, then we can encourage folks to create new Wikidata items about networks or assign existing items to `network` tags. In the meantime, it’s safe enough to add translations to Wikidata items about that have already been tagged with `network` tags. This is already largely done for English, but speakers of other languages can still benefit from translated labels. They’ll see an English fallback until their language has its own label on Wikidata for a given network.

1ec5:
I’ve restored the link from each network to the respective Wikidata item (but with a more subtle color now). If Wikidata has a label in a fallback language but not your most preferred language, a little language indicator appears off to the side, just like on Wikidata.

fallbacks

1ec5:
> I'm just thinking we want to head off the kind of thing that I did, which was to immediately edit wikidata wrongly to add missing network information.

The “Route markers” section gives a lot of exposure to Wikidata’s items about route networks (what it calls “highway systems”). This is good in that it’ll encourage translations that we probably wouldn’t otherwise get for a niche project. However, editing Wikidata directly requires some care, or at least some circumspection. The data model is more nuanced than OSM, but the existing data is often less nuanced due to imports from Wikipedia, ill-considered merges of items, and subsequent bots running rampant.

If a user sees a legend entry described by only the raw `network` value where a human-readable description should go, ideally they’d open an issue in this repository and we’d take care of setting up Wikidata for translation, by making sure there’s an appropriate item tagged with its corresponding `network` tag. However, it’s just as likely that they’d search Wikipedia for an article about the network, land on a _list_ article, go to the linked Wikidata item, and change the label to say what the legend should say. That would be incorrect because the item is about a list of routes, not the network itself. This would be confusing because the rest of the item would continue to be modeled as if it’s a list.

To mitigate this confusion, I spent the last two weeks cleaning up hundreds of Wikidata items, unmerging items where necessary and adding statements about shields and number formats to clarify what each item is supposed to represent. I’ve taken care of most of the national networks and many of the regional route networks supported by this project, as well as some that are not yet supported but obviously established in OSM. There are still a lot of supported networks lacking Wikidata items at the regional and local level, but we can fill those out over time. With these edits in place, I feel a lot more confident about the user experience for contributing translations.

Along the way, I discovered that a number of shields, particularly in Europe, were implemented based on assumptions that appear to be unreliable. I’ve opened a separate issue about removing or reviewing each of these shields.

ZeLonewolf:
Is this a regression or a TO-DO?

image

1ec5:
The labels look fine on my end:

[English Channel](https://zelonewolf.github.io/openstreetmap-americana/#map=9.57/51.0134/1.4775&language=en)

Please open a new issue, especially if you see anything amiss in the console.

ZeLonewolf:
Looks like it just takes a bit to AJAX in the label sometimes.

PR #633, opened 2022-12-23
Great Britain Motorway / Trunk shields

ZeLonewolf:
Partial implementation of #106

This PR implements motorway and trunk shields on Great Britain. Route network data is provided by OpenMapTiles, which uses geometry processing to create a pseudo-network value for motorways (M-routes, highway=motorway) and trunk roads (A-routes, highway=trunk).
 
image

The pseudo-networks of `omt-gb-trunk` and `omt-gb-motorway` are created by OpenMapTiles:

image

Additional work is needed to add primary and/or secondary routes as well as handle the problem described in openmaptiles/openmaptiles#1462


claysmalley (approved):
Looks good to me.

A bit ironic that OpenStreetMap's country of origin becomes a special case with regards to route shields. But I don't see any other way to handle countries where signage depends on the underlying route classification.

PR #635, opened 2022-12-26
Suppress shields for recreational TLA networks

### LINKED ISSUE #631: shield shown for trail that is invisible
The MMT shown on the screenshot, taken [at this location](https://zelonewolf.github.io/openstreetmap-americana/#map=15.33/39.469642/-76.588534) on 2022-12-22 is what I presume [a shield for this Merryman's Mill Trail on OSM](https://www.openstreetmap.org/relation/11117036). I think that if we don't show paths and tracks on Americana, then the refs for them should be hidden too.
image

1ec5:
Reject three-letter-acronym recreational `network` values when deciding whether to render a shield. Also added unit tests for `network` and `ref` validation.

[MMT](https://zelonewolf.github.io/openstreetmap-americana/#map=15/39.46359/-76.58835)

Fixes #631.

### DIFF
--- a/src/js/shield.js
+++ b/src/js/shield.js
@@ -62,11 +62,15 @@ function compoundShieldSize(dimension, bannerCount) {
   };
 }
 
-function isValidRef(ref) {
-  if (ref == null || ref.length == 0 || ref.length > 6) {
-    return false;
-  }
-  return true;
+export function isValidNetwork(network) {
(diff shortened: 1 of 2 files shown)


ZeLonewolf (approved):
Does not seem to impact render or cause any console error messages.  We may want to render trail symbology at some point, but I anticipate having to make a separate layer for that.

PR #637, opened 2022-12-27
Intermittent water fix and refinements

### LINKED ISSUE #634: Intermittent water outline not applying to all intermittent water areas
Intermittent water areas are supposed to look like this:

Screen Shot 2022-12-26 at 12 55 23 PM

https://zelonewolf.github.io/openstreetmap-americana/#map=12.93/36.06253/-114.47416

But some look like this (this one is a river):

before

https://zelonewolf.github.io/openstreetmap-americana/#map=14.78/37.22049/-110.61003

zekefarwell:
This PR is a collection of a few fixes and refinements to intermittent water styling.  I fixed #634 where a solid water line instead of a dashed one was rendering for intermittent river areas.  I slightly increased the intermittent water fill saturation for just a bit more distinction from the green park fill.  I explored moving to an opaque fill but didn't come up with anything satisfactory so left it translucent for now.  A pattern with partial transparency could be a future option.  I made the the dashed outline thinner and darker.  I find this better matches the hairline used for the regular water outline.  I also lightened intermittent waterway lines at high zoom so they match the regular waterway line color and blend in a bit better with intermittent water areas that may be drawn around them.  I'd love to hide them entirely under the areas but the area translucency prevents this.

Before:
Screen Shot 2022-12-26 at 11 43 18 PM
After:
Screen Shot 2022-12-26 at 11 43 26 PM

Before:
Screen Shot 2022-12-26 at 11 45 28 PM
After:
Screen Shot 2022-12-26 at 11 45 24 PM

Before:
Screen Shot 2022-12-27 at 12 02 53 AM
After:
Screen Shot 2022-12-27 at 12 02 49 AM

### DIFF
--- a/src/layer/water.js
+++ b/src/layer/water.js
@@ -87,15 +87,19 @@ export const waterLine = {
   id: "water_line",
   type: "line",
   filter: [
+    "all",
+    ["!=", ["get", "intermittent"], 1],
+    [
     "match",
     ["get", "class"],
-    bigRivers,
-    [">=", ["zoom"], 8],
-    mediumRivers,
-    [">=", ["zoom"], 16],
-    "lake",
-    ["all", ["!=", ["get", "intermittent"], 1], [">=", ["zoom"], 8]],
+      bigRivers,
+      [">=", ["zoom"], 8],
+      mediumRivers,
+      [">=", ["zoom"], 16],
+      "lake",
+      [">=", ["zoom"], 8],
     true,
+  ],
   ],
   paint: {
     "line-color": Color.waterLineBold,
@@ -112,11 +116,7 @@ export const waterLineIntermittent = {
   id: "water_line_intermittent",
   type: "line",
   minzoom: 8,
-  filter: [
-    "all",
-    ["==", ["get", "intermittent"], 1],
-    ["==", ["get", "class"], "lake"],
-  ],
+  filter: ["all", ["==", ["get", "intermittent"], 1]],
   paint: {
     "line-color": Color.waterLine,
     "line-dasharray": [6, 4],
--- a/src/layer/water.js
+++ b/src/layer/water.js
@@ -66,16 +66,15 @@ export const water = {
   id: "water",
   type: "fill",
   paint: {
-    "fill-color": Color.waterFill,
-    "fill-opacity": [
+    "fill-color": [
       "case",
       [
         "any",
         ["==", ["get", "intermittent"], 1],
         ["==", ["get", "brunnel"], "tunnel"],
       ],
-      0.3,
-      1,
+      Color.waterIntermittentFill,
+      Color.waterFill,
     ],
     "fill-outline-color": Color.waterFillTranslucent,
   },
@@ -90,16 +89,16 @@ export const waterLine = {
     "all",
     ["!=", ["get", "intermittent"], 1],
     [
-    "match",
-    ["get", "class"],
+      "match",
+      ["get", "class"],
       bigRivers,
       [">=", ["zoom"], 8],
       mediumRivers,
       [">=", ["zoom"], 16],
       "lake",
       [">=", ["zoom"], 8],
-    true,
-  ],
+      true,
+    ],
   ],
   paint: {
     "line-color": Color.waterLineBold,
--- a/src/layer/water.js
+++ b/src/layer/water.js
@@ -117,8 +117,9 @@ export const waterLineIntermittent = {
   minzoom: 8,
   filter: ["all", ["==", ["get", "intermittent"], 1]],
   paint: {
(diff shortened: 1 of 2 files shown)


ZeLonewolf:
Here is my favorite test area in Lake Powell [Localhost link](http://localhost:1776/#map=13/37.24181/-110.50826), so you can see both regular and intermittent water areas:

image

ZeLonewolf:
I have to say, I was not aware that intermittent river lines were visible _above_ river water areas. If there's an intermittent river area mapped, I think it should cover the river line.

1ec5:
> I have to say, I was not aware that intermittent river lines were visible above river water areas. If there's an intermittent river area mapped, I think it should cover the river line.

This is a consequence of making intermittent water areas translucent. We can’t hide anything under something that’s translucent.

ZeLonewolf:
In that case we should probably make them opaque?

zekefarwell:
> In that case we should probably make them opaque?

I did try making them opaque and it looked great over the neutral background color.

santa-ata-opaque

However, it looked totally weird over the green park fill.

mead-opaque


Translucent shot of the same location for comparison.

mead-translucent


Maybe it would be possible to find the perfect shade of light bluish-greenish-off-whitish that would look right against both the neutral background and the park green.  But as we add more background fill colors the more difficult the problem becomes.  This is definitely an area worthy of further exploration, but it probably makes sense to do that once we have more background fills to test against.

src/layer/water.js
```diff
@@ -112,14 +110,11 @@ export const waterLineIntermittent = {
...
-    ["==", ["get", "intermittent"], 1],
-    ["==", ["get", "class"], "lake"],
-  ],
+  filter: ["all", ["==", ["get", "intermittent"], 1]],
   paint: {
-    "line-color": Color.waterLine,
-    "line-dasharray": [6, 4],
+    "line-color": Color.waterIntermittentOutline,
+    "line-dasharray": [10, 6],
+    "line-width": 0.5,
```
1ec5, line 117: Making the line half as thick causes the dash array to shrink by half, so you’ve compensated by making the dashes a little longer. Did you explore making the dashes a little longer, to fully account for the shrinkage? This could also help make the line more legible since the color is lighter than before.

1ec5 (commented):
Another benefit of an opaque fill is that a seam wouldn’t appear between two `natural=water` `intermittent=yes` areas as it does today. But I agree that we can continue to refine this aspect over time.

zekefarwell:
Lake Powell and Lake Mead good test locations.  Here are some others I've been using.  Lots of good ones in southern California.
http://localhost:1776/#map=11.88/33.91273/-117.58252
http://localhost:1776/#map=13.31/33.83906/-117.84845
http://localhost:1776/#map=15.74/33.771069/-117.889255
http://localhost:1776/#map=13.43/34.27463/-116.80625

src/layer/water.js
zekefarwell: Yes I originally set the dash array to [12, 8] to fully compensate for the thickness change.  This works well at integer zoom levels:

[12, 8] at Zoom 15.0:
z15

[12, 8] at Zoom 16.0
z16

At fractional zoom levels MapLibre stretches the dash array out and it looks significantly different.  The longer dashes and gaps work less well I find.

[12, 8] at Zoom 15.5:
z15 5

[12, 8] at Zoom 15.9:
z15 9


I changed the dash array to [10, 6] in an attempt to make the dashes and gaps at the high end of the fractional zoom levels not quite so long.  Of course, this makes the them a bit too short at the low end.  In the middle they look just right.

[10, 6] at Zoom 15.9:
10,6  at z15 9

[10, 6] at Zoom 15.5:
10,6  at z15 5

[10, 6] at Zoom 15.0:
10,6  at z15


Honestly, I'm not very satisfied with this change.  Just due to the way dash arrays are rendered in fractional zoom levels, they will look wrong some of the time.   We can have them look right at the low end and significantly too long at the high end, or right in the middle by too short at the low end and too long a the high end.  Not sure which is least bad.  If you think the former is better, I'm happy to go with that and switch back to [12, 8].

zekefarwell: > This could also help make the line more legible since the color is lighter than before

My intent was to make the line carry more visual weight, despite being thinner, by darkening and saturating the color.  If you're perceiving it as lighter, perhaps there is more tweaking to do.  It is quite a challenge though, because the line is partially underneath the translucent fill layer.  That lightens whatever line color is chosen somewhat.

1ec5: I hadn’t considered fractional zoom levels. This is yet another thing that mapbox/mapbox-gl-js#12326 fixed… I agree with the compromise you implemented.

1ec5 (approved):

PR #640, opened 2022-12-28
Distinguish Greek national roads from motorways

1ec5:
Greek national roads have a different shield than Greek motorways: white on a blue rectangle. #139 probably missed this distinction because the two main motorways were incorrectly tagged as `network=GR:national` until [today](https://www.openstreetmap.org/changeset/130621801).

[Selianitika](https://commons.wikimedia.org/wiki/File:Selianitika_Iterchange_Olympia_Odos.jpg)

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -3326,12 +3326,18 @@ export function loadShields(shieldImages) {
   );
 
   // Greece
-  shields["GR:motorway"] = shields["GR:national"] = hexagonVerticalShield(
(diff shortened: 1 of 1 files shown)


ZeLonewolf:
Needs a tile update to test

1ec5:
A tile update would only test [this changeset](https://www.openstreetmap.org/changeset/130621801). Other than the two previously mistagged routes, the rest of Greece is up-to-date.

ZeLonewolf (approved):
Looks good as far as I can tell.  I'm seeing all three styles of route.

image

PR #644, opened 2022-12-30
Draw octagon shields mathematically

claysmalley:
Replaces octagon shield SVG files with a draw function.

Screenshot from 2022-12-30 16-08-18


michaelblyons:
Is this one of the shapes that should have "vertical" in its name, lest there be a 🛑 shaped one where the top and bottom are flat?

1ec5:
That might be a good idea. We already have an example of that shape in Indonesia, though that will have to remain image-backed because of the colored border along the top edge.

1ec5 (approved):

ZeLonewolf:
We also might one day like to render stop signs, perhaps with localized text!

PR #645, opened 2023-01-02
Render water labels before protected area labels

ZeLonewolf:
Currently, protected area labels are rendered with higher importance than water bodies.  This tends to result in less-important protected areas winning in a collision with more-important water bodies.  Additionally, protected area labels have more flexibility to reposition themselves compared to water bodies, so water body names should be rendered earlier in the layout.

Before/After (http://localhost:1776/#map=5/25.86/-108.52):

image
image

Before/After (http://localhost:1776/#map=5/8.64/120.05):

image
image

Before/After (http://localhost:1776/#map=5/-10.52/132.77):

image
image


claysmalley (approved):
Looks good to me.

ZeLonewolf:
Thanks for reviewing my PR in a conflict state :joy:

PR #647, opened 2023-01-02
Restore green hexagon shield to test page

claysmalley:
#640 changed the Greek national road shield to a blue rectangle. We were using this network as an example for green vertical hexagons, so the example is changed to `GR:motorway`.


ZeLonewolf (approved):
Sounds good, thanks!

PR #648, opened 2023-01-02
Replace pentagon shield icons with draw function

claysmalley:
This replaces the function that draws home plate shields with a more versatile pentagon function with more parameters. The shield definition functions in `shield_defs.js` are also restructured to avoid having to pass negative values to turn them upside-down.

Screenshot from 2023-01-02 15-20-49

In the wild (Englishtown, NJ):

Screenshot from 2023-01-02 15-01-41

Also featuring a more accurate rendition of the Hungarian motorway shield:

Screenshot from 2023-01-02 15-00-10


ZeLonewolf:
I believe these pentagons are now wider:

image

I don't know if that was intended or a happy accident, but they're so much easier to read now.

ZeLonewolf:
However, the Hungary ones seem to stay the same size:

image

Wondering if horizontal expansion should be a standard thing or if we did something special for pentagons because they squeeze the text so much.

claysmalley:
The wider pentagons were indeed intentional, as was the fixed-width Hungarian shield. Horizontal expansion is already kind of standard practice, as long as it doesn't make the shield less recognizable. Part of the motivation for developing these draw functions is to make variable-width shields easier without having to maintain a library of icon files.

src/js/shield_canvas_draw.js
```diff
@@ -420,9 +442,9 @@ export function hexagonVertical(
   let y0 = lineWidth;
   let y5 = CS - lineWidth;
 
-  let x1 = x0 + drawRadius;
+  // let x1 = x0 + drawRadius;
```
ZeLonewolf, line 445: Did you mean to leave these two comments?

src/js/shield_canvas_draw.js
```diff
@@ -189,16 +190,15 @@ export function trapezoid(
...
+  let angleSign = shortSideUp ? -1 : 1;
+  let sine = Math.sin(angle);
+  let cosine = Math.cos(angle);
+  let tangent = Math.tan(angle);
 
   if (rectWidth == null) {
     var shieldWidth =
       ShieldText.calculateTextWidth(ref, genericShieldFontSize) +
-      (2 + (CS * sine) / 2) * PXR;
+      (2 + ((CS / PXR) * tangent) / 2) * PXR;
```
ZeLonewolf, line 201: If I did this right, this simplifies to

```
2*PXR + CS * tangent / 2
```
Does that expression make sense for what this code is doing?  If so, that should be a bit cleaner to read without the parentheses and repeated terms PXR and 2.

src/js/shield_canvas_draw.js
```diff
@@ -215,22 +215,23 @@ export function trapezoid(
...
 
   let x1 = x0 + (y1 - y0) * tangent;
   let x2 = x1 + drawRadius * cosine;
-  let x3 = x0 + (y2 - y0) * tangent;
-  let x4 = x0 + (y3 - y0) * tangent;
-  let x5 = x3 + drawRadius * cosine;
+  let x3 = x0 + angleSign * (y2 - y0) * tangent;
+  let x4 = x0 + angleSign * (y3 - y0) * tangent;
+  let x5 = x3 + angleSign * drawRadius * cosine;
   // let x6 = width - x5;
```
ZeLonewolf, line 229: These were already here but is there a need for the commented code? and also x10 below.

src/js/shield_canvas_draw.js
```diff
@@ -317,18 +318,27 @@ export function diamond(fill, outline, ref, radius, outlineWidth, rectWidth) {
...
+  let angleSign = pointUp ? -1 : 1;
+  let sine = Math.sin(angle);
+  let cosine = Math.cos(angle);
+  let tangent = Math.tan(angle);
+
   if (rectWidth == null) {
     var shieldWidth =
-      ShieldText.calculateTextWidth(ref, genericShieldFontSize) + 2 * PXR;
+      ShieldText.calculateTextWidth(ref, genericShieldFontSize) +
+      (2 + ((CS / PXR - offset) * tangent) / 2) * PXR;
```
ZeLonewolf, line 341: Same simplification as further up.  Given the same formula, wondering if there's a common aspect that can be refactored out.

src/js/shield_canvas_draw.js
```diff
@@ -341,39 +351,51 @@ export function homePlate(
...
+  // let x2 = x0 + drawRadius1 * halfComplementTangent1 * offsetCosine;
+  let x5 = x4 + drawRadius2 * halfComplementTangent2;
+  let x7 = x8 - drawRadius2 * halfComplementTangent2;
+  // let x10 = x12 - drawRadius1 * halfComplementTangent1 * offsetCosine;
+  let x11 = x12 - angleSign * drawRadius1 * halfComplementTangent1 * sine;
+  let y2 = y3 - angleSign * drawRadius1 * halfComplementTangent1 * cosine;
+  // let y4 = y3 + angleSign * drawRadius1 * halfComplementTangent1 * offsetSine;
+
+  // let x3 = x5 - angleSign * drawRadius2 * sine;
+  // let x9 = x7 + angleSign * drawRadius2 * sine;
```
ZeLonewolf, line 390: Lots of commented code here?

src/js/shield_canvas_draw.js
claysmalley: Makes sense, fixed.

src/js/shield_canvas_draw.js
claysmalley: Fixed.

src/js/shield_canvas_draw.js
claysmalley: Technically not needed, but might be useful to whoever is reading this in the future to see the control points I calculated but ended up not needing. If I get rid of these commented lines, should I decrement the following variable names so no numbers are skipped?

ZeLonewolf: Whatever you think is the most consistent/easy to follow?  In general, commented-out code is a code smell so if there is any there should be a reason.

claysmalley: Commented lines of code removed; variable names decremented.

src/js/shield_canvas_draw.js
claysmalley: Commented lines of code removed; variable names decremented.

src/js/shield_canvas_draw.js
claysmalley: Commented lines of code removed; variable names decremented.

ZeLonewolf (approved):
Everything seems to be in working order.

image

One thing we should revisit later is whether the county route pentagons are too prominent when compared to other networks like Interstates.

PR #662, opened 2023-01-03
Refactor layers ordering code to the layers folder

ZeLonewolf:
This change is a first increment towards improving modularity in the project and is eventually aimed at creating a style.json artifact that can be published as part of the build process.

This PR separates the code that sets the ordering of layers into `src/layer/index.js`.  This improves separation between the layer code in `layers` from the demo map code in `americana.js`, hopefully making the code a bit easier to navigate.  I also updated the documentation to reflect this new organization.

CC/#68


src/americana.js
```diff
@@ -36,210 +21,7 @@ import SampleControl from "openmapsamples-maplibre/OpenMapSamplesControl.js";
...
-    lyrPlace.village,
-    lyrPlace.town,
-    lyrPlace.city,
-    lyrPlace.countryOther,
-    lyrPlace.country3,
-    lyrPlace.country2,
-    lyrPlace.country1,
-    lyrPlace.continent
-  );
+  let layers = Layers.build();
```
1ec5, line 24: This is fine, I suppose, but if the goal is to generate a style JSON file headlessly, then we need to extract not only the surrounding `buildLayers()` function but also `buildStyle()` into a separate file.

src/americana.js
```diff
@@ -36,210 +21,7 @@ import SampleControl from "openmapsamples-maplibre/OpenMapSamplesControl.js";
...
 
   Label.localizeLayers(layers, Label.getLocales());
```
1ec5, line 26: The reason you omitted this statement from `build()` is that `Label.getLocales()` requires access to `navigator`, which is unavailable in a command line environment. So have `buildLayers()` accept the locales as an argument and move this method call to wherever `buildLayers()` is being called.

src/americana.js
ZeLonewolf: It was a bigger untangle to do that because of the code that accesses the URL bar and I've borked the refactor a couple times trying to do it.  Hence the staged approach so the diff doesn't get too unwieldy to review.

src/americana.js
ZeLonewolf: Updated in 1b5630a

src/layer/index.js
```diff
@@ -0,0 +1,235 @@
...
+import * as lyrRail from "./rail.js";
+import * as lyrRoad from "./road.js";
+import * as lyrTransportationLabel from "./transportation_label.js";
+import * as lyrWater from "./water.js";
+import * as lyrBuilding from "./building.js";
+import * as lyrHighwayExit from "./highway_exit.js";
+import * as lyrFerry from "./ferry.js";
+
+/**
+ * Builds the Americana layers property without localization.
```
1ec5, line 23: Need to update this comment.

1ec5 (approved):

src/americana.js
1ec5: Similar to the suggestion below about passing in a locale, `buildStyle()` can accept a base URL.

PR #666, opened 2023-01-05
Format semicolon-delimited value lists in labels

### LINKED ISSUE #665: Pretty-print semicolon delimiters in compound names
When a feature’s `name` property contains a semicolon, the style should replace it with a more presentable character.

## Background

When the language fallback list only includes unsupported languages, such as `vi`, or language codes that cannot be supported, such as `mul`, the style resorts to the `name` property of each feature, which corresponds to the `name` key in OSM. The `name` key typically represents the name in the local language, but sometimes there are multiple names of equal standing in that language or multiple local languages. In regions where official language policies promote multilingualism, OSM communities long ago coalesced around [ad-hoc delimiters](https://wiki.openstreetmap.org/wiki/Multilingual_names), such as a hyphen, solidus, or space, between each name in `name`.

[Belgium and Switzerland](https://zelonewolf.github.io/openstreetmap-americana/#map=3/45.53/8.75&language=mul) [Baltic Sea](https://zelonewolf.github.io/openstreetmap-americana/#map=6/55.897/18.027&language=mul) [Casablanca](https://zelonewolf.github.io/openstreetmap-americana/#map=7/33.544/-7.526&language=mul) [Kowloon](https://zelonewolf.github.io/openstreetmap-americana/#map=14/22.29821/114.18288&language=mul)

## Problem

While mappers have tended to consider these delimiters “good enough”, in most cases they aren’t the only valid delimiters to use in a map context. In fact, these delimiters cause several noticeable deficiencies:

* Chaotic differences between simultaneously visible features with no apparent linguistic or stylistic explanation
* Characters like the hyphen that do a poor job of visually setting text apart
* Poor line breaking, with words getting orphaned on a line otherwise containing text from another language or script
* A jarring inconsistency with the punctuation expected in the speaker’s preferred language or in the context of an American-style map

Meanwhile, in other regions that experience more grassroots multilingualism, or where multiple names need to be listed in a monolingual context, mappers have often applied a [semicolon](https://wiki.openstreetmap.org/wiki/Semi-colon_value_separator), for consistency with non-name keys that require machine readability. These semicolons clearly aren’t intended to be shown to the user verbatim, where they look like a rendering glitch.

[Kaser and New Square](https://zelonewolf.github.io/openstreetmap-americana/#map=12/41.13671/-74.04091&language=mul) [马岔河村、菜园村、刘灿东村、后于口村、王石楼村、李岔河村、岔河新村、富康新村、前鱼口村](https://zelonewolf.github.io/openstreetmap-americana/#map=13/35.90727/115.95732&language=mul)

[Columbus Cincinnati Road or Cincinnati-Columbus Road](https://zelonewolf.github.io/openstreetmap-americana/#map=20.73/39.3028086/-84.3879294/-62.5&language=mul)

There is [a long, winding discussion in the community forum](https://community.openstreetmap.org/t/multiple-delimited-names-in-the-name-tag/6803) about delimiters in `name`. It isn’t clear that there’s consensus to replace ad-hoc delimiters with the semicolon en masse, but there does seem to be some popular acceptance of the existing semicolons, at least in regions where no ad-hoc delimiter is well-established.

## Proposed solution

Ideally, it would be the responsibility of the data consumer to apply an appropriate delimiter in all these cases. However, these punctuation and whitespace characters are too ambiguous to interpret as delimiters; there are far too many individual names that legitimately contain them too. For now, we should focus on pretty-printing the semicolons due to their clear intent. Every `text-field` that refers to a name property should replace occurrences of a semicolon with a more appropriate delimiter:

* A newline in the main part of a point-placed label.
* A bullet in a city label’s local-language gloss, which needs to be more compact.
* A bullet in a line-placed label where GL JS doesn’t support newlines (mapbox/mapbox-gl-js#8575). (In the future, consider breaking out multiple symbol layers to alternate the names spatially without any delimiter.)

Replacing characters in a `text-field` is challenging for reasons specific to this project’s technology stack. The style specification doesn’t include an expression operator to replace a substring within a larger string: mapbox/mapbox-gl-js#4100. Fortunately, it does contain the `index-of` operator for finding the substring, so we can concatenate that occurrence’s prefix and suffix with the replacement string. The style specification also lacks an operator for splitting or looping, so we’d only be able to replace a fixed number of semicolons. Most multiply named features have only a few names, so this shouldn’t be a major problem in practice.

## Alternatives considered

The semicolon delimiters could be pretty-printed when generating the vector tiles, but that would limit our ability to choose different delimiters, for example depending on whether the label is point- or line-placed. Anyways, we’ll eventually need the ability to parse out individual values from this list in order to deduplicate glossed names: https://github.com/ZeLonewolf/openstreetmap-americana/pull/592#discussion_r1035284310.

/ref gravitystorm/openstreetmap-carto#4755

1ec5:
Whenever a feature’s name contains multiple values separated by semicolons, the label now puts each value on a separate line or separates them with an elegant bullet, for up to nine values, honoring the `;;` escape sequence.

This implementation includes a general-purpose polyfill for a string replacement expression operator. We can use it for any purpose in an expression or filter but need to be careful about avoiding a combinatoric explosion in the size of an expression’s code. I’ve included a comprehensive battery of unit tests.

[Kaser](https://zelonewolf.github.io/openstreetmap-americana/#language=mul&map=12/41.13779/-74.0456) Kaser legend
_[Kaser](https://www.openstreetmap.org/node/158859890) and [New Square](https://www.openstreetmap.org/node/158856497), New York_

[Machahe](https://zelonewolf.github.io/openstreetmap-americana/#map=13/35.90759/115.95543&language=mul) Machahe legend
_[Machahe/Caiyuan/Liucandong/Houyukou/Wangshilou/Lichahe/Chahexin/Fukangxin/Qianyukou villages](https://www.openstreetmap.org/node/9816809799) in Liangshan County, Jining City, Shandong, China_

[Viet Heritage Garden](https://zelonewolf.github.io/openstreetmap-americana/#map=15/37.32391/-121.85512&language=mul)
_[Viet Heritage Garden](https://www.openstreetmap.org/way/765864912), San José, California_

[Columbus Cincinnati Road](https://zelonewolf.github.io/openstreetmap-americana/#map=20.91/39.3028071/-84.3879331/-62.9&language=mul)
_[Columbus Cincinnati Road/Cincinnati–Columbus Road](https://www.openstreetmap.org/way/679125385) in West Chester Township, Butler County, Ohio_

[63A&B](https://zelonewolf.github.io/openstreetmap-americana/#map=14/39.1059/-84.28081) 63A&B legend
_Southbound Interstate 275 at [exits 63A/B](https://www.openstreetmap.org/node/246821345) in Union Township, Clermont County, Ohio_

Fixes #665.

### DIFF
--- a/src/constants/label.js
+++ b/src/constants/label.js
@@ -175,7 +175,7 @@ export function replaceExpression(
  * Increasing this constant deepens recursion for replacing delimiters in the
  * list, potentially affecting style loading performance.
  */
-const maxValueListLength = 9;
+const maxValueListLength = 3;
 
 /**
  * Returns an expression interpreting the given string as a list of tag values,
--- a/src/constants/label.js
+++ b/src/constants/label.js
@@ -117,13 +117,134 @@ export function localizeLayers(layers, locales) {
 }
 
 /**
- * The name in the user's preferred language.
+ * Returns an expression that replaces a finite number of occurrences of a
+ * substring expression withing a larger string expression, starting at a given
+ * index.
+ *
+ * This expression nests recursively by the maximum number of replacements. Take
+ * special care to minimize this limit, which exponentially increases the length
+ * of a property value in JSON. Excessive nesting causes acute performance
+ * problems when loading the style.
+ *
+ * The returned expression can be complex, so use it only once within a property
+ * value. To reuse the evaluated value, bind it to a variable in a let
+ * expression.
+ *
+ * @param haystack The overall string expression to search within.
+ * @param needle The string to search for, or an expression that evaluates to
+ *  this string.
+ */
+export function replaceExpression(
+  haystack,
+  needle,
+  replacement,
+  haystackStart,
+  numReplacements = 1
+) {
+  let asIs = ["slice", haystack, haystackStart];
+  if (numReplacements <= 0) {
+    return asIs;
+  }
+
+  let needleStart = ["index-of", needle, haystack, haystackStart];
+  let needleLength =
+    typeof needle === "object" ? ["length", needle] : needle.length;
+  let needleEnd = ["+", needleStart, needleLength];
+  return [
+    "case",
+    [">=", needleStart, 0],
+    [
+      "concat",
+      ["slice", haystack, haystackStart, needleStart],
+      replacement,
+      replaceExpression(
+        haystack,
+        needle,
+        replacement,
+        needleEnd,
+        numReplacements - 1
+      ),
+    ],
+    asIs,
+  ];
+}
+
+/**
+ * Maximum number of values in a semicolon-delimited list of values.
+ *
+ * Increasing this constant deepens recursion for replacing delimiters in the
+ * list, potentially affecting style loading performance.
+ */
+const maxValueListLength = 9;
+
+/**
+ * Returns an expression interpreting the given string as a list of tag values,
+ * pretty-printing the standard semicolon delimiter with the given separator.
+ *
+ * https://wiki.openstreetmap.org/wiki/Semi-colon_value_separator
+ *
+ * The returned expression can be complex, so use it only once within a property
+ * value. To reuse the evaluated value, bind it to a variable in a let
+ * expression.
+ *
+ * @param valueList A semicolon-delimited list of values.
+ * @param separator A string to insert between each value, or an expression that
+ *  evaluates to this string.
+ */
+export function listValuesExpression(valueList, separator) {
+  let maxSeparators = maxValueListLength - 1;
+  // Replace the ;; escape sequence with a placeholder sequence unlikely to
+  // legitimately occur inside a value or separator.
+  const objReplacementChar = "\x91\ufffc\x92"; // https://overpass-turbo.eu/s/1pJx
+  let safeValueList = replaceExpression(
+    valueList,
+    ";;",
+    objReplacementChar,
+    0,
+    maxSeparators
+  );
+  // Pretty-print the ; delimiter.
+  let prettyValueList = replaceExpression(
+    ["var", "safeValueList"],
+    ";",
+    separator,
+    0,
+    maxSeparators
+  );
+  // Replace the placeholder sequence with an unescaped semicolon.
+  let prettySafeValueList = replaceExpression(
+    ["var", "prettyValueList"],
+    objReplacementChar,
+    ";",
+    0,
+    maxSeparators
+  );
+  return [
+    "let",
+    "safeValueList",
+    safeValueList,
+    ["let", "prettyValueList", prettyValueList, prettySafeValueList],
+  ];
+}
+
+/**
+ * The names in the user's preferred language, each on a separate line.
  */
 export const localizedName = [
   "let",
   "localizedName",
   "",
-  ["var", "localizedName"],
+  listValuesExpression(["var", "localizedName"], "\n"),
+];
+
+/**
+ * The names in the user's preferred language, all on the same line.
+ */
+export const localizedNameInline = [
+  "let",
+  "localizedName",
+  "",
+  listValuesExpression(["var", "localizedName"], " \u2022 "),
 ];
 
 /**
@@ -217,7 +338,7 @@ export const localizedNameWithLocalGloss = [
       ["var", "localizedCollator"],
     ],
     // ...just pick one.
-    ["var", "localizedName"],
+    ["format", listValuesExpression(["var", "localizedName"], "\n")],
     // If the name in the preferred language is the same as the name in the
     // local language except for the omission of diacritics and/or the addition
     // of a suffix (e.g., "City" in English)...
@@ -227,7 +348,13 @@ export const localizedNameWithLocalGloss = [
       ["var", "diacriticInsensitiveCollator"]
     ),
     // ...then replace the common prefix with the local name.
-    overwritePrefixExpression(["var", "localizedName"], ["get", "name"]),
+    [
+      "format",
+      overwritePrefixExpression(
+        ["var", "localizedName"],
+        listValuesExpression(["get", "name"], "\n")
+      ),
+    ],
     // If the name in the preferred language is the same as the name in the
     // local language except for the omission of diacritics and/or the addition
     // of a prefix (e.g., "City of" in English or "Ciudad de" in Spanish)...
@@ -237,7 +364,13 @@ export const localizedNameWithLocalGloss = [
       ["var", "diacriticInsensitiveCollator"]
     ),
     // ...then replace the common suffix with the local name.
-    overwriteSuffixExpression(["var", "localizedName"], ["get", "name"]),
+    [
+      "format",
+      overwriteSuffixExpression(
+        ["var", "localizedName"],
+        listValuesExpression(["get", "name"], "\n")
+      ),
+    ],
     // Otherwise, gloss the name in the local language if it differs from the
     // localized name.
     [
@@ -252,7 +385,7 @@ export const localizedNameWithLocalGloss = [
       // bother rendering it.
       ["concat", ["slice", ["var", "localizedName"], 0, 1], " "],
       { "font-scale": 0.001 },
-      ["get", "name"],
+      listValuesExpression(["get", "name"], " \u2022 "),
       { "font-scale": 0.8 },
       ["concat", " ", ["slice", ["var", "localizedName"], 0, 1]],
       { "font-scale": 0.001 },
--- a/test/spec/label.js
+++ b/test/spec/label.js
@@ -272,9 +272,9 @@ describe("label", function () {
     let evaluatedLabelAndGloss = (locales, properties) => {
       let evaluated = evaluatedExpression(locales, properties);
       if (typeof evaluated === "string") {
-        return [evaluated, null];
+        return [evaluated];
       }
-      return [evaluated.sections[0].text, evaluated.sections[4].text];
+      return [evaluated.sections[0].text, evaluated.sections[4]?.text];
     };
 
     let expectGloss = (
@@ -295,11 +295,12 @@ describe("label", function () {
     };
 
     it("puts an unlocalized name by itself", function () {
-      expect(
-        evaluatedExpression(["en"], {
-          name: "Null Island",
-        })
-      ).to.be.eql("Null Island");
+      let evaluated = evaluatedExpression(["en"], {
+        name: "Null Island",
+      });
+
+      expect(evaluated.sections.length).to.be.eql(1);
+      expect(evaluated.sections[0].text).to.be.eql("Null Island");
     });
     it("glosses an anglicized name with the local name", function () {
       let evaluated = evaluatedExpression(["en"], {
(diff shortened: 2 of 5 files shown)


src/constants/label.js
```diff
@@ -117,13 +117,134 @@ export function localizeLayers(layers, locales) {
...
+  ];
+}
+
+/**
+ * Maximum number of values in a semicolon-delimited list of values.
+ *
+ * Increasing this constant deepens recursion for replacing delimiters in the
+ * list, potentially affecting style loading performance.
+ */
+const maxValueListLength = 9;
```
1ec5, line 178: Please compare this branch to main and note any perceptible degradation in style loading speed, tile loading speed, panning and zooming performance, or memory or CPU usage. If it’s severe enough, we can tune this limit downward until we’re satisfied with the balance between handling enough features and maintaining reasonable performance.

src/constants/label.js
```diff
@@ -117,13 +117,134 @@ export function localizeLayers(layers, locales) {
...
+ *
+ * @param valueList A semicolon-delimited list of values.
+ * @param separator A string to insert between each value, or an expression that
+ *  evaluates to this string.
+ */
+export function listValuesExpression(valueList, separator) {
+  let maxSeparators = maxValueListLength - 1;
+  // Replace the ;; escape sequence with a placeholder sequence unlikely to
+  // legitimately occur inside a value or separator.
+  const objReplacementChar = "\x91\ufffc\x92"; // https://overpass-turbo.eu/s/1pJx
```
1ec5, line 198: It’s amazing how much junk there is in `name` that can impersonate a [strip marker](https://www.mediawiki.org/wiki/Strip_marker), but this sequence does the trick for now. If someone opens a pub for text processing geeks and names it “The &#x91;&#xfffc;&#x92; & ;”, I’ll kindly ask someone to offer up a prized password as sacrifice.

src/constants/label.js
```diff
@@ -217,7 +338,7 @@ export const localizedNameWithLocalGloss = [
       ["var", "localizedCollator"],
     ],
     // ...just pick one.
-    ["var", "localizedName"],
+    ["format", listValuesExpression(["var", "localizedName"], "\n")],
```
1ec5, line 341: The return value of this method is so complex that the style-spec package (used by the unit tests) is apparently unable to upcast it to a `format` expression, leading to an error when it sees the direct `format` expression in the else slot below. Interestingly, GL JS has no such problem at runtime. This reassures the expression parser that each branch of the `case` expression will resolve to a `format` expression.

test/spec/label.js
```diff
@@ -272,9 +272,9 @@ describe("label", function () {
     let evaluatedLabelAndGloss = (locales, properties) => {
       let evaluated = evaluatedExpression(locales, properties);
       if (typeof evaluated === "string") {
-        return [evaluated, null];
+        return [evaluated];
       }
-      return [evaluated.sections[0].text, evaluated.sections[4].text];
+      return [evaluated.sections[0].text, evaluated.sections[4]?.text];
```
1ec5, line 277: As a result of the manual cast above, this expression no longer returns a literal string, always a `format` object.

ZeLonewolf:
Test location below.

Performance measurement in Chrome on my laptop (multiple refreshes based on network tab in the Chrome debugger console
Current master (remote github): 1.9-2.1s
Current master (localhost): 1.9-2.1s
This branch (localhost): 2.6-2.9s

Approximately 37% slowdown.

Current/New:
image
image
([localhost link](http://localhost:1776/#map=13/36.6016/119.4771&language=mul))

Cursory testing seems to indicate that these numbers are independent of location (high zoom on an ocean tile yields the same results).

I'm curious to understand if the performance difference is due to _execution_ or _building the style.json_.

1ec5:
> Cursory testing seems to indicate that these numbers are independent of location (high zoom on an ocean tile yields the same results).

Yes, the expression is only evaluated at runtime when loading a tile or reevaluating a symbol layer, so the impact is uniform everywhere. That said, it would be possible to limit this semicolon replacement to certain layers that are more likely to have semicolons in them. Or we could create a separate copy of each layer prefiltered to just the features that contain semicolons in them, but at the cost of additional layers that participate in symbol collision.

> I'm curious to understand if the performance difference is due to execution or building the style.json.

The new code probably isn’t a significant contributor to the time it takes to build the JSON. (It happens in memory; there’s no file involved.) However, the step right after that loads the JSON requires more memory than before. Under the hood, GL JS compiles the expression to a type-safe, object-oriented data structure, which becomes more expensive as the expression becomes more deeply nested. Whenever a new tile needs to be displayed, a Web Worker needs to evaluate more massive expressions. The `case` statements should allow the evaluator to short-circuit most of the expression unless there are semicolons, but it still requires work to ferry around such a large data structure.

With a built-in string replacement operator, the performance concern would essentially go away. The operator’s performance would be similar to the `upcase` and `downcase` operators, which no one bats an eye about. Deduplicating values would require slightly more involved string splitting and array joining operators, but those probably wouldn’t have this performance impact either.

ZeLonewolf:
Pending anyone else's review of the code changes, I think it's a good idea to proceed with the proposed change to handle semicolon-delimited lists.  I'm concerned about the 1/3 performance drop for processing up to 9 semi-colons, though I understand that future improvements to MapLibre may make it possible for use to be more efficient about this. Perhaps we could reduce the 9 down to a number that captures the vast majority of the sane delimiter use cases while ignoring the silly ones that are probably tagging errors.

claysmalley:
Should spaces before and after semicolons be trimmed, or is that considered a tagging error?

Screenshot from 2023-01-05 11-27-30

1ec5:
> Should spaces before and after semicolons be trimmed, or is that considered a tagging error?

I believe that’s a tagging error. In fact, editors such as iD automatically remove the space when you touch the name.

zekefarwell:
> Should spaces before and after semicolons be trimmed, or is that considered a tagging error?

The wiki mentions a single space after a semicolon being valid:
https://wiki.openstreetmap.org/wiki/Semi-colon_value_separator#Space_character_padding

I'm sure handling either `;` or `; ` would make for more complex code, but it could be a strategically good choice if it's doable.  On maps that don't replace semicolons with something nicer looking, mappers will probably find "First Name; Second Name" a bit more tolerable than "First Name;Second Name".

1ec5:
Editors strip out the space in general, but [preserve it for certain syntaxes](https://github.com/openstreetmap/iD/blob/1ee45ee1f03f0fe4d452012c65ac6ff7649e229f/test/spec/util/clean_tags.js#L58-L72) like `opening_hours` and `*:conditional`.

zekefarwell:
> Editors strip out the space in general

Well iD does anyway.  I didn't have any problem including a space here with JOSM: https://www.openstreetmap.org/way/971815925

ZeLonewolf:
For place nodes, [stats](https://overpass-turbo.eu/s/1pLE) on prevalence:

| # Semi-colons | Prevalence |
|---------------|------------|
| 1+ semicolon  | 444        |
| 2+ semicolons | 9          |
| 3+ semicolons | 6          |
| 4+ semicolons | 3          |
| 5+ semicolons | 2          |
| 6+ semicolons | 1          |

1ec5:
For reference, the semicolon delimiter was first documented (without a space) [in 2007](https://wiki.openstreetmap.org/wiki/Special:Diff/60691), when it replaced a recommendation to use a comma.[^comma] There continued to be some uncertainty about the delimiter [until 2008](https://wiki.openstreetmap.org/wiki/Special:Diff/99405).

>> Editors strip out the space in general
> 
> Well iD does anyway.  I didn't have any problem including a space here with JOSM: https://www.openstreetmap.org/way/971815925

You can do [anything you want](https://www.openstreetmap.org/changeset/91011113) in JOSM. 😉 But it too [prefers a semicolon](https://github.com/JOSM/josm/blob/3af7bae967e273dcee423ca0aac04524615d5ba5/src/org/openstreetmap/josm/gui/tagging/TagModel.java#L122-L130) whenever it constructs a value list for you. The main reason space-padded semicolons appear in the database today is because Potlatch used to insert this delimiter when merging two ways. It was [usually a mistake](https://lists.openstreetmap.org/pipermail/talk/2008-May/026611.html) caused by its confusing shortcut for merging ways. Based on that behavior, mappers (including yours truly) used to manually insert the same delimiter for consistency, but it has been rare for a long time: openstreetmap/iD#941.

Documentation about space padding is pretty stale, [dating back](https://wiki.openstreetmap.org/wiki/Special:PermanentLink/567830#Space_character_padding) to when the same page pointed out that characters other than a semicolon were being used as value delimiters. This page is the product of past [edit-warring and also flamewarring](https://lists.openstreetmap.org/pipermail/tagging/2015-January/thread.html#21141) on the tagging list, so people have probably been hesitant to do much cleanup on it since. (The controversy was not about the choice of delimiter but rather about multiple-value lists in the first place.) I [updated the documentation](https://wiki.openstreetmap.org/wiki/Special:Diff/2459831) to reflect longstanding reality.

If there’s a strong need to support the space after the semicolon, then there will be an additional performance hit as long as we build upon this magnificent workaround for the lack of a string replacement operator. This extra padding is more noticeable with the inline treatment than with the newline treatment. We could replace the bullet with a series of spaces, which would make the imbalance imperceptible, but at the cost of making it even more likely for a dual-named label to get collided out.

Given this complexity, I’d suggest tracking support for space-padded semicolons as tail work, which I would consider blocked by a string replacement operator. I’m not opposed to it in principle, since a value’s leading and trailing spaces are always extraneous, but the cost-benefit ratio isn’t superb.

> For place nodes, [stats](https://overpass-turbo.eu/s/1pLE) on prevalence:

Thanks, this argues strongly in favor of setting the limit to just one or two semicolons for now. I suppose it would be difficult to gather similar statistics about ad-hoc delimiters that would be candidates for conversion to semicolons – which is the point of using semicolons in the first place. But out of curiosity, is there a similar distribution on places if we naïvely assume all space-padded slashes and hyphens are intended to be value delimiters?

[^comma]: A comma continues to be used very commonly in `addr:housenumber` to this day as part of the [Karlsruhe schema](https://wiki.openstreetmap.org/wiki/Proposed_features/House_numbers/Karlsruhe_Schema), authored by the same mapper who [originally documented the comma as a more general delimiter](https://wiki.openstreetmap.org/wiki/Special:Diff/49728).

zekefarwell:
> You can do [anything you want](https://www.openstreetmap.org/changeset/91011113) in JOSM. 😉

Indeed 💪💪

> Given this complexity, I’d suggest tracking support for space-padded semicolons as tail work, which I would consider blocked by a string replacement operator. I’m not opposed to it in principle, since a value’s leading and trailing spaces are always extraneous, but the cost-benefit ratio isn’t superb.

Sounds good to me.  Even without explicitly handling the extra space after a semicolon, the bullet delimiter is far superior to to displaying the raw value with the semicolon.  I don't really find the extra space on one side of the bullet all that noticeable.  I just didn't think we ought to necessarily consider a space after a semicolon to be a tagging error.  At least not while the raw, semicolon delimited values are displayed on the osm.org standard layer.

ZeLonewolf:
Query for [all types of objects](https://overpass-turbo.eu/s/1pLT) this time.

* 1 semi-colon 31,953
* 2 semi-colons 2,163
* 3 semi-colon 539
* 4 semi-colon 34
* 5 semi-colon 10
* 6+ semi-colon 6

zekefarwell (approved):
This looks good to me.  I did some performance profiling and there's definitely an increase to JS execution time and jank on initial load and while panning/zooming around the map.  However, performance is already not great and without profiling I'm not sure the increase from this changes is really all that noticeable (at least on my hardware).  Given the stats on numbers of semicolons it seems like dropping `maxValueListLength` to 4 would be perfectly reasonable if that would improve performance.

src/constants/label.js
```diff
@@ -175,7 +175,7 @@ export function replaceExpression(
  * Increasing this constant deepens recursion for replacing delimiters in the
  * list, potentially affecting style loading performance.
  */
-const maxValueListLength = 9;
+const maxValueListLength = 3;
```
1ec5, line 178: Three values separated by two semicolons is plenty enough for the `name`s currently using semicolons. In the future, with a more efficient string replacement implementation, we can add support for more semicolons, which will make this option more attractive to mappers currently using informal delimiters like spaces.

1ec5:
> Should spaces before and after semicolons be trimmed, or is that considered a tagging error?

#670 collapses up to one space following a semicolon. If a mapper does intentionally use <code>;&nbsp;</code> to make the label look nicer in some other renderers, it’ll still work fine in this style. However, iD does strip it out, so if this becomes an important tagging style, iD will need a one-line fix.

PR #667, opened 2023-01-05
Draw escutcheon shields mathematically

claysmalley:
One more draw function down!

Screenshot from 2023-01-05 18-13-21


ZeLonewolf:
I'm looking at the Virginia shields and I'm wondering if I'm seeing a bit of clipping in the horizontal direction on the sides?

image

ZeLonewolf:
Yep, the online version has cleaner edges:
image

claysmalley:
Weird. Nothing looks wrong at 200% zoom, but at 100% the notches are there (though less noticeable than in your screenshot). Something funky is going on with `bezierCurveTo()`.

Screenshot from 2023-01-05 18-30-54
Screenshot from 2023-01-05 18-30-39

claysmalley:
I've adjusted the metrics a little but I can't seem to get the notches to go away.

Screenshot from 2023-01-06 10-20-24

ZeLonewolf (approved):
Did some side-by-side testing and this looks good to go.  It's an improvement over the current shields and I'm not seeing obvious glitches now.

Issue #471, opened 2022-07-02
Gloss names of places in the local language


1ec5:
World political maps and atlases by American publishers typically show well-known places in English but also gloss the name in the local language when it differs from the English name. Typically, the local name is shown in parentheses and/or italics, and sometimes on a separate line in smaller type.

<img src="https://user-images.githubusercontent.com/1231218/177016672-537aedc3-3597-4e11-94b5-eb51a9363d36.png" width="400" alt="Rand McNally">

I’ve seen these glosses on labels of populated places, but rarely on other kinds of features. I suppose there are a variety of reasons:

* Many anglicized city names are less well-known than official English names of countries.
* A waterway may form an international boundary and thus have a whole host of local names, but consistency along the length of the waterway is often prioritized.
* Labels for non-place POIs are already long enough to risk getting collided out by POI labels without getting doubled in length by the gloss.
* Political maps rarely label anything less significant than a populated place.

We should consider glosses for other feature types at higher zoom levels, where collision isn’t as much of a concern, but populated places seems like a good place to start.

To achieve the gloss, we can set `text-field` to [a `format` expression](https://maplibre.org/maplibre-gl-js-docs/example/display-and-style-rich-text-labels/) that includes both `name:en` and `name` when the two properties differ. A [`collator`](https://maplibre.org/maplibre-gl-js-docs/style-spec/expressions/#types-collator) expression can be passed into the string comparison to perform case and diacritic folding.

/ref #100 #428

1ec5:
The ability to choose a base language other than English (#20 #21) would complicate this feature somewhat, but even in English there are some edge cases to consider:

* Since “Montreal” and “Montréal” differ by only diacritics, it’s probably fine to render it as just “Montréal” instead of “Montreal (Montréal)”.
* Hong Kong’s place name labels are bilingual in both English and Chinese, so the expression would need to avoid duplicating the English name. For example, it should be “Kowloon (九龍)”, not “Kowloon (九龍 Kowloon)”.

1ec5:
[...]

Issue #586, opened 2022-11-27
Place names are unavailable in many relevant languages


1ec5:
This style uses vector tiles that contain place names in many languages, but the selection of languages leaves some room for improvement. For a style that’s focused on the Americas, particularly the United States, language support is nevertheless dominated by European languages:

<details>
<summary>Languages represented in United States place POI generated by OpenMapTiles/Planetiler</summary>

* Albanian (sq)
* Amharic (am)
* Arabic (ar)
* Armenian (hy)
* Azerbaijani (az)
* Basque (eu)
* Belarusian (be)
* Bosnian (bs)
* Breton (br)
* Bulgarian (bg)
* Catalan (ca)
* Chinese (zh)
* Corsican (co)
* Croatian (hr)
* Czech (cs)
* Danish (da)
* Dutch (nl)
* English (en)
* Esperanto (eo)
* Estonian (et)
* Finnish (fi)
* French (fr)
* Georgian (ka)
* German (de)
* Greek (el)
* Hebrew (he)
* Hindi (hi)
* Hungarian (hu)
* Icelandic (is)
* Indonesian (id)
* Irish (ga)
* Italian (it)
* Japanese (ja)
* Kannada (kn)
* Kazakh (kk)
* Korean (ko)
* Kurdish (ku)
* Latin (la)
* Latvian (lv)
* Lithuanian (lt)
* Luxembourgish (lb)
* Macedonian (mk)
* Malayalam (ml)
* Maltese (mt)
* Norwegian (no)
* Occitan (oc)
* Polish (pl)
* Portuguese (pt)
* Romanian (ro)
* Romansh (rm)
* Russian (ru)
* Scottish Gaelic (gd)
* Serbian (sr)
* Slovak (sk)
* Slovenian (sl)
* Spanish (es)
* Swedish (sv)
* Tamil (ta)
* Telugu (te)
* Thai (th)
* Turkish (tr)
* Ukrainian (uk)
* Welsh (cy)
* Western Frisian (fy)

</details>

[...]

PR #670, opened 2023-01-07
Deduplicate names in gloss; gloss towns

1ec5:
Replaced the general-purpose function for generating string replacement expressions with a function for generating an expression that specifically scans a string for semicolon delimiters, escaped semicolons, and space padding after a semicolon. Compared to #666, this change decreases the overall minified style JSON size[^size] (a rough proxy for style complexity) from 954,393&nbsp;characters down to just 932,343&nbsp;characters.

With this optimization in place, a city’s local-language name gloss now omits any name that matches the city’s name in the user’s preferred language to avoid repetition. This brings the size back up to 943,759&nbsp;characters. Unfortunately, every multilingual `place=city` is currently tagged with a `name` that contains delimiters other than a semicolon, making this enhancement rather pointless. To make this deduplication worthwhile, this PR also enables a local-name gloss on `town`s too.

Previously, I mentioned in https://github.com/ZeLonewolf/openstreetmap-americana/issues/471#issue-1292150160 that American maps and atlases only gloss major city labels, but we’ve already been glossing every `city` label. Spot-checking shows that label density isn’t suffering too much from the more verbose town labels, and in any case we support zooming in. The one downside to enabling glosses on town labels is that the town label layer has to repeat the massive expression that we’ve been using for cities: we’re back up to 952,745&nbsp;characters&nbsp;– still less than we started out with, but only slightly.

[Kaser and New Square](https://zelonewolf.github.io/openstreetmap-americana/#map=12/41.13986/-74.0432&language=en)

<table>
<thead>
<tr>
<th>Change</th><th>Style JSON size (characters)</th>
</tr>
</thead>
<tbody>
<tr>
<td>replacer</td><td align="right">954,393</td>
</tr>
<tr>
<td>− replacer + scanner</td><td align="right">−22,050</td>
</tr>
<tr>
<td>+ deduplication</td><td align="right">+11,416</td>
</tr>
<tr>
<td>+ town glosses</td><td align="right">+8,986</td>
</tr>
</tbody>
<tfoot>
<tr>
<th scope="row">Net change</th><td align="right">−1,648</td>
</tfoot>
</table>

/ref https://github.com/ZeLonewolf/openstreetmap-americana/pull/592#discussion_r1035284310 https://github.com/ZeLonewolf/openstreetmap-americana/pull/666#issuecomment-1372443684

[^size]: `JSON.stringify(map.getStyle()).length`

### DIFF
--- a/src/constants/label.js
+++ b/src/constants/label.js
@@ -135,59 +135,110 @@ export function localizeLayers(layers, locales) {
  * @param listStart A zero-based index into the list at which the search begins.
  * @param numReplacements The maximum number of replacements remaining.
  */
-function listValueExpression(list, separator, listStart, numReplacements) {
+function listValueExpression(
+  list,
+  separator,
+  valueToOmit,
+  listStart,
+  numReplacements
+) {
   let asIs = ["slice", list, listStart];
   if (numReplacements <= 0) {
     return asIs;
   }
 
   let iteration = numReplacements;
+  let rawSeparator = ";";
   return [
     "let",
     "needleStart" + iteration,
-    ["index-of", ";", list, listStart],
-    "needleEnd" + iteration,
-    ["+", ["index-of", ";", list, listStart], 1],
+    ["index-of", rawSeparator, list, listStart],
     [
       "case",
       [">=", ["var", "needleStart" + iteration], 0],
       // Found a semicolon.
       [
-        "concat",
-        // Start with everything before the semicolon.
+        "let",
+        "value" + iteration,
         ["slice", list, listStart, ["var", "needleStart" + iteration]],
+        "needleEnd" + iteration,
+        ["+", ["var", "needleStart" + iteration], rawSeparator.length],
         [
-          "let",
-          "lookahead" + iteration,
-          // Look ahead by one character.
+          "concat",
+          // Start with everything before the semicolon unless it's the value to
+          // omit.
           [
-            "slice",
-            list,
-            ["var", "needleEnd" + iteration],
-            ["+", ["var", "needleEnd" + iteration], 1],
+            "case",
+            ["==", ["var", "value" + iteration], valueToOmit],
+            "",
+            ["var", "value" + iteration],
           ],
           [
-            "concat",
-            // If the lookahead character is another semicolon, append an unescaped semicolon.
-            // Otherwise, append the passed-in separator.
-            ["match", ["var", "lookahead" + iteration], ";", ";", separator],
-            // Recurse for the next value in the value list.
-            listValueExpression(
+            "let",
+            "lookahead" + iteration,
+            // Look ahead by one character.
+            [
+              "slice",
               list,
-              separator,
-              // Skip past the current value and semicolon for any subsequent searches.
+              ["var", "needleEnd" + iteration],
+              ["+", ["var", "needleEnd" + iteration], rawSeparator.length],
+            ],
+            [
+              "let",
+              // Skip past the current value and semicolon for any subsequent
+              // searches.
+              "nextListStart" + iteration,
               [
                 "+",
                 ["var", "needleEnd" + iteration],
                 // Also skip past any escaped semicolon or space padding.
-                ["match", ["var", "lookahead" + iteration], [";", " "], 1, 0],
+                [
+                  "match",
+                  ["var", "lookahead" + iteration],
+                  [rawSeparator, " "],
+                  rawSeparator.length,
+                  0,
+                ],
               ],
-              numReplacements - 1
-            ),
+              [
+                "case",
+                // If the only remaining value is the value to omit, stop
+                // scanning.
+                [
+                  "==",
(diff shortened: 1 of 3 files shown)


src/constants/label.js
```diff
@@ -130,42 +129,118 @@ export function localizeLayers(layers, locales) {
...
   if (numReplacements <= 0) {
     return asIs;
   }
 
-  let needleStart = ["index-of", needle, haystack, haystackStart];
-  let needleLength =
-    typeof needle === "object" ? ["length", needle] : needle.length;
-  let needleEnd = ["+", needleStart, needleLength];
+  let iteration = numReplacements;
+  let rawSeparator = ";";
```
1ec5, line 151: I spilled a lot of ink in [this forum discussion](https://community.openstreetmap.org/t/multiple-delimited-names-in-the-name-tag/6803) about the need for a reliable delimiter&nbsp;– the semicolon. But what if we pretend that one of the popular ad-hoc delimiters is unambiguous enough to split on instead?

Sad reality | Simulation
----|----
[Alicante before](https://zelonewolf.github.io/openstreetmap-americana/#map=8.65/38.3721/-0.4708&language=en) | [Alicante after](https://zelonewolf.github.io/openstreetmap-americana/#map=8.65/38.3721/-0.4708&language=en)
[Brussels before](https://zelonewolf.github.io/openstreetmap-americana/#map=12/50.86018/4.34901&language=en) | [Brussels after](https://zelonewolf.github.io/openstreetmap-americana/#map=12/50.86018/4.34901&language=en)
[Casablanca before](https://zelonewolf.github.io/openstreetmap-americana/#map=8/33.578/-7.378&language=en) | [Casablanca after](https://zelonewolf.github.io/openstreetmap-americana/#map=8/33.578/-7.378&language=en)
[Hong Kong before](https://zelonewolf.github.io/openstreetmap-americana/#map=10.5/22.3253/114.217&language=en) | [Hong Kong after](https://zelonewolf.github.io/openstreetmap-americana/#map=10.5/22.3253/114.217&language=en)

But alas, we can’t just turn every hyphen or slash or space into a value delimiter:

Sad reality | Sadder simulation
----|----
[Tizi Ouzou and Bordj Bou Arreridjj before](https://zelonewolf.github.io/openstreetmap-americana/#map=7/36.284/4.355&language=en) | [Tizi Ouzou and Bordj Bou Arreridjj after](https://zelonewolf.github.io/openstreetmap-americana/#map=7/36.284/4.355&language=en)
[Reynoldsburg - New Albany Road before](https://zelonewolf.github.io/openstreetmap-americana/#map=14/40.06114/-82.8109/45&language=en) | [Reynoldsburg - New Albany Road after](https://zelonewolf.github.io/openstreetmap-americana/#map=14/40.06114/-82.8109/45&language=en)

ZeLonewolf:
Great work! 

For places with multiple glosses, would it be better if each label has its own parentheses on a separate row?  Just thinking that the open parenthesis on one line and the close parenthesis on the last line looks a touch uneven.

1ec5:
> For places with multiple glosses, would it be better if each label has its own parentheses on a separate row? Just thinking that the open parenthesis on one line and the close parenthesis on the last line looks a touch uneven.

This is probably only hypothetical for now, since so few places with long names or many local languages are using the semicolon in their `name`s. But you’re right that there’s an imbalance. There’s not much we can do about that without GL JS support for [hanging punctuation](https://en.wikipedia.org/wiki/Hanging_punctuation) or [optical margin alignment](https://en.wikipedia.org/wiki/Optical_margin_alignment). A separate pair of parentheses on each line would look busy (really) (no, really really), and we’d still have some long names get word-wrapped onto multiple lines without any opportunities to introduce parentheses.

The only practical way to resolve this imbalance would be to nix the parentheses and rely on font attributes to distinguish the gloss from the main text. Some maps do rely on italics or a horizontal bar to set off the gloss, but I’m not sure it would work as well in this style, especially since we allow the label to appear above the icon.

src/constants/label.js
```diff
@@ -385,7 +442,10 @@ export const localizedNameWithLocalGloss = [
       // bother rendering it.
       ["concat", ["slice", ["var", "localizedName"], 0, 1], " "],
       { "font-scale": 0.001 },
-      listValuesExpression(["get", "name"], " \u2022 "),
+      listValuesExpression(["get", "name"], " \u2022 ", [
```
ZeLonewolf, line 445: Perhaps the bullet character could be extracted to a constant for better readability?

ZeLonewolf:
When localized to Hebrew:
image

I assume this is a tagging error?

1ec5:
“<bdi>שיכון סקווירא</bdi>” is New Square’s name in Yiddish. It has a different name in Hebrew, “<bdi>ניו סקוור</bdi>”. #586 tracks the lack of a dedicated Yiddish name field in the vector tiles.

ZeLonewolf (approved):
This looks awesome, thanks!

PR #671, opened 2023-01-07
Fix conversion from `fill-opacity` to CSS

### LINKED ISSUE #664: Legend displays intermittent lake or pond with gray fill instead of light blue
Maybe because the intermittent fill uses a semi-transparent color value?

Screen Shot 2023-01-03 at 8 13 04 PM

https://zelonewolf.github.io/openstreetmap-americana/#map=16.84/-28.073345/153.423839

1ec5:
Unmultiply RGB components by the alpha component. If `fill-opacity` is unspecified, use `fill-color` directly in CSS. Apply `fill-extrusion-opacity` to the whole feature, not just its fill.

[Tucson](https://zelonewolf.github.io/openstreetmap-americana/#map=14/32.11553/-110.83281)

Fixes #664.

### DIFF
--- a/src/js/legend_control.js
+++ b/src/js/legend_control.js
@@ -340,15 +340,14 @@ export default class LegendControl {
     let fillColor =
       fill?.layer.paint["fill-color"] ||
       fill?.layer.paint["fill-extrusion-color"];
-    if (fillColor) {
-      let opacity =
-        fill?.layer.paint["fill-opacity"] ??
-        fill?.layer.paint["fill-extrusion-opacity"] ??
-        fillColor.a ??
-        1;
-      fillColor = `rgba(${fillColor.r * 255}, ${fillColor.g * 255}, ${
-        fillColor.b * 255
-      }, ${opacity})`;
+    let fillOpacity =
+      fill?.layer.paint["fill-opacity"] ??
+      fill?.layer.paint["fill-extrusion-opacity"];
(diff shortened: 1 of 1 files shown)


ZeLonewolf:
I zoomed to the following location ([localhost link](http://localhost:1776/#map=12/37.25814/-110.45004))

I don't seem to be seeing intermittent water show up here.

image

1ec5:
> I don't seem to be seeing intermittent water show up here.

These are [river areas](https://www.openstreetmap.org/relation/12499805), not lake areas. #632 didn’t add a legend entry for river areas because they would be redundant to river lines. If this is confusing, perhaps there needs to be an ability for a single legend entry to be depicted as both a swatch and a line, each occupying half the table cell.

ZeLonewolf:
Looks good on a lake, then.  Intermittent river areas are probably worth a legend entry, separate from the river area + river line for a normal river.
image

ZeLonewolf (approved):

1ec5:
> Intermittent river areas are probably worth a legend entry, separate from the river area + river line for a normal river.

Yeah, this needs more thought. The legend doesn’t depict river areas at all, regardless of intermittency. It isn’t difficult to add a separate entry for river areas, but the difficulty is what to call it and how to visually distinguish it from the entry for lakes and ponds.

PR #673, opened 2023-01-08
Split localized name on semicolons before gloss

1ec5:
The semicolon-related changes in #666 left out one part of the massive `text-field` expression that inserts the localized name before the local-language gloss. Now non-native `name:*` tags that contain semicolons will also benefit from semicolon splitting.

[San Francisco](https://zelonewolf.github.io/openstreetmap-americana/#map=10/37.8435/-122.3284&language=zh)

### DIFF
--- a/src/constants/label.js
+++ b/src/constants/label.js
@@ -350,22 +350,27 @@ function overwritePrefixExpression(target, newPrefix) {
 function endsWithExpression(target, candidateSuffix, collator) {
   let wordBoundary = " ";
   return [
-    "all",
+    "let",
+    "suffixStart",
+    ["-", ["length", target], ["length", candidateSuffix]],
     [
-      "==",
-      ["slice", target, ["-", ["length", target], ["length", candidateSuffix]]],
-      candidateSuffix,
-      collator,
-    ],
-    [
-      "==",
+      "all",
       [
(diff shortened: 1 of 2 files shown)


1ec5:
This increases the JSON style size to 954,677&nbsp;characters, slightly more than the high-water mark of 954,393 immediately following #666.

1ec5:
If you’re wondering why Chinese has so many triple-named features, it’s because of differences in place naming practices between Mandarin and Cantonese, between Traditional and Simplified Chinese, and among the various countries that use these languages. `name:zh` is more or less a compatibility fallback for more specific tags like `name:zh-Hant-TW` and `name:cmn`. #586 tracks adding specialized support for these more specific language codes, but it’s still valid to request Chinese generically, since the naming differences don’t cut very cleanly along national lines.

1ec5:
I refactored the massive glossed label expression to use more `let` statements, avoiding repetition of the complex result of `listValuesExpression()`. This brings the overall style JSON size down to 943,031&nbsp;characters, more than canceling out the effect of deduplicating names and adding glosses to town labels in #670.

ZeLonewolf (approved):
Looks good to me.  Thanks!

Issue #70, opened 2021-12-28
Deployable style


ZeLonewolf:
The Americana style should be deployable to anyone that wants to self-host.  This includes the following tasks:

- [ ] A style JSON (we'll call it americana.json) is generated at compile time and posted to gh-pages where it is used in favor of the currently dynamically generated object. (#68)
- [ ] Code for generating the style is located in separate folders from the code needed for style runtime (i.e. for shield loading).  The dynamic code is built into a consolidated americana-shields.js that contains all shield generation code and is hosted on gh-pages.  This is currently blocked until shield generation is available after the release of OpenMapTiles 3.13.
- [ ] Sprites are packaged for deployment as artifacts.
- [ ] Documentation is added that's sufficient for a style user to host their own copy, including shields. The Americana runtime would include compiled sprites, the style JSON, and a javascript library for dynamic shield generation.

dschep:
Since this ticket was written, POI have been added which is another obstacle Americana having a deployable `style.json`

PR #590, opened 2022-11-28
Lazily load mapbox-gl-rtl-text plugin

1ec5:
Load the [BSD-licensed](https://github.com/mapbox/mapbox-gl-rtl-text/blob/4ba8fa31f3ce449e8f7dee8191ab3e54a19982e9/LICENSE.md) mapbox-gl-rtl-text plugin from a CDN to ensure that Arabic- and Hebrew-language labels appear with the correct text shaping. The plugin is only loaded when needed.

[Algiers](https://zelonewolf.github.io/openstreetmap-americana/#map=8/36.629/2.998&language=ar)

Fixes #580.

PR #674, opened 2023-01-09
Script to generate style.json

ZeLonewolf:
This PR further refactors the style generation code so that a `style.json` can be generated without access to `window.navigator`. Additionally, I added a standalone script to generate the style JSON.  This can be invoked with:

```
node scripts/generate_style.js > style.json
```

I envision this as a preliminary refactor in support of the goals of #70 as well as a convenience script to generate a style for various development/debugging purposes.

**EDIT**

With the changes based on discussion, this can now be invoked also as:

```
npm run style -l <list of languages> -o <output file>
```
...with the languages defaulting to `mul` and the output file defaulting to stdout if unspecified.


scripts/generate_style.js
```diff
@@ -0,0 +1,11 @@
+"use strict";
```
jleedev, line 1: If you use `import` or `export`, then you syntactically know that it has to be a module and is therefore automatically in strict mode.

scripts/generate_style.js
```diff
@@ -0,0 +1,11 @@
...
+
+import * as Style from "../src/js/style.js";
+
+let style = Style.build(
+  "https://6ug7hetxl9.execute-api.us-east-2.amazonaws.com/data/v3.json",
+  "https://zelonewolf.github.io/openstreetmap-americana/sprites/sprite",
+  ["mul"]
+);
+
+console.log(JSON.stringify(style));
```
jleedev, line 11:
```suggestion
console.log("%j", style);
```

src/americana.js
```diff
@@ -23,27 +22,11 @@ import { default as OpenMapTilesSamples } from "openmapsamples/samples/OpenMapTi
...
-    light: {
-      anchor: "viewport",
-      color: "white",
-      intensity: 0.12,
-    },
-    version: 8,
-  };
+  return Style.build(
+    config.OPENMAPTILES_URL,
+    new URL("sprites/sprite", baseUrl).href,
```
jleedev, line 27: Might (?) work without the href, which is implicitly used whenever a URL is used as a string.

scripts/generate_style.js
```diff
@@ -0,0 +1,11 @@
+"use strict";
+
+import * as Style from "../src/js/style.js";
+
+let style = Style.build(
+  "https://6ug7hetxl9.execute-api.us-east-2.amazonaws.com/data/v3.json",
```
1ec5, line 6: Import config.js to avoid hard-coding this URL in more than one place.

src/js/style.js
```diff
@@ -0,0 +1,27 @@
+"use strict";
+
+import * as Layers from "../layer/index.js";
+
+//Generate style.json
```
jleedev, line 5:
```suggestion
// Generate style.json
```

scripts/generate_style.js
```diff
@@ -0,0 +1,11 @@
...
+  "https://zelonewolf.github.io/openstreetmap-americana/sprites/sprite",
```
1ec5, line 7: If we’re comfortable with hard-coding the sprite base URL here, we might as well do it for the Web code path too, rather than dynamically building the URL based on the containing webpage’s URL.

scripts/generate_style.js
1ec5: Oh, we could remove a lot of `use strict` from this codebase then.

scripts/generate_style.js
```diff
@@ -0,0 +1,11 @@
...
+  ["mul"]
```
1ec5, line 8: This script could take the language code as a [command line argument](https://nodejs.org/en/knowledge/command-line/how-to-parse-command-line-arguments/), falling back to `mul` as a neutral default.

We should document the need to install the [mapbox-gl-rtl-text](https://github.com/mapbox/mapbox-gl-rtl-text/) plugin (#590), especially if the language is set to `mul`.

src/js/style.js
```diff
@@ -0,0 +1,27 @@
...
+export function build(tileURL, spriteURL, locales) {
+  return {
+    id: "streets",
```
1ec5, line 8: I don’t think an ID is even necessary. This ID might’ve come from a GL JS runtime styling example that used Mapbox Streets or something along those lines.

src/americana.js
ZeLonewolf: That gave me
```
Error: sprite: string expected, object found
```
Not sure if there's a JS equivalent to Java's Object.toString()

scripts/generate_style.js
ZeLonewolf: I just blindly pepper it on the top of every .js file because at some point someone told me I should do that :-D

src/americana.js
1ec5: [Yes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString), and [URL overrides it](https://developer.mozilla.org/en-US/docs/Web/API/URL/toString). But `href` is fine too.

jleedev: Ah, never mind then.

(You could say `new URL(…) + ''` e.g. but that would just be less clear.)

scripts/generate_style.js
ZeLonewolf: I haven't fully thought through all the parts of what a "deployable" package would look like. The americana.js file is really becoming the demo_map.js.  The fact that the sprites URL gets baked into the style.json is kind of a problem for deployability. At least it would be one less thing to have to deal with when the repo changes. I guess ultimately we would have to provide some kind of script for people to run to generate the necessary artifacts.

1ec5: #70 has multiple possible answers. Americana’s “demo map” is already deployable as an iframe embedded in a webpage. Exporting `Style.build()` makes this package deployable to a project that uses NPM to build the webpage.

But it looks like you’re working towards an even tighter level of integration, a style JSON that can be dropped into a webpage that already has GL JS installed but doesn’t use NPM. There will be an expectation that the style is _not_ self-contained but rather fully hosted except for the style JSON. In that context, there would probably be little demand for the ability to configure these URLs.

scripts/generate_style.js
ZeLonewolf: Command line arguments for languages with fallback to `mul` implemented in [3c36f42](https://github.com/ZeLonewolf/openstreetmap-americana/pull/674/commits/3c36f42ccb908162180387887e71f793dc78cb6c).

ZeLonewolf: Plugin install requirements documented in 9cca82d.

ZeLonewolf: Plugin install requirements documented in 9cca82d.

src/js/style.js
ZeLonewolf: Fixed in 576064f.  Glad we were able to save a few bytes off of our 1MB style file :-D

.gitignore
```diff
@@ -6,3 +6,4 @@ rebusurance.zip
 config.js
 .parcel-cache
 dist/
+style.json
```
1ec5, line 9: The script doesn’t hard-code this path – it seems to just be a path you happen to be using on the command line. If anything, dist/ (which is already ignored) would be a better home for this JSON file.

ZeLonewolf: Would it be better for the script to dump a file or dump to stdout?

1ec5: It’s a pretty big file, so dumping to a file would make sense. Perhaps this script could take an output path as a command-line option and fall back to stdout if there isn’t one.

ZeLonewolf: The script already uses the command-line option of a varargs list of language codes. Is there a more sophisticated way to handle arbitrary command-line options without reinventing the wheel?

jleedev: I'd suggest [`commander`]:

```
import { Command } from "commander";
const program = new Command();
program
  .option("-l, --locales <locales...>", "language codes")
  .option("-o, --outfile <file>", "output file", "-");
program.parse(process.argv);
console.log("%o", program.opts());
```

```
$ node ./scripts/generate_style.js --help 
Usage: generate_style [options]

Options:
  -l, --locales <locales...>  language codes
  -o, --outfile <file>        output file (default: "-")
  -h, --help                  display help for command
$ node ./scripts/generate_style.js -l it vi 
{ outfile: '-', locales: [ 'it', 'vi', [length]: 2 ] }
```

[`commander`]: https://www.npmjs.com/package/commander

scripts/generate_style.js
ZeLonewolf: I removed the "use strict" from the two files I added.

.gitignore
ZeLonewolf: Added this in a1b136e.  Thanks for the suggestion of commander, it was really easy to work with!

jleedev: Appears to have forgotten dependency change?

ZeLonewolf: What do you mean?

jleedev: `npm i commander`

1ec5 (approved):

Issue #225, opened 2022-03-12
India route shields


1ec5:
India has a system of [distinctive route shields](https://commons.wikimedia.org/wiki/Category:Highway_shield_templates_of_India) for national highways and state highways.

It isn’t currently possible to render route shields for any of these routes because the `ref`s include redundant prefixes like “NH” and “SH”. The prefixes are incorrect, because they shouldn’t appear as part of the main legend on the shield, and they aren’t part of the route’s numeric name. (No one would say “State Highway SH14”.)

Here’s a background image to use for national highways:

![shield40_in_national](https://user-images.githubusercontent.com/1231218/158014104-359623c8-a27f-4968-9272-2fb1b104b2c0.svg)

claysmalley:
[At least one National Highway](https://en.wikipedia.org/wiki/National_Highway_766EE_(India)) has a 5-character ref:

<img width="200" alt="NH766EE-IN" src="https://user-images.githubusercontent.com/1732117/178502239-d38e18b4-965b-4005-b781-fe1e5692ffe9.svg" />

1ec5:
> Another problem is that there are two tagging schemes for national highways: some are tagged `network=IN:NH` while others are tagged `network=IN:NH:<state>`, qualified by a two-letter state code.

A mapper in the OpenStreetMap India Telegram chat [explained](https://t.me/OSMIndia/22036) that `network=IN:NH` was the older tag, but after the Ministry of Road Transport and Highways began renumbering national highways in 2016, they used different `network` values for the new numbers, leaving `IN:NH` as the tag for historic national route relations. `IN:NH` is also used on superrelations representing current national highways that cross state boundaries, but superrelations are irrelevant to this project since OpenMapTiles ignores them.

The new values contain state abbreviations because they wanted to be able to compare OSM’s coverage of national highways with official and commercial sources and break down the statistics by state. So the overly complicated `network` values are a shortcut to avoid having to add an area filter to Overpass queries.

1ec5:
> A mapper in the OpenStreetMap India Telegram chat [explained](https://t.me/OSMIndia/22036) that `network=IN:NH` was the older tag, but after the Ministry of Road Transport and Highways began renumbering national highways in 2016, they used different `network` values for the new numbers, leaving `IN:NH` as the tag for historic national route relations.

The historic route relations are tagged with keys other than `ref`, such as `ref:old`, so these relations are already being omitted correctly.

[...]

PR #675, opened 2023-01-09
Add National Highway shields of India

claysmalley:
Adds support for `network=IN:NH`. Partial implementation of #225.

Screenshot from 2023-01-09 09-26-39
Screenshot from 2023-01-09 09-22-15
Screenshot from 2023-01-09 09-25-39


ZeLonewolf:
The padding seems a bit thick on the top margin, perhaps scoot it north by a pixel or two? Take a look at where the `ref=111` example almost bangs into the curves.

ZeLonewolf:
Does this fix #225 or is it only a partial implementation?

1ec5 (approved):

1ec5:
It’s at most one-third of an implementation.

ZeLonewolf (approved):

Issue #467, opened 2022-06-28
Saudi Arabia highway shields


1ec5:
Saudi Arabia has a three-tiered system of highway shields. The shield icons will be an interesting design challenge.

[<img src="https://upload.wikimedia.org/wikipedia/commons/d/d6/Saudi_Arabia_-_Highway-65.svg" width="64" alt="65">](https://commons.wikimedia.org/wiki/File:Saudi_Arabia_-_Highway-65.svg) [<img src="https://upload.wikimedia.org/wikipedia/commons/4/45/Saudi_Arabia_-_City_road-509.svg" width="64" alt="509">](https://commons.wikimedia.org/wiki/File:Saudi_Arabia_-_City_road-509.svg) [<img src="https://upload.wikimedia.org/wikipedia/commons/d/d5/Saudi_Arabia_-_Branch_Road-2222.svg" width="64" alt="2222">](https://commons.wikimedia.org/wiki/File:Saudi_Arabia_-_Branch_Road-2222.svg)

One national highway has been one mapped so far, tagged [`network=SA:national`](https://www.openstreetmap.org/relation/11366015). (There’s also a redundant partial route tagged [`network=SA:highway`](https://www.openstreetmap.org/relation/8249853).) No city or branch roads have been mapped as relations yet.

So far, routes have only been tagged with the route number in Western European numerals, even though the signs only bear Eastern Arabic numerals.

ZeLonewolf:
I'm inclined to think that the use of western numerals in the `ref` tag is a tagging error when the signage on the ground is using eastern numerals.

michaelblyons:
> I'm inclined to think that the use of western numerals in the `ref` tag is a tagging error when the signage on the ground is using eastern numerals.

I concur. If both are present and either is sufficient to mark the sign (like Iran, it sounds like?) I'm fine with Western numbers. But in the examples above from Saudi Arabia, I imagine it's more helpful to mark the text that appears on the sign. Even if my brain reads the tertiary highway above as "gamma gamma gamma gamma,"\* it'll be more helpful than having to look up which digits correspond to the ones on the sign.

\* Chamelion

1ec5:
[...]

PR #676, opened 2023-01-11
India national expressway shield

1ec5:
Added an India national expressway shield mostly identical to the national highway shield. Added a function to convert European numerals to Roman numerals, which are the distinguishing visual element of this shield.

[Ghaziabad](https://zelonewolf.github.io/openstreetmap-americana/#map=13/28.70246/77.52094)

Working towards #225.

### DIFF
--- a/src/js/shield.js
+++ b/src/js/shield.js
@@ -345,6 +345,34 @@ function getRouteDef(id) {
   };
 }
 
+/**
+ * Reformats an alphanumeric ref as Roman numerals, preserving any alphabetic
+ * suffix.
+ */
+export function romanizeRef(ref) {
+  let number = parseInt(ref, 10);
+  if (isNaN(number)) {
+    return ref;
+  }
+
+  let roman =
+    "M".repeat(number / 1000) +
+    "D".repeat((number % 1000) / 500) +
+    "C".repeat((number % 500) / 100) +
+    "L".repeat((number % 100) / 50) +
+    "X".repeat((number % 50) / 10) +
+    "V".repeat((number % 10) / 5) +
+    "I".repeat(number % 5);
+  roman = roman
+    .replace("DCCCC", "CM")
+    .replace("CCCC", "CD")
+    .replace("LXXXX", "XC")
+    .replace("XXXX", "XL")
+    .replace("VIIII", "IX")
+    .replace("IIII", "IV");
+  return roman + ref.slice(number.toString().length);
+}
+
 export function generateShieldCtx(id) {
   var routeDef = getRouteDef(id);
   var shieldDef = getShieldDef(routeDef);
@@ -361,6 +389,13 @@ export function generateShieldCtx(id) {
   // Handle special case for manually-applied abbreviations
   if (routeDef.ref === "" && shieldDef.refsByWayName) {
     routeDef.ref = shieldDef.refsByWayName[routeDef.wayName];
+  }
+
+  // Convert numbering systems. Normally alternative numbering systems should be
+  // tagged directly in ref=*, but some shields use different numbering systems
+  // for aesthetic reasons only.
+  if (routeDef.ref && shieldDef.numberingSystem === "roman") {
+    routeDef.ref = romanizeRef(routeDef.ref);
   }
 
   var ctx = generateBlankGraphicsContext(shieldDef, routeDef);
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -3372,10 +3372,13 @@ export function loadShields(shieldImages) {
       bottom: 7,
     },
   };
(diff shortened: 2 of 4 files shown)


src/js/shield_defs.js
```diff
@@ -3385,6 +3385,10 @@ export function loadShields(shieldImages) {
       bottom: 7,
     },
   };
+  shields["IN:NE"] = {
+    ...shields["IN:NH"],
+    numberingSystem: "roman",
+  };
```
1ec5, line 3391: I considered adding an “NE” banner up top, similar to the “TWP” disambiguation for some Ohio township routes, because this shield looks so similar to the NH shield at a glance. However, the Roman numerals are probably sufficient.

src/js/shield_defs.js
```diff
@@ -3385,6 +3385,10 @@ export function loadShields(shieldImages) {
       bottom: 7,
     },
   };
+  shields["IN:NE"] = {
+    ...shields["IN:NH"],
+    numberingSystem: "roman",
```
1ec5, line 3390: This touches on #465, but I think the NE shields are different than the other cases in that issue. It seems to be common to refer to the expressways by [European numerals](https://en.wikipedia.org/wiki/Expressways_of_India#National_Expressways) in writing, but shields or the gantry name signs mentioned in https://github.com/ZeLonewolf/openstreetmap-americana/issues/225#issuecomment-1375884357 mostly use Roman numerals. (I’ve come across photos of a few signs that use European numerals, but only as part of an [entrance portal](https://www.shutterstock.com/vi/image-photo/delhi-india-june-18-2022-welcome-2169899597).)

src/js/shield.js
```diff
@@ -314,6 +314,34 @@ function getRouteDef(id) {
...
+    "X".repeat((number % 50) / 10) +
+    "V".repeat((number % 10) / 5) +
+    "I".repeat(number % 5);
+  roman = roman
+    .replace("DCCCC", "CM")
+    .replace("CCCC", "CD")
+    .replace("LXXXX", "XC")
+    .replace("XXXX", "XL")
+    .replace("VIIII", "IX")
+    .replace("IIII", "IV");
```
1ec5, line 341: There are [NPM packages](https://www.npmjs.com/package/romans) for converting to Roman numerals, which we could use instead of rolling our own. Either way, I’m thinking of this as a temporary polyfill until more browsers support numbering systems in [`Intl.NumberFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat).

ZeLonewolf: So `ref` has an Arabic number, but the IRL sign has a roman numeral??

1ec5: That’s correct. I’m asking for confirmation in the India Telegram chat, but so far it seems like the shield is the only context in which someone would use Roman numerals.

ZeLonewolf: Somewhere we discussed countries that use Eastern Arabic Script numerals in their highway shields and I thought the expectation was that the "as shown on the sign" value is what would go in the `ref` tag.  I'm wondering if we're doing the right thing for mapper feedback by doing the roman numeral conversion.

1ec5: That discussion started in #465 and continued in https://github.com/ZeLonewolf/openstreetmap-americana/issues/467#issuecomment-1169170636. But I think the national expressways pose a different sort of challenge. If it turns out that “II” should only appear in a shield, while “2” should appear in text, then we’d be doing a disservice to non-renderers by encouraging mappers to essentially put ASCII art into `ref`.

1ec5: Going back to the subject of this thread, any opinion on pulling in a dependency versus relying on this homegrown implementation?

ZeLonewolf: Oh, I'd defer to people that know how to program on that.

src/js/shield_defs.js
1ec5: I added an “NE” banner:

[Ghaziabad](https://zelonewolf.github.io/openstreetmap-americana/#map=11/28.6911/77.535)

src/js/shield.js
claysmalley: I think this implementation is okay until we can use `Intl.NumberFormat`. Fingers crossed that it won't be long.

claysmalley (approved):
Looks good to me. Ideally we wouldn't need our own function to convert number formats, but I think this is okay as a proof of concept.

ZeLonewolf:
I'm looking for a test location, in order to confirm that this still works in #714.

1ec5:
I always link my screenshots to test locations. 🙂

ZeLonewolf:
Ohhhh

PR #683, opened 2023-01-12
Filter out more transparent road, rail features from legend

### LINKED ISSUE #682: Blank light rail legend entry
At [this location](https://zelonewolf.github.io/openstreetmap-americana/#map=13/38.92278/-77.41692), the legend shows an empty entry for light rail:

image

1ec5:
The legend should never show an entry based on a transparent feature. Unfortunately, some of the layers powering the road and rail entries set `line-opacity` to a `step` expression that can resolve to 0 at some zoom levels. This is normally an antipattern in that it forces the renderer to do more for no apparent reason, but it’s probably necessary for minimizing the number of layers. This PR just prevents the entries from showing up in the legend, causing confusion.

Fixes #682.

### DIFF
--- a/src/js/legend_control.js
+++ b/src/js/legend_control.js
@@ -164,6 +164,14 @@ export default class LegendControl {
     let mainFeature = features[0];
     if (!mainFeature) return;
 
+    // No point in illustrating an invisible line.
+    if (
+      mainFeature.layer.type === "line" &&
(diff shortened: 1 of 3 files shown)


ZeLonewolf (approved):
Works as advertised, thanks for the fix!

PR #685, opened 2023-01-13
Make pentagon shields symmetrical

claysmalley:
I made a mistake in calculating some control points and it ended up in ever-so-slightly lopsided pentagons. This should make them truly symmetrical.

before, after, diff:
old-5 old-980
new-5 new-980
diff-5 diff-980


ZeLonewolf (approved):
LGTM!

PR #687, opened 2023-01-13
Ability to parameterize shield generator

ZeLonewolf:
This PR allows a style deployer to pass parameterized options to the shield generator code.  As an initial capability for this framework, I've included the option to generate non-deterministic shield artwork

### DIFF
--- a/src/americana.js
+++ b/src/americana.js
@@ -59,8 +59,17 @@ map.on("styledata", function (event) {
   ShieldDef.loadShields(map.style.imageManager.images);
 });
 
+function getOptionsFromURL() {
+  const params = new URLSearchParams(window.location.hash);
+  return {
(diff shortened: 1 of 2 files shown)


michaelblyons:
It is genuinely difficult to internalize that there's a legitimate real-world use case for the phrase "shield generator."

ZeLonewolf:
Well, it's entirely feasible that someone might want to make a map with shields that are stylistically different.

michaelblyons:
I know, I know. I was making what I hoped was identifiable as a joke. This is the first time I've seen someone use the phrase "shield generator" who wasn't writing science fiction about, say, a raid on the moon of Endor. I should have been less circumspect.

Carry on. 🙂

PR #688, opened 2023-01-14
Right-click context menu

ZeLonewolf:
This adds a basic right-click context menu to the map, where we can start to build out feature inspection tools.

Initially, this menu links to the existing osm.org view/edit pages.

image


ZeLonewolf:
Realistically, this needs a complete do-over.

Issue #435, opened 2022-06-18
Community anchor institutions


1ec5:
The map should have point of interest icons for community anchor institutions. These are mostly public-sector points of interest that define a typical populated place.

Every map has a different selection of community anchor institutions, but some common ones on American maps are city halls, courthouses, police stations, fire stations, hospitals (being added in #387), schools (already tracked in #78), libraries, and post offices. Sometimes community centers, museums, and banks are also included. A rule of thumb is that the name of a community anchor institution is somewhat unimportant on the map, because it’s usually named after the community (or vice versa).

Some maps have dedicated icons for each kind of institution, while others pair the generic POI icon with an abbreviated label. Community anchor institutions appear on many maps that don’t include retail POIs in general, so I think it would be reasonable to introduce them at a lower zoom level than the general POI layer. At these lower zoom levels, it would be unnecessary to label the POI by its `name:en`.

<img width="280" alt="Engels (Huff) Hamilton County 1997" src="https://user-images.githubusercontent.com/1231218/174456035-c90a2298-b515-49dc-bca8-c663ca2078f4.png"> <img width="268" alt="Rand McNally Dallas 1994" src="https://user-images.githubusercontent.com/1231218/174456064-848d920e-b084-4cc9-93a0-8014ad166901.png"> <img width="213" alt="Rand McNally Atlanta 2000" src="https://user-images.githubusercontent.com/1231218/174456153-9f2204e9-1e12-40ed-ad12-28f50e73c872.png">

Among these POI types, post offices will be trickier. We’d want to distinguish U.S. Post Offices from storefronts for delivery services (The UPS Store, FedEx Office, and more obscure services). Even though the USPS may be of diminished importance these days, their rural locations in particular are of principal importance to the communities they serve, more than any other POI. Thanks to the [name suggestion index](https://nsi.guide/index.html?t=operators&k=amenity&v=post_office&tt=usps), either `operator:type=public` or `operator:wikidata=Q668687` could be used to make this distinction. However, I don’t think either key is currently recognized by the OpenMapTiles schema.

zekefarwell:
That list looks good.  I'd add:

- Churches
- Parks
- Airports 

Churches may require a variety of icons for different religions, but they are often quite prominent buildings that serve as landmarks even for those who never set foot inside.

[...]

PR #689, opened 2023-01-14
Add school icons

### LINKED ISSUE #78: Schools
Even though this project is currently focused on the lower zoom levels with large features like roads, I think it’s time to start thinking about what we’d need for proper rendering of points of interest at higher zoom levels, because the tagging system for some points of interest may fit American cartographic needs even less well than for transportation features.

For sure, the style would at some zoom levels mark schools, which are important community institutions that serve as a point of reference for other places around town. The description below is focused on what a reasonable end result might look like, but it’s likely that we’d need to focus on coverage of certain tagging schemes to make the suggested treatments work.

## Icon

Conventionally, street- and community-level maps of the U.S. depict schools as pitch-roofed buildings with a pennant-shaped flag on top.

Typically, community institutions (such as schools, post offices, hospitals, and police and fire departments) are marked by only a simple icon at middle zoom levels where there wouldn’t be enough room for a full label. However, some maps pack a bit more information in school icons:

* Some place a small letter inside the building to indicate the school’s classification – typically “E” for elementary, “M” for middle school, and “H” for high school, but other initials are possible.
* Some highlight middle and high schools with a darker fill than elementary schools to emphasize the relative size and importance.
* Some place more than one flag on the roof to flag an institution of higher education.
* Some replace the pennant with a religious symbol such as a cross for religious schools. (The color or letter distinguishes the religious school icon from the church icon.)

  Champion (Mobil) Buffalo 1989 Seeger Cincinnati 1991 Rand McNally Atlanta 2000 Mapsco Fort Worth 1992 legend Universal Florida 1995

(Some maps in my collection use the flag-on-house symbol for a fairgrounds, and some use a filled square for educational institutions, but those seem to be in the minority.)

## Label

I recall seeing some maps that use the label instead of the icon to indicate a school’s classification. For example, it might say “(Grades 1–8)” in fine print beneath the school name.

## Fill

Many maps fill school buildings in a darker or more brilliant color than other buildings, to give them more prominence.

Some maps also fill schoolgrounds, particularly the ones that are at a high enough scale that they don’t need to show an icon.

wmisener:
Adds school POI icons. Fixes #78, and makes progress on #435.

In #78, a wide variety of US road maps were shown, a sizeable proportion of which displayed schools as some variation of a flat-roofed building topped by a pennant-shaped flag and flagpole. The icon I propose adheres to this design, although there's room for stylistic choice regarding the location, size, and direction of the flag.
poi_school

I colored the icon "infrastructure blue" to adhere to the [Contributing Guide](https://github.com/ZeLonewolf/openstreetmap-americana/blob/main/CONTRIBUTING.md#points-of-interest), with the specified 2px-wide knockout. In the future, a separate color could be considered, especially in conjunction with an area fill if one is desired.
Location: San Francisco, CA ([localhost link](http://localhost:1776/#map=16.2/37.777946/-122.489602))
Screen Shot 2023-01-13 at 10 55 29 PM

Right now I've selected it to appear at zoom 15, the same as hospitals. Browsing around, it may be better for both of these icons to appear even earlier, but a decision can wait until there are more POIs to compare to.
Location: Beverly Hills, CA ([localhost link](http://localhost:1776/#map=15.53/34.061426/-118.412763))
Screen Shot 2023-01-13 at 10 53 48 PM

This PR does not render colleges/universities or preschools. It also does not differentiate between different grade levels, as proposed in #78. It's unclear if such differentiation is possible with current tagging practices, but if so, there's probably room in the "building" icon, or to the right of the flagpole, for a letter. There's also no distinction between public/private/religious, though that could also be added in the future, perhaps through inverting the fill or changing the flag to a different symbol.
Location: Salinas, CA ([localhost link](http://localhost:1776/#map=15/36.66654/-121.66473))
Screen Shot 2023-01-13 at 10 54 16 PM

Also adds a legend entry, which is just "School".
Location: Kingman, AZ ([localhost link](http://localhost:1776/#map=15.01/35.19116/-114.06062))
Screen Shot 2023-01-13 at 10 59 37 PM

Interaction with other features:
Location: New York, NY ([localhost link](http://localhost:1776/#map=16.18/40.715686/-73.997471))
Screen Shot 2023-01-14 at 12 16 08 AM

This is my first PR, and I think the first PR for POIs after the original (#387), so I hope I haven't missed anything!

### DIFF
--- a/icons/poi_school.svg
+++ b/icons/poi_school.svg
@@ -1,6 +1,5 @@
 <svg width="20" height="20" xmlns="http://www.w3.org/2000/svg">
- <g style="display:inline">
-  <path style="display:inline;fill:#f9f5f0;fill-opacity:1;stroke:#fff;stroke-width:0;stroke-linecap:round;paint-order:stroke markers fill" d="M 9,0 A 2.0002,2.0002 0 0 0 8.2304688,0.15429688 l -6,2.50000002 a 2.0002,2.0002 0 0 0 0,3.6914062 L 6.1992188,8 H 6 a 2.0002,2.0002 0 0 0 -2,2 v 8 a 2.0002,2.0002 0 0 0 2,2 h 8 a 2.0002,2.0002 0 0 0 2,-2 V 10 A 2.0002,2.0002 0 0 0 14,8 H 13 V 2 A 2.0002,2.0002 0 0 0 11,0 Z"/>
-  <path style="display:inline;fill:#003f87;fill-opacity:1;stroke:#fff;stroke-width:0;stroke-linecap:round;paint-order:stroke markers fill" d="M 9.0000001,1.9999999 3,4.5 9.0000001,6.9999993 Z M 6,10 v 8 h 8 V 10 Z M 9,2 v 8 h 2 V 2 Z"/>
+ <g style="display:inline;stroke:#000;stroke-opacity:0">
+  <path style="display:inline;fill:#003f87;fill-opacity:1;stroke:#f9f5f0;stroke-width:2;stroke-linecap:round;stroke-linejoin:round;stroke-dasharray:none;stroke-opacity:1;paint-order:stroke markers fill" d="M 11,1 3,4 9,6 Z M 5,9 V 19 H 15 V 9 Z M 9,6 v 3 h 2 V 1 Z"/>
  </g>
 </svg>
--- a/icons/poi_school.svg
+++ b/icons/poi_school.svg
@@ -1,57 +1,6 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<svg
-   width="20"
-   height="20"
-   version="1.1"
-   id="svg4"
-   sodipodi:docname="poi_school.svg"
-   inkscape:version="1.2 (dc2aeda, 2022-05-15)"
-   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
-   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
-   xmlns="http://www.w3.org/2000/svg"
-   xmlns:svg="http://www.w3.org/2000/svg">
-  <defs
-     id="defs8" />
-  <sodipodi:namedview
-     id="namedview6"
-     pagecolor="#505050"
-     bordercolor="#eeeeee"
-     borderopacity="1"
-     inkscape:showpageshadow="0"
-     inkscape:pageopacity="0"
-     inkscape:pagecheckerboard="0"
-     inkscape:deskcolor="#505050"
-     showgrid="true"
-     inkscape:zoom="18.847001"
-     inkscape:cx="7.587414"
-     inkscape:cy="8.7812379"
-     inkscape:window-width="1309"
-     inkscape:window-height="713"
-     inkscape:window-x="0"
-     inkscape:window-y="23"
-     inkscape:window-maximized="0"
-     inkscape:current-layer="g706">
-    <inkscape:grid
-       type="xygrid"
-       id="grid632"
-       originx="0"
-       originy="0" />
-  </sodipodi:namedview>
-  <g
-     id="g706"
-     style="display:inline">
-    <path
-       id="path941"
-       style="display:inline;fill:#f9f5f0;fill-opacity:1;stroke:#ffffff;stroke-width:0;stroke-linecap:round;paint-order:stroke markers fill"
-       sodipodi:type="inkscape:offset"
-       inkscape:radius="2"
-       inkscape:original="M 9 2 L 3 4.5 L 9 7 L 9 10 L 6 10 L 6 18 L 14 18 L 14 10 L 11 10 L 11 2 L 9 2 z "
-       d="M 9,0 A 2.0002,2.0002 0 0 0 8.2304688,0.15429688 l -6,2.50000002 a 2.0002,2.0002 0 0 0 0,3.6914062 L 6.1992188,8 H 6 a 2.0002,2.0002 0 0 0 -2,2 v 8 a 2.0002,2.0002 0 0 0 2,2 h 8 a 2.0002,2.0002 0 0 0 2,-2 V 10 A 2.0002,2.0002 0 0 0 14,8 H 13 V 2 A 2.0002,2.0002 0 0 0 11,0 Z"
-       inkscape:label="path941" />
-    <path
-       id="path6157"
-       style="display:inline;fill:#003f87;fill-opacity:1;stroke:#ffffff;stroke-width:0;stroke-linecap:round;paint-order:stroke markers fill"
-       d="M 9.0000001,1.9999999 3,4.5 9.0000001,6.9999993 Z M 6,10 v 8 h 8 V 10 Z M 9,2 v 8 h 2 V 2 Z"
-       sodipodi:nodetypes="cccccccccccccc" />
-  </g>
+<svg width="20" height="20" xmlns="http://www.w3.org/2000/svg">
+ <g style="display:inline">
+  <path style="display:inline;fill:#f9f5f0;fill-opacity:1;stroke:#fff;stroke-width:0;stroke-linecap:round;paint-order:stroke markers fill" d="M 9,0 A 2.0002,2.0002 0 0 0 8.2304688,0.15429688 l -6,2.50000002 a 2.0002,2.0002 0 0 0 0,3.6914062 L 6.1992188,8 H 6 a 2.0002,2.0002 0 0 0 -2,2 v 8 a 2.0002,2.0002 0 0 0 2,2 h 8 a 2.0002,2.0002 0 0 0 2,-2 V 10 A 2.0002,2.0002 0 0 0 14,8 H 13 V 2 A 2.0002,2.0002 0 0 0 11,0 Z"/>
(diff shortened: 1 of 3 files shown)


ZeLonewolf:
Thanks, this looks great! I appreciate the retro school icon from old maps. Really fits the "Americana" theme in the traditional sense.

I did notice that the knockout on the icon seems to be thicker than that of the other icons, and I wonder if the contributing guidelines need to be tweaked.

Below are the settings I used in Inkscape on the coffee cup, and I want to specifically point out the "Order" section with the six buttons.  I'm not sure how I arrived at 2.555 pixels but I'm pretty sure if I'd been more competent about graphic design, I'd have aligned the icon to the grid in a more sensible way:

image

claysmalley:
Shields only have a 1px border, so maybe we should stick with a 1px knockout.

Should we use this POI icon for `college` and `university` as well?

ZeLonewolf:
1px sounds right, so if that 2.5555 were actually 2, that would result in the 1px knockout.

claysmalley:
It's the other way around. It should be 1px in the file, and it'll show up as 2px at 200% zoom.

My workflow when resizing an icon is to first disable the stroke, then scale the icon to the desired dimensions subtracted by 1px, then re-add the 1px stroke.

Screenshot from 2023-01-14 14-15-58

ZeLonewolf:
I meant that if you change the overlap order to the top-center option, half of the stroke width is covered by the fill.  So my 2.555px stroke was actually 1.25px ish.

claysmalley:
You're right, my bad :smile: 

Workflow is pretty much the same in that case, just use 2px instead of 1px.

wmisener:
Here's a version with a 1px-wide knockout (i.e., the stroke width is 2px, but the fill covers half of it), does it seem better?
poi_school_1pxknockout

Since shrinking the knockout width left extra room to get to 20px, I made the building square bigger, but kept the flag the same size. I think I like the proportions better now, but don't feel that strongly if others want a bigger flag.

Some screenshots
New York
Screen Shot 2023-01-14 at 11 46 48 AM

Salinas
Screen Shot 2023-01-14 at 11 47 00 AM

Visually, this smaller knockout seems like it's closer to the text and shield knockouts, and less than the cafe and bar ones.

wmisener:
> Should we use this POI icon for `college` and `university` as well?

We could. I left them out for now, since a lot of maps differentiate them. A few posted in #78 add a second flag, but there are other icons available, like a mortarboard cap or something. Higher education institutions tend to be more important in a city-wide orientation sense than schools, so they should probably appear at lower zooms as well. Maybe I'll open a separate issue.

ZeLonewolf:
It is less thickness for sure than the coffee/bar icons, but based on the discussion above, it sounds like that's because the coffee/bar knockouts are wrong :)

ZeLonewolf:
I'm happy leaving higher education for a separate PR/discussion, it doesn't have to be tackled here.

icons/poi_school.svg
```diff
@@ -0,0 +1,6 @@
+<svg width="20" height="20" xmlns="http://www.w3.org/2000/svg">
+ <g style="display:inline">
+  <path style="display:inline;fill:#f9f5f0;fill-opacity:1;stroke:#fff;stroke-width:0;stroke-linecap:round;paint-order:stroke markers fill" d="M 9,0 A 2.0002,2.0002 0 0 0 8.2304688,0.15429688 l -6,2.50000002 a 2.0002,2.0002 0 0 0 0,3.6914062 L 6.1992188,8 H 6 a 2.0002,2.0002 0 0 0 -2,2 v 8 a 2.0002,2.0002 0 0 0 2,2 h 8 a 2.0002,2.0002 0 0 0 2,-2 V 10 A 2.0002,2.0002 0 0 0 14,8 H 13 V 2 A 2.0002,2.0002 0 0 0 11,0 Z"/>
+  <path style="display:inline;fill:#003f87;fill-opacity:1;stroke:#fff;stroke-width:0;stroke-linecap:round;paint-order:stroke markers fill" d="M 9.0000001,1.9999999 3,4.5 9.0000001,6.9999993 Z M 6,10 v 8 h 8 V 10 Z M 9,2 v 8 h 2 V 2 Z"/>
```
1ec5, line 4: Bring the top of the flagpole to a sharper point. This may technically distort the pennant a little, but as an optical illusion, the flagpole won’t look curved at actual size.

Pengor:
Nice work! I'm not sure that the specific feature name (i.e. the name of the school) needs to be shown in the legend though, I think simply showing the icon and "School" is sufficient and saves space

ZeLonewolf:
> Nice work! I'm not sure that the specific feature name (i.e. the name of the school) needs to be shown in the legend though, I think simply showing the icon and "School" is sufficient and saves space

That is a generic capability of the label, not a specific feature implemented by this PR.

icons/poi_school.svg
wmisener: I'm not sure I understand what you mean, something like this?
poi_school_1pxknockout_point

wmisener: Maybe easier to see in inkscape
Screen Shot 2023-01-14 at 3 48 17 PM

1ec5:
> I'm not sure that the specific feature name (i.e. the name of the school) needs to be shown in the legend though, I think simply showing the icon and "School" is sufficient and saves space

#691

icons/poi_school.svg
1ec5: That’s better, but we could even do away with the bevel:

sharp

Pengor:
> That is a generic capability of the legend, not a specific feature implemented by this PR.

Ah, gotcha, thanks!

icons/poi_school.svg
wmisener: Ah understood, that is kind of nice I think
Screen Shot 2023-01-14 at 4 12 02 PM

src/layer/poi.js
```diff
@@ -7,6 +7,7 @@ var iconDefs = {
   hospital: "hospital",
   medical: ["doctors", "clinic"],
   parking: "parking",
+  school: "school",
```
1ec5, line 10: Kindergartens are unlikely to have very distinct icons from other schools, so we could match `kindergarten` at the same time.

1ec5, line 10: > This PR does not render colleges/universities or preschools.

In the absence of a separate entry for colleges and universities, calling this entry “school” would imply that it also includes colleges and universities. Maybe we should match `college` and `university` at the same time for now, then split them out once we have a distinct icon or treatment for them.

ZeLonewolf:
I believe we also need to make additions to our taginfo file.

src/layer/poi.js
ZeLonewolf, line 10:
```suggestion
  school: ["kindergarten", "school", "college", "university"],
```

src/layer/poi.js
```diff
@@ -34,7 +35,7 @@ export const poi = {
     [
       "match",
       ["get", "subclass"],
-      "hospital",
+      ["hospital", "school"],
```
ZeLonewolf, line 38:
```suggestion
      ["hospital", ...iconDefs.school],
```

src/layer/poi.js
```diff
@@ -108,4 +111,9 @@ export const legendEntries = [
     layers: [poi.id],
     filter: ["==", ["get", "subclass"], iconDefs.parking],
   },
+  {
+    description: "School",
+    layers: [poi.id],
+    filter: ["==", ["get", "subclass"], iconDefs.school],
+  },
```
ZeLonewolf, line 118:
```suggestion
  {
    description: "School",
    layers: [poi.id],
    filter: ["in", ["get", "subclass"], ["literal", iconDefs.school]],
  },
```

src/layer/poi.js
wmisener: As I mentioned in the OP, I do think that down the line these features should render separately, probably in icon but definitely in zoom level. But at this stage, I can see the rationale in having the spirit of getting more categories rendering at all.

wmisener: Kindergartens are now included

wmisener: Colleges and universities are now included

wmisener:
I've added colleges, universities, and kindergartens to the PR. I'll open separate issues to track rendering these differently than primary/secondary schools.

> I believe we also need to make additions to our taginfo file.

I *think* I've done this correctly? Someone should probably double check...

ZeLonewolf:
> > I believe we also need to make additions to our taginfo file.
> 
> I _think_ I've done this correctly? Someone should probably double check...

There isn't really a way to test it beyond checking taginfo the next day, so generally if it doesn't blow up the linter, all we can do is go by eyeball and post a change later if something is wrong.  In any case, if our taginfo JSON has a problem, it's not a big deal if our project isn't listed in taginfo for a day!

wmisener:
Test renderings:
Preschool: Kigala Preschool ([localhost link](http://localhost:1776/#map=16.17/34.023767/-118.464327))
Screen Shot 2023-01-15 at 9 22 50 AM

College: City College of San Francisco ([localhost link](http://localhost:1776/#map=15.2/37.724285/-122.452126))
Screen Shot 2023-01-15 at 9 21 31 AM

University: University of California, Los Angeles ([localhost link](http://localhost:1776/#map=15.13/34.070675/-118.446188))
Screen Shot 2023-01-15 at 9 20 48 AM

ZeLonewolf (approved):
So many schools!

Thanks for this work, the result looks great. There is going to be some tail work I think around prioritizing POIs and tweaking zoom levels, but I think we need to get more icons on the map before we can really start to do that work. So I think this is good to go.

image

wmisener:
Preschools are tracked in #693, colleges and universities are tracked in #694.

PR #690, opened 2023-01-14
Add low-zoom tileserver source

ZeLonewolf:
This PR adds a low-zoom tileserver source.

I've added a second .mbtiles source to the [AWS tileserver](https://6ug7hetxl9.execute-api.us-east-2.amazonaws.com/) which is rendered using openmaptiles-tools from a bleeding-edge openmaptiles schema.  This mechanism allows us to test out the style against new features which have been added to openmaptiles but aren't yet available in planetiler.

This specific source (low-zoom) is filtered to show administrative boundaries, water features, and some place labels.  Because of the time/resources needed for openmaptiles-tools rendering, it's necessary to be selective in what we render. My motivation for this particular extract was to test out the rendering of large water-body labels.  There's a few glitches in the tile data, such as some missing boundaries, which seems to be an artifact of osmium tags-filter.  I described this process in a recent OpenStreetMap [diary entry](https://www.openstreetmap.org/user/ZeLonewolf/diary).

image

If there is other "in openmaptiles but not yet released" functionality that anyone would like me to add, please make a request by all means.  If we need to test unreleased high-zoom functionality, we can do that by means of a geographic filter (for example, choosing one country or state to render to full zoom).


ZeLonewolf:
I have a better idea

PR #697, opened 2023-01-16
Shield for German Bundesstraße

ZeLonewolf:
The German Bundesstraße (second-level network) is signed with yellow signs and black borders:
image

It appears to have [significant proliferation](https://overpass-turbo.eu/s/1qbI) through the country, with some of these relations having hundreds of versions. For example, the [route relation for B85](https://www.openstreetmap.org/relation/2050993) has more than 200 versions:
image

A German-language community forums [discussion](https://community.openstreetmap.org/t/network-an-fernstrassen/7089/29) seems to support tagging this network as `network=DE:national`.

image


claysmalley:
The `ref` values seem to have the `B` prefix despite not appearing on signage. However, the map style on [openstreetmap.de](https://openstreetmap.de/karte/), which processes `ref` from ways, does include this prefix. It's unclear whether the prefix in the relation's `ref` is just an oversight, or German readers truly expect it to be on the map.

ZeLonewolf:
It's standard practice to include network information in the ref, for example `ref=I-95` would be typical for US Interstate 95 way refs. So I think it's expected that the German style would show the full value. I'm not sure if the prefix being on the map is expected, or just a tagging error that replicates way-style tagging to relations, but either way I'm fine to let Germans figure that one.

What I'm not clear on is whether this network value properly corresponds to the symbology in this PR.

Mashin6:
The prefix is probably essential when displaying the ref in a generic rectangular shield in order to clearly identify which network is being displayed. However on a graphically accurate shield this becomes redundant and I would say even distracting.

claysmalley:
> I'm not sure if the prefix being on the map is expected, or just a tagging error that replicates way-style tagging to relations, but either way I'm fine to let Germans figure that one.

Fair enough. I concur with Martin that the prefixes probably don't belong in relations, though that's an issue for the Germans to handle.

> What I'm not clear on is whether this network value properly corresponds to the symbology in this PR.

The colors are correct, but Bundesstraßen have a fixed-width rectangular shield. Let's set the width to 34px like we've done with other ones.

ZeLonewolf:
Looks good at 34px:

image

claysmalley (approved):
Looks good!

PR #699, opened 2023-01-16
Segregate shield generation code and compile for separate library

ZeLonewolf:
This PR modularizes the shield generation code as follows:
* Code is moved to its own folder, `src/shieldjs`
* Modify the build script in `scripts/build.js` to produce a separate shield.js. This produces dist-shieldjs with a consolidated shield.js.

This is intended to be an incremental step towards packaging the project as described in #68 and #70.

Additional work needed past this PR:
* An upgrade to one of the CI libraries is needed in #698 to support deployments to github pages.
* Improve modularity with the common libraries (currently, unnecessary code is brought in from common files)
* Publish artifacts to gh-pages
* Publish library to npm

PR #701, opened 2023-01-16
Convert roundedRectangle to JSON definition

ZeLonewolf:
Unblocks #699 

This PR implements an incremental step towards #118 by converting the `roundedRectangle()` function into a JSON definition.  This still leaves in place the existing function pointers for other draw functions so that the conversion can be done incrementally and in stages.  Ultimately the goal is that the shield definitions can be entirely implemented as a JSON, which allows for the possibility of a shield generator that can work on native as well as the existing one for the web.

Instead, the shield definitions would implement something that looks like this:
```
draw: {
      drawFunc: "roundedRectangle",
      params: {
        fillColor: "white",
        strokeColor: "black",
        textColor: "black",
        rectWidth: 20,
      },
    },
```

This would then get extended to other drawing functions. Additionally, I added:

```
export function registerDrawFunction(name, fxn) {
  drawFunctions[name] = fxn;
}
```

This function would allow some other implementer the ability to define their own custom shield drawing functions (in addition to the built-in ones that will also call this function internally).

This approach will require standardizing all the draw functions to take in a `ref` and `options` object, where the `options` contains any color/style variations that need to be passed in.

CC/#577 

If the team is satisfied with this general approach, my intent is to apply this to the other drawing functions, and then move on to the text layout functions.  That will remove the major blockers to having a pure-JSON shield definition.

### DIFF
--- a/src/js/shield_canvas_draw.js
+++ b/src/js/shield_canvas_draw.js
@@ -125,14 +125,13 @@ export function blank(ref) {
   return Gfx.getGfxContext({ width: width, height: CS });
 }
 
-export function roundedRectangle(
-  fill,
-  outline,
-  ref,
-  radius,
-  outlineWidth,
-  rectWidth
-) {
+function roundedRectangle(ref, params) {
+  let fill = params.fillColor == undefined ? "white" : params.fillColor;
+  let outline = params.strokeColor == undefined ? "black" : params.strokeColor;
+  let radius = params.radius == undefined ? 1 : params.radius;
+  let outlineWidth = params.outlineWidth == undefined ? 1 : params.outlineWidth;
+  let rectWidth = params.rectWidth == undefined ? null : params.rectWidth;
+
   if (rectWidth == null) {
     var shieldWidth =
       ShieldText.calculateTextWidth(ref, genericShieldFontSize) + 2 * PXR;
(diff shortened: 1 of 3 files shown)


src/js/shield_canvas_draw.js
```diff
@@ -125,14 +125,13 @@ export function blank(ref) {
...
-  outline,
-  ref,
-  radius,
-  outlineWidth,
-  rectWidth
-) {
+function roundedRectangle(ref, params) {
+  let fill = params.fillColor == undefined ? "white" : params.fillColor;
+  let outline = params.strokeColor == undefined ? "black" : params.strokeColor;
+  let radius = params.radius == undefined ? 1 : params.radius;
```
claysmalley, line 131: Undefined corner radius should probably default to `0`.

claysmalley (changes requested):
Allegheny County belts and Branson routes are broken:

Screenshot from 2023-01-16 13-18-48
Screenshot from 2023-01-16 13-19-32

ZeLonewolf:
Oops, guess I forgot about other draw functions that call roundedRectangle!

src/js/shield_defs.js
```diff
@@ -69,15 +69,15 @@ function roundedRectShield(
...
   radius = radius ?? 2;
   return {
-    backgroundDraw: (ref) =>
-      ShieldDraw.roundedRectangle(
-        fillColor,
-        strokeColor,
-        ref,
-        radius,
-        1,
-        rectWidth
```
ZeLonewolf, line 79: @claysmalley this is where the original code defaulted to a 1px radius, so I just moved that downstream.

ZeLonewolf:
Fixes in place for Branson and PA dot routes.  Looks like those were the only internal users of the rounded rectangle function.

claysmalley:
Once the default behavior is changed to no corner radius, everything should be good to go.

src/js/shield_defs.js
ZeLonewolf: Ha, all the more reason why the options object is cleaner.  Done.

claysmalley (approved):
Looks good! This is a helpful starting point for JSONifying the rest of the draw functions.

PR #706, opened 2023-01-17
Refactor circle/ellipse draw functions to JSON

ZeLonewolf:
A followup from #701.  This PR converts circle and ellipse draw functions to JSON.


claysmalley (approved):
Looks good to me.
Screenshot from 2023-01-16 20-31-08

michaelblyons:
Does it not look like there's more padding above the number than below?

ZeLonewolf:
Yes, it does.  There might be an issue with padding or text constraint.  Please open a bug.

ZeLonewolf:
Partial implementation of #118

PR #707, opened 2023-01-17
Upgrade deploy scripts

### LINKED ISSUE #698: Upgrade CI to upload-artifact v3
We should upgrade to v3 of https://github.com/actions/upload-artifact.

Among other features, it allows for multiple directories to be used as the upload source, which will be helpful in modularizing the codebase.

ZeLonewolf:
Fixes #698 (in theory)

This PR upgrades the deploy script based on some dodgy instructions I found on the Internet.  There's a very real possibility this won't work.

I want to upgrade the deploy script to take advantage of new functionality that allows us to upload multiple artifacts.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,6 +1,8 @@
-# Taken from https://github.com/marketplace/actions/deploy-to-github-pages
+# Taken from https://github.com/actions/upload-pages-artifact/blob/main/action.yml
+# and https://github.com/marketplace/actions/deploy-to-github-pages
 
 name: Build and Deploy
+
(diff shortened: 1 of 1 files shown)


jleedev:
If I'm reading this correctly, it's a copy paste of the "tar" step of the action so that it can be customized later?

Upgrading to the latest version of the action makes sense, but I would hold off on rewriting the step until it's actually used for something specific.

jleedev:
Also looks like the latest version supports specifying multiple paths: https://github.com/actions/upload-artifact#upload-using-multiple-paths-and-exclusions

So this would be unnecessary.

PR #708, opened 2023-01-17
Refactor remaining shape draw functions to JSON

ZeLonewolf:
This continues the refactor started in #701 and #706 by converting the remaining draw functions to JSON.  Additionally, I moved the two custom drawing functions to their own file, which gives a nice demonstrator of how to add your own custom shields.  This removes the dependency of shield.js on color.js, which was the major blocker on #699.

### DIFF
--- a/src/js/custom_shields.js
+++ b/src/js/custom_shields.js
@@ -1,0 +1,73 @@
+"use strict";
+
+import * as Color from "../constants/color.js";
+import * as ShieldDraw from "./shield_canvas_draw.js";
+
+// Special case for Allegheny, PA Belt System, documented in CONTRIBUTE.md
+export function paBelt(params) {
+  let ctx = ShieldDraw.roundedRectangle({
+    fillColor: Color.shields.white,
+    strokeColor: Color.shields.black,
+    outlineWidth: 1,
+    radius: 2,
+    rectWidth: 20,
+  });
+
+  let fillColor = params.fillColor;
+
+  let lineWidth = 0.5 * ShieldDraw.PXR;
+  let diameter = ShieldDraw.CS / 3 - lineWidth;
+  ctx.beginPath();
+  ctx.arc(
+    ShieldDraw.CS / 2,
+    ShieldDraw.CS / 2,
+    diameter,
+    0,
+    2 * Math.PI,
+    false
+  );
+
+  ctx.fillStyle = fillColor;
+  ctx.strokeStyle = Color.shields.black;
+  ctx.fill();
+
+  ctx.lineWidth = lineWidth;
+  ctx.stroke();
+  return ctx;
+}
+
+// Special case for Branson color-coded routes, documented in CONTRIBUTE.md
+export function bransonRoute(params) {
+  var ctx = ShieldDraw.roundedRectangle({
+    fillColor: Color.shields.green,
+    strokeColor: Color.shields.white,
+    outlineWidth: 1,
+    radius: 2,
+    rectWidth: 20,
+  });
+
+  let fillColor = params.fillColor;
+
+  let lineWidth = 0.5 * ShieldDraw.PXR;
+  let x = 0.15 * ShieldDraw.CS + lineWidth;
+  let width = 0.7 * ShieldDraw.CS - 2 * lineWidth;
+
+  let y = 0.4 * ShieldDraw.CS + lineWidth;
+  let height = 0.45 * ShieldDraw.CS - 2 * lineWidth;
+
+  ctx.beginPath();
+  ctx.rect(x, y, width, height);
+
+  ctx.fillStyle = fillColor;
+  ctx.strokeStyle = Color.shields.white;
+  ctx.fill();
+
+  ctx.lineWidth = lineWidth;
+  ctx.stroke();
+  return ctx;
+}
+
+export function loadCustomShields() {
+  ShieldDraw.registerDrawFunction("branson", bransonRoute);
+  ShieldDraw.registerDrawFunction("paBelt", paBelt);
+}
--- a/src/js/custom_shields.js
+++ b/src/js/custom_shields.js
@@ -28,7 +28,7 @@ export function paBelt(params) {
   );
 
   ctx.fillStyle = fillColor;
-  ctx.strokeStyle = Color.shields.black;
+  ctx.strokeStyle = params.strokeColor;
   ctx.fill();
 
   ctx.lineWidth = lineWidth;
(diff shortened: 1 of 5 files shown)


src/js/shield_defs.js
```diff
@@ -246,6 +246,15 @@ function diamondShield(fillColor, strokeColor, textColor, radius, rectWidth) {
...
+      drawFunc: "diamond",
+      params: {
+        fillColor: fillColor,
+        strokeColor: strokeColor,
+        radius: radius,
+        rectWidth: rectWidth,
+      },
+    },
     backgroundDraw: (ref) =>
       ShieldDraw.diamond(fillColor, strokeColor, ref, radius, 1, rectWidth),
```
claysmalley, line 259: Should `backgroundDraw` still be here?

src/js/shield_defs.js
```diff
@@ -487,6 +392,16 @@ function hexagonVerticalShield(
...
+      drawFunc: "hexagonVertical",
+      params: {
+        offset: offset,
+        fillColor: fillColor,
+        strokeColor: strokeColor,
+        radius: radius,
+        rectWidth: rectWidth,
+      },
+    },
     backgroundDraw: (ref) =>
```
claysmalley, line 405: Another redundant `backgroundDraw`.

src/js/shield_defs.js
```diff
@@ -578,17 +493,17 @@ function octagonVerticalShield(
...
-        ref,
-        radius,
-        1,
-        rectWidth
-      ),
+    canvasDrawnBlank: {
+      drawFunc: "octagonVertical",
+      params: {
+        offset: offset,
+        angle: 0,
```
claysmalley, line 500: Why is `angle` set to 0 instead of `angleInRadians`?

src/js/shield_defs.js
```diff
@@ -340,19 +296,18 @@ function pentagonUpShield(
...
-        strokeColor,
-        ref,
-        radius1,
-        radius2,
-        1,
-        rectWidth
-      ),
+    canvasDrawnBlank: {
+      drawFunc: "pentagon",
+      params: {
```
claysmalley, line 301: Set `pointUp` to `true`, or else this shield will be upside-down.

src/js/shield_defs.js
```diff
@@ -209,17 +210,16 @@ function trapezoidUpShield(
...
-        fillColor,
-        strokeColor,
-        ref,
-        radius,
-        1,
-        rectWidth
-      ),
+    canvasDrawnBlank: {
+      drawFunc: "trapezoid",
+      params: {
```
claysmalley, line 215: Set `shortSideUp` to `true` or else this shield will be upside-down.

src/js/custom_shields.js
```diff
@@ -0,0 +1,73 @@
...
+    ShieldDraw.CS / 2,
+    ShieldDraw.CS / 2,
+    diameter,
+    0,
+    2 * Math.PI,
+    false
+  );
+
+  ctx.fillStyle = fillColor;
+  ctx.strokeStyle = Color.shields.black;
```
claysmalley, line 31: How come `strokeColor` is no longer a parameter and this is set to black?

src/js/shield_defs.js
ZeLonewolf: I'm going to change the default for pentagon instead, since all of our pentagons point up at this point.  We can still set `pointUp=false` if we need to override that behavior.

ZeLonewolf:
Fixed issues as noted.  Also removed the backgroundDraw handling code!

src/js/shield_defs.js
claysmalley: Will that affect home plate shields, which use the same draw function but still point downward?

src/js/custom_shields.js
claysmalley: Are you sure? I see white used as an outline for the [green, blue and purple belts](https://github.com/ZeLonewolf/openstreetmap-americana/pull/708/files#diff-21d2ca32bb400fce6149c2ca25ed3366df195a198f357764d46f7ae811835f29L4158-L4160).

src/js/custom_shields.js
```diff
@@ -0,0 +1,73 @@
...
+  let width = 0.7 * ShieldDraw.CS - 2 * lineWidth;
+
+  let y = 0.4 * ShieldDraw.CS + lineWidth;
+  let height = 0.45 * ShieldDraw.CS - 2 * lineWidth;
+
+  ctx.beginPath();
+  ctx.rect(x, y, width, height);
+
+  ctx.fillStyle = fillColor;
+  ctx.strokeStyle = Color.shields.white;
```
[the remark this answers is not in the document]
claysmalley: We're using a green perimeter on the [yellow route](https://github.com/ZeLonewolf/openstreetmap-americana/pull/708/files#diff-21d2ca32bb400fce6149c2ca25ed3366df195a198f357764d46f7ae811835f29R4047).

src/js/custom_shields.js
ZeLonewolf: Interesting, I'm thinking that was actually a bug though?  Shouldn't they all be black bordered?

image

claysmalley: It was a stylistic choice to improve color contrast: https://github.com/ZeLonewolf/openstreetmap-americana/pull/356#issuecomment-1144137548

Same thing with the Branson routes. We could revisit this decision, but that'd be appropriate for a separate PR.

src/js/shield_canvas_draw.js
```diff
@@ -711,5 +656,12 @@ export function registerDrawFunction(name, fxn) {
 }
 
 //Built-in draw functions (standard shapes)
-registerDrawFunction("roundedRectangle", roundedRectangle);
+registerDrawFunction("diamond", diamond);
 registerDrawFunction("ellipse", ellipse);
+registerDrawFunction("escutcheon", escutcheon);
+registerDrawFunction("hexagonVertical", hexagonVertical);
+registerDrawFunction("hexagonHorizontal", hexagonHorizontal);
+registerDrawFunction("octagonVertical", hexagonVertical);
```
claysmalley, line 664: Shouldn't both of these be `octagonVertical`?

ZeLonewolf: Yup!  Fixed.

src/js/custom_shields.js
ZeLonewolf: Perfect, thanks for the dive on that.  I'll edit this to preserve the current behavior.

ZeLonewolf:
Updated Branson / Pittsburgh shields:

image

image

claysmalley (approved):
Looks good!

ZeLonewolf:
Partial implementation of #118

PR #710, opened 2023-01-17
Node.js 17 → 18

jleedev:
17 is end of life, 18 is the LTS version of 17.


ZeLonewolf (approved):
May the odds be ever in your favor.

PR #711, opened 2023-01-18
Improve diamond shield draw function

claysmalley:
`diamond()` as previously implemented left empty pixels along the edges when `radius > 0`. This tweaks the math to ensure that the shield always reaches the edges of the canvas, giving the text a little bit more room.

before/after/diff
old-5 new-5 diff-5
(highlight with cursor to see canvas edges)


ZeLonewolf (approved):
Looks good!

image

PR #713, opened 2023-01-18
Add triangle shield draw function

claysmalley:
This one is even harder to fit text in than diamond shields. Fortunately, it's used by just one network that we know of, which only has up to 3 digits in refs. If we discover another route network with 4 or more digits on triangle shields, we'll need to write another text constraint function.

Screenshot from 2023-01-18 13-17-19
[Screenshot from 2023-01-18 13-23-42](http://localhost:1776/#map=12.33/36.54503/-82.11864)


ZeLonewolf (approved):
Looks good

image

PR #714, opened 2023-01-19
Convert shield sprite definitions to JSON

ZeLonewolf:
This PR continues the implementation of #118 by converting the current sprite sheet references to plain JSON strings.

image


claysmalley:
[Shield test](http://localhost:1776/shieldtest.html) appears to be broken.

src/americana.js
```diff
@@ -58,9 +58,7 @@ export const map = (window.map = new maplibregl.Map({
 
 CustomShields.loadCustomShields();
 
-map.on("styledata", function (event) {
-  ShieldDef.loadShields(map.style.imageManager.images);
-});
+ShieldDef.loadShields();
```
1ec5, line 61: It’s no longer necessary to defer compiling the shield definitions until the style requests a missing image, but maybe it could be deferred until the [style data begins loading](https://maplibre.org/maplibre-gl-js-docs/api/map/#map.event:styledataloading) instead.

src/js/shield_defs.js
```diff
@@ -632,15 +632,12 @@ function banneredShield(baseDef, modifiers) {
   };
 }
 
-export function loadShields(shieldImages) {
+export function loadShields() {
```
1ec5, line 635: A couple of the scripts rely on passing in mocked versions of `shieldImages`. They no longer need to pass anything in, but instead they should map `spriteBlank` to a full image URL afterwards.

https://github.com/ZeLonewolf/openstreetmap-americana/blob/d35b7e3629ad9cf57f65183eb7666c21d2aea134/scripts/taginfo.js#L14-L20 https://github.com/ZeLonewolf/openstreetmap-americana/blob/d35b7e3629ad9cf57f65183eb7666c21d2aea134/scripts/status_map.js#L25-L31

ZeLonewolf:
> [Shield test](http://localhost:1776/shieldtest.html) appears to be broken.

Fixed this in 34446aa.

src/americana.js
ZeLonewolf: I find that if I wrap these loads, it breaks shieldtest.

src/js/shield_defs.js
ZeLonewolf: > instead they should map spriteBlank to a full image URL afterwards.

Can you elaborate?

1ec5: Somehow or another, these scripts need to end up with absolute URLs to shield images. Previously, they relied on a bit of a hack where `loadShields()` would happen to stick absolute URLs in each definition because the mock “sprite sheet” consisted of absolute URLs. But now we can just insert some code after the call to `loadShields()` that goes through each entry in the returned object, replacing its `spriteBlank` with an absolute URL.

src/americana.js
1ec5: It probably has something to do with shieldtest only waiting for the `load` event:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/067a48ff546ef482022cf062049247cd7f749134/src/shieldtest.js#L13

While it isn’t incorrect to build the shield definitions on page load, it’s suboptimal that shieldtest would be picky about when this happens. I don’t think shieldtest is actually making use of the map at all; it should just call `loadShields()` directly without importing americana.js or waiting for any map event.

1ec5: Actually it’s using the map to get the spritesheet to pass into `generateBlankGraphicsContext()`. I guess this is arguably necessary for covering the full shield rendering code path, but it’s mildly annoying that all the non-shield-related map code needs to come along for the ride.

ZeLonewolf: shieldtest.js separated from americana.js in 4a960e4.  Now we no longer have to take the layers along for the ride.

ZeLonewolf: I tried changing the load functions to:

```
map.on("styledataloading", function () {
  CustomShields.loadCustomShields();
  ShieldDef.loadShields();
});
```
Shields do **not** load when I do this, either in americana.js or shieldtest.

ZeLonewolf: If I change it to the "sourcedata" event, that works, presumably because that happens prior to style loading.  Does that meet the intent?  In any case, the shield definition loading should be a lightweight event so I'm not sure why it's important to defer it.

src/js/shield_defs.js
ZeLonewolf: Okay, I think we have a working solution in 0e32599.

ZeLonewolf: In fact, I just discovered that we don't need to parse the sprite JSON at all, so I scrubbed that in cdbef18

src/americana.js
1ec5: I was thinking that it might be useful to defer it based on any user preferences we decide to vary the style by, but we could always do that explicitly rather than relying on GL JS’s event system. As long as this step is lightweight, we can leave it as is.

scripts/status_map.js
```diff
@@ -28,7 +28,7 @@ let shieldImageURLs = Object.fromEntries(
     `https://raw.githubusercontent.com/ZeLonewolf/openstreetmap-americana/main/icons/${sprite}.svg`,
   ])
 );
```
1ec5, line 30: `shieldImageURLs` is unused and can be removed safely. This script only cares about the keys of `shields` anyways.

scripts/taginfo.js
```diff
@@ -30,29 +23,35 @@ function addNetworkTags(project, sprites) {
     }
     description += ".";
 
-    let icon = definition.backgroundImage || definition.norefImage;
-    if (Array.isArray(icon)) {
-      icon = icon[0];
```
1ec5, line 35: This version of the `icon` code was fine, except that it needs to refer to `spriteBlank` or `norefImage` and set `icon` to a full URL:

```js
icon = `https://raw.githubusercontent.com/ZeLonewolf/openstreetmap-americana/main/icons/${icon[0]}.svg`;
```

src/shieldtest.js
```diff
@@ -1,9 +1,33 @@
...
+    sprite: new URL("sprites/sprite", baseUrl).href,
+  },
+}));
+
+CustomShields.loadCustomShields();
+ShieldDef.loadShields();
+
+map.on("styleimagemissing", function (e) {
+  Shield.missingIconHandler(map, e);
+});
```
1ec5, line 30: Ah, much simpler!

src/americana.js
ZeLonewolf: I put these two calls back inline.  All they do is construct JSON objects so that seems safe.

scripts/status_map.js
ZeLonewolf: Fixed in 9bd2acf

1ec5 (approved):

PR #715, opened 2023-01-19
Use OMT rankings for POI deconfliction

ZeLonewolf:
OpenMapTiles maintains POI "ranks" in the tiles so that more important POI categories are shown instead of less-important ones if their icons collide.

The POI ranking logic can be found here:
https://github.com/openmaptiles/openmaptiles/blob/master/layers/poi/class.sql

For example, in [New York City](http://localhost:1776/#map=17/40.734981/-73.99163), note how the "Parsons school of design" collides out a nearby coffeeshop as the zoom goes down one level:

image
image


wmisener:
`fast_food` not ranked most important? I thought this was Americana! (Just kidding, looks like a sensible system to implement here!)

ZeLonewolf:
Tongue in cheek maybe, but the code to reorder them is not that complicated.

ZeLonewolf:
oh nope, no it wouldn't.  That OMT rank formula gets shoved into a cell partition.

claysmalley (approved):
Looks better than the status quo. Let's go with it.

PR #593, opened 2022-11-30
Draw diamond and horizontal hexagon shields mathematically

claysmalley:
Adds functions to draw diamond shields and horizontal hexagon shields.

Screenshot from 2022-11-30 15-57-53
Screenshot from 2022-11-30 15-58-38
Screenshot from 2022-11-30 18-59-45
Screenshot from 2022-11-30 19-00-06

PR #721, opened 2023-01-21
change Mackinac Island shield to fixed-width

claysmalley:
Michigan State Trunkline Highway 185 is the ring road of Mackinac Island, on which motorized vehicles are banned (except for a handful of emergency vehicles). It has a unique shield due to its non-motorized status. In #593 the SVG icon was replaced by a draw function, but erroneously set the shield to variable-width. This changes it back to fixed-width, and also removes the rounded corners to match the real-life shield.

before/after:
Screenshot from 2023-01-21 14-28-29
Screenshot from 2023-01-21 14-31-08

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -4080,7 +4080,13 @@ export function loadShields(shieldImages) {
   };
 
   shields["US:MI"].overrideByRef = {
-    185: diamondShield(Color.shields.brown, Color.shields.white),
+    185: diamondShield(
+      Color.shields.brown,
+      Color.shields.white,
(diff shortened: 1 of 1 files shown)


ZeLonewolf (approved):
This certainly fixes the problem, though the text is quite small.  I wonder if we might make the shields bigger by approximately square root of two to compensate for the shape.

I also wonder whether diamond shields should always default to sharp corners and fixed width.

In any case, this is an improvement from current state.

claysmalley:
Once I figure out how `shield_text.js` works I'm going to try adding text constraint functions for triangles and diamonds.

PR #725, opened 2023-01-24
Update esbuild; replace live reload server

jleedev:
Known issues for me that this fixes:

- Esbuild's node layer would occasionally leave the background process around when an exception terminated the build script; it seems that the newer version is more careful about this. This is good to prevent an untethered cadre of esbuild processes dutifully watching file changes and building in the background.
- The API is more careful to be responsible for a single build at a time.
- The create-serve package had some incompatibilities with some versions of node and would occasionally fail to detect the local ip address.
- Since the same process is both building and serving, there's no race while a build artifact is being replaced on disk and the reloading page sees a 404 error.

New features:

- Esbuild has an EventSource endpoint and can specifically tell you which files changed. The docs suggest clever tricks to inject a new CSS resource without reloading the page, for example. If the change event only touches the source map but not the script, the client code could avoid reloading.
- But we're just doing the boring thing and reloading whenever anything at all changes, to be clear.
- There's a cool bundle size analyzer.
- There's now a `copy` file loader that could be used to have esbuild be responsible for moving index.html around, which would allow it to participate in live reloading.

Noteworthy change:

- This uses the esbuild serve endpoint now, rather than serving from the file system. This means that every single request is a fresh build, even for static files that esbuild doesn't care about.
- Might want to think about unifying the config.js with the live reload conditional into a single build-time configuration area.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,8 +17,7 @@
       "devDependencies": {
         "@maplibre/maplibre-gl-style-spec": "^17.0.1",
         "chai": "^4.3.7",
-        "create-serve": "^1.0.1",
-        "esbuild": "^0.15.15",
+        "esbuild": "^0.17.4",
         "events": "^3.3.0",
         "maplibre-gl": "^2.1.9",
         "mocha": "^10.1.0",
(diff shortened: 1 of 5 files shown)


ZeLonewolf:
`npm install` required to test

ZeLonewolf (commented):
Works for me on Ubuntu, like to get folks on other platforms to verify also.

adamfranco:
Seems to work fine for me on OSX on Apple M1 (arm64):

```
$ npm --version
9.3.0
$ node --version
v16.17.1
``` 

Changes seem to be tracked fine.

ZeLonewolf (approved):

PR #726, opened 2023-01-24
Add config variable to outline shield text padding

claysmalley:
Adds a debug variable in `config.js` to display a rectangle on shields representing the padding values passed from the shield definition.

Screenshot from 2023-01-24 13-30-41

### DIFF
--- a/src/configs/config.aws.js
+++ b/src/configs/config.aws.js
@@ -7,12 +7,15 @@ const OPENMAPTILES_URL =
   "https://6ug7hetxl9.execute-api.us-east-2.amazonaws.com/data/v3.json";
 
 /*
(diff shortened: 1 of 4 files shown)


ZeLonewolf:
It does seem to work.

I note the box overlaps the bottom pixel row of the text on the Interstate shields:
image

Is that an expected output?

ZeLonewolf:
In other words, should the bbox be one pixel wider so the constrained part is entirely inside the bounds?

claysmalley:
The rectangle is already completely outside the text constraint. What you're seeing is a result of `southHalfEllipseTextConstraint`, and I'm not entirely sure how it works or why it results in text outside the bounds defined by the padding values.

ZeLonewolf (approved):

PR #727, opened 2023-01-25
Generate sample taginfo graphic for all networks

ZeLonewolf:
Currently, taginfo graphics are linked to pre-drawn icons only.  This PR generates graphics for all networks for publication on github-pages.  I implemented a hash of the shield configuration to ensure that networks with common graphics link to a single image.  Output graphics are low quality (the drawing process does some intermediate rasterization before producing the SVG), however, this is sufficient for the purpose of taginfo.

#polishing_the_cannonball


1ec5 (commented):
While I agree that crisp SVG images are more than functionally necessary for taginfo, we should probably double-check how these graphics will actually look in taginfo on both 1× and 2× screens. Taginfo automatically resizes the image to fit in a 16&nbsp;px&nbsp;× 16&nbsp;px box, so some of the wider or taller shields’ borders may turn into mush.

scripts/taginfo.js
```diff
@@ -2,22 +2,57 @@
...
+  ctx.canvas.width = bounds.width;
+  ctx.canvas.height = bounds.height;
+  return ctx;
+});
+
+//Hash function to elminiate duplicate shield graphics
+/*
+    Released into public comain by bryc (github.com/bryc)
+    Source: https://github.com/bryc/code/blob/master/jshash/experimental/cyrb53.js
+*/
```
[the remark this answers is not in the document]
1ec5: Yes, it would produce JSON text, but you could strip out all the non-alphanumeric characters using a regular expression. Even better, take those raw stringified JSON strings and use them as keys in a dictionary, mapping them to unique, incrementing integers; use those integers as the actual file names.

ZeLonewolf:
#737 should improve (hopefully) the quality of graphics produced for taginfo.

ZeLonewolf:
This PR now includes such ridiculous phrases as:
> Roads carrying routes in this network are marked by mahogany fishhead-shaped shields with white borders.

:man_cook:

scripts/taginfo.js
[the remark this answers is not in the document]
ZeLonewolf: Updated to use incrementing integers as suggested.

scripts/taginfo.js
```diff
@@ -19,27 +53,75 @@ function addNetworkTags(project) {
...
+            } else {
+              prettyShapeName = "rectangle";
+            }
+            break;
+          case "hexagonVertical":
+            prettyShapeName = "vertical hexagon";
+            break;
+          case "hexagonHorizontal":
+            prettyShapeName = "horizontal hexagon";
+            break;
```
claysmalley, line 105: Other cases to consider:

* `"octagonVertical"` → `"vertical octagon"`
* `"ellipse"` → `"circle"` if `rectWidth == 20`
* `"pentagon"` → `"home plate"` if `angle == 0`

ZeLonewolf: Thank you, these description additions are excellent.

scripts/taginfo.js
```diff
@@ -2,14 +2,47 @@
...
+import * as CustomShields from "../src/js/custom_shields.js";
+import * as skia from "skia-canvas";
+import namer from "color-namer";
+import { mkdir } from "node:fs/promises";
+
+await mkdir("dist/shield-sample", { recursive: true });
+
+//Headless graphics context
+Gfx.setGfxFactory((bounds) => {
+  let canvas = new skia.Canvas(bounds.width, bounds.height);
```
1ec5, line 16: Wow, I was griping about potential image quality issues on 2× displays when downscaling PNGs, but I didn’t imagine that you were going to pull in Skia for a very correct fix. Thank you! If GL JS ever supports native vector graphics as sprites, we’ll be in a better position to set them up at build time.

scripts/taginfo.js
```diff
@@ -19,27 +52,86 @@ function addNetworkTags(project) {
...
+            if (shapeDef.params.angle == 0) {
+              prettyShapeName = "home plate";
+            }
+        }
+        description += `${prettyFillColor} ${prettyShapeName}-shaped shields with ${prettyStrokeColor} borders`;
+      } else {
+        description += "shields";
+      }
+
       if (definition.modifiers && definition.modifiers.length > 0) {
```
1ec5, line 122: If a programmatically generated shield has modifiers, do we need to output a separate sample image for it? For prefabricated images, we’re just using the base shield image, which tends to downscale better than a shield with stacked modifiers would:

US:TX:Loop:Express

Consider doing the same for the programmatic shields: delete `modifiers` before stringifying.

scripts/taginfo.js
```diff
@@ -19,27 +52,86 @@ function addNetworkTags(project) {
...
+          shapeDef.params.strokeColor,
+          "black"
+        );
+
+        switch (shapeDef.drawFunc) {
+          case "roundedRectangle":
+            if (shapeDef.params.radius == 10) {
+              prettyShapeName = "pill";
+            } else {
+              prettyShapeName = "rectangle";
```
1ec5, line 96: Nit: “Rectangular shields” would read a little better than “rectangle-shaped shields”.

scripts/taginfo.js
ZeLonewolf: I was thinking that we would _add_ modifiers to the pre-generated shield graphics, but I see now that that's a terrible idea.

scripts/taginfo.js
ZeLonewolf: All set - cleaned up these names to be adjectives.

1ec5 (approved):
“Americana does the right thing with this tag” is pretty much all taginfo needs to know. Great job with this PR!

PR #729, opened 2023-01-26
Add performance monitoring to shieldtest

ZeLonewolf:
This PR adds a new column to shieldtest.html that lists the performance of each shield.  This will allow us to assess whether we introduce any performance changes when modifying shield code.

Example:

image

### DIFF
--- a/src/shieldtest.js
+++ b/src/shieldtest.js
@@ -2,7 +2,6 @@
 
 import { map } from "./americana.js";
 import * as shield from "./js/shield.js";
-import { shields } from "./js/shield_defs.js";
 import * as gfx from "./js/screen_gfx.js";
 
 const once = (emitter, name, { signal } = {}) =>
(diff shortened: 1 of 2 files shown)


src/shieldtest.js
```diff
@@ -2,7 +2,6 @@
 
 import { map } from "./americana.js";
 import * as shield from "./js/shield.js";
-import { shields } from "./js/shield_defs.js";
```
ZeLonewolf, line 5: This was just an unused import.

src/shieldtest.js
```diff
@@ -246,13 +245,25 @@ for (let network of networks) {
...
   }
+  let perfEntries = performance.getEntriesByName(`${network}`);
+  var perfDuration = 0;
+  for (let perf of perfEntries) {
+    perfDuration += perf.duration;
+  }
+  let shieldRate = Math.round((1000 * perfEntries.length) / perfDuration);
+  row
+    .insertCell()
+    .appendChild(document.createTextNode(`${shieldRate} shields/sec`));
```
jleedev, line 268:
```suggestion
    .append(`${shieldRate} shields/sec`);
```

ZeLonewolf: That reports in console:
```
Uncaught TypeError: Failed to execute 'appendChild' on 'Node': parameter 1 is not of type 'Node'.
    at shieldtest.js:268:6
```

jleedev (approved):
Nice.

I should learn the performance mark/entries stuff instead of manually using performance.now() :)

PR #732, opened 2023-01-26
add fishhead shield draw function

claysmalley:
Adds a shield to draw the "fishhead" shape of State Routes in Australia and State Highways in New Zealand.

Screenshot from 2023-01-26 15-23-27
Screenshot from 2023-01-26 15-21-13
Screenshot from 2023-01-26 15-23-12


ZeLonewolf:
Seems reasonable - I think we should push #729 and #714 before adding the next draw function.

ZeLonewolf (approved):
Looks good here.

image

image

image

ZeLonewolf:
Just merged these changes into #737, and the results still look pixel-equivalent :+1:

PR #735, opened 2023-01-27
Convert text constraint definitions to JSON

### LINKED ISSUE #118: Extract loadShields objects into JSON

ZeLonewolf:
Fixes #118 and moves us towards #146 and #70.

This PR converts shield text constraints to JSON, which allows the shield definitions to be pure JSON.  There is still some work we could do to reduce repetition in the JSON structure, but there's no longer any function pointers in the definition.

I also added npm scripts to generate the shield text definitions out to JSON at the command line.


claysmalley:
Did something change with `southHalfEllipseTextConstraint`? Before/after/diff:

old-h201 new-h201 diff-h201
old-a26a7 new-a26a7 diff-a26a7

ZeLonewolf:
Yep, that looks pretty wrong

ZeLonewolf:
So the right hand graphic is a correct south-half ellipse and the original is wrong.  So somehow I fixed the bug...

I think for fishhead we actually want a rounded rect constraint for fishhead, because of its shape, which I changed here:

image

claysmalley (approved):

PR #736, opened 2023-01-27
Add style.json to gh-pages

### LINKED ISSUE #68: Pre-generate style.json

ZeLonewolf:
Fixes #68 

This PR generates the style.json and hosts it on GitHub pages.  This artifact is provided for development and experimentation purposes; however, I envision it will eventually be part of an end-user deployable style in the spirit of #70.

I also removed the old code that was copying a maptiler key into the old maptiler config.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -26,7 +26,7 @@ jobs:
         run: |
           npm ci --include=dev
           cp src/configs/config.aws.js src/config.js
-          npm run build
+          npm run build style
       - name: Upload 🏗
         uses: actions/upload-pages-artifact@v1
         with:
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -26,7 +26,8 @@ jobs:
         run: |
           npm ci --include=dev
           cp src/configs/config.aws.js src/config.js
-          npm run build style
+          npm run build
+          npm run style
       - name: Upload 🏗
         uses: actions/upload-pages-artifact@v1
         with:
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -17,9 +17,8 @@ jobs:
       - name: Install and Build 🔧
         run: |
           npm ci --include=dev
-          sed 's/<your MapTiler key>/53iZvB2drcamS0Ge0xiD/g' src/configs/config.maptiler.js > src/config.js
-          npm run build
-        # MapTiler key 53iZvB2drcamS0Ge0xiD only allows requests from zelonewolf.github.io
+          cp src/configs/config.aws.js src/config.js
+          npm run build style
       - name: Test 🧪
         run: |
           npm test
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -17,7 +17,6 @@ jobs:
       - name: Install and Build 🔧
         run: |
           npm ci --include=dev
-          cp src/configs/config.aws.js src/config.js
           npm run build
           npm run style
       - name: Test 🧪
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -18,7 +18,8 @@ jobs:
         run: |
           npm ci --include=dev
           cp src/configs/config.aws.js src/config.js
-          npm run build style
+          npm run build
+          npm run style
       - name: Test 🧪
         run: |
           npm test
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -104,7 +104,9 @@ and re-running `npm install`.
 Environment specific settings go in the untracked file `config.js`. Copy from one of
 the templates in the configs/ folder `config.*.js` and rename it `config.js` in
 the src/ root. The variables in this file can then be changed without the risk of
-accidentally comitting to the main repo.
+accidentally comitting to the main repo. By default, the repository `config.js` points
+to a development tile server operated by the project maintainers. It is free to use
+for development purposes but service is not garaunteed.
 
 You can create a new copy of the config file by running `npm run config`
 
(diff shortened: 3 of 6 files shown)


jleedev:
Can I suggest instead having a checked-in `src/config.js` so that the tree is buildable without extra steps?

```js
export {default} from "./configs/config.aws.js";
```

.github/workflows/deploy.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         run: |
           npm ci --include=dev
           cp src/configs/config.aws.js src/config.js
-          npm run build
+          npm run build style
```
jleedev, line 29: This passes a positional argument of `style` to `build`, which only works because the command is `run-s`.

```suggestion
          npm run build
          npm run style
```

ZeLonewolf:
> Can I suggest instead having a checked-in `src/config.js` so that the tree is buildable without extra steps?
> 
> ```js
> export {default} from "./configs/config.aws.js";
> ```

I made this change in 865affe.  I had no idea you could do that.

.github/workflows/test-build.yml
```diff
@@ -17,9 +17,8 @@ jobs:
       - name: Install and Build 🔧
         run: |
           npm ci --include=dev
-          sed 's/<your MapTiler key>/53iZvB2drcamS0Ge0xiD/g' src/configs/config.maptiler.js > src/config.js
-          npm run build
-        # MapTiler key 53iZvB2drcamS0Ge0xiD only allows requests from zelonewolf.github.io
+          cp src/configs/config.aws.js src/config.js
+          npm run build style
```
jleedev, line 21:
```suggestion
          npm run build
          npm run style
```

CONTRIBUTING.md
```diff
@@ -104,7 +104,9 @@ and re-running `npm install`.
 Environment specific settings go in the untracked file `config.js`. Copy from one of
```
jleedev, line 104: > untracked file

Ah yes, that was why.

I'm sure it's fine.

ZeLonewolf: Uh. :sweat_smile:

jleedev: Generally speaking this could be done by having a `local.config.js` which is still ignored, and the build script will read it if present.

esbuild's `inject` option might work.

I'd be fine to leave it and just avoid coming changes accidentally, or you could back it out.

ZeLonewolf: You'd still have to explicitly `git <whatever> -f` to accidentally commit changes, no?

jleedev: No, it's easy to do `git add -u` or `git commit -a` on autopilot and make a change. gitignore only keeps the file untracked; once a file is tracked it doesn't get ignored.

There's skip-worktree or assume-unchanged. Not sure exactly, but the file could have a comment up top saying how to instruct git to ignore changes to it.

I'm presuming that most users will use the default and only occasionally switch to a different source.

michaelblyons: Can you set a CI task that fails if someone adds certain file names? Then if they screw up, it's not a problem.

Other related things are .gitattributes merge behavior (which may only affect conflicts), smudge filters, and the skip-worktree/assume-unchanged mentioned above.

The annoyance with the latter that I kept encountering is that if anyone *else* did it wrong, everyone who did it right had to un-skip, merge, and re-skip. And if anyone forgot to re-skip, it would start the cycle again.

ZeLonewolf: I'm not quite sure how to do
`import config from "./config.js";`
but make it be "and if not there, grab it from `config.default.js` ?

jleedev: I got it to work. Pushing up a commit…

scripts/build.js
```diff
@@ -1,35 +1,60 @@
...
-  });
+    define: {
+      ...localConfig?.define,
+      ...buildOptions?.define,
+    },
+  };
+  return esbuild[key](options)
+    // esbuild will pretty-print its own error messages;
+    // suppress node.js from printing the exception.
+    .catch(() => process.exit(1));
```
jleedev, line 50: Note: moved this around to more carefully catch promise rejections from esbuild, but errors elsewhere in this function will not be swallowed, and _synchronous_ exception from esbuild.build will also throw if it doesn't like the options we passed.

src/config.js
```diff
@@ -1 +1,14 @@
-export { default } from "./configs/config.aws.js";
+// The build script injects this global variable if "local.config.js"
+// exists in the project root (alongside package.json). That file is in
+// gitignore and so can be modified without making changes to this file.
+
+const importConfig = () => {
+  if (typeof CONFIG_PATH !== "undefined") {
+    return import(CONFIG_PATH);
+  } else {
+    return import("./configs/config.aws.js");
```
jleedev, line 9: Important thing here is that syntactically these are still imports of string literals (since CONFIG_PATH gets replaced by a string literal before being built), so they get bundled normally and do not emit import() calls.

src/config.js
```diff
@@ -1 +1,14 @@
-export { default } from "./configs/config.aws.js";
+// The build script injects this global variable if "local.config.js"
+// exists in the project root (alongside package.json). That file is in
+// gitignore and so can be modified without making changes to this file.
```
jleedev, line 3: Well, should be added to gitignore :)

CONTRIBUTING.md
jleedev: Things such as git hooks, index bits, attributes, filters, etc. etc. are interesting in principle but they have to be configured separately from cloning, only the simplest of attributes are built-in to git. So I don't think they're ideal for this case.

src/config.js
ZeLonewolf: .gitignore added in cf49208

Ultimately I'd like to come up with a scheme where the demo map has a user-selectable list of tileservers that a user can swap out on the fly.  But I think this does the job for now.

scripts/build.js
ZeLonewolf: I wish I could say I understand how this works, but it certainly does work.

jleedev (approved):

PR #737, opened 2023-01-28
Remove intermediate rasterization steps in shield drawing code

ZeLonewolf:
While investigating #727, I discovered that all of the intermediate draw functions involved in shield generation was causing unnecessary rasterization and prevented us from producing vector samples of the tiles.

This PR removes all intermediate canvas creations in the draw code.  Instead, we now create a canvas object onces, pre-computed for the correct size, and then draw each component on the sole canvas.  The `colorLighten` function is now done at pixel copy time. By inspection, this does appear to speed up shield drawing.  In addition, I added the Pittsburgh and Branson special routes to the shield test grid.

Hopefully this both simplifies the code AND makes things a bit more performant.

I do need someone to confirm that everything is sane in both 1x and 2x+ displays.

image
image


src/js/shield.js
```diff
@@ -6,36 +6,42 @@ import * as ShieldDef from "./shield_defs.js";
...
+  dest[destOffset + 1] = Math.max(source[sourceOffset + 1], colorLighten[1]); //Green
+  dest[destOffset + 2] = Math.max(source[sourceOffset + 2], colorLighten[2]); //Blue
+  dest[destOffset + 3] = source[sourceOffset + 3]; //Alpha
+}
 
-  for (var i = 0; i < shield.data.data.length; i++) {
-    imgData.data[i] = shield.data.data[i];
-  }
+function loadSprite(ctx, shield, bannerCount, verticalReflect, colorLighten) {
+  var imgData = ctx.createImageData(shield.data.width, shield.data.height);
```
jleedev, line 19: `var` has function scope, `let` has block scope.

Generally speaking we should never use `var`.

ZeLonewolf: Fixed in 8922d1a

package.json
```diff
@@ -26,6 +26,7 @@
   },
   "dependencies": {
     "@basemaps/sprites": "^6.28.1",
+    "color-rgba": "^2.4.0",
```
jleedev, line 29: Just FYI, another library I'd recommend: https://www.npmjs.com/package/d3-color

src/js/shield.js
```diff
@@ -6,36 +6,44 @@ import * as ShieldDef from "./shield_defs.js";
...
+import rgba from "color-rgba";
 
-function loadShield(ctx, shield, bannerCount, verticalReflect) {
-  var drawCtx = Gfx.getGfxContext(shield.data);
-  var imgData = drawCtx.createImageData(shield.data.width, shield.data.height);
+function loadPixel(source, dest, sourceOffset, destOffset, colorLighten) {
+  dest[destOffset] = Math.max(source[sourceOffset], colorLighten[0]); //Red
+  dest[destOffset + 1] = Math.max(source[sourceOffset + 1], colorLighten[1]); //Green
+  dest[destOffset + 2] = Math.max(source[sourceOffset + 2], colorLighten[2]); //Blue
+  dest[destOffset + 3] = source[sourceOffset + 3]; //Alpha
```
jleedev, line 15: This could probably be done natively via https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/globalCompositeOperation

src/js/shield.js
```diff
@@ -6,36 +6,44 @@ import * as ShieldDef from "./shield_defs.js";
...
-    ctx.drawImage(
-      drawCtx.canvas,
-      0,
-      bannerCount * ShieldDef.bannerSizeH +
-        ShieldDef.topPadding -
-        drawCtx.canvas.height -
-        2 * ShieldDraw.PXR
-    );
-    ctx.restore();
+    //4 bytes/px, copy in reverse vertical order.
```
jleedev, line 35: This feels a little weird, you're reflecting the image vertically and also adjusting the color lightness at the same time.

Is it done in one operation for performance reasons? What is being reflected and why?

src/js/shield.js
```diff
@@ -137,72 +145,51 @@ function textColor(shieldDef) {
...
+function drawShield(ctx, shieldDef, routeDef) {
+  let bannerCount = getBannerCount(shieldDef);
+  let yOffset = bannerCount * ShieldDef.bannerSizeH + ShieldDef.topPadding;
 
-  return Gfx.getGfxContext(compoundBounds);
+  //Shift canvas to draw shield below banner
+  ctx.translate(0, yOffset);
+  let drawFunc = getDrawFunc(shieldDef);
+  drawFunc(ctx, routeDef.ref);
+  ctx.translate(0, -yOffset);
```
jleedev, line 167: Better to bracket these with [save]…restore:

[save]: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/save

```suggestion
  ctx.save();
  ctx.translate(0, yOffset);
  let drawFunc = getDrawFunc(shieldDef);
  drawFunc(ctx, routeDef.ref);
  ctx.restore();
```

src/js/shield_canvas_draw.js
```diff
@@ -16,22 +16,27 @@ const minGenericShieldWidth = 20 * PXR;
 const maxGenericShieldWidth = 34 * PXR;
 const genericShieldFontSize = 18 * PXR;
 
-function ellipse(params, ref) {
+export function computeWidth(rectWidth, ref) {
```
claysmalley, line 19: It looks like this function is conflating all shields in how they calculate width. Before/after:

old-pentagon old-trapezoid old-hexagon
new-pentagon new-trapezoid new-hexagon

src/js/shield.js
```diff
@@ -382,47 +369,60 @@ export function generateShieldCtx(map, id) {
...
+    }
+  } else {
+    width = shieldArtwork.data.width;
+    height = shieldArtwork.data.height;
+  }
+
+  height += bannerCount * ShieldDef.bannerSizeH + ShieldDef.topPadding;
+
+  //Generate empty canvas sized to the graphic
+  let ctx = Gfx.getGfxContext({ width: width, height: height });
```
jleedev, line 404:
```suggestion
  let ctx = Gfx.getGfxContext({ width, height });
```

src/js/shield.js
ZeLonewolf: The problem I had is that a straight compositeOperation lighten also seems to also color the transparent background, which I don't want. That's why I originally had a multi-step series of composite operations that created an intermediate canvas.

src/js/shield.js
ZeLonewolf: At present, the vertical flipping is done for the Hawaii shield only, and I didn't change the existing logic. There's shields in Japan and Taiwan that are an upside-down version of this same shape. Since there's only one shield that uses this feature, we could certainly just make a graphic for it.

Since the lighten transformation is a pixel-twiddling operation on the background graphic, and we were already doing pixel copies, I figured I'd just add the logic at the time of copy. I applied the color lightness to both the flip and no-flip versions more or less just for completeness, since there aren't any shields that are flipped AND recolored.

While the code looks low level, it does seem to be performant. If we unwound the flipping capability it would simplify the code at the expense of one extra graphic (and the hope that we don't identify other shields that are upside-down variants of something.

image

src/js/shield_canvas_draw.js
ZeLonewolf: Good catching. I fixed this in 1388532, and switching between two tabs (remote and branch) seem to show no pixel variations.

jleedev (approved):

PR #739, opened 2023-01-29
Omit synthetic networks from taginfo project file

1ec5:
Taginfo says Americana supports some `network=omt-*` tags that aren’t actually real tags in OpenStreetMap. These values are synthesized by OpenMapTiles for the United Kingdom and Ireland because these countries color-code road networks but only designate numbered roads, not numbered routes. This PR replaces these automatically generated entries with UK- and Ireland-specific entries associated with the appropriate `highway=*` values.

omt


ZeLonewolf (approved):
Thanks!  Works as advertised.

PR #742, opened 2023-01-29
Provide basic sample usage

ZeLonewolf:
The purpose of this PR is to provide a sample that we can show to people interested in deploying their own copy of the Americana style, without all the bells, whistles, and development tools currently deployed to the demo map. This adds a `samples/` folder with a single `basic.html` that shows the usage of a simple styled map (but without highway shields).

Once #699 is complete, the next steps would be to add to this basic demo by pointing to a JSON shield definition as well as a javascript shield generation library. At that point, we would have a deployable style as envisioned in #70 that any end-user could deploy and customize as desired.


src/samples/basic.html
```diff
@@ -0,0 +1,51 @@
...
+      #attribution-logo {
+        position: absolute;
+        bottom: 10px;
+        z-index: 100;
+      }
+    </style>
+  </head>
+  <body>
+    <div id="map"></div>
+    <div id="attribution-logo"></div>
```
jleedev, line 40: What is this?

ZeLonewolf: Ooh yeah, that's a leftover container from when we had to display a maptiler logo in order to use their tileserver.  We can safely scrub that (e1c7162)

ZeLonewolf:
Pending #699

PR #748, opened 2023-01-31
Add Ireland to shield status map

1ec5:
#696 added support for route networks in Ireland that don’t exist in OSM but are instead synthesized by OpenMapTiles using a different `network` syntax. This PR generalizes the script that generates the shield status map to handle all such networks, not just Great Britain’s.


ZeLonewolf (approved):
Fancy.  Good thing I chose to use the country codes in the pseudo-network values.

PR #749, opened 2023-01-31
Added Asian Highway Network shields

### LINKED ISSUE #122: Asian Highway shields

1ec5:
Added shields for the Asian Highway Network.

Unlike the United Nations’s Economic Commission for Europe (UNECE), which manages the international E road network, the Economic and Social Commission for Asia and the Pacific (UNESCAP) has yet to harmonize Asian Highway route signage in each of its member countries. As a result, a single route can alternate among multiple different appearances along its length. We need to preserve some of the distinctions in this style, because for instance some countries reserve the color blue for AHs while others reserve blue for non-AHs.

[China–Vietnam](https://zelonewolf.github.io/openstreetmap-americana/#map=14/21.97405/106.7146) [China–North Korea](https://zelonewolf.github.io/openstreetmap-americana/#map=12/42.56967/130.52053&language=en) [North Korea–South Korea](https://zelonewolf.github.io/openstreetmap-americana/#map=12/37.9099/126.6856) [India–Myanmar](https://zelonewolf.github.io/openstreetmap-americana/#map=14/24.25441/94.29611)

Fixes #122.


src/js/shield_defs.js
```diff
@@ -3238,6 +3238,32 @@ export function loadShields() {
...
+      },
+    },
+    white: roundedRectShield(Color.shields.white, Color.shields.black),
+    white_id: roundedRectShield(
+      Color.shields.white,
+      Color.shields.green,
+      Color.shields.black
+    ),
+  };
+  shields["AH"] = shields["AsianHighway"];
```
1ec5, line 3265: [This wiki page](https://wiki.openstreetmap.org/wiki/Special:PermanentLink/2349783#Tags_for_the_relation) specifies `network=AsianHighway` for Asian Highway routes (along with `int_ref`, grr), but a significant minority of routes are tagged `network=AH` instead. I can’t find any discussion of using `AH` anywhere, so it’s probably just a lack of awareness about the `AsianHighway` standard. A two-letter code like `AH` would be problematic anyways, because it would be easily confused with a (nonexistent) ISO&nbsp;3166-1 country code.

features/az-cn-ir.geojson
```diff
@@ -0,0 +1,247 @@
...
+      [58.54611921520129, 37.69756305911791],
+      [58.405232856419275, 37.62976366376465],
+      [56.109734380422594, 38.4137452248973],
+      [54.65218822340282, 37.44910617782409],
+      [54.62249826191993, 37.449327082986386],
+      [53.211798427691946, 36.90999719120106],
+      [48.885303542763126, 42.06064324046332],
+      [48.415437922722475, 41.63699828114585]
+    ],
+    [
```
1ec5, line 72: This polygon has two outer rings. It should be a multipolygon geometry instead. GL JS has no problem with this weirdly-formed polygon, but for instance GitHub’s GeoJSON previewer makes a mess of it.

src/js/shield_defs.js
claysmalley: Sounds like a strong case against supporting `network=AH`. I would rather leave it out to encourage mappers to use a standard tagging scheme.

1ec5: Indeed, I’m definitely planning to remove support for it before merging. For now, though, it is necessary for demonstrating some of the cross-border differences in shield designs, because they happen to coincide with cross-border differences in how some route relations have been tagged. (I suspect it’s an oversight rather than anything intentional.)

PR #750, opened 2023-01-31
Force shield canvas to cpu.

jleedev:
Enable `willReadFrequently` on the shield drawing canvas, which (is a hint to) move rendering from the GPU to CPU.

Doesn't make a noticeable difference on many systems, but my chromebook is showing a consistent 4x speedup, from 8 ms to 2 ms, by avoiding the overhead of going to the GPU.


ZeLonewolf:
This is massively faster on my laptop:

image

ZeLonewolf (approved):

1ec5:
Mixed results but not a big difference either way on my ten-year-old MacBook Pro:

before
after

Issue #514, opened 2022-07-26
Road label kerning on curved roads


wmisener:
Road name labels appear to be designed to follow road curves. This provides a nice effect, unless the road is too sharply curved. Then there can be strange gaps between and/or letters which appear to "collide" with their adjacent letters. Examples (all taken from [here](https://zelonewolf.github.io/openstreetmap-americana/#14.04/34.03781/-118.41526)):
National Blvd, which turns 90 degrees at this intersection. Note d in Boulevard turns sideways:
<img width="264" alt="Screen Shot 2022-07-25 at 11 28 02 PM" src="https://user-images.githubusercontent.com/58491489/180939478-16033490-6b9d-4d3d-a8ed-d1735814bc45.png">
Duxbury Circle, note large gaps between b and u, r and y, and r and c:
<img width="96" alt="Screen Shot 2022-07-25 at 11 28 45 PM" src="https://user-images.githubusercontent.com/58491489/180939788-686ad17f-c9ca-4f01-af96-880b47b6d8a6.png">
Dannyhill Drive, letters run into each other:
<img width="123" alt="Screen Shot 2022-07-25 at 11 28 51 PM" src="https://user-images.githubusercontent.com/58491489/180939998-6e27ec09-e18b-44a6-b67c-b0a3ff944deb.png">

Potentially related (though it can also be spun off to a separate issue if appropriate), at similar zoom levels (z just above 14), adjacent road labels sometimes appear to collide. See [here](https://zelonewolf.github.io/openstreetmap-americana/#14.19/37.69647/-122.4714):
Niantic Avenue and Willits Street:
<img width="135" alt="Screen Shot 2022-07-25 at 11 32 26 PM" src="https://user-images.githubusercontent.com/58491489/180940684-ba70384b-0d7c-4c2e-a230-2e14e4825b32.png">
Bruno Avenue and Mission Street:
<img width="93" alt="Screen Shot 2022-07-25 at 11 32 44 PM" src="https://user-images.githubusercontent.com/58491489/180940709-c6c2fab8-2596-40eb-8694-92d7899bf4e3.png">

1ec5:
> Road name labels appear to be designed to follow road curves. This provides a nice effect, unless the road is too sharply curved. Then there can be strange gaps between and/or letters which appear to "collide" with their adjacent letters.

[...]

PR #753, opened 2023-02-01
Label both sides of international boundaries

### LINKED ISSUE #52: Boundary edge labels
When zoomed in beyond the level where a boundary shows a centrally-placed label, a label should be drawn along the edges.  In the case of adjacent areas, labels should be drawn on either side.

Here is an example from openstreetmap-carto:

image

Boundary labels should be drawn for administrative boundaries and protected areas.

1ec5:
Added a label layer on either side of a country boundary. Each label states the country’s name in the user’s preferred language.

The vector tiles only contain country codes, not localized names. To perform this conversion, I hard-coded a table mapping ISO 3166-1 alpha-3 codes to alpha-2 codes. When localizing the style, each alpha-3 code is mapped to the preferred language’s name for the country based on its alpha-2 code. When the user prefers American English, this PR adds 9,296&nbsp;bytes to the style JSON size for a total of 955,295&nbsp;bytes.

[Canada – United States](https://zelonewolf.github.io/openstreetmap-americana/#map=5/49.14/-106.45&language=en) [Canada – United States](https://zelonewolf.github.io/openstreetmap-americana/#map=6/44.378/-82.649&language=en) [Estados Unidos – México](https://zelonewolf.github.io/openstreetmap-americana/#map=8/31.618/-108.097&language=es) [Egypt – Sudan](https://zelonewolf.github.io/openstreetmap-americana/#map=6/21.8/34.839&language=en) [Indonesia – Papua New Guinea](https://zelonewolf.github.io/openstreetmap-americana/#map=5/-3.98/141.16&language=en) [インドネシア〜〜パプアニューギニア –](https://zelonewolf.github.io/openstreetmap-americana/#map=5/-3.98/141.16&language=ja)

Working towards #52.


jleedev:
Labels of differing length sometimes disagree on whether to appear. I'd suggest playing with `text-max-angle` but that probably wouldn't fix anything.

Curvy bit of United Kingdom/Ireland, lots of Ireland label but only a few United Kingdom label

ZeLonewolf:
I'm delighted by the absurdity of installing a lookup table in the style JSON. Is it also an option to have the country code perform a live lookup into wikidata and install the right text and would there be any advantage in doing that?

Obviously the "right" solution is proper encoding of left/right values for all levels of boundaries on the server side, but this is a pretty neat hack for the data we currently have.

jleedev:
Obvious differences include:

- Côte d’Ivoire / Ivory Coast
- Congo - Kinshasa / Democratic Republic of the Congo

(Not the sort of CI/CD problem I'm experienced with)

jleedev:
> Is it also an option to have the country code perform a live lookup into wikidata

```sparql
SELECT ?country ?countryLabel ?iso3166_12 ?iso3166_13 WHERE {
  ?country wdt:P297 ?iso3166_12;
    wdt:P298 ?iso3166_13.
  SERVICE wikibase:label { bd:serviceParam wikibase:language "[AUTO_LANGUAGE],en". }
}
```

1ec5:
> I imagine this needs a dot on the I, but I don't know how that would apply in general

This is upstream issue: mapbox/mapbox-gl-js#3999. It also affects the existing state/provincial labels, which use the same `text-transform` property.

> Is it also an option to have the country code perform a live lookup into wikidata and install the right text and would there be any advantage in doing that?

Any lookup needs to happen by the time we build the layers. I don’t think it would be appropriate to block style loading on a Wikidata Query Service query. We could run a query at build time for all the country names in all the languages, then check the results into this repository, but that’s a lot of data to maintain in here.

`Intl.DisplayNames` is the same API we’re using to localize country names in the “Route markers” section of the legend and language names in the language picker. These days, most browsers get the names from the Unicode Consortium’s CLDR project, but some might maintain the translations themselves.

This reliance on `Intl.DisplayNames` doesn’t scale beyond country names, so further fixing #52 would require more than putting ISO 3166-2 region codes in the tiles; it would be more sensible to include `name:*` tags in the tiles at that point.

1ec5:
> Specifically, I presume that the Intl.DisplayNames is fully populated with strings in the given language and so it is correct to uppercase-manipulate them as such,

This is a good point. I’m already massaging the names for presentational purposes, by doubling up spaces to simulate a `word-spacing` effect. It would be trivial to uppercase the strings in a locale-aware manner at the same time.

ZeLonewolf:
> it would be more sensible to include `name:*` tags in the tiles at that point.

I agree that would be the desired approach, however I'm not sure if it's technically feasible in the current SQL approach. Higher admin levels might have to wait until OMT switches over to planetiler.

ZeLonewolf:
If you can figure out how this boundary code works, and how it can be extended to support osm-carto style rendering... please tell me :sweat: 

https://github.com/openmaptiles/openmaptiles/blob/master/layers/boundary/boundary_name.sql

1ec5:
Does it make sense to land boundary edge labels for countries only until we figure out the gnarlier subnational issue upstream? Or would it be confusing to label only one level of boundary in the meantime?

ZeLonewolf:
I think that having country boundaries labeled with a kludge is still better than no labels at all.  Easy enough to undo the kludge later.

adamfranco:
Eventually it would be great to see boundary labels for other admin levels and other types of boundaries. I find them very useful when zoomed in enough to lose context of the greater region enclosed by the boundary. For example, [this view of a wildness area nested withing a National Forest, but briefly sharing a border](https://www.openstreetmap.org/#map=16/43.8521/-72.9847) is pretty tough to tease out visually without having those boundary labels that OSM Carto places.

1ec5:
> For example, [this view of a wildness area nested withing a National Forest, but briefly sharing a border](https://www.openstreetmap.org/#map=16/43.8521/-72.9847) is pretty tough to tease out visually without having those boundary labels that OSM Carto places.

Boundary edge labels may be more feasible for protected areas, which the tiles represent as polygons with localized names. “Concurrencies” between protected areas and other protected areas are less common, but one challenge would be preventing neighboring protected areas’ labels from colliding each other out. I think this would be a fine feature to explore in a separate PR, since the constraints are quite different and wouldn’t need to be rolled back when we have a more robust server-side solution for administrative boundaries.

1ec5:
> > Specifically, I presume that the Intl.DisplayNames is fully populated with strings in the given language and so it is correct to uppercase-manipulate them as such,
> 
> This is a good point. I’m already massaging the names for presentational purposes, by doubling up spaces to simulate a `word-spacing` effect. It would be trivial to uppercase the strings in a locale-aware manner at the same time.

Fixed:

[Türkiye – Suriye](https://zelonewolf.github.io/openstreetmap-americana/#map=8/36.842/38.514&language=tr)

1ec5:
> Labels of differing length sometimes disagree on whether to appear. I'd suggest playing with `text-max-angle` but that probably wouldn't fix anything.

Every time this happens, it seems to be because one label curves inexplicably, colliding with the label on the other side of the boundary. This happens identically whether `text-max-angle` is set to 0 or 180.

[Clones](https://zelonewolf.github.io/openstreetmap-americana/#map=14/54.12535/-7.3072&language=en)

1ec5:
I set `text-ignore-placement` to `true`, which theoretically allows other symbols to overlap with these labels, but in practice just allows more of the left-hand-side labels to appear without getting collided out by the right-hand-side labels. Other symbols should be unaffected because we aren’t using `text-allow-overlap` anywhere else. This change doesn’t fully fix the problem, but it does get us closer.

`text-ignore-placement` | [Clonoony Logh](https://zelonewolf.github.io/openstreetmap-americana/#map=13/54.14062/-7.31141&language=en)
----|----
`false` | Before
`true` | After

`text-ignore-placement` | [Galloon Island](https://zelonewolf.github.io/openstreetmap-americana/#map=13/54.14128/-7.40462&language=en)
----|----
`false` | Before
`true` | After

`text-ignore-placement` | [Eritrea](https://zelonewolf.github.io/openstreetmap-americana/#map=7/14.353/38.473&language=en)
----|----
`false` | Before
`true` | After

src/layer/boundary.js
```diff
@@ -242,6 +243,72 @@ export const country = {
...
+    "symbol-placement": "line",
+    "text-font": ["OpenHistorical Bold"],
+    "text-size": {
+      stops: [
+        [3, 6],
+        [7, 10],
+      ],
+    },
+    "text-field": getCountryName(["get", "adm0_l"]),
+    "text-offset": [0, -1],
```
1ec5, line 291: Setting this offset causes glyphs to bunch up and spread out awkwardly at any bend in the geometry:

[Mackinac](https://zelonewolf.github.io/openstreetmap-americana/#map=13/46.50721/-84.34302&language=en)

This is due to an upstream issue: mapbox/mapbox-gl-js#3718. It affects any label offset away from the geometry, such as road name labels: #514. I’ve tried to mitigate the bunching somewhat by applying a 0.1&nbsp;em letter spacing and limiting `text-max-angle` to 30&nbsp;degrees, but pushing these settings further would start to have visual tradeoffs or reduce the density of these labels.

ZeLonewolf (approved):
Looks like this is as good as it's going to get without fixing the upstream rendering issue.  Consider filing a maplibre-gl-js bug report if the issue hasn't been documented on the maplibre side.

src/layer/boundary.js
1ec5: maplibre/maplibre-gl-js#2170

1ec5:
> Every time this happens, it seems to be because the collision boxes of one label curve toward the geometry, colliding with the label on the other side of the boundary.

maplibre/maplibre-gl-js#2171

Issue #692, opened 2023-01-15
Points of Interest


ZeLonewolf:
This ticket tracks the comprehensive rendering of POI icons. Please link POI tickets here for tracking, and let's identify a subset of POIs that we view as must-have features for a "1.0" version of Americana. This is the comprehensive list of `class`/`subclass` values derived from OpenMapTiles. We can also choose not to render certain values.

- [ ] shop
  - [ ] accessories
  - [ ] antiques
  - [ ] beauty
  - [ ] bed
  - [ ] boutique
  - [ ] camera
  - [ ] carpet
  - [ ] charity
  - [ ] chemist
  - [x] coffee #387
  - [ ] computer
  - [ ] convenience
  - [ ] copyshop
  - [ ] cosmetics
  - [ ] garden_centre
  - [ ] doityourself
  - [ ] erotic
  - [ ] electronics
  - [ ] fabric
  - [ ] florist
  - [ ] frozen_food
  - [ ] furniture
  - [ ] video_games
  - [ ] video
  - [ ] general
  - [ ] gift
  - [ ] hardware
  - [ ] hearing_aids
  - [ ] hifi
  - [ ] ice_cream
  - [ ] interior_decoration
  - [ ] jewelry
  - [ ] kiosk
  - [ ] locksmith
  - [ ] lamps
  - [ ] mall
  - [ ] massage
  - [ ] motorcycle
  - [ ] mobile_phone
  - [ ] newsagent
  - [ ] optician
  - [ ] outdoor
  - [ ] perfumery
  - [ ] perfume
  - [ ] pet
  - [ ] photo
  - [ ] second_hand
  - [ ] shoes
  - [ ] sports
  - [ ] stationery
  - [ ] tailor
  - [ ] tattoo
  - [ ] ticket
  - [ ] tobacco
  - [ ] toys
  - [ ] travel_agency
  - [ ] watches
  - [ ] weapons
  - [ ] wholesale
- [ ] town_hall
  - [ ] townhall
  - [ ] public_building
  - [ ] courthouse
  - [ ] community_centre
- [ ] golf
  - [ ] golf
  - [ ] golf_course
  - [ ] miniature_golf
- [ ] fast_food
  - [ ] fast_food
  - [ ] food_court
- [ ] park
  - [ ] park
  - [ ] bbq
- [ ] bus
  - [ ] bus_stop
  - [ ] bus_station
- [ ] railway
  - [ ] halt
  - [ ] tram_stop
[...]

PR #755, opened 2023-02-01
add aerialways

### LINKED ISSUE #500: Aerialways
In the US, the average person might associate aerialways with ski resorts or scenic mountain lookouts, but aerial gondolas are increasingly becoming integral parts of public transit networks in mountainous cities around the world. In South America alone, aerial gondolas function as feeders for the rail-based transit networks in Medellín and Caracas, and the backbone of the transit network in La Paz, Bolivia. At least one aerial gondola in the US is oriented towards public transit: the Portland Aerial Tram, in Oregon.

Most US ski resort maps show aerialways as thick red lines with square caps. TriMet's official map shows the Portland Aerial Tram as a dark gray dashed line:

Screenshot from 2022-07-17 21-49-08

claysmalley:
Fixes #500

Breckenridge, Colorado, US:
[Screenshot from 2023-02-01 18-02-59](http://localhost:1776/#map=12.18/39.47382/-106.0675) [Screenshot from 2023-02-01 18-04-01](http://localhost:1776/#map=15.16/39.477636/-106.069137)

La Paz/El Alto, Bolivia:
[Screenshot from 2023-02-01 18-04-50](http://localhost:1776/#map=12.11/-16.50734/-68.14187) [Screenshot from 2023-02-01 18-05-58](http://localhost:1776/#map=15.33/-16.499038/-68.165718)

Aletsch Arena, Valais/Wallis, Switzerland:
[Screenshot from 2023-02-01 18-11-12](http://localhost:1776/#map=12.07/46.37868/8.05526) [Screenshot from 2023-02-01 18-12-22](http://localhost:1776/#map=14.79/46.38791/8.05734)

Medellín, Colombia:
[Screenshot from 2023-02-01 18-13-35](http://localhost:1776/#map=12.32/6.29307/-75.54735)


ZeLonewolf:
The great American inspiration: James Niehues:
image

Definitely works better against a snow map.  I suppose there's no way to differentiate skiing arialways from transportation ones.

The Bolivia case seems to get especially cluttered.  I wonder if perhaps it would benefit from a strong knockout and a softer line color.

claysmalley:
Adjusted the width so that the knockout casing shows up better:

Screenshot from 2023-02-01 19-38-17
Screenshot from 2023-02-01 19-39-23

src/js/legend_config.js
```diff
@@ -40,6 +41,10 @@ export const sections = [
     name: "Railroads",
     entries: RailLayers.legendEntries,
   },
+  {
+    name: "Aerialways",
```
1ec5, line 45: I think “aerialway” is only a real word in OSM English. Both aerial trams and chair lifts count as _aerial lifts_.

ZeLonewolf: Chairlifts or gondolas on the east coast.

ZeLonewolf: Tram sounds like some quaint en_GB thing to me.

claysmalley: _Aerial lift_ sounds good to me as a general term.

The two transit-oriented aerial lifts I know of in the United States have _tram_ in their names:
* [Portland Aerial Tram](https://en.wikipedia.org/wiki/Portland_Aerial_Tram)
* [Roosevelt Island Tramway](https://en.wikipedia.org/wiki/Roosevelt_Island_Tramway)

I think "aerial tramway" in US English is reasonably understood to be a transit-oriented `aerialway=gondola` or `aerialway=cable_car`. But at a ski resort, I'd definitely call that a gondola.

jleedev:
Shucks, `landuse=winter_sports` isn't in the tiles.

ZeLonewolf:
Yeah, a little background on that:
https://github.com/openmaptiles/openmaptiles-skiing

wmisener:
The bright red makes these features very visually prominent, and thus by implication very important (though the width adjustment helped make them a bit more muted). That prominence makes sense enough on a map of a single ski resort, especially with their snow-white backgrounds, but I'm not sure it's right for a general purpose map. I don't know if these aerialways are that much more important to a typical user than, e.g., railways, which render much more subtly. This especially strikes me in the examples in urban settings.

On Americana so far, red has been reserved for freeways and trunk roads, so this would change/dilute the meaning of the color red in the style. The current casing style also makes them look quite similar to a trunk-expressway IMO. I wonder if it would be more consistent for these features to be purple, the designated "transportation" color (although I think airports are the only thing rendered in that class so far), as has been toyed with for busways in #477.

1ec5:
I agree that it would be confusing to overload red to mean either an important road or something very different, especially since the two can cross. There’s a limit to how much inspiration we can draw from a site map of any kind, whether a ski resort, golf course, or shopping mall, since a site map can afford to deemphasize anything off-site considerably. If we need to maintain a connection to ski resort maps, perhaps pink (with a white casing) would be sufficiently distinct from a road, but it would need to be desaturated a bit to avoid overwhelming the map.

claysmalley:
Desaturated a little. Here's a sample of red, pink and purple:

Screenshot from 2023-02-02 20-32-52
Screenshot from 2023-02-02 20-35-15
Screenshot from 2023-02-02 20-36-28

Screenshot from 2023-02-02 20-33-22
Screenshot from 2023-02-02 20-35-28
Screenshot from 2023-02-02 20-36-20

Screenshot from 2023-02-02 20-34-31
Screenshot from 2023-02-02 20-35-43
Screenshot from 2023-02-02 20-36-37

quincylvania:
Color aside, I think these might benefit from rendering little dots for the endpoints, any intermediate stations, and maybe at higher zooms the [`aerialway=pylon`](https://wiki.openstreetmap.org/wiki/Tag:aerialway%3Dpylon) nodes. This would indicate the distinctive cable infrastructure and allow lines to stand out while having a thinner stroke. Like you could probably go down to one centered stroke here (showing just stations at this zoom, not poles):

Artboard

osm-carto takes a nod at this approach but doesn't use the actual locations of the poles, just a pattern, which I personally find a little misleading and silly-looking at high zooms.

Screenshot 2023-02-02 at 8 43 56 PM

claysmalley:
Aerialway stations tracked in #692. I think matching the color of aerialway lines is a good idea.

Unfortunately `aerialway=pylon` isn't present in the tiles, otherwise it would definitely be implemented in this PR.

adamfranco:
+1 for the purple option. I like how it fits the transit theme and also its visual weight. It's distinctive enough to be able to follow through the myriad grayscale lines of Bolivia without jumping out too much.

adamfranco:
Also, at very high zooms the knockout might be dropped to let the wires stand alone as they cross over buildings and other features at something approximating a realistic width. With pylons (at some point) that effect could be really neat.

claysmalley:
Changed to airport purple. Here are some aerialways near airports for context:

[Screenshot from 2023-02-03 15-24-18](http://localhost:1776/#map=13.71/45.40468/6.62885)
[Screenshot from 2023-02-03 15-24-55](http://localhost:1776/#map=12.09/-16.51363/-68.17508)
[Screenshot from 2023-02-03 15-25-41](http://localhost:1776/#map=15.25/-16.499222/-68.174327)

claysmalley:
Line gap now significantly widens at very high zoom:

[Screenshot from 2023-02-03 16-11-16](http://localhost:1776/#map=18.44/-16.5122554/-68.1537056)

ZeLonewolf:
The purple is definitely an improvement over the red.  What is the distinction between the double-thin lines and the strong dashed lines?

claysmalley:
Dashed line, feet touch the ground. Two parallel lines, you're five stories up.

claysmalley:
Note in the Courchevel example above, there are two drag lifts (dashed lines) that pass under road bridges. Ideally, `layer` would be exposed here, but for most cases we can assume drag lifts are at `layer=0` and aerial lifts are at `layer=6`.

ZeLonewolf:
For inspiration, an areial tram on SwissTopo.



I like that it's desatured and thinner. Wondering if we can try that but with our purple hue.  Also consider long-dashes.

claysmalley:
All aerialways now transition from single lines to double lines at zoom 15:

[Screenshot from 2023-02-06 13-23-11](http://localhost:1776/#map=12.24/45.40672/6.63046)
[Screenshot from 2023-02-06 13-24-13](http://localhost:1776/#map=16.48/45.414391/6.633179)
[Screenshot from 2023-02-06 13-25-15](http://localhost:1776/#map=12.57/-16.50343/-68.13996)
[Screenshot from 2023-02-06 13-26-03](http://localhost:1776/#map=17.29/-16.512452/-68.153709)
[Screenshot from 2023-02-06 13-27-11](http://localhost:1776/#map=14.98/40.75851/-73.95927)
[Screenshot from 2023-02-06 13-28-04](http://localhost:1776/#map=15.82/40.760001/-73.962796)
[Screenshot from 2023-02-06 13-29-18](http://localhost:1776/#map=17.82/40.760797/-73.963585)

src/layer/aerialway.js
```diff
@@ -0,0 +1,144 @@
...
+  "source-layer": "transportation",
+};
+
+export const liftCasing = {
+  id: "lift_casing",
+  type: "line",
+  paint: {
+    "line-color": Color.backgroundFill,
+    "line-width": casingLineWidth,
+    "line-gap-width": casingLineGapWidth,
```
1ec5, line 92: The legend doesn’t correctly implement `line-gap-width`, making this legend entry look misleading: #770.

[Roosevelt Island Tramway](https://zelonewolf.github.io/openstreetmap-americana/#map=17.82/40.761363/-73.962493)

claysmalley:
In case anyone was worried that drag lifts would clash with political boundaries:

[Screenshot from 2023-02-07 20-24-18](http://localhost:1776/#map=13.85/45.92273/7.72628)

zekefarwell:
Of the three colors shown in screenshots so far, the pink looks best to my eyes.  I also find the endpoint dots Quincy suggested also help set the aerialways apart from other liner features.  Something like that seems like a good idea.

The low zoom screenshots with a single line rather than a double were looking more chairlift-like to me.  I wasn't sure why until I went looking at ski resort trail maps and remembered that they generally just use a single line.   Maybe a single line would work well at higher zoom levels too?  

image
image
image
image
image

Or maybe there are other types of maps that commonly show aerial lifts as double lines?

claysmalley:
> Of the three colors shown in screenshots so far, the pink looks best to my eyes.

There seems to be some conflicting opinions on the color. I do like the purple, though I suspect whatever color we use here is likely to become the public transit color. Purple would make train stations and bus stops hard to find within airports.

> I also find the endpoint dots Quincy suggested help set the aerialways apart from other liner features.

Aerialways always start and end at stations, which are POIs that I intend to implement separately. #692

> The low zoom screenshots with a single line rather than a double were looking more chairlift-like to me. I wasn't sure why until I went looking at ski resort trail maps and remembered that they generally just use a single line. Maybe a single line would work well at higher zoom levels too?

Cables are typically 15-20 feet apart in real life. There aren't many high-zoom examples of aerialways on maps, but at a zoom level where indoor features could potentially show up, it looks clunky to me if this gap isn't present. I could probably adjust the threshold to a higher zoom level, though.

[Screenshot from 2023-02-07 22-24-07](http://localhost:1776/#map=17.29/-16.512452/-68.153709)
[Screenshot from 2023-02-06 13-26-03](http://localhost:1776/#map=17.29/-16.512452/-68.153709)

1ec5:
Regarding the color, I think the bubblegum pink or the red has a certain flair that fits with the _Pop Art_ labeling. Makes it feel more sporty. But a more muted purple is fine too. The important thing is that no one would mistake it for a kind of road,[^black] and I think purple accomplishes that.

[^black]: Ironically, if we had chosen black or gray for runways, as many print maps do, we probably could’ve gotten away with making aerialways the same color. Readers wouldn’t have as much of an expectation that the color indicates the kind of conveyance.

1ec5:
By the way, the screenshot in https://github.com/ZeLonewolf/openstreetmap-americana/pull/755#issuecomment-1421944578 seems to be showing an aerialway name of the format `Line: Terminus ↔ Terminus`, which is a controversial part of the PTv2 tagging scheme that places non-name data inside `name`. Is this a common occurrence, and are we expected to hide `name`s used in this manner?

claysmalley:
> By the way, the screenshot in https://github.com/ZeLonewolf/openstreetmap-americana/pull/755#issuecomment-1421944578 seems to be showing an aerialway name of the format `Line: Terminus ↔ Terminus`, which is a controversial part of the PTv2 tagging scheme that places non-name data inside `name`. Is this a common occurrence, and are we expected to hide `name`s used in this manner?

I haven't seen that outside of Bolivia. I agree that the termini shouldn't be in the name, but that is a tagging issue for local communities to solve. I wouldn't bother with writing extra code to hide names that conform to suspicious PTv2 tagging.

ZeLonewolf:
While red would be great on a ski-specific map, I think having purple be the general-purpose transit color makes sense as visual language. And the doubling at high zoom is a nice touch.

The one thing that stood out to me is that the font descents obscure the lines a bit.  Are we okay with that or should the text be offset a bit more?

image

claysmalley:
I added a text offset, though it seems to be offsetting in the opposite direction half the time and I'm not sure why:

Screenshot from 2023-02-26 10-34-22

claysmalley:
One last change: magenta. Because if we're gonna need to repurpose this color for public transit, it can't be the airport purple, since airports aren't necessarily passenger transport facilities. Also, I'm gay for transit :rainbow_flag::aerial_tramway:

I think this is as good as it's going to get for now. Tail work:

- [ ] aerialway stations (part of #692)
- [ ] aerialway pylons not present in tiles
- [ ] `text-offset` is applied in the wrong direction on lines that appear right-to-left when `line-gap-width > 0`. Am I doing something wrong or is this a Maplibre bug?

[Screenshot from 2023-03-05 21-30-38](http://localhost:1776/#map=14.74/40.75881/-73.96054)

[Screenshot from 2023-03-05 21-48-20](http://localhost:1776/#map=12.47/-16.50377/-68.16689)

[Screenshot from 2023-03-05 21-49-12](http://localhost:1776/#map=16.17/-16.499224/-68.164653)

[Screenshot from 2023-03-05 21-50-38](http://localhost:1776/#map=12.41/39.47371/-106.06839)

[Screenshot from 2023-03-05 21-52-17](http://localhost:1776/#map=15.68/45.413041/6.631128)

ZeLonewolf:
Regarding:
> text-offset is applied in the wrong direction on lines that appear right-to-left when line-gap-width > 0. Am I doing something wrong or is this a Maplibre bug?

Can you summarize what's different between aerialway rendering and road rendering that would cause this bug to appear in one case and not the other?

claysmalley:
`text-offset` is defined in `src/layer/transportation_label.js`, and it only applies to aerialways so far. When the direction of a line as mapped on OSM goes right-to-left across the screen, the offset seems to work in the opposite direction as expected.

`line-gap-width` may be a red herring, as it's only present in `src/layer/aerialway.js`.

1ec5:
Here’s one of those test areas above:

text-keep-upright: true

Setting `text-keep-upright` to `false` illustrates the problem:

text-keep-upright: false

Essentially, we’re expecting `text-keep-upright` to flip the symbol onto the other side of the line, but all it does is flip the text itself around.

One workaround would be to set `text-anchor` to `center`, centering the text over just one of the lines:

No text-offset

With a wider knockout, it could do a good job of emphasizing that there are two parallel lines as opposed to a casing.

1ec5:
> Essentially, we’re expecting text-keep-upright to flip the symbol onto the other side of the line, but all it does is flip the text itself around the text-anchor, which is currently set to bottom.

Moreover, I think this GL JS behavior is pretty reasonable. For example, the boundary edge labels in #52 shouldn’t flip across the line based on `text-offset`, because that would overlap with the other side’s label. If we wanted a different behavior in the renderer, it would have to be some other `text-anchor` value that’s lower than `bottom`.

claysmalley:
Setting `text-anchor` to `center` and increasing the offset seems to have the effect we're looking for:

[Screenshot from 2023-03-07 16-51-00](http://localhost:1776/#map=17.08/-16.51237/-68.15383)

[Screenshot from 2023-03-07 16-53-26](http://localhost:1776/#map=13.65/45.38969/6.56954)

[Screenshot from 2023-03-07 16-57-32](http://localhost:1776/#map=15.09/36.53631/138.34612)

ZeLonewolf (approved):
Looks good.  I think it's time to get this change in the map!

PR #757, opened 2023-02-02
increase minimum width of diamond shields

claysmalley:
The minimum size for diamond shields should be 24x24.

Before/after [(localhost)](http://localhost:1776/#map=12.9/10.50966/-66.907):
Screenshot from 2023-02-01 20-53-27
Screenshot from 2023-02-01 20-53-51


ZeLonewolf (approved):
Looks good!

PR #759, opened 2023-02-03
Icons and descriptions in taginfo

wmisener:
Added icon links and a more detailed description of the icon for schools, and added taginfo entries for bars, pubs, cafes, hospitals, doctors, clinics, and parking POI categories. I think these were accidentally omitted in the pull which added them (#387).


1ec5 (approved):
Thanks!

PR #761, opened 2023-02-03
Guard against incompatible expression

jleedev:


src/js/util.js
```diff
@@ -23,6 +23,9 @@ export function layerClone(def, id) {
 //Make a clone of a layer definition, with a filter added
 export function filteredClone(def, filterStep, idSuffix) {
   var clone = layerClone(def, def.id + idSuffix);
+  if (!["all", "any"].includes(clone.filter[0])) {
+    throw new TypeError("Unlikely filter");
+  }
```
ZeLonewolf, line 28: This makes sense, though I can't think of any case where we'd use this on anything besides "all"

ZeLonewolf (approved):

PR #762, opened 2023-02-03
Validate the output style json

jleedev:
The map will refuse to start with errors anyway, but this will cause them to show up in the checks.


ZeLonewolf:
Will this also cause the build to fail if there are errors?

jleedev:
Yes.

I expect that would only be an issue if the two npm packages happen to diverge, but not if their releases go correctly.

ZeLonewolf (approved):
Looks good to me!

PR #764, opened 2023-02-04
Fix lightening of anti-aliased pixels

quincylvania:
The `colorLighten` property is a cool way to reduce the spritesheet by changing shield colors at runtime, but it has issues with anti-aliased pixels that aren't strictly white or black.

Here is a 1x shield without `colorLighten`. This is how it should look.

Screenshot 2023-02-04 at 1 50 04 PM
Screenshot 2023-02-04 at 3 03 48 PM

Here's the shield with a reusable black sprite and `colorLighten` set to brown.

Screenshot 2023-02-04 at 1 47 20 PM
Screenshot 2023-02-04 at 3 00 45 PM

Here it is with this PR. Indistinguishable from rendering with a brown SVG.

Screenshot 2023-02-04 at 2 55 59 PM
Screenshot 2023-02-04 at 3 02 38 PM

With this fix, I suggest keeping all monochrome POI icons and adjusting the colors in the definitions.


claysmalley (approved):
Looks good! I was wondering what caused `colorLighten` to behave this way; glad that you figured it out.

Old/new/diff, New Brunswick primary routes:

old-nb new-nb diff-nb

ZeLonewolf:
Thanks for figuring out how to do this!

PR #765, opened 2023-02-05
Add Natchez Trace Parkway shield

quincylvania:
This one was fun. Supporting historic roads like this helps emphasize the "Americana" aspect of the project. In any case, a 444-mile limited-access route deserves some love. This might be a case where we'd want to show the name label as well as the shield at high zooms, not sure if that's a blocker.

https://www.openstreetmap.org/relation/3300032
https://en.wikipedia.org/wiki/Natchez_Trace_Parkway
https://commons.wikimedia.org/wiki/File:Natchez_Trace_Parkway_Logo.gif

The shield was a bit tricky to condense, but I think focusing on the rider captures the essence. I ignored the brown outer border present on some signs since three layers of borders is too many, and the yellow makes it pop more.

shield_us_nps_ntp

2x:

Screenshot 2023-02-05 at 4 12 15 PM

1x:

Screenshot 2023-02-05 at 4 12 24 PM

Screenshot 2023-02-05 at 4 12 56 PM

Screenshot 2023-02-05 at 4 12 36 PM

Pics from the web:

Natchez_Trace_Parkway_Logo
Screenshot 2023-02-05 at 4 18 24 PM
s-l300-2150528874
4047654091_8d070143da_n-2528586032


1ec5 (approved):
As in https://github.com/ZeLonewolf/openstreetmap-americana/pull/562#discussion_r991796731, a tan/cream background might be more faithful to the original signage, but white helps with contrast. I agree with focusing on the rider; the tree (beyond these few leaves) would’ve been difficult to discern anyways.

quincylvania:
Related to the possible shield restrictions raised in #1101, NPS has [also registered](https://www.federalregister.gov/documents/2000/03/28/00-7563/official-trail-marker-for-the-natchez-trace-national-scenic-trail) the shield for the "Natchez Trace National Scenic Trail". While this is technically separate from the Natchez Trace Parkway shield, the core graphic is identical. Not sure if this is something to worry about.

PR #538, opened 2022-08-15
Adjust shield padding for Roboto Condensed metrics

claysmalley:
Small adjustments to the padding of a bunch of shield icons to make sure both Arial Narrow and Roboto Condensed choose the right width variant icon, and otherwise look good.

PR #768, opened 2023-02-06
Consistent shield text layout

### LINKED ISSUE #625: Shield labels don't fit properly with Safari
Labels don't look correct when using Safari on iPhone and Mac.

See https://osmus.slack.com/archives/C01V02K52UX/p1647394819815939

IMG_1994

### LINKED ISSUE #834: Two-digit Interstate shields are too wide
Sometime in the last month, the text sizing code changed in such a way that two-digit Interstate shields are too wide and their numerals are too large, at least on macOS where a condensed font is being applied.

[Indianapolis](https://zelonewolf.github.io/openstreetmap-americana/#map=12/39.6868/-86.18271)

quincylvania:
Closes #625.
Closes #834 

I've identified the following inconsistencies between Safari and other browsers:
- Text has different y-placement when [`textBaseline`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/textBaseline) is set to anything other than `alphabetic`. This means we need to use `alphabetic` for everything.
- [`measureText`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/measureText) values can vary by over 15%, regardless of `textBaseline`. We can't rely on this function.
- [`fillText`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/fillText) will round off the `y` coordinate in some browsers. We need to always pass in an integer for consistent results.

I tried replacing `measureText` by rendering the text to a canvas and counting the pixels, but that proved unruly. The simplest and most consistent solution is to replace the drawing context entirely and just hardcode the height and width of text based on the font size. This gives us the most control anyway, though it may require more tweaking, especially if we wanted to change fonts.

One caveat is that this won't work well with characters that go above or below the "M" height, such as accent marks or lowercase letters like "g" or "q". My North American brain doesn't know of any `ref` tags that would include those, but we should keep an eye out.

I also removed the `valign` property from the constraints. Vertical alignment can be handled simply by setting the padding correctly.

**Safari before, 1x:**

Screenshot 2023-02-06 at 2 07 17 PM

**Firefox before, 1x:**

Screenshot 2023-02-06 at 2 09 56 PM

**Safari after, 1x (Firefox is essentially the same):**

Screenshot 2023-02-06 at 2 06 09 PM


jleedev:
Some regression on Firefox Android:

[](https://zelonewolf.github.io/openstreetmap-americana/#map=11.4/40.3468/-80.1618/0/4)

Two digit interstate is fixed, but three-digit PA is now too big.

I think this is because Firefox refuses to see the Roboto Condensed system font.

Is there a way to avoid making the font bigger in these? The issue in this case was using too large a font on the two digit interstates.

1ec5:
> One caveat is that this won't work well with characters that go above or below the "M" height, such as accent marks or lowercase letters like "g" or "q". My North American brain doesn't know of any ref tags that would include those, but we should keep an eye out.

France and Italy have some “bis” and “ter” routes, but I don’t know if it goes up to “quater” anywhere.

On the topic of exotic characters in route numbers, the opinion in #465 was that `ref` should be in the local writing system. Eastern Arabic numerals can be found in the Middle East, and Taiwan uses Chinese characters instead of A/B/C.

> I think this is because Firefox refuses to see the Roboto Condensed system font.

Relevant to #538 perhaps?

ZeLonewolf:
On Chrome/Web

Looks like the interstate text wants to scoot up by a pixel.

image

ZeLonewolf:
> France and Italy have some “bis” and “ter” routes, but I don’t know if it goes up to “quater” anywhere.

Italy absolutely has qtr suffixed routes as I've driven on them.  However, they're not well mapped in route relations.

ZeLonewolf:
SS7qtr in Caserta province, Italy:

image

zekefarwell:
> Oh dear. I'd say we want to keep the fonts consistent across all platforms/browsers if at all possible.

Yes, ideally 😬.  I got started exploring shield font and width cross browser standardization in this draft PR: https://github.com/ZeLonewolf/openstreetmap-americana/pull/544

It also seemed to me like we could simplify the code significantly if we just picked the shield blank width based on the number of characters rather than measuring the pixel width of the text.

1ec5:
> It also seemed to me like we could simplify the code significantly if we just picked the shield blank width based on the number of characters rather than measuring the pixel width of the text.

Other styles take this approach, such as openstreetmap-carto and Mapbox Streets. But measuring the text has allowed us to achieve a very tight appearance by comparison, at least when it works. The old shield renderer used to have a special case for refs that start with 1, since it’s so narrow. But if we don’t control the fonts well, then we’d have to have many such special cases.

zekefarwell:
> measuring the text has allowed us to achieve a very tight appearance by comparison, at least when it works

It hasn't ever worked in Safari. I'd say predictable consistent behavior between browsers is more important, but that's just me.

ZeLonewolf:
If we use @zekefarwell's approach of just tying shield selection to ref length, there is a setting in the canvas text draw that lets you stretch the text in the horizontal direct.  So, you could force-fit things if there's a wider character.

1ec5:
> It hasn't ever worked in Safari.

This PR is all about fixing #625.

quincylvania:
> If it results in a regression in Firefox on Android, then maybe we need some browser detection.

I'd much rather look into fixing the font differences. The style shouldn't look different on different devices, and the font is not an insignificant part of the style.

zekefarwell:
> there is a setting in the canvas text draw that lets you stretch the text in the horizontal direct. So, you could force-fit things if there's a wider character

Yup, that's exactly what #544 does.  

> maybe we need some browser detection. That’s par for the course in a Web application

The main problem is just that we rely on a system font stack.  If we ship a web font that should keep things much more consistent across browsers (regardless of font choice)

claysmalley:
> One caveat is that this won't work well with characters that go above or below the "M" height, such as accent marks or lowercase letters like "g" or "q". My North American brain doesn't know of any ref tags that would include those, but we should keep an eye out.

In Armenia, `Մ` and `Հ` appear in `ref` as prefixes to numbers. Same with `ს` and `შ` in Georgia. On real-life signage, these letters seem to match the height of numerals, but some of these letters are taller than numerals in certain fonts.

quincylvania:
> If we ship a web font that should keep things much more consistent across browsers (regardless of font choice). Of course web fonts can be heavy, so that's a consideration too.

Shipping fonts is standard for modern websites, and I'd say is especially warranted in cases where style is forefront (e.g. maps). Without including a specific known font, we can't be sure that the map will look good or even legible for all users, no matter how adaptive the code might be. Here's a case in Ukraine where Americana is broken because the font doesn't have the Cyrillic `М` character.

Screenshot 2023-02-07 at 5 56 37 PM

Indeed, font files are relatively large assets and should be kept to a minimum. But once they're loaded once you're good to go.

claysmalley:
> Here's a case in Ukraine where Americana is broken because the font doesn't have the Cyrillic М character.

The CSS `font-family` list for shield text is `"sans-serif-condensed", "Arial Narrow", sans-serif`. Right now, we don't explicitly list any fonts with non-Latin characters, but if your device's default sans-serif font has Cyrillic characters like mine, it should show up.

Screenshot from 2023-02-07 19-17-16

ZeLonewolf:
If I understand the problem correctly, if we used a web font rather than a system font, it would solve the browser/system font measurement problem. ([stackoverlow](https://stackoverflow.com/questions/6117553/external-font-on-html5-canvas)). That would seem a lot cleaner than storing character widths in code?

quincylvania:
@ZeLonewolf I do think we should bundle web fonts (or use hosted) to guarantee consistent results across platforms. I don't think that will fix Safari's height measurement issues, but we might be able to get consistent widths without hardcoding them.

quincylvania:
Upon further testing, it looks like the `width` of `measureText` is probably consistent enough across browsers, it's the height that's the issue. I removed the hardcoded character widths. This should fix some issues with other fonts, but I still think we should use a known web font.

1ec5:
> I still think we should use a known web font.

Agreed. If nothing else, I’ve had it up to <sup>here</sup> with the uncondensed shield font in @ZeLonewolf’s screenshots. 😉

ZeLonewolf:
A web font can be substituted by:

1. Adding the font in the head of index.html and shieldtest.html to be something like this:
```html
<link rel="preconnect" href="https://fonts.googleapis.com" />
<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
<link href="https://fonts.googleapis.com/css2?family=Barlow+Condensed:wght@700&display=swap" rel="stylesheet">
```
Note that the final link will differ based on what font and weight is selected, but Google Fonts will give you the code

2. At the top of screen_gfx.js, change the font variable:
```javascript
const fontFamily = 'Name of Font';
```
Hopefully, we can get to an attractive and consistent visual between the switch to a web font and the tweaks already here. Barlow Condensed was suggested in [Slack](https://osmus.slack.com/archives/C01V02K52UX/p1676860862598029?thread_ts=1675729207.628889&cid=C01V02K52UX) but I'm ambivalent about which font is chosen (and it's not hard to tweak later).

1ec5:
There’s a lot of wiggle room for the particular font, as long as it’s condensed. The normal-pitch fonts that otherwise come up in these conversations just seem too wide to look serious. Roboto Condensed has come up as a suggestion; that would be reasonable. A condensed font would allow us to choose a thinner weight. This isn’t necessary for now, but if we ever decide to shrink the basic shield size smaller than 20×20, then it would become necessary.

ZeLonewolf:
I did notice one artifact - the "A 562" appears to be offset to the south for no obvious reason. This is still better than the current rendering and I'd be ok with logging this as a bug (might have to do with the space in the ref?).  But curious nonetheless.
image

I'm not totally convinced about eliminating the vertical alignment factor but on the interstate shields, I guess it does work to center it vertically. Just concerned about making best use of the biggest block of space. However, I can't argue with the visual results, which seem quite reasonable:
image

quincylvania:
> I did notice one artifact - the "A 562" appears to be offset to the south for no obvious reason.

This might be a rounding issue, or something weird with the returned text width. I can look into it once we pick a font.

> I'm not totally convinced about eliminating the vertical alignment factor

I didn't understand what this was for when you can just set the alignment with `padding`.

ZeLonewolf:
> > I'm not totally convinced about eliminating the vertical alignment factor
> 
> I didn't understand what this was for when you can just set the alignment with `padding`.

I'll attempt to draw out what that code was doing (or intending to do), and then we can decide if it's needed or needs to be scrapped.

ZeLonewolf:
The graphic below describes what's happening for a shield using padding and elliptical text constraint (which has a "middle" vertical alignment). The shield artwork is not shown, but imagine that this is all drawn on top.

The maroon is the padding, which constrains everything.

From that, the label is logically placed in the center, and then is expanded proportionately until the corners of the text bounding box hit the ellipse.

image

Below is a similar example, but this time using a southHalfEllipseConstraint. That label starts with top alignment and then expands downward until it hits the half-ellipse limit.

image

For elliptical constraints there's really no such thing as vertical alignment, you draw the text in that bounding box and that's that.

For rectangular constraints (or rounded rectangles), you have a choice to make if the text is wide:
image

In that case -- do you snap the text to the middle or the top or the bottom? That's a style question that may differ based on shield artwork. The alignment parameter I believe was intended to give that option, but I'm not sure if any shield actually uses that in practice.

So after all this discussion - if you're confident that the design intents above are satisfied that I'm good with dropping the alignment code.

ZeLonewolf:
IMO, the Overpass 500wt font used in @zekefarwell's attempt in #544 looks and feels like the superior font for shields, and if we use horizontal scaling, it solves what @1ec5 is going for in the desire for a more condensed font. Is it possible to take the code changes here, and the font selection and horizontal scaling from there?

quincylvania:
> I'm also perfectly happy if someone else wants to pull the ideas from that branch into this one instead.

@zekefarwell I'll take a look and see what I can integrate here! I've been waiting on #801 before continuing work here, though maybe I don't need to wait.

zekefarwell:
I don't think you need to wait.  The font stack used for drawing shield graphics is entirely different than the one used by MapLibre.  Different font formats and contexts.

ZeLonewolf:
Agreed, #801 is be completely independent of the work here.

quincylvania:
> I'll take a look and see what I can integrate here!

I pulled in the changes from #544 and made a few tweaks of my own.

The shield sprite and the font size are now selected based on the ref character count instead of the text width for all 1-, 2-, and 3-character refs. We avoid overflow by setting the `maxWidth` property of `fillText` to automatically condenses wider text.

I'd say the Overpass font is an improvement, but as a wide font, results in wildly different-looking fonts when condensed and uncondensed text are near each other. I'd recommend we pick a condensed base font such that the `fillText` magic is more subtle.

Screenshot 2023-03-12 at 1 39 58 PM
Screenshot 2023-03-12 at 1 33 16 PM

quincylvania:
I decided I really don't like the squished text affect so I removed it. However, I still think shields tend to look better when font size is consistent based on the number of characters.

Screenshot 2023-03-13 at 7 25 59 PM
Screenshot 2023-03-13 at 7 25 31 PM

src/js/shield_defs.js
```diff
@@ -1080,10 +1072,10 @@ export function loadShields() {
     textLayout: textConstraint("southHalfEllipse"),
     textColor: Color.shields.white,
     padding: {
-      left: 4,
-      right: 4,
+      left: 5,
+      right: 5,
       top: 6,
```
ZeLonewolf, line 1077: Reducing the top padding by 1px seems to give a better result:
```suggestion
      top: 5,
```
image

ZeLonewolf:
This looks good to me on Ubuntu Chrome and Firefox.  Need someone to test on Mac/Safari.

1ec5:
Did we decide at some point not to use a condensed font anymore? I really think a condensed font makes the shields look more polished, especially the bowl-shaped shields that grow horizontally by a lot with a normal-pitch font. Earlier in this discussion there was a consensus about using a condensed font, but simulating a condensed font by squishing the text horizontally doesn’t quite work.

src/index.html
```diff
@@ -97,6 +103,9 @@
 
   <body>
     <div id="map"></div>
+    <p style="font-family: Overpass; font-weight: 600; visibility: hidden">
+      <b style="font-weight: 700">Invisible</b> text so the font will load early
+    </p>
```
1ec5, line 108: Do we need to set the ARIA role to [`presentation`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/presentation_role) or apply a `speak: none` property to force screen readers to skip this text?

src/js/screen_gfx.js
```diff
@@ -2,9 +2,14 @@
 
 import rgba from "color-rgba";
 
-const fontFamily = '"sans-serif-condensed", "Arial Narrow", sans-serif';
-export const shieldFont = (size) => `bold ${size}px ${fontFamily}`;
-export const fontSizeThreshold = 12;
+const shieldFontFamily =
+  'Overpass, "sans-serif-condensed", "Arial Narrow", sans-serif';
```
1ec5, line 6: While we’re at it loading a Web font from Google Fonts, I’m partial to either Roboto Condensed or Ubuntu Condensed. It already looks pretty good (and indistinguishable from the previous Arial Narrow) even before the various dimensional tweaks in the shield definitions.

Roboto Condensed

src/js/shield_text.js
```diff
@@ -243,15 +213,17 @@ export function drawBannerText(ctx, text, bannerIndex) {
   ctx.fillStyle = "black";
 
   ctx.font = Gfx.shieldFont(textLayout.fontPx);
-  ctx.textBaseline = "top";
+  ctx.textBaseline = "alphabetic";
```
1ec5, line 216: The top pixel or so of each banner is cut off:

[Downingtown](https://zelonewolf.github.io/openstreetmap-americana/#map=15/40.00801/-75.6983)

ZeLonewolf:
With the merger of #699, these changes will need to get worked into the shield library.

ZeLonewolf:
Hopefully resolved by #1010

ZeLonewolf:
Supplanted by #1010. Thanks for working on this!

Issue #744, opened 2023-01-30
UI chrome should be localizable


1ec5:
Virtually all of the _content_ in the Americana application is localizable, and much of it is already translated. However, major parts of the UI are still hard-coded in English:

* [ ] GL JS controls[^gljs]
* [ ] Samples and Legend buttons
* [ ] Map attribution[^attribution]
* [ ] Language switcher
* [ ] Legend entries (other than the “Route markers” section)
* [ ] “Fork me on GitHub”
* [ ] Search box

To the extent that we need to maintain our own translation tables, there’s an implementation in https://github.com/ZeLonewolf/openstreetmap-americana/pull/215#issuecomment-1060067427 that we can lift pretty easily. For the time being, I don’t think we’ll have enough translatable strings to require a translation management system such as Transifex or Translatewiki.net, but that can happen after we have basic infrastructure in place for translations.

[^gljs]: When configuring the map, we can set the `locale` option to provide our own translations of [these strings](https://github.com/maplibre/maplibre-gl-js/blob/862e92bc32673671612ebe12a9f8e9e0cdd35d72/src/ui/default_locale.ts#L2-L18). Unfortunately, GL JS doesn’t provide translations of its own: https://github.com/mapbox/mapbox-gl-js/issues/12109#issuecomment-1195748154.
[^attribution]: Ideally, the metadata associated with each tileset would come with localized attribution strings, just as the tilesets provide localized versions of each feature’s name. However, the TileJSON format only supports a very basic format for attribution: mapbox/tilejson-spec#20.

PR #773, opened 2023-02-08
Group shields in legend by highway system

### LINKED ISSUE #771: Tennessee state route shields have repetitive entries in legend
Tennessee’s primary and secondary state route shields have separate entries in the legend, but since these are two shields for the same network, Wikidata has [only a single item associated with both](https://www.wikidata.org/wiki/Q98791713#P1282) `network=US:TN:primary` and `network=US:TN:secondary`, [resulting in a redundant entry](https://osmus.slack.com/archives/C01V02K52UX/p1675779305556649):

[McMinnville](https://zelonewolf.github.io/openstreetmap-americana/#map=11/35.6927/-85.798&language=en)

The legend should combine these rows into a single row with two shields on it, just as it does in other cases where there are multiple shields within the same shield definition.

1ec5:
The legend now incorporates Wikidata labels into its “Route markers” section while generating the section, instead of patching in the labels afterwards. If the Wikidata Query Service response hasn’t come back by the time the section is generated, the legend blows away the section and regenerates it once it receives the response.

When generating the section, entries are now combined with other entries that correspond to the same highway system item in Wikidata. This improves the display of highway systems that haven’t been modeled in as much detail in Wikidata. More detailed highway systems continue to get separate rows with separate labels.

[Murfreesboro](https://zelonewolf.github.io/openstreetmap-americana/#map=12/35.86777/-86.37516&language=en) [Guayama](https://zelonewolf.github.io/openstreetmap-americana/#map=15/17.97549/-66.10174&language=es)

Fixes #771.

### DIFF
--- a/src/js/legend_control.js
+++ b/src/js/legend_control.js
@@ -11,6 +11,15 @@ import * as maplibregl from "maplibre-gl";
 
 const maxPopupWidth = 30; /* em */
 
+/**
+ * Wikidata labels are normally lowercased so that they can appear in any
+ * context. Convert them to sentence case for consistency with the rest of the
+ * legend.
+ */
+function toSentenceCase(lowerCase, locale) {
+  return lowerCase[0].toLocaleUpperCase(locale) + lowerCase.substring(1);
+}
+
 export default class LegendControl {
   onAdd(map) {
     this._map = map;
@@ -72,13 +81,12 @@ export default class LegendControl {
     this.close();
 
     let contents = this.getContents();
-    let rows = contents.querySelectorAll(".legend-row");
     this._popup.setDOMContent(contents);
 
     let anchorCoordinate = this._map.unproject(anchor);
     this._popup.setLngLat(anchorCoordinate).addTo(this._map);
 
-    this.prettifyNetworkLabels(rows);
+    this.completeNetworkLabels();
 
     document.getElementById("legend-container").scrollTop = 0;
   }
@@ -107,6 +115,9 @@ export default class LegendControl {
     for (let data of this.sections) {
       let section = this.getSection(data);
       if (!section) continue;
+      if (data.id) {
+        section.id = `legend-section-${data.id}`;
+      }
 
       let container = template.getElementById("legend-container");
       container.appendChild(section);
@@ -143,7 +154,7 @@ export default class LegendControl {
     }
     if (!rows.length) return;
 
-    template.querySelector("tbody").replaceChildren(...rows);
+    template.querySelector(".legend-row-container").replaceChildren(...rows);
     if (!data.source) {
       template.querySelector("tfoot").remove();
     }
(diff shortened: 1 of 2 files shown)


ZeLonewolf:
Yessss
image

Issue
image

At this location, console errors:
http://localhost:1776/#map=10/40.6983/-73.958

What happened here?
[image](http://localhost:1776/#map=6/38.141/-93.347)

Issue
[image](http://localhost:1776/#map=13/36.64379/-93.27921)

1ec5:
This PR is specifically about grouping shields that belong to the same highway system but not the same `network` – subnetworks, in other words.

> Yessss image

This was implemented as part of #632. (Compare with The King’s Highway in Ontario.)

> Issue image

This is an unrelated regression in the shield drawing code: #781.

> At this location, console errors: http://localhost:1776/#map=10/40.6983/-73.958

Fixed.

> What happened here? [image](http://localhost:1776/#map=6/38.141/-93.347)

Whenever the legend encounters instances of a network with and without a `ref`, it shows both, since the two are likely to differ visually. None of the routes in the Oklahoma Turnpike system are numbered per se, but the Indian Nation Turnpike was incorrectly tagged as State Highway 375, which is concurrent with the turnpike. This tagging error was fixed in [changeset 131,759,548](https://www.openstreetmap.org/changeset/131759548); the fix is awaiting a tile refresh.

We could suppress the duplicate shield whenever the shield definition’s `notext` property is `true`, but I think the duplication is a good way of spotting this particular class of tagging errors, which was historically quite common before any renderer consumed road route relations.

> Issue [image](http://localhost:1776/#map=13/36.64379/-93.27921)

Also #781.

src/js/legend_control.js
```diff
@@ -569,9 +600,46 @@ export default class LegendControl {
...
+      if (locale.match(/^\w+/)?.[0] !== locales[0].match(/^\w+/)?.[0]) {
+        let languageTag = document.createElement("span");
+        languageTag.className = "language";
+
+        let languageNames = new Intl.DisplayNames(locales, {
+          type: "language",
+        });
+        languageTag.textContent = languageNames.of(locale);
+        descriptionCell.appendChild(document.createTextNode(" "));
+        descriptionCell.appendChild(languageTag);
```
jleedev, line 636:
```suggestion
        descriptionCell.append(" ", languageTag);
```

ZeLonewolf:
I concur with leaving #781 issues as a separate work item, and this will help highlight the bug when testing a fix for it.

src/js/legend_control.js
```diff
@@ -107,6 +115,9 @@ export default class LegendControl {
     for (let data of this.sections) {
       let section = this.getSection(data);
       if (!section) continue;
+      if (data.id) {
+        section.id = `legend-section-${data.id}`;
```
jleedev, line 119: This whole flow is a little hard to read: The `sections` property is assigned by the caller, one of its items has an `id` of `"shields"`, and then 569 lines later we query the document for this computed value.

Anyway, LGTM. Something to think about for a future refactor.

jleedev (approved):

src/js/legend_control.js
1ec5: The shield section is different from the rest of the legend in numerous ways, to the point that it requires a completely separate code path to generate it in the first place. I suppose it would be worth trying to consolidate the code paths somehow.

One of the differences is that the descriptions get replaced by Wikidata labels asynchronously. Originally, I intended to implement Wikidata labeling functionality for the whole legend, but I reconsidered in favor of eventually maintaining localizations of the static legend entries in this repository: #744. That work will most likely require us to assign an ID to each of the sections and each of their entries anyways.

Previously, all the description rows’ DOM elements, in every section, were obtained from a `querySelectorAll()` call in `open()` then plumbed through to `prettifyNetworkLabels()`. But there’s nothing particularly special about this selector query. Making the element unique in the DOM allows us to query for it at any time. Instead of consolidating the shield section’s code path with the other section generation code, we could split it out into a separate class in a separate file, making it clearer that the shield section is more involved and independent from the rest of the legend.

1ec5: #782 tracks refactoring `LegendControl` to be more digestible.

PR #774, opened 2023-02-08
Simulate `line-gap-width` in legend

### LINKED ISSUE #770: Lines with gaps appear filled in legend
A road under construction is depicted as a double line at high zoom levels, but in the legend, the double line appears filled in:

[A-35](https://zelonewolf.github.io/openstreetmap-americana/#map=22/45.09307357/-73.08376347&language=fr)

This is because the road construction line layer specifies the `line-gap-width` paint property:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/820a1fe4c206c60cb32a12ea9ff83cef4be94e49/src/layer/construction.js#L48-L56

The legend currently uses this property to calculate the overall width of the line feature for sizing the generated SVG image, but it doesn’t actually separate the line into two lines:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/b21979c3cb15066d362936c3abe53425e93b51d3/src/js/legend_control.js#L381-L389

1ec5:
Line layers that specify the `line-gap-width` paint property now appear as a double line in the legend. The SVG line is now clipped to a `[`-shaped polygon.

[A-35](https://zelonewolf.github.io/openstreetmap-americana/#map=22/45.09306201/-73.08378183&language=fr)

Fixes #770.

### DIFF
--- a/src/js/legend_control.js
+++ b/src/js/legend_control.js
@@ -402,7 +402,7 @@ export default class LegendControl {
     let height = Math.max(...lineWidths);
 
     let svg = cell.querySelector("svg");
-    svg.style.height = `${Math.ceil(height)}px`;
+    svg.style.height = `${height}px`;
 
     for (let feature of lineFeatures) {
       let line = document.createElementNS("http://www.w3.org/2000/svg", "line");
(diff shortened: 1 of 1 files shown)


ZeLonewolf:
Looks like there's something not quite right in the double-line styling a couple zooms out:

[image](http://localhost:1776/#map=18/45.093792/-73.083212&language=fr)

1ec5:
Fixed:

A-35

ZeLonewolf:
I'm seeing the following console error if I press the legend button too quickly (before everything loads):
```
legend_control.js:706 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'querySelector')
    at Xl.completeNetworkLabels (legend_control.js:706:67)
```

This can be pulled out as tail work if it's unrelated to this PR.

ZeLonewolf (approved):
Otherwise, it looks good!

1ec5:
> I'm seeing the following console error if I press the legend button too quickly (before everything loads):

Fixed in #784.

Issue #27, opened 2021-07-10
Add layering for stacked tunnels


ZeLonewolf:
This image from lower Manhattan shows that stacked tunnels are a problem, in addition to stacked bridges:
![image](https://user-images.githubusercontent.com/3254090/125148281-1a73a180-e0ff-11eb-965f-7e882466b24e.png)

ZeLonewolf:
The NYC example is no longer a problem, but I'm not sure if that's because it's a toll vs free overlap.  Is this still a problem anywhere?
[![image](https://user-images.githubusercontent.com/3254090/215367148-2864306b-e8af-46d9-885c-29cb55e4b1a4.png)](https://zelonewolf.github.io/openstreetmap-americana/#map=15/40.70287/-74.01093)

ZeLonewolf:
Blocked by maplibre/maplibre-gl-js#2108

ZeLonewolf:
This example

ZeLonewolf:
This example in Sydney, Australia has values of `layer` going down to at least `layer=-5`:

[![image](https://user-images.githubusercontent.com/3254090/215377701-cd899789-09fe-4da6-be34-ecf89dde84de.png)](https://zelonewolf.github.io/openstreetmap-americana/#map=18/-33.87439/151.218122)

PR #778, opened 2023-02-09
Add layer sorting for roads within a common layer

ZeLonewolf:
This is a partial fix for #27.

This PR sorts features in the vertical direction by `layer`, _when they are in the same style layer_. This resolves a handful of layer corner cases not currently supported, however this is not a general solution for layered roads, which needs a solution to maplibre/maplibre-gl-js#2108. However, all `motorway` are now in the same layer, so in the case of overlapping motorway features, they will now sort correctly. This includes overlapping bridge/tunnel motorway and motorway_link features to any layer level. However, this does not address overlaps between motorways and other highway classes such as trunk/primary/etc.

In the screenshots below, in Boston, there is a `layer=-2` tunnel _underneath_ a `layer=-1` tunnel.

Before:
[image](https://zelonewolf.github.io/openstreetmap-americana/#map=18/42.364902/-71.05877)

After:
[image](http://localhost:1776/#map=18/42.364902/-71.05877)

Because `line-sort-key` takes a number, I changed the existing class deconfliction selector to use fractional values and add in the `layer` value, which is always a whole number.

This takes the total style size from `944,239` to `954,931`.

I suspect that some of the code in `layerSortKey` is actually unnecessary since we use separate layers, however I'd like to leave that investigation for a separate PR.

### DIFF
--- a/src/layer/road.js
+++ b/src/layer/road.js
@@ -32,6 +32,7 @@ const tunDashArray = [
 const getBrunnel = ["get", "brunnel"];
 const getClass = ["get", "class"];
 const getExpressway = ["coalesce", ["get", "expressway"], 0];
+const getLayer = ["coalesce", ["get", "layer"], 0];
 const getRamp = ["coalesce", ["get", "ramp"], 0];
 const getToll = ["coalesce", ["get", "toll"], 0];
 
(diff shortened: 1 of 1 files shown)


claysmalley (approved):
Looks good!

Before/after:

[Screenshot from 2023-02-11 18-32-35 Screenshot from 2023-02-11 18-32-52](http://localhost:1776/#map=13.95/40.70121/-74.01544)

PR #783, opened 2023-02-09
Specify shield width in custom functions

### LINKED ISSUE #781: Extra whitespace in Allegheny County belt, Branson color route shields
The Allegheny County belt system and Branson color route shields contain extra whitespace to the right of the shield:

[shieldtest](https://zelonewolf.github.io/openstreetmap-americana/shieldtest.html)

This is pushing the shield away from the road centerline and adding unsightly padding to the legend, as seen in https://github.com/ZeLonewolf/openstreetmap-americana/pull/773#issuecomment-1423577937.

ZeLonewolf:
Fixes #781

This PR allows the author of a custom shield drawing function (PA "dot" shields, Branson shields) to specify the width of the drawing canvas that the shield will be drawn onto.

[image
](http://localhost:1776/#map=13/40.45428/-79.97001&language=en)


src/js/custom_shields.js
```diff
@@ -68,6 +68,6 @@ export function bransonRoute(ctx, params) {
 }
 
 export function loadCustomShields() {
-  ShieldDraw.registerDrawFunction("branson", bransonRoute);
-  ShieldDraw.registerDrawFunction("paBelt", paBelt);
+  ShieldDraw.registerDrawFunction("branson", bransonRoute, 20 * ShieldDraw.PXR);
+  ShieldDraw.registerDrawFunction("paBelt", paBelt, 20 * ShieldDraw.PXR);
```
1ec5, line 72: `ShieldDraw.computeWidth()` should be responsible for multiplying by the pixel density, like it does for the other shapes.

ZeLonewolf: Fixed in de26f6a

src/js/shield_canvas_draw.js
```diff
@@ -693,14 +697,21 @@ export function draw(name, ctx, options, ref) {
 //Register draw functions
 const drawFunctions = {};
 
+//Register fixed width values
+const fixedWidthDefinitions = {};
```
1ec5, line 701: Ideally, these parallel structures would be folded into a single constant structured along these lines:

```js
{
  branson: {
    draw: function () { … },
    width: 20,
  },
};
```

The current code is fine, but consider that refactor down the line if we end up attaching more metadata to these custom drawing functions.

1ec5 (approved):

PR #784, opened 2023-02-10
Fix error when opening legend

1ec5:
This PR fixes a regression from #773 that produces an error when opening the legend: https://github.com/ZeLonewolf/openstreetmap-americana/pull/774#issuecomment-1424847267. If there’s no “Route markers” section because no route markers are visible in the current viewport, then there are no pending rows and no need to regenerate them with Wikidata labels.

### DIFF
--- a/src/js/legend_control.js
+++ b/src/js/legend_control.js
@@ -704,8 +704,9 @@ export default class LegendControl {
     if (!networkMetadata) return;
 
     let section = document.getElementById("legend-section-shields");
-    let rowContainer = section.querySelector(".legend-row-container");
-    let pendingRows = rowContainer.querySelectorAll('[data-pending="true"]');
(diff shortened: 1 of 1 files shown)


src/js/legend_control.js
```diff
@@ -686,8 +686,9 @@ export default class LegendControl {
     if (!networkMetadata) return;
 
     let section = document.getElementById("legend-section-shields");
```
1ec5, line 688: We could pull this statement up to the top of the method and short-circuit fetching the metadata from Wikidata. But even if that query runs unnecessarily the first time the user opens the legend, it avoids a flash of raw tag values when opening the legend again in the presence of shields. I think this is a better user experience worth the cost of a relatively small request.

ZeLonewolf (approved):
Looks good to me

PR #785, opened 2023-02-10
Bump maplibre to 2.4.0

ZeLonewolf:
This PR upgrades maplibre to 2.4.0, the latest stable version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "chai": "^4.3.7",
     "esbuild": "^0.17.4",
     "events": "^3.3.0",
-    "maplibre-gl": "^2.1.9",
+    "maplibre-gl": "^2.4.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -41,7 +41,7 @@
     "chai": "^4.3.7",
     "esbuild": "^0.17.4",
     "events": "^3.3.0",
-    "maplibre-gl": "^2.1.9",
+    "maplibre-gl": "^2.4.0",
```
1ec5, line 44: This is fine, though it should be unnecessary as long as the package isn’t breaking backwards compatibility within the v2._x_ series. The `^` means NPM would resolve to the latest available version before v3.0.0.

1ec5 (approved):

PR #786, opened 2023-02-10
Use map.style.getImage() instead of internal maplibre variables

ZeLonewolf:
Based on initial discussions in maplibre/maplibre-gl-js#2159, this PR changes the sprite sheet access from the internal variable map.style.imageManager.images[id] to map.style.getImage(id). There is no map.getImage(id) function at the top level, so this is the most top-level function we can call to extract images.

### DIFF
--- a/src/js/shield.js
+++ b/src/js/shield.js
@@ -377,15 +377,12 @@ export function romanizeRef(ref) {
   return roman + ref.slice(number.toString().length);
 }
 
-export function getDrawnShieldBounds(id) {
-  let routeDef = getRouteDef(id);
-  let shieldDef = getShieldDef(routeDef);
-
+export function getDrawnShieldBounds(shieldDef, ref) {
   let width = Math.max(
     ShieldDraw.CS,
     ShieldDraw.computeWidth(
       shieldDef.canvasDrawnBlank.params,
-      routeDef.ref,
+      ref,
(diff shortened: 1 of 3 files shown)


ZeLonewolf:
This PR is blocked by maplibre/maplibre-gl-js#2164. It cannot be merged until we can utilize this fix.

ZeLonewolf:
Ready for review.  This change will allow us to be compatible with maplibre 3.0.0.  After that cutover, we'll want to change the access method to `Map.getImage()` as implemented in maplibre/maplibre-gl-js#786

src/js/shield.js
```diff
@@ -377,11 +377,23 @@ export function romanizeRef(ref) {
   return roman + ref.slice(number.toString().length);
 }
 
-export function generateShieldCtx(map, id) {
-  return generateSpriteCtx(map.style.imageManager.images, id);
+export function getDrawnShieldBounds(id) {
+  let routeDef = getRouteDef(id);
+  let shieldDef = getShieldDef(routeDef);
```
1ec5, line 382: Both of this function’s call sites already have `shieldDef` and `routeDef` in context that they can pass in. That would make it a little easier to unit test this function.

ZeLonewolf: Good call, fixed in bd85e51

1ec5 (approved):

PR #787, opened 2023-02-12
Remove unneeded sort attributes

ZeLonewolf:
This PR simplifies the sort key used in various road classes.  The sort key is used to determine what in order to draw features _within a single style layer_. In general, the only key that makes sense to sort is `layer`, as this has a clear definition of ordering in OSM. Sorting comes into play at certain zooms where the thickness of road lines may cause overlaps.

For motorways, the only attributes that make sense for sorting are ramps, and toll roads, with toll roads on top, then non-toll roads, and then ramps.

For lower classifications, the only attribute that we need to consider sorting on is to have expressways appear above other classes of road.

It's pointless to sort by highway classification because these are already filtered into separate layers.

This change saves 23Kb of style JSON.

This change should result in no difference in rendering. Tested in Boston, NYC, and the Sam Houston Tollway.

Before:
```
Total layer size 954,931 bytes
road(201)........................753,810 bytes
```
After:
```
Total layer size 931,071 bytes
road(201)........................729,950 bytes
```

### DIFF
--- a/src/layer/road.js
+++ b/src/layer/road.js
@@ -92,32 +92,21 @@ const opacity = [
   1,
 ];
 
-const layerSortKey = [
+const motorwaySortKey = [
   "+",
   getLayer,
-  ["*", -0.28, getRamp],
-  [
-    "*",
-    0.04,
-    [
-      ...classSelector,
-      "motorway",
-      0.06,
-      "trunk",
-      0.05,
-      "primary",
-      0.04,
-      "secondary",
-      0.03,
-      "tertiary",
-      0.02,
-      "minor",
-      0.01,
-      0,
-    ],
-  ],
-  ["*", 2, getExpressway],
-  getToll,
+  0.1,
+  ["*", -0.1, getRamp],
+  ["*", 0.2, getToll],
+];
+
+const expresswaySortKey = [
+  "+",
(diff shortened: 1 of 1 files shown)


claysmalley:
> It's pointless to sort by highway classification because these are already filtered into separate layers.

Don't `roadSimpleFill` and `roadSimpleCasing` render multiple classes of road in one layer?

ZeLonewolf:
Yes, those are combined, however they're still separated by `layer`. So all the road casings within a layer, and all the road fills will render together, which is the correct behavior - they're black and white and render with intersections.

claysmalley (approved):

1ec5:
> It's pointless to sort by highway classification because these are already filtered into separate layers.

I verified that we aren’t regressing #524 either.

ZeLonewolf:
> > It's pointless to sort by highway classification because these are already filtered into separate layers.
> 
> I verified that we aren’t regressing #524 either.

That's handled by this separate sort attribute block in the highway shield layer:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/435091bc31c9784f1f9db09179837caab0c870df/src/layer/highway_shield.js#L60-L74

PR #798, opened 2023-02-18
Add Great River Road shield

### LINKED ISSUE #769: Great River Road shield
The [Great River Road](https://en.wikipedia.org/wiki/Great_River_Road) straddles the Mississippi River through ten states (and formerly two Canadian provinces). It has a distinctive shield featuring a steamboat inside a pilotwheel:

[Shield](https://commons.wikimedia.org/wiki/File:GreatRiverRoad.svg)

[Wisconsin](https://commons.wikimedia.org/wiki/File:GreatRiverRoadWI_Signs.jpg)

The route is mapped as a series of route relations. The [route superrelation](https://www.openstreetmap.org/relation/456651) and the relations for the national route through [Iowa](https://www.openstreetmap.org/relation/1784814) and [Wisconsin](https://www.openstreetmap.org/relation/1783095) are tagged with `network=US:AB:NSB`, indicating that they’re part of the America’s Byways system. None of the other relations are tagged with a network.

I think these relations should be part of a network specifically about the well-signposted Great River Road, rather than America’s Byways, which is secondary. Something like `US:GRR` or `US:MRPC` would be appropriate.

The official transportation map of each state other than Iowa simplifies the shield to just the pilotwheel:

Iowa DOT 2010
Illinois DOT 2005–06
MoDOT 2009
Mississippi DOT 2008
LaDOTD 2002

quincylvania:
Closes #769. I tried this as a wheel with a white halo. I think it looks decent on thin roads but I'm not sure about thick roads. It might look better circumscribed in a square, as it appears on signage. Feedback is appreciated.

Screenshot 2023-02-17 at 8 41 49 PM
Screenshot 2023-02-17 at 8 45 04 PM
Screenshot 2023-02-17 at 8 44 29 PM


ZeLonewolf:
Looks a little silly on the map, looks a little silly in real life.  I say it works.

Hopefully some day these will be clickable.

1ec5:
Would it look more recognizable as a ship’s wheel with a different number of spokes? Some of the official state transportation maps use only eight, even if the real shield depicts twelve. Removing the white ring separating the spokes from the wheel could also help. Finally, some of the examples in https://github.com/ZeLonewolf/openstreetmap-americana/issues/769#issuecomment-1435486459 cut a hole in the middle of the wheel; not sure if that would help or not.

quincylvania:
I tweaked the icon a bit and I'm happy with it now. We are running into the issue where circles look smaller than squares of the same width, which is even more pronounced due to the spokes on the wheel. Maybe we should allow shields to overflow the 20x20px box for cases like this.

2x:

Screenshot 2023-02-18 at 10 25 27 AM

1x:

Screenshot 2023-02-18 at 10 25 10 AM

1ec5 (approved):
This is looking great at 2×. At 1× it still kind of looks like a cogwheel to me, but I guess that can’t be helped at this size. For what it’s worth, we do already make exceptions to the basic icon size, for example for diamonds (North Carolina, Michigan) or especially tall shields (India).

PR #799, opened 2023-02-18
Add Merritt Parkway shield

### LINKED ISSUE #758: Merritt Parkway shield
Driving the Merritt today, I noticed these wonderfully ogre-shaped shields. This is the only Connecticut parkway with its own shield from what I can tell. I just added the route to OSM, so it'll have to wait for the next data refresh.

https://www.openstreetmap.org/relation/15425309
https://en.wikipedia.org/wiki/Merritt_Parkway
https://commons.wikimedia.org/wiki/File:Merritt_Pkwy_Shield.svg

Merritt_Pkwy_Shield svg
Screenshot 2023-02-02 at 6 17 06 PM
920x920

Old design: 
CT19560154i1-2837067163

quincylvania:
Closes #758.

This one required some code rejiggering because there are multiple routes with `network=US:CT:Parkway` that aren't distinguished by `ref`. While this could be changed in OSM, I think it makes sense for all the state's parkways to have the same `network` tag, and there aren't signed abbreviations that would be commonly-accepted under `ref`. The situation is different from Kentucky's and New York's parkways since we're trying to show a specific sprite, not add text.

The code changes essentially allow `refsByWayName` to be used alongside `overrideByRef`, letting overrides key off of a way's name. I'm open to other ways of achieving the result.

Another special case in this PR is the leaf color on the shield. Americana hasn't established a standard shade of light green, and dark green is too low contrast against the blue background.

1x:

Screenshot 2023-02-18 at 11 45 47 AM

2x:

Screenshot 2023-02-18 at 11 59 50 AM
Screenshot 2023-02-18 at 11 59 31 AM

The Merritt Parkway and the Wilbur Cross Parkway meet at the Housatonic River. Only the Merritt is signed with a shield. Both are State Route 15.

Screenshot 2023-02-18 at 11 56 18 AM


ZeLonewolf:
I've driven the Merritt many times on my way to NYC, and I was surprised to learn that there are other parkways in Connecticut. Before we push ahead on special casing, let's make sure the tagging is consistent with that of other states with named parkways.

1ec5:
> Another special case in this PR is the leaf color on the shield. Americana hasn't established a standard shade of light green, and dark green is too low contrast against the blue background.

If you’re interested in consistency with MUTCD signs, Caltrans has standardized on teal, green (but not the usual guide sign green), and aqua on its Adopt-a-Highway signs:

[S32-2](https://commons.wikimedia.org/wiki/File:MUTCD-CA_S32-2.svg) [S32-3](https://commons.wikimedia.org/wiki/File:MUTCD-CA_S32-3.svg) [S32-4](https://commons.wikimedia.org/wiki/File:MUTCD-CA_S32-4.svg)

src/js/shield_defs.js
```diff
@@ -4126,6 +4130,16 @@ export function loadShields() {
...
+  shields["US:CT:Parkway"].refsByWayName = {
+    "Merritt Parkway": "MP",
+  };
+
+  shields["US:CT:Parkway"].overrideByRef = {
+    MP: {
+      spriteBlank: "shield_us_ct_parkway_merritt",
+    },
+  };
+
```
ZeLonewolf, line 4142: Syntax-wise, this seems a little kludgy. Is there any way we can say something more simple like:

```json
"Merritt Parkway": "shield_us_ct_parkway_merritt"
```

quincylvania: I added an `overrideByWayName` property to enable this. I left the code that allows `refsByWayName` and `overrideByRef` to work together since there are still cases that might be useful.

ZeLonewolf (approved):
Code changes look good, and I concur with the approach to do a direct name mapping.  That will make our eventual shield generation library (#699) more useful as a general-purpose utility.  Connecticut's tagging feels a bit sketchy but I think this is a reasonable solution that expands our ability to handle special cases.

Net increase in style size is 101 bytes.

image

I would like one of the more artistically-minded maintainers to weigh in on the graphics but I think this is otherwise good to go.

quincylvania:
For the design, I simplified the ogre shield so it looks pronounced at 1x. The flower/leaf component I enlarged and vertically-aligned, but did not simplify. The graphic is a complex natural motif but sticks to three color fills, so I think simple downscaling gives the impression of the image better than an abstraction would.

jleedev (approved):

Issue #613, opened 2022-12-10
Render CJK glyphs locally


1ec5:
Viewing the map in Chinese, Japanese, or Korean requires an inordinate number of requests for fontstack glyph PBFs because Chinese, Japanese, and Korean characters are spread out over so many 255-codepoint ranges. The map takes longer than usual to load, consumes more memory, and can feel more sluggish. This also affects viewing East Asia in any language, due to the city name glosses added in #592.

GL JS has a [`localIdeographFontFamily`](https://maplibre.org/maplibre-gl-js-docs/example/local-ideographs) option for rendering CJK glyphs client-side on demand, in a manner not unlike how we’re rendering route shields. All we need to do is specify some font names that are likely to be on the system.

We have the opportunity to prioritize, say, a Japanese font when the user prefers Japanese labels. This is important because, in many cases, Unicode has conflated multiple mutually unintelligible characters from each CJKV country into the same Unicode codepoint. Because we try to show labels in the user’s preferred language, rather than the local language, we know which language to target worldwide, allowing us to overcome the challenges in gravitystorm/openstreetmap-carto#2208 https://github.com/ZeLonewolf/openstreetmap-americana/issues/428#issuecomment-1159544924 in terms of font selection.

/ref OpenHistoricalMap/issues#258

1ec5:
`localIdeographFontFamily` is actually enabled by default with `sans-serif` as the font, so this issue would only track tailoring the font fallback list to the user’s preferred language. The browser’s `sans-serif` font might well resolve to a font designed for the user’s preferred language, so the only case where we might have a problem is if the user manually overrides the preference with a `language=` for a different CJK langauge. Quite the edge case, so I’m closing this issue as works-for-me unless we find that it’s actually a big deal.

1ec5:
> The browser’s `sans-serif` font might well resolve to a font designed for the user’s preferred language

Confirmed in Firefox 108.0.1 on macOS by setting `intl.accept_languages` in about:config and observing the [Yellow Sea](https://www.openstreetmap.org/node/305639182) label. The name is 黄海 in both Chinese and Japanese, but 海 is one of the unified characters that has a different form in Japanese than in Traditional or Simplified Chinese or Korean:

Chinese (`zh`) | Japanese (`ja`)
----|----
[...]

PR #801, opened 2023-02-19
Render our own font stack

ZeLonewolf:
This PR adds the ability to generate our own fonts and customize which glyphs are serviced by which fonts. Supports 173 languages.

Note that:
* CJK glyphs will fall back to system defaults
* Han Unification:
  * CJK Noto Fonts are bundled but not yet used.  Additional work is needed to test for area and from that apply the appropriate CJK font. However, the default seems attractive enough
  * Additional work is need to identify other areas that need geography-specific fonts, e.g., see this discussion about the correct font to use in Pakistan: gravitystorm/openstreetmap-carto#4644.
* Ligatures still don't seem to render how they should, but this is beyond the scope of this PR and may be an upstream issue.

[image
](http://localhost:1776/#map=11/17.9674/102.6655&language=mul)
[image](http://localhost:1776/#map=7/37.46/127.63&language=mul)


ACKNOWLEDGEMENTS.md
```diff
@@ -2,5 +2,5 @@
 
 - The stylized highway shields are inspired by "[Rebusurance](https://github.com/1ec5/rebusurance/)" by Minh Nguyễn.
 - Initial layer code is derived from Oliver Wipfli's [openmaptiles-starter](https://github.com/wipfli/openmaptiles-starter/) style.
-- The font stack was [packaged by the OpenHistoricalMap project](https://github.com/OpenHistoricalMap/map-styles/).
+- The font stack was [packaged from Google Fonts](https://fonts.google.com/).
```
1ec5, line 5: Link to the [Google Fonts attribution page](https://fonts.google.com/attribution) for good measure.

scripts/fonts.json
```diff
@@ -0,0 +1,402 @@
...
+    "Noto Sans Georgian": ["regular", "700"],
+    "Noto Sans Gurmukhi": ["regular", "700"],
+    "Noto Sans Hebrew": ["regular", "700"],
+    "Noto Sans Kannada": ["regular", "700"],
+    "Noto Sans Lao": ["regular", "700"],
+    "Noto Sans Malayalam": ["regular", "700"],
+    "Noto Sans Myanmar": ["regular", "700"],
+    "Noto Sans Tamil": ["regular", "700"],
+    "Noto Sans Telugu": ["regular", "700"],
+    "Noto Sans Thai": ["regular", "700"]
```
1ec5, line 20: Note that the SIL Open Font License allows the font’s designer to specify a “Reserved Font Name” that others can’t use as the font’s name when distributing the font or any subset of it. According to the [Google Fonts attribution page](https://fonts.google.com/attribution), some of the available fonts do specify a reserved font name, but none of them are listed here. I think we should be fine in any case, because all the stylesheet knows is that it’s called “Americana”.

src/layer/water.js
```diff
@@ -206,7 +206,7 @@ export const waterPointLabel = {
   filter: ["all", ["==", ["geometry-type"], "Point"]],
   layout: {
     "text-field": Label.localizedName,
-    "text-font": ["Open Sans Bold Italic"],
+    "text-font": ["Americana-Bold-Italic"],
```
1ec5, line 209: Eventually, if gl-native’s local glyph rendering functionality expands beyond CJK, the system may need to match the specified fontstack’s weight and style based on words like “Bold” and “Italic” in the name, but the hyphens would foil that. For now, it doesn’t matter though.

/ref https://github.com/mapbox/mapbox-gl-native/issues/7862#issuecomment-613760767 https://github.com/mapbox/mapbox-gl-native/pull/16253#issuecomment-618730298

scripts/fonts.json
```diff
@@ -0,0 +1,402 @@
...
+  },
+  "glyph-ranges": {
+    "arabic": [
+      [1536, 1791],
```
1ec5, line 24: Document whether these are open or closed ranges (inclusive or exclusive of the latter codepoint).

quincylvania:
Ligatures are affected by the [context.textRendering](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/textRendering) property on Chrome. We should probably set this regardless of whether it fixes the problem, as leaving it on `auto` could lead to inconsistent results across browsers.

CONTRIBUTING.md
```diff
@@ -89,6 +89,12 @@ and `package-lock.json` files. Run this command to install them in the project d
 
 `npm install`
 
+### Fonts
+
+In order to build the project's [fonts](scripts/FONTS.md), install the `google-font-installer` package into npm with:
+
+`npm -g google-font-installer`
```
claysmalley, line 96:
```suggestion
`npm install -g google-font-installer`
```

claysmalley:
Huge step in the right direction! Well done.

Screenshot from 2023-02-19 08-36-57

The name of Providence in Hindi is प्रोविडेंस. A couple issues here:

* Ligatures are not forming. प्र (U+092A, U+094D, U+0930) is showing up in its non-combined form, प्‌र.
* Combining characters are appearing in the wrong visual order. व (U+0935) plus the vowel "i" forms वि (U+0935, U+093F). The combining vowel U+093F is joining with the following character instead.

Similar issues appear in a variety of South and Southeast Asian scripts. Seems that complex text layout is not being handled.

ZeLonewolf:
Can we determine whether the noted Brahmic script ligature issues are newly introduced or an upstream issue that we need to pursue in other libraries?

Perhaps something along the lines of shieldtest could be generated for font scripts.

ZeLonewolf:
Looks like newly introduced:
Screenshot_20230219_121734_Chrome

ZeLonewolf:
Ligature issues seem to be related to https://github.com/google/fonts/issues/1335

1ec5:
> Can we determine whether the noted Brahmic script ligature issues are newly introduced or an upstream issue that we need to pursue in other libraries?

This is a longstanding limitation of GL JS. The homegrown text layout engine in GL JS doesn’t have any data or logic around text shaping: https://github.com/maplibre/maplibre-gl-native/discussions/778 mapbox/mapbox-gl-native#7774 mapbox/mapbox-gl-native#7528.

To the extent the OHM fontstack appears to support Indic text shaping, it’s because of a hack that relies on negative glyph metrics. This is commonplace for Khmer, but for Indic, I think it might explain why the glyphs look like 8-bit and have haphazard spacing.

ZeLonewolf:
Okay, so the expected script is

image

Therefore the old way is still wrong.

claysmalley:
Where did you find that? The Wikidata entry for Providence spells it a different way, प्रोविडेंस.

ZeLonewolf:
I took that screen grab off the Sanskrit Wikipedia page for Providence:
https://sa.wikipedia.org/wiki/%E0%A4%AA%E0%A5%8D%E0%A4%B0%E0%A4%BE%E0%A4%B5%E0%A4%BF%E0%A4%A1%E0%A5%87%E0%A4%A8%E0%A5%8D%E0%A4%B8%E0%A5%8D

ZeLonewolf:
I'll pull the string direct from the database so we're doing a proper comparison.

ZeLonewolf:
Database:
image

Noto Sans Devanagari:
image

Fontspace.com unicode inspector:
image

ramSeraph:
> Where did you find that? The Wikidata entry for Providence spells it a different way, प्रोविडेंस.

I think it depends on how they decided to transliterate the word.. both of them sound correct phonetically.  Then again, I am not a native Hindi speaker.

ZeLonewolf:
See: https://github.com/wipfli/maplibre-canvas-labels

CONTRIBUTING.md
```diff
@@ -89,6 +89,12 @@ and `package-lock.json` files. Run this command to install them in the project d
 
 `npm install`
 
+### Fonts
+
+In order to build the project's [fonts](scripts/FONTS.md), install the `google-font-installer` package into npm with:
+
+`npm install -g google-font-installer`
```
1ec5, line 96: Should this be a formal dependency?

ZeLonewolf: How do I specify a global dependency formally?

ZeLonewolf: Ah ok, I wasn't familiar with `npm exec`

package.json
```diff
@@ -30,18 +33,21 @@
...
     "openmapsamples": "github:adamfranco/OpenMapSamples",
     "openmapsamples-maplibre": "github:adamfranco/OpenMapSamples-MapLibre",
+    "sync-request": "^6.1.0",
     "tokenfield": "^1.5.2"
   },
   "devDependencies": {
     "@basemaps/sprites": "^6.32.1",
     "@maplibre/maplibre-gl-style-spec": "^17.0.1",
     "chai": "^4.3.7",
-    "color-name": "^2.0.0",
```
ZeLonewolf, line 42: Accidental mis-name of color-namer

ZeLonewolf:
I have fixed the font issues for Tifinagh:
image

However, I now see that this doesn't run on Mac as currently implemented.

ZeLonewolf:
PR updated to install build_pbf_fonts via cargo.

scripts/fonts.js
```diff
@@ -0,0 +1,143 @@
...
+
+  const ttfFile = `build/font/ttf/${stack}.ttf`;
+  fs.writeFileSync(ttfFile, outputBuffer);
+  console.log(`Built ${ttfFile}`);
+}
+
+//Binary download.  Temporary hack until https://github.com/stadiamaps/build_pbf_glyphs/issues/4 is resolved
+const pbfBuilderFilename = "~/.cargo/bin/build_pbf_glyphs";
+
+execSync(
```
jleedev, line 128: Not well documented, but the Promise version of `exec` allows removing all of this boilerplate.

https://nodejs.org/dist/latest-v19.x/docs/api/child_process.html#child_processexeccommand-options-callback

package.json
```diff
@@ -30,20 +35,22 @@
     "color-namer": "^1.4.0",
     "color-rgba": "^2.4.0",
     "events": "^3.3.0",
+    "fonteditor-core": "^2.1.11",
     "maplibre-gl": "^2.4.0",
     "openmapsamples": "github:adamfranco/OpenMapSamples",
     "openmapsamples-maplibre": "github:adamfranco/OpenMapSamples-MapLibre",
+    "sync-request": "^6.1.0",
```
jleedev, line 42: Node.js has `fetch` now, and for prior versions you can turn it on with a flag or use `undici` from npm.

scripts/fonts.js
```diff
@@ -0,0 +1,143 @@
...
+    }
+    const variants = fontSpec[fontFamily].join(",");
+    execSync(
+      //Requires google-font-installer
+      `npm exec -- gfi download "${fontFamily}" -v ${variants} -d ${destFolder}`,
+      (error, stdout, stderr) => {
+        if (error) {
+          console.log(`error: ${error.message}`);
+          console.log(
+            `Is google-font-installer present? Install with "npm install google-font-installer`
```
jleedev, line 26: This is sort of superfluous in that we should always assume that the dependencies were installed.

scripts/fonts.js
```diff
@@ -0,0 +1,143 @@
...
+}
+
+const downloadLockFolder = "download/font-lock";
+const downloadFolder = "download/font";
+const distFontFolder = "dist/fonts";
+const ttfFontFolder = "build/font/ttf";
+
+[downloadFolder, ttfFontFolder, distFontFolder, downloadLockFolder].forEach(
+  (folder) => {
+    fs.mkdirSync(folder, { recursive: true }, (e) => {});
```
jleedev, line 52: `mkdirSync` doesn't take a callback.

scripts/fonts.js
```diff
@@ -0,0 +1,143 @@
...
+import { execSync } from "child_process";
+import request from "sync-request";
+
+function createRangeArray(start, end) {
+  return Array.from({ length: end - start + 1 }, (_, i) => start + i);
+}
+
+function loadGoogleFonts(fontSpec, destFolder) {
+  for (const fontFamily in fontSpec) {
+    const downloadLockFile = `${downloadLockFolder}/${fontFamily}.lock`;
```
jleedev, line 13: This feels like working around a shortcoming in this library, that it knows how to compute the destination filename but it won't do so until you ask it to actually download the font. I guess it works, though.

jleedev:
Hooray, more unnecessary platform dependencies.

```
error: failed to run custom build command for `pbf_font_tools v2.3.0`           
                                                                                
Caused by:                                                                      
  process didn't exit successfully: `/tmp/cargo-install06GbW8/release/build/pbf_
font_tools-fadfaad88948b465/build-script-build` (exit status: 101)              
  --- stderr                                                                    
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error 
{ os: "linux", arch: "arm" }', /root/.cargo/registry/src/index.crates.io-1cd6603
0c949c28d/pbf_font_tools-2.3.0/build.rs:6:41                                    
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace 
warning: build failed, waiting for other jobs to finish...                      
error: failed to compile `build_pbf_glyphs v1.3.0`, intermediate artifacts can b
e found at `/tmp/cargo-install06GbW8`                                           
```

Since apparently allowing "apt-get install protobuf-compiler" would be asking too much.

ZeLonewolf:
> Hooray, more unnecessary platform dependencies.

The goal is convert .ttf fonts into .pbf and this was the first one I figured out how to use.  It does run in the Mac/Linux runners.  I could explore other options like tiny-sdf or node-fontnik.

scripts/fonts.js
ZeLonewolf: oh, I suppose I could do something like strip out the spaces and check for matching files now that I look at this

ZeLonewolf: Oh you mean like not using the Google font Downloader at all

package.json
```diff
@@ -12,15 +12,20 @@
     "code_format": "run-s code_format:prettier code_format:svgo",
     "code_format:prettier": "prettier --write --list-different .",
     "code_format:svgo": "svgo -q -f icons/",
-    "clean": "shx rm -rf dist",
+    "clean": "run-s clean-download clean-build",
+    "clean-download": "shx rm -rf download",
+    "clean-build": "shx rm -rf dist build",
+    "fonts:install_glyph_builder": "cargo install --git https://github.com/stadiamaps/build_pbf_glyphs build_pbf_glyphs",
```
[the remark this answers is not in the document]
ZeLonewolf: I was following the instructions at https://github.com/stadiamaps/build_pbf_glyphs
How should it be?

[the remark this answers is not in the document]
jleedev: `cargo install build_pbf_glyphs@1`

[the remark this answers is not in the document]
ZeLonewolf: Fixed in 36d63ce

scripts/fonts.js
ZeLonewolf: I re-wrote this in e1073d9.

Okay, actually in the interest of full disclosure, I didn't actually understand your comment so I fed it to ChatGPT and it rewrote it for me.  Hope we got it right...

package.json
ZeLonewolf: I removed sync-request in b8e6763.  It turns out it wasn't actually in use, so it must have crept in while I was evaluating alternatives.

scripts/fonts.js
ZeLonewolf: Removed in 7d5a2f6

scripts/fonts.js
ZeLonewolf: Okay, fixed in 854ec9e.

ZeLonewolf:
I've addressed the code review comments but unclear if there's an outstanding `cargo` issue that needs to be resolved.

scripts/fonts.js
jleedev: Lol, lmao.

I would do `const exec = promisify(child_process.exec)` rather than at every call site.

jleedev (dismissed):
Looks good. Issues for later:

- Perhaps make use of github action caching for the build tools that need to be compiled every time.
- Make use of the npm workspaces feature so that it could be possible to not build the font stack and just use the published version, somehow.

The pbf_font_tools build shouldn't be a blocker for anyone's day to day usage, someone else can chime in if it is, and the fix is trivial.

1ec5 (commented):
It’s hard to overstate how big of an improvement Noto represents over Unifont. I’m looking forward to being able to cobble together additional font styles across writing systems for different themes within the map style.

package.json
```diff
@@ -12,15 +12,20 @@
...
     "presprites": "shx rm -rf dist/sprites",
     "sprites": "node scripts/sprites.js",
     "status_map": "node scripts/status_map.js",
     "taginfo": "node scripts/taginfo.js",
-    "start": "run-s clean sprites serve",
+    "start": "run-s clean-build sprites fonts serve",
     "serve": "node scripts/serve.js",
     "build:code": "node scripts/build.js",
-    "build": "run-s clean sprites build:code taginfo status_map",
+    "build": "run-s clean sprites fonts build:code taginfo status_map",
```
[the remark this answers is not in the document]
jleedev: https://blog.rust-lang.org/2023/03/09/Rust-1.68.0.html

You're just in time to try this out:

```
CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
```

ramSeraph:
I have no horse in this race,  but since I am getting notifications( and can't apparently shut up :) ),  I am wondering, why not use https://github.com/maplibre/font-maker, I think that font-maker can be bundled as a pure npm package( something which hasn't been done yet though ) and run directly without introducing a rust dependency. Just a suggestion, you can ignore it. `font-maker` also doesn't depend on system installed freetype lib.

ramSeraph:
There is a command line tool( normal platform dependant executable ) which requires compilation with gcc/clang.. there is another option to build a wasm bundle from the code, which probably can be wrapped in some javascript and published to npm.. which hasn't been done yet( you can probably request @bdon for that ). I will check if it is something that can done without publishing to npm.

ZeLonewolf:
What is needed here is a cross-platform utility that can convert .ttf and .otf files to .pbf fontstacks, while font-maker appears to be web-based and platform-dependent. While I would love to avoid the expensive rust compilation step, it's not clear from the font-maker documentation how I would integrate it into the workflow I have here.

package.json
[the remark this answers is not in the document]
ZeLonewolf: On my system, a clean build does not seem to re-build the rust package?

ramSeraph:
I think it can be platform independent( all node/js engine supported platforms including browsers ), it is just that the packaging for the node part hasn't been written yet and it is only currently showcasing the browser compatibility part. Let me see if I can hack together some POC to prove that. 

But, don't wait on me, if I manage to do it, you can probably use it later.

jleedev:
It should be "pretty easy" to use node-canvas and tiny-sdf to render these.

package.json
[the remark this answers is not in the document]
1ec5: I don’t know if it actually rebuilds anything, but it does spend several minutes fetching the crates.io index. That registry configuration option sounds promising.

ZeLonewolf:
Sample code below courtesy ChatGPT.

Is this what you had in mind?

```javascript
const canvas = require('canvas');
const tinySDF = require('tiny-sdf');
const fontnik = require('fontnik');
const fs = require('fs');

// Load the TTF font file using node-canvas
const font = new canvas.Font('MyFont', './MyFont.ttf');

// Set up the tiny-sdf options
const options = {
  size: 24, // size of each glyph
  buffer: 3, // amount of padding around each glyph
  radius: 8, // radius of the SDF
  cutoff: 0.25, // cutoff value for the SDF
  fontFamily: 'MyFont', // name of the font family
};

// Generate signed distance field glyphs for the font using tiny-sdf
const glyphs = tinySDF(options);

// Create a new canvas using node-canvas
const canvas = canvas.createCanvas(options.size, options.size);

// Loop through each glyph and draw it onto the canvas
for (const glyph of glyphs) {
  // Clear the canvas
  canvas.clearRect(0, 0, options.size, options.size);

  // Set the font and fill style
  canvas.font = `${options.size}px ${options.fontFamily}`;
  canvas.fillStyle = '#fff';

  // Draw the glyph onto the canvas
  canvas.fillText(glyph.char, glyph.x, glyph.y);

  // Export the canvas as a PNG image
  const buffer = canvas.toBuffer();

  // Convert the PNG image to a .pbf fontstack using fontnik
  const fontstack = fontnik.fromPng(buffer);

  // Save the .pbf fontstack file to disk
  fs.writeFileSync(`${glyph.char}.pbf`, fontstack);
}
```

ramSeraph:
> Sample code below courtesy ChatGPT.
> 
> Is this what you had in mind?
> 
> ```js
> const canvas = require('canvas');
> const tinySDF = require('tiny-sdf');
> const fontnik = require('fontnik');
> const fs = require('fs');
> 
> // Load the TTF font file using node-canvas
> const font = new canvas.Font('MyFont', './MyFont.ttf');
> 
> // Set up the tiny-sdf options
> const options = {
>   size: 24, // size of each glyph
>   buffer: 3, // amount of padding around each glyph
>   radius: 8, // radius of the SDF
>   cutoff: 0.25, // cutoff value for the SDF
>   fontFamily: 'MyFont', // name of the font family
> };
> 
> // Generate signed distance field glyphs for the font using tiny-sdf
> const glyphs = tinySDF(options);
> 
> // Create a new canvas using node-canvas
> const canvas = canvas.createCanvas(options.size, options.size);
> 
> // Loop through each glyph and draw it onto the canvas
> for (const glyph of glyphs) {
>   // Clear the canvas
>   canvas.clearRect(0, 0, options.size, options.size);
> 
>   // Set the font and fill style
>   canvas.font = `${options.size}px ${options.fontFamily}`;
>   canvas.fillStyle = '#fff';
> 
>   // Draw the glyph onto the canvas
>   canvas.fillText(glyph.char, glyph.x, glyph.y);
> 
>   // Export the canvas as a PNG image
>   const buffer = canvas.toBuffer();
> 
>   // Convert the PNG image to a .pbf fontstack using fontnik
>   const fontstack = fontnik.fromPng(buffer);
> 
>   // Save the .pbf fontstack file to disk
>   fs.writeFileSync(`${glyph.char}.pbf`, fontstack);
> }
> ```

I think you just sold me on trying chatGPT... that almost looks correct. "canvas.createCanvas" was the major glaring error I can see.

ramSeraph:
A question for @1ec5 , would sticking to tinysdf( or equivalent ) on both server and client would lead to better mixing of CJK and non CJK fonts? Which approach has more information needed for label placement `tinysdf` or `node-fontnik`?

1ec5:
> would sticking to tinysdf( or equivalent ) on both server and client lead to better mixing of CJK and non CJK fonts? Which approach has more/correct information needed for label placement `tinysdf` or `node-fontnik`?

In principle, both are capable of font metrics for non-complex text. TinySDF can even handle variable-width (non-CJK) glyphs as of mapbox/tiny-sdf#24, but there are still some limitations, for example mapbox/tiny-sdf#34. Mixing CJK and non-CJK is mostly a nonissue, apart from getting the baselines to match. But I’m not sure the status quo does that well, and anyways a good style would increase the text size of any CJK for legibility.

ZeLonewolf:
> I think you just sold me on trying chatGPT... that almost looks correct.

Don't get excited, supervising an AI to write code to the point of it working is actually more work than you might imagine :laughing:

package.json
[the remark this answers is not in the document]
ZeLonewolf: I tried putting that environment variable in the `fonts:install_glyph_builder` task and it made no discernible difference on my system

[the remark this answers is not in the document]
1ec5: I added a configuration file opting into the sparse protocol (which will soon be the default anyways). It does seem to bring installation times back to normal.

scripts/fonts.json
```diff
@@ -0,0 +1,450 @@
...
+      [64285, 64331]
+    ],
+    "hiragana": [[12352, 12447]],
+    "kanbun": [[12688, 12703]],
+    "kannada": [[3200, 3327]],
+    "kanji": [[19968, 40959]],
+    "katakana": [
+      [12448, 12543],
+      [12784, 12799]
+    ],
```
1ec5, line 105: Can we add [Noto Sans Khmer](https://fonts.google.com/noto/specimen/Noto+Sans+Khmer) for Khmer, which is prominent in Southeast Asia?

[Phnom Peng](https://zelonewolf.github.io/openstreetmap-americana/#map=7/11.782/104.944&language=en)

Are there any other scripts of national languages that we should account for?

package.json
1ec5, line 28: As with the `start` script above, building the project shouldn’t toss the downloaded fonts. This change will eliminate the build overhead I was complaining about earlier:

```suggestion
    "build": "run-s clean-build sprites fonts build:code taginfo status_map",
```

scripts/fonts.json
```diff
@@ -0,0 +1,450 @@
...
+    "M PLUS Rounded 1c": ["regular", "700"],
+    "Noto Sans": ["regular", "700", "italic", "700italic"],
+    "Noto Sans Arabic": ["regular", "700"],
+    "Noto Sans Armenian": ["regular", "700"],
+    "Noto Sans Bengali": ["regular", "700"],
+    "Noto Sans Devanagari": ["regular", "700"],
+    "Noto Sans Ethiopic": ["regular", "600"],
+    "Noto Sans Georgian": ["regular", "700"],
+    "Noto Sans Gurmukhi": ["regular", "700"],
+    "Noto Sans Hebrew": ["regular", "700"],
```
1ec5, line 13: Consider [Noto Rashi Hebrew](https://fonts.google.com/noto/specimen/Noto+Rashi+Hebrew) as an analogue to Noto Sans Italic.

scripts/fonts.json
```diff
@@ -0,0 +1,450 @@
+{
+  "font-families": {
+    "Gothic A1": ["regular", "700"],
+    "M PLUS Rounded 1c": ["regular", "700"],
+    "Noto Sans": ["regular", "700", "italic", "700italic"],
+    "Noto Sans Arabic": ["regular", "700"],
```
1ec5, line 6: Consider [Noto Naskh Arabic](https://fonts.google.com/noto/specimen/Noto+Naskh+Arabic) as an analogue to Noto Sans Italic.

scripts/fonts.json
```diff
@@ -0,0 +1,450 @@
...
+    "Noto Sans Telugu": ["regular", "700"],
+    "Noto Sans Thai": ["regular", "700"]
+  },
+  "bundle-font-stacks": {
+    "Noto Sans HK": ["regular", "700"],
+    "Noto Sans JP": ["regular", "700"],
+    "Noto Sans KR": ["regular", "700"],
+    "Noto Sans SC": ["regular", "700"],
+    "Noto Sans TC": ["regular", "700"]
+  },
```
1ec5, line 30: I’m pretty sure the user will never see these fonts, because GL JS is configured to use the browser’s default “sans-serif” CJK font by default: #613.

ZeLonewolf:
With the change in d611a08, this should solve the cross-platform issues.

From a strategic perspective, pbf font stacks are a "temporary" technology until maplibre-gl can support a more robust mechanism that can properly support ligatures etc. This is a high priority for the maplibre project: is:open https://github.com/maplibre/maplibre/issues/193

Despte the bounty, it may be some time before a change of this complexity is implemented.  In the meantime, we can at least have our broken ligatured scripts at least have a nice font :grimacing: and work under the expectation that in the future we'll be able to directly link web fonts.

1ec5 (commented):
To be clear, my concern about build times was limited to the overhead of downloading the crates.io index (for a single Rust dependency) and downloading the individual Noto fonts from scratch on each build. It wasn’t specifically about the node-fontnik pipeline; whatever works works, at least until GL JS has a better story around fonts.

I think we’ve solved the crates.io overhead. As for the Noto overhead, it’s only when running the `build` script, but not when running the `start` script. My suggestion below addresses that issue too.

I added some suggestions of Noto fonts that are analogous to Noto Sans Italic that we could use as a quick fix. However, Google Fonts doesn’t provide analogues in each of the writing systems we’re trying to support. We’d either need to swap in a font from a different source or synthesize an oblique font.

scripts/fonts.json
ZeLonewolf: I included these in order to allow a developer working on Han Unification to have multiple CJK fonts already rendered to work with.  They are not bundled into any of the Americana-* fontstacks, they're each rendered out to their own fontstack.

scripts/fonts.json
ZeLonewolf: Khmer script added in 8459f0a:

image

scripts/fonts.json
1ec5: OK, these are by far the largest fonts that would be downloaded when installing the package, but I guess we can live with that as long as it only happens once.

scripts/fonts.json
ZeLonewolf: Noto Rashi Hebrew:

[image](http://localhost:1776/#map=14/31.59067/34.54142&language=he)

scripts/fonts.json
ZeLonewolf: I've added this font to the download, but whenever I try to use it, the script errors out with a duplicate code point error.  I've verified that I don't have any overlapping ranges so I'm not sure at this point how to move it forward.

ZeLonewolf: The code point range at issue is within the `[1536, 1791]` range.

ZeLonewolf: I can't even run the Naskh Arabic font through the pipeline by itself.  There seems to be something fundamentally wrong with this font that breaks this workflow.

scripts/fonts.json
claysmalley: Canadian Aboriginal Syllabics, Gujarati, Odia (Oriya), Punjabi, Sinhala, Tibetan

ZeLonewolf: Can you offer some examples of test locations where those scripts are tagged in the main `name` tag in order to test?

ZeLonewolf: [ ] Canadian Aboriginal Syllabics
[ ] Gujarati
[ ] Odia
[ ] Punjabi (should be covered by existing Gurmukhi script)
[ ] Sinhala
[ ] Tibetan

claysmalley: You're right about Punjabi; I overlooked Gurmukhi.

Gujarati and Odia are the predominant languages of the Indian states of Gujarat and Odisha, respectively. Being in India, `name=*` labels are in English, so these won't appear outside of `name:gu=*` and `name:or=*`, neither of which are present in OpenMapTiles.

I can remember Sinhala being used for `name=*` in the part of Sri Lanka where it's spoken (and Tamil in the Tamil-speaking area), but the local community seems to have shifted consensus to English recently.

* Canadian Aboriginal Syllabics: [Iqaluit, Nunavut :canada:](https://zelonewolf.github.io/openstreetmap-americana/#map=11.77/63.7458/-68.5242)
* Tibetan: [Thimphu :bhutan:](https://zelonewolf.github.io/openstreetmap-americana/#map=9.54/27.4657/89.6317/0/8)

ZeLonewolf: If they're not in the name tag and not an an OMT supported language, we'll never see it appear. Probably possible to do an overpass query to find examples. I can add them in for completeness of course but it'll be untested and this isn't the first time I've mangled unicode ranges.

scripts/fonts.json
1ec5: Hopefully we aren’t running into that classic Mapbox assumption that there’s a one-to-one correspondence between Unicode codepoints and glyph indices. We could compare this font against Noto Sans Arabic in a font editor and spot-check whether there are any discrepancies.

/ref https://github.com/maplibre/maplibre-gl-native/discussions/778#discussioncomment-5051041

ZeLonewolf: That does appear to be the case, though in this instance, the issue is manifesting in the `fonteditor-core` library.

scripts/fonts.json
1ec5: A solution regarding Han Unification should probably retain the current client-side font solution rather than switching to server-side fontstacks, which carry significant performance penalties. (Aside from wasteful bandwidth usage, it’s very easy to hog memory and overflow buffers just by looking at East Asia when using server-side CJK fontstacks.)

The client-side font infrastructure is already mostly ready for deunified CJK fonts, except that GL JS only maintains a single instance of TinySDF that it initializes as soon as it needs to render any glyphs, and TinySDF only applies fonts to the canvas once at initialization. I suspect that addressing either limitation would be trivial.

ZeLonewolf: If it's not useful, I can drop these fonts.

scripts/fonts.json
ZeLonewolf: Tibetan added in 549918b:

image

ZeLonewolf: Sinhala added in 6abaf19:

image

ZeLonewolf: I am currently rendering a new planet with more languages which should allow us to freely test some of these scripts (and also make the map more international)

ZeLonewolf: Tibetan on top with the new tileserver update:
image

ZeLonewolf: Adding Oryia script in 2facda0:
[image](http://localhost:1776/#map=6/24.605/82.398&language=or)

ZeLonewolf: Adding Gujarati in 131e4c5:
[image](http://localhost:1776/#map=5/40.86/-81.09&language=gu)

ZeLonewolf:
I've added the additional scripts that were requested in the code reviews (this necessitated a planet rebuild with new languages added). I believe this satisfies all issues with the exception of whether to also include Han Unification fonts.

scripts/fonts.json
claysmalley: Looks good. There are problems with ligatures and combining characters in all of the above screenshots, but that seems to be downstream of maplibre/maplibre-gl-js#50.

ZeLonewolf: Agreed - the solution that that will likely require a different font technology than PBF fontstacks.

scripts/FONTS.md
```diff
@@ -0,0 +1,8 @@
+The Americana fontstack is built from [Google Fonts](https://fonts.google.com/).
+
+In order to add fontstack support, modify the `fonts.json` file as follows:
+
+1. Add font-family and variant information to the `font-families` section. The font-family is the name of the font as listed on Google Fonts, e.g. "Noto Sans". Use `gfi download "<name of font>"` to get a full list of the variants. The variant is everything to the right of the dash in the filename, so if a file is named `NotoSans-700.ttf`, the variant is `700`, though it will be listed in Google as something like "Bold 700". The `gfi` command requires you to install the `google-font-installer` package into npm with `npm install -g google-font-installer`.
```
1ec5, line 5: Not sure if it should go here or in CONTRIBUTING.md, but we could document some editorial guidelines to aid in choosing fonts in non-Western writing systems. For example, once we get around to implementing custom CJK fonts, either on the server side via a fontstack or (preferably) on the client side via standard Web fonts, we’d want to pair Western regular sans-serif with [gothic](https://en.wikipedia.org/wiki/East_Asian_Gothic_typeface) and Western regular serif with [Ming](https://en.wikipedia.org/wiki/Ming_%28typefaces%29), and Western italic sans-serif with [regular](https://en.wikipedia.org/wiki/Regular_script#In_computing), based on what Western and CJK texts do in analogous situations.

That said, one of the outgrowths of this PR may end up being a separate project specifically about fonts, where we’d naturally explore and elaborate on these ideas.

src/layer/water.js
```diff
@@ -137,7 +137,7 @@ const labelPaintProperties = {
 const labelLayoutProperties = {
   "symbol-placement": "line",
   "text-field": Label.localizedNameInline,
-  "text-font": ["OpenHistorical Italic"],
+  "text-font": ["Americana-Italic"],
```
1ec5, line 140: There’s some interest in replacing this italic sans-serif font with an italic serif font for natural features like waterways and waterbodies. Let’s track that as tail work, since it may require some research into analogues in other writing systems.

1ec5 (approved):
Piggybacking on @jleedev’s original approval, I think we’ve since made good strides in the build system aspects of this PR, as well as the internationalization aspects. It’s good enough to ship, but let’s ticket out the various larger issues that were raised during review.

scripts/FONTS.md
claysmalley: Tracking Unicode ligatures and combining characters in #827

PR #804, opened 2023-02-23
Dual color replacement in sprites

quincylvania:
Currently, color replacement is limited to lightening black pixels, and white pixels are static. This PR adds a `colorDarken` property that works alongside `colorLighten` to unlock complete replacement of colors in any two-tone sprite. This minimizes redundant SVG files and thus makes shield management easier. We could also use it for runtime POI halo colors until we move that functionality to the style.


ZeLonewolf:
Cool idea! Is this idea extendable to the Pittsburgh colored routes?

[image](http://localhost:1776/#map=12/40.42157/-79.97202)

Or the Branson, Missouri colored routes?

[image](http://localhost:1776/#map=13/36.64282/-93.26872)

quincylvania:
@ZeLonewolf Those shields appear to be tricolor (shape+bg+outline), so they wouldn't work out of the box.

src/js/screen_gfx.js
```diff
@@ -38,58 +38,79 @@ export function getPixelRatio() {
...
+
+export function copyPixel(source, dest, sourceOffset, destOffset) {
+  //Red
+  dest[destOffset] = source[sourceOffset];
+  //Green
+  dest[destOffset + 1] = source[sourceOffset + 1];
+  //Blue
+  dest[destOffset + 2] = source[sourceOffset + 2];
+  //Alpha
+  dest[destOffset + 3] = source[sourceOffset + 3];
```
ZeLonewolf, line 59: Can we just use the spread operator here?

```javascript
let a = [1,2,3];
let b = [...a];
a[0] = 999;
console.log(a);
console.log(b);
```

ZeLonewolf: Ah ok maybe not the spread operator since it's from a specific position, but I assume there's a simpler way to array copy.

quincylvania: Probably. I'm inclined to prioritize readability over cleverness unless there's a performance gain.

src/js/poi.js
```diff
@@ -19,7 +19,7 @@ export function missingIconLoader(map, e) {
   var height = sourceSprite.data.height;
 
   var ctx = Gfx.getGfxContext({ width, height });
-  Gfx.copyImageData(ctx, sourceSprite, 0, false, color);
+  Gfx.copyImageData(ctx, sourceSprite, 0, false, color, null);
```
ZeLonewolf, line 22: Is this needed?

ZeLonewolf:
Might be an issue with `norefImage`:

[image](http://localhost:1776/#map=12/40.59353/-79.88428)

src/js/poi.js
quincylvania: Not really, just left it as a reminder to add POI halo coloring later. We can remove it though.

quincylvania:
> Might be an issue with norefImage

@ZeLonewolf Good catch. This seems to be the only conflicting case, so I just updated the SVG. We'll need to keep an eye out in the future though.

ZeLonewolf:
That was an interesting solution, and one I didn't really expect. PA turnpike is the only network with `noRefImage` + color changing (and there's a bunch of other networks with `noRefImage` that could also have been modeled as `spriteBlank` + `notext`). The problem with this approach is that we lose the ability that we previously had to specify an alternate no-ref shield that's outside of the color space specified in the base shield. Suppose tomorrow that Pennsylvania changes the graphic of their unnumbered PA turnpike to Ben Franklin's head -- before this change we could just change the graphic, and now we can't do that (unless we do something really weird like make a green-color-inverse of Ben's head!)

quincylvania:
@ZeLonewolf Yes, I understood the hypothetical issues here. What solution would be better?

This was an existing though unrealized problem with just `colorLighten`. It can get messy to have one shield definition that applies different shield sprites for different situations like this. If we ignore `colorLighten` and `colorDarken` for all `noRefImage` sprites, well then that might be unexpected in some cases. And we don't want to add `noRefColorLighten` and `noRefColorDarken`.

ZeLonewolf:
My thinking is something along these lines, but I'm not wedded to this logic flow if there's a better way to do it:
norefshield drawio

I also haven't looked at the impact this change has on our taginfo graphics, e.g:
https://taginfo.openstreetmap.org/tags/network=US%3APA%3ATurnpike#projects

I suspect the taginfo script needs to change to handle re-colored shields to retain this behavior:
image

quincylvania:
Okay, I resolved the `norefImage` issue by converting the property to an entire alternate shield definition called `noref`. This lets us define stuff like colors and `verticalReflect` specific to the alternate sprite. I think this is more intuitive and it actually makes the drawing code cleaner.

```
shields["US:PA:Turnpike"] = {
    spriteBlank: ["shield_us_pa_2", "shield_us_pa_3"],
    textColor: Color.shields.white,
    padding: {
      left: 3,
      right: 3,
      top: 5,
      bottom: 5,
    },
    colorLighten: Color.shields.white,
    colorDarken: Color.shields.green,
    noref: {
      spriteBlank: "shield_us_pa_turnpike_noref",
      colorLighten: Color.shields.white,
      colorDarken: Color.shields.green,
    },
  };
```

For taginfo sprites, I have it save preview SVGs with the colors swapped out. This might not work in all cases but it gets us most of the way there.

ZeLonewolf (approved):
This looks great.  Thanks!

Issue #694, opened 2023-01-15
Render universities/colleges distinctly from schools


wmisener:
Colleges (`amenity=college`) and universities (`amenity=university`) are tagged differently than schools, and they play different roles in the community fabric. They're often larger, sometimes with extensive campuses, and therefore serve both as major destinations themselves and as very important locations for city-wide orientation, sometimes defining the character of a city region or even a whole town.

As of https://github.com/ZeLonewolf/openstreetmap-americana/pull/689, colleges and universities are rendered the same as all other schools, including preschools, which may prove confusing. It would make sense to render them similarly to schools, but a distinct icon could also work. One idea could be to a "C" or "U" inside the school building, similarly to how some of the map samples in #78 differentiate elementary/middle/high schools. Other possibilities include a larger building with multiple flags (also shown in #78) or some sort of mortarboard cap icon (a common higher education signifier, e.g. in iD). Samples welcome!

I think it would also make sense for them to appear at much lower zoom than schools do (currently z15), since they are often major landmarks and thus deserve to be shown at a city-wide scale. Just from eyeballing, I think z11 or z12 would be appropriate. Due to their often-large extent, it might be appropriate to consider an area rendering as well.

adamfranco:
I like the idea of the mortarboard. It's a pretty distinctive shape and a "C" or "U" in the building box may be harder to  translate for those who read different scripts. While mortarboards come out of a particular cultural context, doing image searches shows them to be pretty broadly applied around the world:

- [China](https://www.google.com/search?q=china+university+graduation&client=firefox-b-1-d&sxsrf=AJOqlzXBBWyKC7Bsb6yHe6oicflk96CXNg:1673983932984&source=lnms&tbm=isch&sa=X&ved=2ahUKEwiPmr2NrM_8AhX6jYkEHcJ5DIwQ_AUoAXoECAEQAw&biw=1304&bih=790&dpr=1)
[...]

PR #807, opened 2023-02-26
Add town hall icon

### LINKED ISSUE #794: Render town halls
Town halls are important city landmarks, so rendering them would make progress on Americana's representation of community institutions (#435) and POIs generally (#692). A fairly common representation of town halls are as Neoclassical columned buildings, topped by a rectangular flag, which represents their governmental purpose and distinguishes them from museums, which can be represented similarly in some maps. Two OSM-adjacent examples of this genre of icon are Carto and Temaki:
Screen Shot 2023-02-16 at 9 21 42 PM Screen Shot 2023-02-16 at 9 21 19 PM

Many American road maps use simple squares to denote these and other important places of interest, but I think we can take a richer approach, as we already have for other POI icons.
Screen Shot 2023-02-16 at 9 21 19 PM

I personally actually quite like the temaki icon, so that or a similar design could work well.

wmisener:
Renders icons for `amenity=townhall`. Fixes #794, makes progress on #435 and #692.

I tried a couple of the ideas put forward in #794. First, one inspired by the temaki icon, but modified to be 20x20 px and with some other adjustments: poi_town_hall. I like this one, though maybe the waving flag is a bit large?

I also tried a rectangular flag (or maybe, a flag waving in a stiff wind), which I also like. poi_town_hall_straightflag

Finally, a "clocktower"/Back to the Future style building, as [suggested](https://github.com/ZeLonewolf/openstreetmap-americana/issues/794#issuecomment-1435073178) by @1ec5: poi_town_hall_clockbuilding. I think this could work, but it's not to my personal taste as much. I thought combining a clock and the columns would be too fussy, but would be willing to try it if someone thought it was worth a shot.

Samples, from Culver City, CA ([localhost link](http://localhost:1776/#map=16.57/34.021305/-118.395587)):
Screen Shot 2023-02-25 at 3 09 41 PM Screen Shot 2023-02-25 at 3 14 15 PM Screen Shot 2023-02-25 at 3 17 11 PM

Right now the icons appear at z15, the same as schools and hospitals, though these features are probably important enough to render at z<15. Legend entry is "Town hall".
Screen Shot 2023-02-25 at 3 46 50 PM

More samples:
Davenport, IA ([localhost link](http://localhost:1776/#map=16.31/41.52258/-90.579347)):
Screen Shot 2023-02-25 at 3 23 22 PM

Falmouth, MA ([localhost link](http://localhost:1776/#map=15.08/41.55155/-70.62052)):
Screen Shot 2023-02-25 at 3 49 40 PM

Atlanta, GA ([localhost link](http://localhost:1776/#map=15.5/33.747261/-84.395116)):
Screen Shot 2023-02-25 at 3 53 14 PM


1ec5:
Thanks for trying out multiple approaches. I think what I find suboptimal about the (otherwise beautiful) columnar icons is that they include more literal detail compared to the abstract and minimalistic school icon. It’s the difference between a typical modern digital map, which aims for consistency with a search UI or even non-map UIs, and the genre of community maps that has [its own visual language](https://www.e-education.psu.edu/geog486/node/594), one that schoolchildren continue to be taught. I think we’ll find that, over time, as we add more POI icons, the school icon will look increasingly out of place compared to something like Temaki’s paper and pencil, even though the symbol in #689 is a perfectly appropriate choice for this style.

[Symbol iconicity](https://www.e-education.psu.edu/geog486/node/594)

/ref https://github.com/ZeLonewolf/openstreetmap-americana/issues/794#issuecomment-1445238645

ZeLonewolf:
> Obviously, there are only so many basic shapes, not enough for the full range of POI types we’ll eventually want to annotate on the map. But the special subset of community anchor institutions that we show at lower zoom levels would look more important the more they hew to basic shapes.

Thanks for spelling this out in such clear terms. This idea of abstract vs iconic is probably something we should think more about and try to express in terms of design guidelines. Do we want to deliberately constrain the diversity of symbology for POIs?  Or do you view an American style map as more towards the abstract in design while keeping a full range of symbol diversity?

1ec5:
> Do we want to deliberately constrain the diversity of symbology for POIs? Or do you view an American style map as more towards the abstract in design while keeping a full range of symbol diversity?

The correct answer is probably somewhere in the middle. We want Americana to make OSM look like a serious map worthy of respect, but a little playfulness isn’t out of the question. Abstract symbols only work as long as they are few in number, but too many detailed icons can be difficult to differentiate too.

At the extreme, [squares for everything](https://github.com/ZeLonewolf/openstreetmap-americana/issues/794#issue-1588795871) scales well and would be familiar in a way, but it would be so uninteresting that people wouldn’t have the opportunity or motivation to add more kinds of POIs to the map. On the other hand, if you search Google Images for examples of “community map symbols”, you’ll find plenty of elementary school worksheets and lesson plans about map reading skills that include very detailed cartoon icons of community anchor institutions, bordering on [pictorial maps](https://www.davidrumsey.com/luna/servlet/s/90htkr).

One way to look at it is that this PR constrains the symbology by ruling out more familiar, abstract icons. We’ll need to redesign the school icon so it can “compete” with the other community anchor icons. That’s a shame, because it’s the most recognizable of any traditional POI icon, and it provides a convenient way for us to distinguish between different school classifications once that becomes possible. Other obvious icons for police/fire/doctors/campground may end up looking too simple too. Long-term, if we add [landmark icons](https://www.justinobeirne.com/google-maps-adds-landmark-icons) and brand icons (#809) to the map, we’ll actually want the usual POI icons to be relatively understated, so that these fancier icons can draw the reader’s attention.

A useful exercise would be to observe the map at a zoom level where we only show community anchor institutions.

wmisener:
I also thank you for this useful framing of the core design issue here. While more obscure features should by necessity be iconic, there could definitely be a place for more abstract shapes for higher importance icons. My inclination is definitely toward more iconic icons, which I find more intuitive, but I'm open to a variety of solutions. The legend allows Americana a bit more freedom than other maps might have in this respect.

I'm not totally sure I see an iconic POI here necessitating a redesign for the school icon, or ruling out other simple icons. Some are well-known enough to stand on their own, and/or common enough that people will pick up on it quickly even if unfamiliar. Compare that to a rare POI like a town hall that, by definition, occurs once per town. To me, that rarity requires it to be at least moderately self-explanatory. So I don't see them in conflict per se, though maybe there's a stylistic middle ground.

> A useful exercise would be to observe the map at a zoom level where we only show community anchor institutions.

Right now, that's z15, though it should probably really be z13 or so. The only POIs rendered at that level currently are schools and hospitals.

wmisener:
Some more samples, still looking at Culver City, CA, whose city hall happen to be conveniently close to a hospital and some schools.

Here's the star icon, modified from the state capital star already in use:
Screen Shot 2023-02-25 at 11 40 33 PM

I appreciate the simplicity, but I'm not personally convinced by this rendering. To me the star looks like a generic POI marker, maybe a non-specific attraction. If I had to guess, I'd think it was a police station due to the similarity to a sheriff's badge. I also find the logic is a bit muddled when relating this to capital cities: most places that get this icon, which is to say most towns, would not be cities represented by black stars. Meanwhile, the actual capitol buildings, denoted by their city getting a black star, are not always tagged `amenity=townhall` and thus wouldn't get a blue star icon themselves.

wmisener:
Here's another icon riffing off of the clock tower [Minh suggested](https://github.com/ZeLonewolf/openstreetmap-americana/issues/794#issuecomment-1445238645):
Screen Shot 2023-02-25 at 11 42 46 PM

Might need some tweaks, I think this design direction has potential as a way to preserve geometric simplicity while still giving an intuitive feel. I think keeping the flag is worthwhile but would be open to making it wavy if desired.

wmisener:
Now that I've made the above icon, I'm realizing that maybe I think it's intuitive because this is Culver City Hall! Haha
image

ZeLonewolf:
I think the flags should all point the same way. The school already decided that the wind is blowing from the east.

ZeLonewolf:
I think the flags should all point the same way. The school already decided that the wind is blowing from the east.

ZeLonewolf:
At the end of this discussion, it would be good if we could add our new POI design understanding to the POI section of the contributors guide.

1ec5:
Thanks for humoring my suggestions; it’s good to see where they fall apart.

> I also find the logic is a bit muddled when relating this to capital cities: most places that get this icon, which is to say most towns, would not be cities represented by black stars. Meanwhile, the actual capitol buildings, denoted by their city getting a black star, are not always tagged amenity=townhall and thus wouldn't get a blue star icon themselves.

Good point. I’m not entirely convinced that people would be confused by a star suddenly appearing within a non-capital city at a higher zoom level – as you say, this might ultimately happen for police stations anyways. But maybe I’m implicitly relying on the fact that the name already includes “city hall”. We can’t necessarily assume that’s the case everywhere in the world, even in English.

> Compare that to a rare POI like a town hall that, by definition, occurs once per town. To me, that rarity requires it to be at least moderately self-explanatory.

What if we lower the stakes? Some digital maps, such as Apple Maps, use the same icon for all government office POIs, not only a government headquarters building but also various department offices. Already we have the issue that county headquarters are tagged either `amenity=courthouse` or `amenity=townhall` `townhall:type=county`, depending on whether the county’s executive and judicial functions are colocated in the same building.

If we use the same icon for all government office POIs, then we open up more possibilities. It not really a community anchor institution, so the argument about “basic” POIs having abstract icons goes away. By the same token, we could use any symbol and the user would be more likely to associate it with government.

> Now that I've made the above icon, I'm realizing that maybe I think it's intuitive because this is Culver City Hall! Haha

Yes, I think we’re bringing our own biases as to what a “typical” town or city hall might look like. It’s ironic because there has never been a “typical” architecture for town or city halls in general across the U.S. In many places, they’re actually pretty nondescript office buildings (looking at you, New Orleans City Hall). But to a New Englander, it must seem strange that a town hall icon _doesn’t_ take the shape of a one-room schoolhouse. For the record, I was biased towards a fictional city near Culver City named Hollywood. 😉

What stereotype best captures “Americana”? My personal stereotype of these facilities differs between _town_ halls and _city_ halls. When in doubt, I figure that map symbols should tend toward the small-town feel, just like the maps they teach kindergarteners to read.

It’s worth noting that the traditional school symbol is part of a motif of symbols that consist of a finial (or “hood ornament”, if you will) atop a square (representing a building). The other obvious example is places of worship, for which print maps, particularly topographic maps, top the square with a cross or other religious symbol. Once again, this assumes the reader’s familiarity with typical architecture, but it’s successful to the extent that one can distinguish a cross from a pennant. Maybe this is where the clock tower symbol comes from as well.

We gravitate towards a flag as a town hall’s finial for the same reason we use it as a school’s finial: every town hall should have a flag raised above it, right? We can’t have two icons looking too similar, so this is where you’ve opted to make the building more detailed than a simple square.

Are there any other finials besides flags and clocks that would communicate the governmental nature of a government building? Going out on a limb, we could top the square with a star. A literal interpretation of that icon would fit in better in the PRC than in the U.S., but would it be interpreted literally if we add other icons that follow this finial motif?

zyphlar:
I'm a big fan of abstract in the sense of "what can a person expect to DO at this amenity?" which is a big criterion for how OSM is structured. Sure America might love its stars-with-circles-are-the-government map motifs, which works great for capital cities, but when something is tagged in OSM the distinguishing factor is almost always the amenity a user can expect to find there, vastly over what kind of building or what ownership structure it may exist in.

So that's why we have a latte mug for cafe and a burger for fast food and a painting for museums, even though all three of those amenities can be and often are in buildings that could look abstractly like a schoolhouse.

So in that vein, my suggestion is an apple-pencil-ruler or graduation cap for a school, a gavel for a courthouse, and a podium for a town hall.

wmisener:
>If we use the same icon for all government office POIs, then we open up more possibilities. It not really a community anchor institution, so the argument about “basic” POIs having abstract icons goes away. By the same token, we could use any symbol and the user would be more likely to associate it with government.

I agree that we eventually want to adhere to a consistent design language, which will allow us to make new icons as variations on a theme. In my mind, I was thinking of rendering courthouses distinctly since they have broadly-understood iconology associated with them (like the scales of justice, or a gavel). But I see the reasoning of rendering them similarly or the same as town halls, keeping the "government" theme intact. That's even before considering the usage of `amenity=courthouse` for county administrative seats, which I was unaware of.

>The other obvious example is places of worship, for which print maps, particularly topographic maps, top the square with a cross or other religious symbol. Once again, this assumes the reader’s familiarity with typical architecture, but it’s successful to the extent that one can distinguish a cross from a pennant. Maybe this is where the clock tower symbol comes from as well.

Places of worship are of course a whole other can of worms, but my initial inclination is that a building with a religious symbol "finial" works well for a Christian cross, moderately for an Islamic star and crescent, and poorly for most other religious symbols (Jewish star of David, Buddhist wheel, Indian om, etc), and so might not be the best stylistic choice. But I don't want to sidetrack the conversation here, so further discussion is probably best reserved for a new issue.

>We gravitate towards a flag as a town hall’s finial for the same reason we use it as a school’s finial: every town hall should have a flag raised above it, right?

Maybe this is what you mean, but I don't see the flags' symbolism in these features as merely imitative of any real flags being flown: the flag represents the public/governmental nature of the objects. I could see a similar flag motif used for other similar facilities, like ranger stations and embassies, when those are added. I think a school having a modified flag is sensible enough, given that public schools are quasi-governmental, but common enough to be distinguished.

wmisener:
>So that's why we have a latte mug for cafe and a burger for fast food and a painting for museums, even though all three of those amenities can be and often are in buildings that could look abstractly like a schoolhouse.

I'll note that this style, while indeed rendering cafes with latte mugs, does not yet render fast food restaurants or museums at all, and I'm not sure which style you're referring to that renders art museums with paintings, maybe Maki? Carto renders art museums as a columned neoclassical building, and art galleries as an artist's palette.

I personally am not sure what a podium symbol would look like, and I don't think I've ever seen such a symbol used for town halls or similar features on an American map. But if you feel up to prepping a test visual I'd be open to trying it. Rendering universities and colleges with a mortarboard icon to distinguish them from primary and secondary schools was suggested in #694, and I've been meaning to prepare a pull request to implement the idea.

zyphlar:
I'm just pulling random examples out of thin air, no idea if it's implemented in any particular style. My suggestion for houses of worship is a praying person, again because it's just so easy for schoolhouses, courthouses, and churches to look identical otherwise.

wmisener:
>I think the flags should all point the same way. The school icon already decided that the wind is blowing from the east.

A wind out of the east: poi_town_hall_towerbuilding
Screen Shot 2023-02-26 at 10 08 25 PM
This feels like the wrong way for me, since usually flags are depicted with the hoist side to the left. Maybe we should switch the school flag to blow the other direction 😆. But I don't really feel strongly about this either way.

>Going out on a limb, we could top the square with a star.

Here's my attempt at a star-topped building poi_town_hall_starbuilding
Screen Shot 2023-02-26 at 10 33 16 PM

The first thing that popped into my mind looking at this icon is a military officer's rank insignia or epaulet, making me think star symbology might be appropriate for military facilities. Maybe a more square building would mitigate this. As a town hall, it is perhaps a bit post-revolution, but I suppose it could work in the framework of a consistent design language.

My personal preference of the options I've attempted is some variation of a building with a square flag atop it, like poi_town_hall_towerbuilding or its reflection, as I find it the best balance of simple but still conveying the meaning.

1ec5:
> So that's why we have a latte mug for cafe and a burger for fast food and a painting for museums, even though all three of those amenities can be and often are in buildings that could look abstractly like a schoolhouse.

In that case, we might have some contention between the $ icons for town halls and banks, your tax dollars being hard at work. 😉

Both the finial-on-building motif and the object-found-there motifs have their limits. This is why the same maps that symbolize schools with the traditional flag-on-square also symbolize the post office with an envelope, the police station with a sheriff’s badge, and a fire station with flames. Conversely, the same maps that symbolize ice cream parlors with ice cream cones also symbolize museums as classical buildings with Doric columns. A very typical bank icon shelters the almighty dollar under a roof, lest it get rained on.

I wouldn’t expect us to choose one motif exclusively, but I don’t think the school icon should stand apart from the rest, clearly evoking a motif that we don’t follow through on.

> That's even before considering the usage of amenity=courthouse for county administrative seats, which I was unaware of.

This is an artifact of different regional forms of government. No one in California would equate a county courthouse to a seat of county government, because the superior courts in that state are always separate from county offices. But many states historically collocated all county government into a courthouse, in rural and urban counties alike. Some like Kentucky and Texas still vest executive authority in the county judge.[^louisiana] So the term “county courthouse” is understood to be analogous to “city hall” or “statehouse”.[^statehouse]

> The first thing that popped into my mind looking at this icon is a military officer's rank insignia or epaulet, making me think star symbology might be appropriate for military facilities. Maybe a more square building would mitigate this. As a town hall, it is perhaps a bit post-revolution, but I suppose it could work in the framework of a consistent design language.

Yes, I think it would be clearer with a narrower building. I would suggest the same of your preferred icon in https://github.com/ZeLonewolf/openstreetmap-americana/pull/807#issuecomment-1445291890. If we want the icon to depict something architectural and also include a flag as a finial, then let’s focus on the most distinctive part of the façade rather than reproducing the whole building. I’m pretty sure this is the concept behind the clock tower icon we rejected earlier; it was meant to focus on the clock tower of a city hall rather than represent a freestanding clock tower.

Your original design included a pediment, while later iterations included a smaller but more grandiose cupola. Both are traditional architectural elements of the halls of power. If we reduce the building to a triangle or semicircle topped by a flag, I think it would satisfy both requirements of intuitiveness and simplicity. If a triangle, then the flag could be offset to one side to avoid confusion with the school icon.

[^louisiana]: Curiously, a Louisiana parish is headed by a _police jury_, which led the GNIS import to mistag county headquarters as `amenity=police`, even though the police are always headquartered elsewhere.
[^statehouse]: Some statehouses also house the governor’s office and supreme court. A map would give the statehouse more prominence than the governor’s mansion, in any case.

adamfranco:
2 possible arguments for flags oriented for a wind from the west:

1. The [prevailing winds](https://en.wikipedia.org/wiki/Prevailing_winds) in the continental US are from the west. [Local winds may come from any direction](http://hint.fm/wind/), but there will be more from the west than the east. 
2. Since English is read left-to-right it makes sense to begin reading the flag with the pole on the left.

ZeLonewolf:
I don't have any preference on which way the flags are pointing, I just think they should be in the same direction.

wmisener:
Got a bit busy, but back to this town hall icon. It seems like the west-to-east blowing flag is more popular, maybe if this icon ends up with a flag, we can switch the school icon's pennant separately (as far as I recall the direction was chosen arbitrarily there).

Narrower buildings:
Star-topped: poi_town_hall_starbuilding
Tower-topped: poi_town_hall_towerbuilding

Making the aspect ratio of the tower a little longer (inspired by Rontgen, see https://osmus.slack.com/archives/C01V02K52UX/p1677796021787889): poi_town_hall_talltowerbuilding
Domed building (inspired by the semi-circle suggestion): poi_town_hall_domebuilding

Of these, I kind of like the dome for its simplicity. Does it translate well enough across towns and cities, not all of which may have domed buildings?

1ec5:
> Making the aspect ratio of the tower a little longer

I think the flag would have to be _much_ smaller, almost proportional to the building, in order to be recognizable as a flag atop a building. As it is, it still looks like an icon for a monument perhaps; I’m not confident that a user who encounters this icon independently would interpret it correctly.

> Of these, I kind of like the dome for its simplicity. Does it translate well enough across towns and cities, not all of which may have domed buildings?

I do like it for its simplicity, though its weakness is that smaller towns are less likely to have domes. (They’re expensive.) A dome would make the icon more suitable as a general government building icon, due to its association with state capitols and county courthouses, if we decide to use it for that later on.

A pediment (triangle) would be more directly associated with town halls big and small. However, my earlier suggestion of a simple triangle could be problematic: it might look like a flag atop a hill. If we go this direction, perhaps it could be a squat triangle atop a square that’s only _slightly_ wider. Again, I think a silhouette of the building would suffice.

wmisener:
> A dome would make the icon more suitable as a general government building icon, due to its association with state capitols and county courthouses, if we decide to use it for that later on.

Unfortunately, I'm not sure a tagging standard for such objects exists. I've seen state capitols tagged a number of ways (including `amenity=townhall`), and even the US Capitol is tagged `office=government`.

wmisener:
>I think the flag would have to be much smaller

Here's an attempt at a smaller flag: poi_town_hall_talltowerbuilding
I'm still not sure this screams "government building" to me personally.

>If we go this direction, perhaps it could be a squat triangle atop a square that’s only slightly wider.

I like this idea. Here's an attempt: poi_town_hall_pedimentbuilding

Possibly a good direction. It maybe ends up looking a bit like a hill though. Could try it with a little clock too: poi_town_hall_pedimentbuilding_clock

You could make the triangle wider, but I feel like it loses some of its definition as a shape, like it's a slightly off version of a full home-plate pentagon: poi_town_hall_pedimentbuilding

Similar design without a flag: poi_town_hall_pedimentbuilding_noflag

I do personally like the flag motif. However, I do recognize that being pixel-aligned necessitates that a centered flagpole be fairly thick, 2px, which limits the scale of everything else. If anything, I feel like the triangle should be even smaller if there isn't a flag: poi_town_hall_pedimentbuilding_noflag

Perhaps the dome could benefit from a slightly wider base as well: poi_town_hall_domebuilding

Anyways, while this is a lot iterating, I hope it's clarified something about the desired attributes of the symbols desired for this style. My personal preference probably remains a building with a tower topped by a flag (something like poi_town_hall_towerbuilding, maybe with a slightly smaller flag). But I don't mind the triangle-topped or domed ones either.

1ec5:
>> A dome would make the icon more suitable as a general government building icon, due to its association with state capitols and county courthouses, if we decide to use it for that later on.
> 
> Unfortunately, I'm not sure a tagging standard for such objects exists. I've seen state capitols tagged a number of ways (including `amenity=townhall`), and even the US Capitol is tagged `office=government`.

Indeed, I mean that a domed icon would be suitable as an icon for government in general, hence `office=government`, `amenity=townhall`, and even `amenity=courthouse`. Different tags in this group would still have different minimum zoom levels. But it sounds like you’d prefer to call out town halls specifically as a separate category from government in general.

> Anyways, while this is a lot iterating, I hope it's clarified something about the desired attributes of the symbols desired for this style.

Yes, thanks for iterating further; I really appreciate your patience! I definitely see us getting closer to a good spot. Here’s my unvarnished first impression of each of these iterations:

* poi_town_hall_talltowerbuilding Without enough definition between the tower and flagpole, the whole icon looks like a flagpole on a pedestal. The flagpole does need to be thinner, at the cost of looking fuzzier at 1×.
* poi_town_hall_pedimentbuilding I agree that it isn’t clearly enough a building. Perhaps making the pediment squatter would help.
* poi_town_hall_pedimentbuilding_clock Unfortunately, the clock doesn’t really come out; it looks more like an unfilled triangle atop the base, which somehow reminds me of a tent.
* poi_town_hall_domebuilding Although I think a pediment would be more broadly applicable than a dome for town halls specifically, I really appreciate this icon for being both simple and recognizable. The proportions are quite pleasant too. I’d like to see if we can somehow make the pediment idea work, but if not, this would be a very good alternative in my opinion.
* poi_town_hall_towerbuilding The flag would have to be a _lot_ smaller for the proportions to work out. If we were to shine a light on the front of this icon, I’m 100% sure it would be an oversized flag of Texas.

I tried some of my own medicine and came up with squat pediment, thin flagpole. Yes, the flagpole is slightly offset to the left, but the imbalance isn’t discernible at 2×, let alone 1×. It’s basically the White House to poi_town_hall_domebuilding’s Capitol Hill. My only concern with this version is that it starts to look too similar to a school icon.

ZeLonewolf:
This one:
poi_town_hall_pedimentbuilding
Kind of looks like a circus or army tent to me.

wmisener:
The offset flagpole trick opens up some design options:
poi_town_hall_towerbuilding The tower style, with a less Texas-sized flag
poi_town_hall_pedimentbuilding Longer/squatter base of the building. Fairly similar to  poi_town_hall_pedimentbuilding_smallflag (same as in https://github.com/ZeLonewolf/openstreetmap-americana/pull/807#issuecomment-1475441424 but with the outline added)
poi_town_hall_domebuilding Dome with a smaller flag

wmisener:
Some test renderings of these four:
Tower
Screen Shot 2023-03-19 at 6 46 35 PM
Pediment low/long building
Screen Shot 2023-03-19 at 6 48 00 PM
Pediment tall/squarer building
Screen Shot 2023-03-19 at 6 48 57 PM
Dome
Screen Shot 2023-03-19 at 6 45 13 PM

Hm, these make me think I went too far the other way with the flags, they look too puny next to the school pennants. Those could be adjusted, but since we're going for iconography here I think the national flags should be a bit bigger.

1ec5:
I’d favor thinning out the school icon’s flagpole and perhaps shrinking its flag too.

wmisener:
Here's a couple samples with the flags a bit bigger.

Screen Shot 2023-03-19 at 10 51 51 PM

Screen Shot 2023-03-19 at 10 49 18 PM

I personally find these more attractive than the smaller flag versions. They feel more robust to me against the other symbols, and I'm not so bothered by them appearing not to scale. But that probably comes down to a personal preference.

I certainly agree that the flags should end up similar in size to the school pennants, so if smaller flags are opted for, it should be paired with shrinking the school pennants (along with switching their direction).

1ec5:
The flag size only really bothered me on versions like poi_town_hall_talltowerbuilding and poi_town_hall_towerbuilding with a much narrower base. The larger size is fine on the stouter versions. I did argue for a narrower building in https://github.com/ZeLonewolf/openstreetmap-americana/pull/807#issuecomment-1447268100 for consistency with the school icon and potentially a future place of worship icon. But of all the constraints I’ve put on you, this is the most fungible, simply because a wider building conveys importance, which is appropriate for a town hall or other governmental building.

ZeLonewolf:
Of the options so far, the original and this one:
image

Seem to be the clearest.

wmisener:
OK, do you specifically mean the dome with the smaller flag there, or just the dome shape generally? I think I find the large flag more to my taste.
More samples (Scranton, PA, [localhost link](http://localhost:1776/#map=15.53/41.411779/-75.657017)):

Dome, large flag poi_town_hall_domebuilding
Screen Shot 2023-04-04 at 10 48 46 PM

Dome, small flag poi_town_hall_domebuilding_smallflag
Screen Shot 2023-04-04 at 10 55 55 PM


Original columned building poi_town_hall_straightflag
Screen Shot 2023-04-04 at 10 58 15 PM

ZeLonewolf:
> OK, do you specifically mean the dome with the smaller flag there, or just the dome shape generally?

Oh, I meant the dome shape generally.  I do agree that the larger flag is a bit nicer.  Also bear in mind that mine is just one opinion of several, which carries no extra weight just because the project happens to be presently located in my github space...

1ec5:
“Dome, large flag” is quite fine. The thinner flagpole makes it even better. I cannot think of a better icon for all government building POIs once we get to it. (Those POIs would be filtered to come in at higher zoom levels than town halls, so the distinction will be clear.) I think we came close with a pediment, but it is hard to shake the association with a tent of some sort.

wmisener:
I've updated this PR to include the domed icon: poi_town_hall. Not to let the other artwork go to waste, I think I might propose a flagless version of the original columned-building icon I suggested here as the icon for museums (#820).

Test render ([localhost link](http://localhost:1776/#map=15.11/31.999495/-102.076878)):
Screen Shot 2023-04-05 at 10 23 11 AM

Note that in the version from a couple weeks ago, the circle got slightly wonky, I think I had inadvertently altered the bezier curves. That's fixed in the current icon.

src/layer/poi.js
```diff
@@ -53,6 +53,14 @@ var iconDefs = {
     color: Color.poi.infrastructure,
     description: "School",
   },
+  townhall: {
+    classes: {
+      town_hall: ["townhall"],
+    },
+    sprite: "poi_town_hall",
+    color: Color.poi.infrastructure,
+    description: "Town hall",
```
1ec5, line 62: As far as I can tell, American maps tend to label this as “city hall” rather than “town hall”, maybe because formally, every state has cities but only some have towns (while others have villages etc.). Or we could split the difference with “City or town hall”.

There’s still the minor issue of `amenity=townhall` being used for [county headquarters](https://taginfo.openstreetmap.org/tags/townhall%3Atype=county#overview) that aren’t courthouses, but that won’t be as much of a problem once we use the same icon for government buildings and rename this entry to “Government office” or somesuch.

1ec5 (approved):

src/layer/poi.js
wmisener: I'd be fine with "City hall" or "City or town hall" in the legend. I went with town only because that's closer to the OSM-ism for it.

I don't see other uses of the tag as too much of a blocker. The defined use seems predominant, and I view a lot of those oddball ones as coming from OSM's lack of an `amenity` tag for government centers besides `amenity=townhall`. Heck, even the state capitol of California is tagged `amenity=townhall`: https://www.openstreetmap.org/relation/20084. The `townhall:type=county` only has 84 uses in the database anyways, compared to ~150k `amenity=townhall` uses total, and many of those are outside the US, where the definition of a county can be different.

ZeLonewolf: I would prefer something like city/town hall rather than something more technical like "municipal administration building" :grimacing: 

I would suggest that there are far more town halls than city halls in existence!

wmisener: Unscientific overpass queries: 2687 `amenity=townhall` in the US contain "Town Hall" in the name ([link](https://overpass-turbo.eu/s/1tqB)), 4284 `amenity=townhall` in the US contain "City Hall" in the name ([link](https://overpass-turbo.eu/s/1tqC)). Seems to come from many states having few "towns", so even places with small populations are still technically "cities". The "town halls" appear concentrated in a few states, mostly in the Northeast but also noticeably SC, WI, and MD, that have lots of "towns", whereas "City Halls" are more evenly distributed (even places with lots of towns have some cities). This seems to support naming the icon "City hall" or maybe "City/town hall"

ZeLonewolf: Good evidence, I stand corrected.

1ec5: > I would prefer something like city/town hall rather than something more technical like "municipal administration building" 😬

Oh yeah, as long as this entry corresponds to `amenity=townhall`, it should be some kind of “hall”. But I think it’s only a matter of time before we reuse this icon for `office=government` (which includes state capitols), at which point we’d generalize this legend entry to correspond to either `amenity=townhall` or `office=government` (“City hall or other government office”).

ZeLonewolf: > But I think it’s only a matter of time before we reuse this icon for office=government (which includes state capitols), at which point we’d generalize this legend entry to correspond to either amenity=townhall or office=government (“City hall or other government office”).

That's actually a really good idea, and is aligned with our concept for simpler icons.  We could keep the town/city halls at lower zooms than the other government offices might show up at.

wmisener: I switched the description to "City hall". Figured more concise was better rather than going with "City or town hall". I think people will understand it applies to other municipality types.

ZeLonewolf:
Looks like conflicts need to be resolved from the recent class/subclass updates in the POI layer.

wmisener:
Was a conflict with the bus stops being added to the same zoom level.

ZeLonewolf (approved):
Looks good, thanks!

image

Hopefully we can make the wind blow in the same direction at some point, but it's not a showstopper.

PR #808, opened 2023-02-26
Through railways include branch lines

1ec5:
We got some private feedback that the legend’s “Mainline track” entry is misleading because it also includes railroad tracks tagged `usage=branch`. OpenMapTiles doesn’t classify railways by `usage=*`, and this style doesn’t visually distinguish railways to the extent that this key would be very useful. (It could be useful to a more rail-centric style, however.) Though a little more verbose, “Mainline or branch line track” seems to cover our bases well enough. It doesn’t mention heavy rail, but this is implied by the presence of separate entries for light rail etc.


claysmalley:
Was this feedback from a North American, or someone from elsewhere? I understand "mainline track" to be `railway=rail` with `service` empty. Mainline track can have any `usage=*`, including `branch`, `industrial` or `military`.

1ec5:
Yes, this was feedback from a North American, albeit someone deeply familiar with OSM’s railway tagging scheme. I’m not well-versed in rail in general, so “mainline” was the best I could come up with at the time in #632.

wmisener:
Not a rail expert, but would "Main or branch line track" still convey the same meaning a little more concisely? Or does "mainline" have to be one word?

wmisener:
One thing that would be neat (way beyond the scope of this PR) would be to have an option to "translate" the legend into OSM tags, so you could see which tags the style interprets into which symbols. So in this case, rather than "Mainline track" it could say "`railway=rail`". Would be useful for experienced OSM users coming to the style, and as mapper feedback. I have no idea if this is feasible technically since everything passes through OMT first, but I suppose we already do some of the legwork by making taginfo entries for every symbol.

claysmalley:
Perhaps we could call `railway=rail` "conventional rail"? This technically excludes high-speed rail, though we don't have any of that in the Americas yet.

1ec5:
> Not a rail expert, but would "Main or branch line track" still convey the same meaning a little more concisely? Or does "mainline" have to be one word?

Sure, we can factor out “line”.

> One thing that would be neat (way beyond the scope of this PR) would be to have an option to "translate" the legend into OSM tags, so you could see which tags the style interprets into which symbols. So in this case, rather than "Mainline track" it could say "`railway=rail`". Would be useful for experienced OSM users coming to the style, and as mapper feedback. I have no idea if this is feasible technically since everything passes through OMT first, but I suppose we already do some of the legwork by making taginfo entries for every symbol.

Ideally, it would be up to the OSM Wiki to better associate raw tags with real-world terminology, and up to OpenMapTiles to better document how it translates raw OSM tags to its own schema. We do bake in assumptions about these translations into our taginfo project file, but it’s fragile; baking in assumptions about OSM tag semantics at the same would be even more fragile. Compounding matters, these entries don’t correspond one-to-one to OMT properties or even Americana’s own style layers. It’s very reader-centric, as any legend would be.

> Perhaps we could call `railway=rail` "conventional rail"? This technically excludes high-speed rail, though we don't have any of that in the Americas yet.

The point of this entry is to distinguish “main” tracks in the sense of not being surrounded by lots of other tracks. That’s what we’ve styled with a little more emphasis than spurs, yard tracks, etc. The spur/yard entry is also conventional rail.

ZeLonewolf (approved):
Current terminology sounds good to me based on discussion above.

PR #816, opened 2023-03-10
Add Newfoundland and Labrador shields

### LINKED ISSUE #280: Newfoundland and Labrador shields
Newfoundland and Labrador has a simple black-on-white rectangular shield.

Newfoundland and Labrador has only one route network (other than the Trans-Canada Highway), tagged `network=CA:NL:R`. The `:R` at the end is unnecessary and inconsistent with other provinces that only have one route network. A suggested tagging scheme that follows guidelines would be `network=CA:NL`. We should hold off on supporting NL shields until this tagging issue is resolved one way or another.

claysmalley:
Fixes #280.

Waiting for tile update before testing is possible.

[Screenshot from 2023-03-09 22-23-43](http://localhost:1776/#map=13.18/47.44842/-53.54476)
[Screenshot from 2023-03-09 22-24-28](http://localhost:1776/#map=12.18/52.85477/-67.10672)
[Screenshot from 2023-03-09 22-25-27](http://localhost:1776/#map=10.22/53.287/-60.3738)


ZeLonewolf:
Is there something special about the 26px width compared to other similar rectangle shields that use 30px?

claysmalley:
I felt it matched the proportions of the real-life shield a little better. The default 34px width for rectangles looked too wide. Though a variable-width rectangle seems okay.

[Screenshot from 2023-03-22 15-53-59
Screenshot from 2023-03-22 15-54-45
](http://localhost:1776/#map=12.12/47.43306/-53.55136)

ZeLonewolf:
I agree that the 34px width looks too wide, and I'm not sure why that's the default.  Rhode Island (just to choose an arbitrary example here) has two different fixed-width signs (one for 1-2 digit routes and another for 3-digit routes) and we match that with variable width signs.  From [what I see on Wikipedia about these highways](https://en.wikipedia.org/wiki/List_of_Newfoundland_and_Labrador_highways), NFL+L also does the same, so I'd think we'd want to match that behavior with the variable width.

claysmalley:
> I agree that the 34px width looks too wide, and I'm not sure why that's the default.

Some European networks use a wide fixed-width rectangle. Not as common in North America.

> Rhode Island (just to choose an arbitrary example here) has two different fixed-width signs (one for 1-2 digit routes and another for 3-digit routes) and we match that with variable width signs. From [what I see on Wikipedia about these highways](https://en.wikipedia.org/wiki/List_of_Newfoundland_and_Labrador_highways), NFL+L also does the same, so I'd think we'd want to match that behavior with the variable width.

I'm hesitant to make shield design assumptions based on SVGs created by Wikipedians, rather than real-life photographs. Actual signage appears to be a rectangle of constant proportions, but much closer to square than the European-style rectangles. However, I don't think the rectangle width is a necessary design element to recognize the route network, and I think most Newfoundlanders would probably agree. I've gone ahead and changed the shield to a variable-width rectangle.

ZeLonewolf:
The few examples I could find:
image
image

ZeLonewolf:
In other words, it seems entirely random.

ZeLonewolf (approved):
Looks good to me!

PR #829, opened 2023-03-20
Make font path absolute

ZeLonewolf:
2nd attempt to fix the broken font link - just force a relative URL.

We can come up with a scheme to make this properly relative later, but the immediate priority is to have a working map.

### DIFF
--- a/src/js/style.js
+++ b/src/js/style.js
@@ -4,7 +4,8 @@ import * as Layers from "../layer/index.js";
 export function build(tileURL, spriteURL, locales) {
   return {
     name: "Americana",
-    glyphs: "fonts/{fontstack}/{range}.pbf",
+    glyphs:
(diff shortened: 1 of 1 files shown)


src/js/style.js
```diff
@@ -4,7 +4,8 @@ import * as Layers from "../layer/index.js";
 export function build(tileURL, spriteURL, locales) {
   return {
     name: "Americana",
-    glyphs: "fonts/{fontstack}/{range}.pbf",
+    glyphs:
+      "https://zelonewolf.github.io/openstreetmap-americana/fonts/{fontstack}/{range}.pbf",
```
1ec5, line 8: If you want to make the fonts host-neutral, then the function would need to take the glyph URL as a parameter, just like it takes the tile and spritesheet URLs. This parameterization only benefits the command-line style JSON generation tool anyhow.

1ec5 (approved):

Issue #790, opened 2023-02-14
Render railway stations


wmisener:
Railway stations can be major landmarks in big cities and small towns alike across America, making them in some sense community anchor institutions (#435) and certainly a POI the project should render (#692). After the rendering of railroad tracks (#101) and some discussion of highlighting passenger routes (#745), I thought I'd spin off this issue to discuss and track rendering the stations specifically.

These icons should probably be purple, befitting their status as transportation hubs. Different maps use different symbologies. There are quite a few examples of different choices in the track samples of #101, some of which I'll re-post here. European OSM maps like Carto use a simple square. This befits a rail station's importance, and a simple symbol along the track is similar to what is used on some American maps, at least at a state-wide scale level.
<img width="371" src="https://user-images.githubusercontent.com/1231218/151247585-1878917d-ad53-4b8d-a5d3-eff2dd21d039.jpg">

When zoomed further in to a city scale, I'm personally thinking that a plain square is not sufficient/desirable. One possibility is to riff on the MUTCD rail shield: 
![image](https://upload.wikimedia.org/wikipedia/commons/a/a5/MUTCD_I-7.svg)
A shielded train station symbol would have a nice synergy with a shielded bus station rendering proposed in #436. The MUTCD symbol contains a platform, but other maps use just the train on tracks, like iD/temaki:
<img width="371" src="https://user-images.githubusercontent.com/58491489/218838742-bec352dc-4ea5-48b7-a58a-30fe2c532f8b.png">
Even an old-timey steam train symbol would probably be recognizable as a station icon, if a bit hokey:
<img width="371" src="https://user-images.githubusercontent.com/1231218/151246910-0aca2cff-3f11-4696-8581-7ed181d9ac02.jpg">

In the future, we could even consider rendering the rail operator's brand symbol. On US maps, for example, the Amtrak logo is a fairly ubiquitous symbol for rail stops. But a non-specific default should probably be established first.
<img width="500" src="https://user-images.githubusercontent.com/1231218/151249099-f9684e0c-1e6d-4b7a-b854-b7e03bdf1aa4.jpg">

1ec5:
> One possibility is to riff on the MUTCD rail shield

This symbol isn’t unique to the MUTCD. Similar symbols appear in wayfinding signs at airports, train stations, and bus stations. Some public transportation agencies use these symbols instead of fancy logos to represent their individual services.

[...]

PR #830, opened 2023-03-22
Update transport POI color to mauve

claysmalley:
> >The intent was for aerialway magenta to be used for public transit features. Airports are not necessarily passenger facilities, but aerial lifts almost always are.
>
> @claysmalley do you think you could add the shade you used to the [Contributing Guidelines](https://github.com/ZeLonewolf/openstreetmap-americana/blob/main/CONTRIBUTING.md#color-scheme-1)? Maybe the purple should be in a new "Air" category, with magenta replacing it for "Transport"

_Originally posted by @wmisener in https://github.com/ZeLonewolf/openstreetmap-americana/issues/790#issuecomment-1476999639_


ZeLonewolf (approved):
Looks good, no breakage.  Thanks!

PR #831, opened 2023-03-24
Adjust line width on detailed shields

claysmalley:
Just cleaning up nitpicks on various shield icons. Very thin lines are made thicker, and edges are moved closer to pixel boundaries.

The icon for DC Route 295 now has a full outline of DC and a white text halo, instead of a cutout outline.

The Kansas Turnpike now just has a T in the middle, instead of KTA (following the same logic as the New Jersey Turnpike icon).


ZeLonewolf (approved):
Looks good!
image
image
image
image
image
image
image

Issue #811, opened 2023-03-04
John Hunt Morgan Heritage Trail shields


1ec5:
The John Hunt Morgan Heritage Trail is a pair of historical tour routes documenting Morgan’s Raid through Indiana and Ohio during the American Civil War. The routes are marked with standard trailblazers and reassurance signs in Indiana and Ohio. The signs point out each turn in the route and occasionally point to an interpretive sign. (There’s a corresponding route in Kentucky too, but I can’t find any evidence of it being signposted beyond the interpretive signs themselves.)

The shields consist of the state shape in white, outlined in black, with the bottom half in black. Inscribed on the state outline are “John Hunt Morgan Heritage Trail”, “1863”, and the path Morgan’s raiders took through the state.

[![Harshaville Bridge](https://user-images.githubusercontent.com/1231218/224485054-27918856-9fa4-468f-8e13-84fd086b7e6c.png)](https://www.mapillary.com/app/?pKey=170481458324192)

So far, I’ve mapped only part of Ohio’s route as [a route relation](https://www.openstreetmap.org/relation/15504667) tagged `network=US:OH:JHMHT`, from the Indiana state line through Hamilton and Clermont counties. This is only a fraction of the route, which stretches almost to the Pennsylvania state line.

By highway route standards, this is a pretty unusual route. Like scenic routes, it’s very indirect in places. It’s signposted only in the general direction that Morgan’s raiders took, not in the other direction. Yet the route also backtracks and loops in some cases as it makes excursions off the main road to each historical site and interpretive sign. This requires the `signed_direction=yes` tag on the relation and a `forward` or `backward` role on each member. As of iD v2.25.1, any accidental gaps or forks in the relation cause iD to scramble the relation order, so extreme care must be taken when remodeling intersections or roundabouts along the route: openstreetmap/iD#8578 openstreetmap/iD#8415.

1ec5:
[...]

PR #832, opened 2023-03-24
Add shield for John Hunt Morgan Heritage Trail in Ohio

1ec5:
Added a crude shield for the John Hunt Morgan Heritage Trail in Ohio.

The shield is based on the three-digit Ohio state route shield, which is wide enough to fit the word “Morgan” in type that’s large enough to be legible at 2×. It probably still looks like mush in 1×, but it isn’t super important that the user be able to read the name, as long as it still looks like condensed red text. The text is set in Noto Serif, converted to paths for portability. I omitted the year 1863 because there wasn’t much room for it amid the wiggle that’s supposed to represent the route’s path through southern Ohio.

[Camargo](https://zelonewolf.github.io/openstreetmap-americana/#map=12/39.2078/-84.29893)

Working towards #811.


claysmalley:
Looks a little wonky with only the left edge removed. The source image has ⅔ of the top edge of Ohio missing—we should either keep the full border, or match the source.

I think this shield is still unique and recognizable without the text. Consider a version with all text omitted, but the full Ohio border restored.

1ec5:
You’re right, the northern border looks sort of cantilevered over the text, which is weird. With the border with Michigan removed:

shield_us_oh_jhmht

This narrower version based on the two-digit Ohio state route shield has less legible text, but I think the state shape comes through better and it somehow feels more rustic like the original:

shield_us_oh_jhmht

With the text removed, it does look less cluttered, but I don’t think it’s as recognizable without some complication up top:

shield_us_oh_jhmht

With the whole border restored, I think it’ll be too much of a departure from the original, though I can still see the resemblance if I already know what it’s supposed to represent:

shield_us_oh_jhmht

The other factor here is that Indiana’s John Hunt Morgan Heritage Trail has a similar shield based on the Indiana state shape, but with the red “MORGAN” even more prominent: https://github.com/ZeLonewolf/openstreetmap-americana/issues/811#issuecomment-1483203014.

ZeLonewolf:
It's unclear from the discussion whether this is ready for review or if work is pending.

1ec5:
I’d like to get some feedback on the options I presented in https://github.com/ZeLonewolf/openstreetmap-americana/pull/832#issuecomment-1483204419. For better or worse, I’m leaning toward the options that include the name “Morgan”, even if illegibly. The textless options resemble the Ohio state route shield too closely, which could confuse the reader. Red is part of the shield’s identity. Meanwhile, the Indiana Morgan Trail’s shape will be even more dependent on that name being there.

claysmalley:
I think it clashes too much with the map's aesthetic if the shield border isn't complete. I'm not opposed to text on the shield, though I still think it would look recognizable without it.

"JHM" seems okay to me. #245 and #562 are precedents for using initialisms even when they aren't present on the ground.

ZeLonewolf:
I hadn't seen it elsewhere, but here's the real-world shield apparently:

image

ZeLonewolf:
Link to live preview: https://preview.ourmap.us/pr/832/#map=12/39.20055/-84.27415

1ec5:
#811 has some more photos of the real-world shields. However, I’m trying to reduce the shields to the bare minimum, while maintaining recognizability, to avoid infringing on the copyright of whichever groups or agencies designed these logos. (Neither the Ohio MUTCD nor the Indiana MUTCD specifies these signs, so their original designs aren’t necessarily in the public domain.)

1ec5:
I narrowed the shield to match the two-digit Ohio state route shield’s dimensions and abbreviated the text to say “JHM” instead of “MORGAN”. I also truncated the border at Toledo instead of having it meet Indiana. The border remains a little thicker than that of the state route shields; a thinner border would look like a bug given the cantilevered shape, and it introduces a bit of the whimsy that I guess the tourism office was going for with the actual design.

[Milford](https://zelonewolf.github.io/openstreetmap-americana/#map=12/39.2078/-84.29893) [Goshen](http://localhost:1776/#map=12/39.22082/-84.18709)

ZeLonewolf:
The white edge on the northwest corner of the shield sure is jarring.  I wonder if we would be better to sacrifice a bit of faith to the original by just having the boundary go all the way around.

claysmalley:
I still don't like the incomplete border. The white-on-white contrast looks like a graphical error more than an intentional design choice. Is it necessary to retain the cutout when text no longer overlaps it?

The real-life Indiana variant has no border at all on the white part, and we'll probably need to add a border anyway when implementing it.

1ec5:
The main reason for an incomplete border is actually just to leave enough room for the text inside, but with the abbreviation that’s no longer necessary.

[Camp Dennison](https://zelonewolf.github.io/openstreetmap-americana/#map=12/39.2078/-84.29893) [Goshen](https://zelonewolf.github.io/openstreetmap-americana/#map=12/39.22479/-84.18939)

ZeLonewolf:
The complete border is definitely an improvement!

claysmalley (approved):
Looks great! Thanks for sticking it out through all the critique.

PR #833, opened 2023-03-24
Update Holmes County shield

### LINKED ISSUE #810: Holmes County–shaped shields in Holmes County, Ohio
Last June, to mitigate safety concerns caused by hapless tourists unfamiliar with Amish Country’s road numbering scheme, Holmes County, Ohio, [abandoned](https://thebargainhunter.com/news/news/signs-signs-everywhere-signs) the state-standard white square shields in favor of distinctive green shields bearing the county’s outline.

Diagram

[CR 349](https://www.facebook.com/permalink.php?story_fbid=pfbid0UuVej43UeUkLm28N3V9e7BGgc1nNnbXeHm1UdGwSBC6AicbnxCdtR1zSDKsr1G3Ul&id=100064690898107)

We’ll need to design a new shield blank in the shape of the county, use it in a new definition for `US:OH:HOL`, and remove the existing definition:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/4351e9806a59febb1efb50ded8c3445c378fb36b/src/js/shield_defs.js#L2444

The following Holmes County townships no longer need to have a “TWP” on top, because the county’s shields no longer look identical to the townships’ shields when simplified. Their entries should be moved up to the array that currently lists Holmes County:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/4351e9806a59febb1efb50ded8c3445c378fb36b/src/js/shield_defs.js#L2563-L2568

1ec5:
Replaced the generic white shield in Holmes County, Ohio, with a shape of the county in green, simplified slightly:

[Berlin](https://zelonewolf.github.io/openstreetmap-americana/#map=12/40.56352/-81.79596)

Fixes #810.

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -2491,6 +2491,12 @@ export function loadShields() {
     "WYA",
     "COS:Jackson",
     "FAI:Greenfield",
+    "HOL:Berlin",
+    "HOL:Clark",
+    "HOL:Knox",
+    "HOL:Monroe", // No black border in reality, but a border is needed for contrast.
+    "HOL:Paint",
+    "HOL:Salt_Creek",
     "JEF:Knox",
     "LOG:Bokescreek",
     "LOG:Pleasant",
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -2453,7 +2453,6 @@ export function loadShields() {
     "GAL",
     "HAS",
     "HOC",
-    "HOL",
     "KNO",
     "LAW",
     "LIC",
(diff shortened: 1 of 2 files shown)


src/js/shield_defs.js
```diff
@@ -2572,12 +2581,6 @@ export function loadShields() {
     ["ASD", "TWP"],
     ["ATH", "Trimble"],
     ["FAI", "Violet"],
-    ["HOL", "Berlin"],
-    ["HOL", "Clark"],
-    ["HOL", "Knox"],
-    ["HOL", "Monroe"], // No black border in reality, but a border is needed for contrast.
-    ["HOL", "Paint"],
-    ["HOL", "Salt_Creek"],
```
claysmalley, line 2580: These townships appear to be deleted instead of moved. Is this intentional? From #810:

> The following Holmes County townships no longer need to have a “TWP” on top, because the county’s shields no longer look identical to the townships’ shields when simplified. Their entries should be moved up to the array that currently lists Holmes County:

> The county engineer [plans to offer](https://thebargainhunter.com/news/news/signs-signs-everywhere-signs) townships the same sign minus the county outline. If and when that happens, we would migrate the townships’ entries to the array for green square shields. There would still be no need for a “TWP” banner in this style, because the township shields would be a different shape than the county shields.

While new signage is planned, I don't see a reason to remove support for existing signage.

1ec5: Whoops, that was a mistake, I added them back in.

claysmalley (approved):

ZeLonewolf:
Good to go or do we need a tile update?

1ec5:
This is just an artwork update. It’s good to go.

PR #838, opened 2023-03-26
Add Georgian uppercase

### LINKED ISSUE #837: All-caps Georgian labels are invisible
When the preferred language is set to Georgian, any layer that contains all-capital letters appears blank, namely, state/provincial labels and boundary edge labels.

[English](https://zelonewolf.github.io/openstreetmap-americana/#map=5/41.23/-81.35&language=en) [Georgian](https://zelonewolf.github.io/openstreetmap-americana/#map=5/41.23/-81.35&language=ka)
_English versus Georgian_

In practice, the Georgian alphabet is unicameral, but capital letters are used for decorative purposes, such as the reason why we’re uppercasing these labels.

ZeLonewolf:
Fixes #837


1ec5 (approved):
I can’t test this because we’re still hard-coding an absolute path to the deployed glyphs: https://github.com/ZeLonewolf/openstreetmap-americana/pull/829#discussion_r1142708915.

Issue #836, opened 2023-03-25
Language picker menu omits some supported languages


1ec5:
The language selection control is limited to a hard-coded list of languages, even though the vector tiles now include many more languages as of https://github.com/ZeLonewolf/openstreetmap-americana/issues/586#issuecomment-1483620850 and you can technically enter any valid language code via a URL hash parameter.

https://github.com/ZeLonewolf/openstreetmap-americana/blob/e26d76a1af1248b7380d93e4c0c9f9d3c8082aa7/src/js/language_label.js#L42-L112

We should synchronize this list with the actual list used to generate the tiles. (We can leave out some deprecated OSM codes like `ja_kana`.) Additionally, the control should allow the user to input any unrecognized but valid code, based on [`Intl.supportedValuesOf()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/supportedValuesOf).

1ec5:
> We should synchronize this list with the actual list used to generate the tiles.

The actual list:

https://github.com/ZeLonewolf/planetiler-scripts/blob/d510a746c64c1c61a10899aba4428dd5ca2924ae/aws/03_render_planet.sh#L18

ZeLonewolf:
That's an outdated location for the languages, although it's probably the same list.

See here:

https://github.com/osmus/tileservice/blob/95185b65ef085383aee4f2ad5eb99bcb568a8a04/renderer/render_once.sh#L51

1ec5:
The OSMUS Tileservice no longer hard-codes a list of supported languages. Instead it exposes every valid language code that’s in OSM. If we want a comprehensive language picker, we’d need to derive the list of languages from taginfo or perhaps a taginfo project file published as part of osmus/tileservice#22. But we can still hard-code a long list if we aren’t concerned about comprehensiveness.

PR #841, opened 2023-03-27
add Noto Sans Cherokee to font stack

claysmalley:
Adds support for Cherokee script.

[Screenshot from 2023-03-26 19-40-53](http://localhost:1776/#map=11.82/35.4714/-83.30396&language=chr%2Cen)

Note: lowercase Cherokee is included here, though technical limitations prevent Unicode characters beyond `U+FFFF` from being rendered. However, lowercase Cherokee is a relatively recent invention that hasn't seen widespread adoption yet, and nearly all Cherokee labels on OSM remain in all caps.

### DIFF
--- a/scripts/fonts.json
+++ b/scripts/fonts.json
@@ -8,6 +8,7 @@
     "Noto Sans Armenian": ["regular", "700"],
     "Noto Sans Bengali": ["regular", "700"],
     "Noto Sans Canadian Aboriginal": ["regular", "700"],
+    "Noto Sans Cherokee": ["regular", "700"],
     "Noto Sans Devanagari": ["regular", "700"],
     "Noto Sans Ethiopic": ["regular", "600"],
     "Noto Sans Georgian": ["regular", "700"],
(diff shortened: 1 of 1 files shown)


ZeLonewolf:
> though technical limitations prevent Unicode characters beyond U+FFFF

What's the limitation?  I've certainly got glyphs beyond that specified in the script definitions.

ZeLonewolf:
Would it make sense to add the language to the language switcher also?

1ec5:
> > though technical limitations prevent Unicode characters beyond U+FFFF
> 
> What's the limitation? I've certainly got glyphs beyond that specified in the script definitions.

maplibre/maplibre-gl-js#2307

1ec5:
> Would it make sense to add the language to the language switcher also?

This is being tracked in #836.

scripts/fonts.json
```diff
@@ -61,6 +62,10 @@
       [6928, 6975],
       [71216, 71231]
     ],
+    "cherokee": [
+      [5024, 5119],
+      [43888, 43967]
```
1ec5, line 67: Thanks for including this range. I’m reading conflicting things about whether a case distinction is meaningful in real-world Cherokee usage, but it shouldn’t be a problem to have this extra glyph PBF available in case lowercase characters ever appear in OSM.

1ec5 (approved):

PR #842, opened 2023-03-29
Change icon stroke width to 1px

claysmalley:
Some existing icons have irregular stroke widths, probably by mistake. This PR changes stroke widths of these icons to 1px as recommended by the style guidelines.

The previously missing 1px stroke around the airplane and "square dot" icons is also added here.

In the hospital and "health cross" icons, the edges of the inner polygons are moved to pixel boundaries for sharper contrast. No stroke is added.


Pengor (changes requested):
Looks like a lot of these SVGs have paths with two `stroke-width` tags now (one under style that overrides the other), I'd suggest removing the one that gets overridden so it's clearer. Otherwise I think this looks good 👍

Pengor (approved):

PR #843, opened 2023-03-29
Add diamond and triangle shield text constraints

claysmalley:
`diamondShield` and `triangleDownShield` currently use the `ellipse` and `southHalfEllipse` text constraints, respectively, which aren't good matches for the underlying shapes. This PR creates new text constraint functions for diamond- and triangle-shaped shields.

Before/after:

Screenshot from 2023-03-29 13-56-27
Screenshot from 2023-03-29 13-55-35


ZeLonewolf (approved):
This looks good.

Let me handle the merge so I can make the appropriate changes in #699 at the same time.

Issue #436, opened 2022-06-18
Bus stops


bgo-eiu:
Blocked by #387 

As busways (#308) and guided busways (#413) have been brought up, it seems appropriate to open a separate bus stop issue as well. (In line with this, bus lanes would probably end up being a different topic too, since these each would require different implementations.)

Based on some conversations on Slack, I think it would be helpful to clarify the role of bus features within a street map. Many of the first reference points that come to mind for print maps of bus transit in America are schematic transit maps, which typically show some degree of simplification of a route or network, and may take on an abstract form or show an overlay on a geographically accurate street grid. However, these types of maps often benefit from de-emphasizing other features in some way, and can be considered specialized. It's not the case that bus passengers are a niche audience, but that these maps only serve a particular use case even for bus passengers. Most people who rely on public transport also use some form of generalized street map, because it contains the details they need to make connections to and from specific stops, context for wayfinding, and the shape of the street network. These features get omitted from schematic transit maps, because that including that information density is incompatible with the goal of highlighting the general shape and connections in a network at a glance. To that end, fulfilling the role of a good street map useful for public transit users seems like a fitting role for Americana.

The better examples of transit maps do make some important and interesting design decisions that I still think would be helpful to draw influence from even if it doesn't make sense to replicate their role directly. They are also some of the easier to find reference points for print maps of bus-related features.  Maps which [distinguish the frequent network from the coverage network](https://humantransit.org/basics/the-case-for-frequency-mapping) can make a big difference in how usable they are for example. Unfortunately bus transit is historically neglected despite quietly constituting a very substantial portion of local travel in American cities, so there are not as many examples of good street maps for bus passengers, or even good general guide maps featuring bus stops. It is common to find maps which make some really unhelpful or questionable decisions in how they show bus stops and/or routes.

For symbology, I quite like the idea of a silhouette of the MUTCD standard bus icon. Many transit agencies simply use the icon at stops in place of a custom logo, and many custom logos are just minimal variants to it, so it has a broad recognizability. It is also common for signposts for park and rides along roadways to feature this icon, so it seems like it would integrate well with the existing route sign aesthetic. A previous older PR took a similar approach at #17 that got me thinking about this.
![image](https://user-images.githubusercontent.com/100172442/174459442-5795ed39-3e24-4436-98bc-85545e5ef40c.png)

[...]

PR #845, opened 2023-03-30
Add bus and rail POIs for stops and stations

Pengor:
## Description

This adds bus stops/stations and rail (including subway and tram) stops/stations. I think there are a lot of directions we could go with the design of the icons for public transportation but a simple head-on view of a bus and a train made sense to me as being both recognizable and consistent with iconography used by various transportation agencies.

I tried to keep the two icons differentiated enough to be recognizable, but the nature of them both being head-on views of a larger vehicle makes them inherently look rather similar.

In the future I could see additional icons for trams/streetcars and subways, but in many places in North America the distinction between light rail, subways, and trams isn't always crystal clear to those unfamiliar with a transit system, so I don't think it's necessary at this point in time. If we eventually have clickable POIs then that info could alternatively be surfaced at that point.

Addresses #436 and #790.

I went with making these appear starting at z16 as the density of bus stops and train stations is approximately equivalent to the density of cafe and bar POIs in downtown areas I referenced while working on this (Seattle and NYC).

### Before (1x display density):
transit_1x_before

### After (1x display density):
transit_1x_after

### Before (2x display density):
transit_2x_before

### After (2x display density):
transit_2x_after

Fun fact: Third Avenue in Seattle (shown in these renders) is supposedly the busiest bus corridor in North America!


icons/poi_rail.svg
claysmalley, line 1: We'll need to find a way to stylistically distinguish stations from stops in both bus and railway features. OsmAnd does this by giving stations (e.g. bus stations, light rail stations, train stations) circular icons, and stops (e.g. bus stops, tram stops) square icons.

I think the bus icon is fine as it is, and we could adopt a similar backgroundless style for tram stops. But I'd like to see bus stations visually distinguished somehow, perhaps with a circle background, and we can apply that same visual element to railway stations.

wmisener: I agree that this would be a good feature (unsurprisingly, since I mentioned a similar tack in https://github.com/ZeLonewolf/openstreetmap-americana/issues/790#issuecomment-1475432421).

wmisener:
Thanks for putting this together! I think the icons are nice: those bus headlights are quite stylish, and while I think down the road it would be best to distinguish light rail, subways, and conventional rail, this icon is nicely ambiguous for the time being IMO.

My first impression from your screenshots is that the bus icons seem quite large. They have a lot of visual weight in my eyes, which I'm not sure matches their importance (and this comes from someone who takes the bus every day to work!). I wonder if it would be better if the bus icons were smaller than 20x20px, so they'd be smaller than the cafes and whatnot. I'm not sure this style has ever done that before so maybe that needs some discussion as a general stylistic choice. I think train stations can stay the same size as they are, though as Clay suggested it might be good to distinguish tram stops from other stations by inscribing the icon in a circle for stations.

scripts/taginfo_template.json
```diff
@@ -471,6 +471,62 @@
...
+      "icon_url": "https://raw.githubusercontent.com/ZeLonewolf/openstreetmap-americana/main/icons/poi_rail.svg"
+    },
+    {
+      "key": "railway",
+      "value": "train_station_entrance",
+      "object_types": ["node"],
+      "description": "Train stations are marked by an icon depicting the front of a train.",
+      "doc_url": "https://openmaptiles.org/schema/#poi",
+      "icon_url": "https://raw.githubusercontent.com/ZeLonewolf/openstreetmap-americana/main/icons/poi_rail.svg"
+    },
```
wmisener, line 521: It looks like you've added `railway=subway_entrance` and `railway=train_station_entrance` to the taginfo file, but I don't think they're being rendered in the PR?

src/layer/poi.js
```diff
@@ -89,7 +103,12 @@ export const poi = {
       ["get", "subclass"],
       ["hospital", ...iconDefs.school.subclasses],
       15,
-      [...iconDefs.bar.subclasses, ...iconDefs.coffee.subclasses],
+      [
+        ...iconDefs.bar.subclasses,
+        ...iconDefs.bus.subclasses,
+        ...iconDefs.coffee.subclasses,
+        ...iconDefs.railway.subclasses,
+      ],
```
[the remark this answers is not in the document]
Pengor: Agreed, updated in ccbd6ebas you suggested (except bus stations don't show up in the tiles until z14) 👍

icons/poi_rail.svg
Pengor: I think this makes sense, I've drawn up some circle-enclosed icons and will commit these alongside changes to render icons by subclass instead of class:

poi_bus_circle
poi_rail_circle

scripts/taginfo_template.json
Pengor: My understanding of the OpenMapTiles schema is that `railway=subway_entrance` translates to to a poi of `class=railway` & `subclass=subway` while `railway=train_station_entrance` translates to a poi of `class=railway` & `subclass=station` but I can double check

Pengor: Upon double checking I think you're right, those two tags end up as `class=entrance` and it's a combination of `railway=station` & `station=subway` that gets translated to `class=railway` & `subclass=subway`

Pengor: Fixed in 759f77b

Pengor:
Updated now that separate POI icons can be rendered by subclass. There are icons for bus and rail stations enclosed in a circle. I think these are sufficiently different looking from one another but happy to incorporate additional feedback!

### Examples
Light rail stations, bus stops, and streetcar stops:
Downtown_Seattle_Transit

Bus station and light rail station:
Bus_and_train_stations

Bus station and bus stops:
Bus_stops_and_station

icons/poi_rail.svg
Pengor: Fixed in 8b7012b

claysmalley:
Looking pretty good altogether.

Just a nitpick—I think the bus and tram icons would look a little better if they were shrunk to match the size of the inner part of the station icons. Otherwise the stops kind of visually dominate the stations, the opposite of what I'd expect.

Pengor:
Shrank the "stop" icons for bus and rail to match the size of the inner parts of the circular icons:

image

image

PR #848, opened 2023-04-02
refactor POI layer to use both class and subclass

claysmalley:
Unblocks #845.


src/layer/poi.js
```diff
@@ -3,50 +3,76 @@ import * as Color from "../constants/color.js";
...
+      hospital: ["hospital"],
+    },
     sprite: "poi_hospital",
     color: Color.poi.infrastructure,
     description: "Hospital",
   },
   medical: {
-    subclasses: ["doctors", "clinic"],
+    classes: {
+      hospital: ["doctors", "clinic"],
```
Pengor, line 32:
```suggestion
      hospital: ["clinic"],
      doctors: ["doctors"],
```
I think this suggestion better matches the OpenMapTiles Schema

Pengor (changes requested):
I tested this out and it definitely worked for setting POI icons by subclass, I just noticed one mismatch with the OpenMapTiles schema. We may also want to have someone else review this who's better with javascript because I'm not entirely sure of how this may affect performance

Pengor (approved):

Issue #466, opened 2022-06-28
Arab Mashreq International Road Network shields


1ec5:
The [Arab Mashreq International Road Network](https://en.wikipedia.org/wiki/Arab_Mashreq_International_Road_Network) traverses several countries in the Middle East, analogous to and overlapping with the Asian Highway (#122) and E road network (#69). Its shields are white against a blue background.

[<img src="https://upload.wikimedia.org/wikipedia/commons/4/4f/Mashreq_road_40.svg" width="64" alt="M40">](https://commons.wikimedia.org/wiki/File:Mashreq_road_40.svg)

So far, three of these routes have been mapped as route relations, tagged `network=Arab Mashreq International Road Network`. Ordinarily we’d balk at the lack of a country prefix or spaces in the `network` value, but these expectations would be relaxed for multinational route networks.

The letter “M” appears below the route number, but as far as I can tell, “M” is typically used as a prefix in road names and numbers, as an integral part of the route number. It appears in `ref` values variously as a [prefix](https://www.openstreetmap.org/relation/3018653) or a [suffix](https://www.openstreetmap.org/relation/8248429). Given the prominent size of the “M”, it might be OK to consider it a simple case of line wrapping.

1ec5:
> Its shields are white against a blue background.

According to the [MUTCD for the Gulf Cooperation Council](https://www.scribd.com/document/595841964/GCC-MUTCD-Final-20140929) revision 4.0 (p. 64 [80]), published in 2014, the shield should be white with a black outline. The blue backdrop comes from the surrounding guide sign, akin to how most shields in the U.S. have green behind them when mounted on a guide sign. The addendum of national deviations doesn’t mention anything about shields.

![Mashreq Route Emblem I 1-1 (may also be shown as a white emblem on a blue-background rectangular sign). Emblems are sized for major urban routes and other routes with speeds over 80 km/h, and may be larger for overhead guide sign applications.](https://github.com/user-attachments/assets/b1753c20-d84c-4e63-a1e6-c1dfbe89b7f2)

Revision 4.0 was published in 2014. The Dutch Wikipedians who uploaded the shields to Wikimedia Commons may have been working from an earlier draft, or they may have mistaken the guide sign backing for an integral part of the shield.

PR #851, opened 2023-04-08
add Saudi Arabia national shields

claysmalley:
Roads in Saudi Arabia are signed solely in Eastern Arabic numerals. This PR contains code to convert digits to Eastern Arabic.

Currently, [only one national route](https://www.openstreetmap.org/relation/11366015) is mapped with `network=SA:national`, and all other routes are missing either the `network=*` tag or a relation altogether. But `SA:national` clearly represents a coherent network, and this could encourage local mappers to complete the tags for other routes.

[Screenshot from 2023-04-08 14-08-58](http://localhost:1776/#map=9.18/24.67/46.752)

Screenshot from 2023-04-08 14-15-36
Screenshot from 2023-04-08 14-16-09


src/js/screen_gfx.js
```diff
@@ -2,7 +2,9 @@
 
 import rgba from "color-rgba";
 
-const fontFamily = '"sans-serif-condensed", "Arial Narrow", sans-serif';
+const naskhFonts =
+  '"Noto Naskh Arabic", KacstNaskh, "Sakkal Majalla", "DecoType Naskh Regular"';
```
claysmalley, line 6: I picked a few Naskh fonts that are available by default on most operating systems. I'm not 100% sure these are all the best choices, but they look close enough to the style as shown on Saudi shields.

src/js/shield.js
```diff
@@ -382,8 +399,15 @@ export function generateShieldCtx(map, id) {
   // Convert numbering systems. Normally alternative numbering systems should be
   // tagged directly in ref=*, but some shields use different numbering systems
   // for aesthetic reasons only.
-  if (routeDef.ref && shieldDef.numberingSystem === "roman") {
-    routeDef.ref = romanizeRef(routeDef.ref);
+  if (routeDef.ref) {
+    switch (shieldDef.numberingSystem) {
+      case "arab":
+        routeDef.ref = latinToArabicDigits(routeDef.ref);
+        break;
```
claysmalley, line 406: `"arab"` was chosen here to match [Unicode CLDR](https://github.com/unicode-org/cldr/blob/main/common/bcp47/number.xml#L15).

src/js/screen_gfx.js
ZeLonewolf: Legibility doesn't look good on my system:

image

Why not use the packaged Americana fontstacks that have Arabic?

claysmalley: Shields are rendered client-side using fonts available to the browser. We would have to re-engineer shields to render text from a PBF fontstack instead of standard web formats.

ZeLonewolf: Oh duh good point.  Don't you need some kind of `<link>` tag to use the Google Noto fonts?

src/js/shield.js
ZeLonewolf: Should it be `arabicizeRef()`? :grin:

src/js/screen_gfx.js
ZeLonewolf: Something like
```
<link href='https://fonts.googleapis.com/css?family=Noto Sans' rel='stylesheet'>
```

claysmalley: Not if it's already installed on the system, as with Android. But downloading and displaying Noto on all systems is an option.

ZeLonewolf: Is there a reason we wouldn't? I assume it's just a one time download into browser cache.

claysmalley: I suppose not. Though we might as well include Noto Sans Armenian as well, if consistency is what we're going for.

ZeLonewolf: I guess we'll need whatever scripts would get rendered into shields.

src/js/shield.js
1ec5: Either that or `arabizeRef()`, yes.

src/js/shield.js
```diff
@@ -382,8 +399,15 @@ export function generateShieldCtx(map, id) {
   // Convert numbering systems. Normally alternative numbering systems should be
   // tagged directly in ref=*, but some shields use different numbering systems
   // for aesthetic reasons only.
-  if (routeDef.ref && shieldDef.numberingSystem === "roman") {
-    routeDef.ref = romanizeRef(routeDef.ref);
+  if (routeDef.ref) {
+    switch (shieldDef.numberingSystem) {
+      case "arab":
+        routeDef.ref = latinToArabicDigits(routeDef.ref);
```
1ec5, line 405: > Roads in Saudi Arabia are signed solely in Eastern Arabic numerals. This PR contains code to convert digits to Eastern Arabic.

In #676, I included a Roman numeral conversion function because the Roman numerals were purely decorative. In plain text (and even some signs), the route numbers are in Western Arabic numerals. By contrast, Eastern Arabic is the local language’s numeral system. Would it be more appropriate to render the number however it’s tagged in `ref`, allowing mappers the flexibility to shunt the transliterated number to a subkey such as `ref:Latn`?

/ref https://github.com/ZeLonewolf/openstreetmap-americana/issues/467#issuecomment-1169234746

src/js/shield_defs.js
```diff
@@ -3687,6 +3687,20 @@ export function loadShields() {
     },
   };
 
+  // Saudi Arabia
+  shields["SA:national"] = {
```
1ec5, line 3691: > Currently, [only one national route](https://www.openstreetmap.org/relation/11366015) is mapped with `network=SA:national`, and all other routes are missing either the `network=*` tag or a relation altogether. But `SA:national` clearly represents a coherent network, and this could encourage local mappers to complete the tags for other routes.

The one `network=SA:national` route is what Saudi Arabia calls a “highway” route, as opposed to a “city road” or “branch road”. All three classifications are part of a nationwide system. Part of Highway 40 has been mapped redundantly as [`network=SA:highway`](https://www.openstreetmap.org/relation/8249853). If we could get a local mapper to confirm their preference, that might also have the effect of encouraging more build-out of these route relations.

src/js/shield.js
```diff
@@ -303,6 +303,23 @@ export function getRouteDef(id) {
...
+    .replaceAll("0", "\u0660")
+    .replaceAll("1", "\u0661")
+    .replaceAll("2", "\u0662")
+    .replaceAll("3", "\u0663")
+    .replaceAll("4", "\u0664")
+    .replaceAll("5", "\u0665")
+    .replaceAll("6", "\u0666")
+    .replaceAll("7", "\u0667")
+    .replaceAll("8", "\u0668")
+    .replaceAll("9", "\u0669");
```
1ec5, line 321: This can be written more succinctly as:

```suggestion
  ref.replaceAll(/[0-9]/g, m => String.fromCharCode(0x0660 + parseInt(m)));
```

src/js/shield.js
ZeLonewolf: - This is a philosophical question ineeded. Off the cuff I would say the ref should have the Eastern Arabic numbers in them, which we would render with the right font.  This PR takes a more resilient approach and papers over the difference at the expense of not exposing what we might agree is an error.

ZeLonewolf: It would be worth asking the folks that implemented the scheme though I'm not sure who to ask.

1ec5: Note that most of Saudi Arabia’s highways are also part of the Arab Mashreq International Road Network, which is signposted in Western Arabic numerals: #466.

icons/shield_sa_national.svg
```diff
@@ -0,0 +1,56 @@
...
...5324 17.65749,6.6516302 17.75338,6.8468004 17.516677,6.9552645 17.343123,7.0368255 17.207302,7.3997479 16.985047,7.508199 16.853309,7.7099414 16.66423,7.9366774 16.625785,8.1185758 16.497557,8.3155226 16.247834,8.1929588 16.019946,8.3246583 15.779121,8.5928635 15.48286,8.2518563 15.197778,8.3180283 15.078781,8.4626691 14.572212,8.6589364 14.655156,8.2633123 14.5005,8.1783003 14.264413,8.2734113 14.3739,7.9603644 14.463139,7.747792 14.561661,7.4824917 14.387665,7.2811401 14.305115,7.0836348 14.526737,6.7783622 14.202765,6.6051108 13.995404,6.4983198 13.783931,6.5730848 13.73848,6.9094551 13.608026,6.9542481 13.706253,7.7807895 13.58612,7.306844 13.498098,7.0060554 13.134601,6.8243725 13.246226,6.4680507 13.259149,6.121132 13.224122,6.1436112 12.98783,5.8459692 12.776669,5.7747115 12.67084,5.5302316 12.496535,5.4115344 12.366755,5.1907696 12.017655,5.0697559 12.000264,4.7897077 11.855838,4.5757747 11.739645,4.2199266 11.364906,4.3218688 10.98494,4.3812401 10.959643,3.7519039 10.549254,3.8545496 9.9411817,3.7641064 9.3036658,3.7462157 8.713742,3.6110542 7.6917476,2.825552 6.7062758,1.9857938 5.5718481,1.3645029 5.2146445,1.0783532 4.7379204,1.1134174 4.3137644,1 Z" id="polyF1S187P4"/>
```
1ec5, line 55: If we simplify this shape a bit, I think it would become more feasible to tweak it to better accommodate the numerals at a slightly larger font size. I don’t have any concrete suggestions, since this is a pretty irregular shape, but maybe play around with it a bit.

ZeLonewolf:
It does seem a bit better with the web fonts but not sure how legible it would be to an Arabic speaker.  Do we think this looks OK or should we try to enlarge or skew the shape a bit?

image

ZeLonewolf:
Shieldtest output:

image

1ec5:
Looks like Latin text in that screenshot has regressed: it’s a serif font instead of a sans-serif font.

ZeLonewolf:
Needs to resolve conflicts.

Issue #391, opened 2022-06-11
Shields of Switzerland


claysmalley:
In Switzerland, motorways have a white-on-red hexagonal shield and main roads have a white-on-blue variable-width rectangular shield.

Most numbered routes in Switzerland are covered by `route=road` relations, but `network=*` tagging is not country-prefixed. There are a number of `network=*` values in Switzerland, with limited documentation. `motorway` clearly corresponds to motorways, but main roads do not have similar relation coverage with consistently tagged `network=*` values. A decision would need to be made on whether this tagging scheme is sufficient, and whether it is appropriate mapper feedback to support it.

<img src="https://user-images.githubusercontent.com/1732117/173170614-3170dc26-0ea4-4adb-8fcd-053426ae07c6.svg" alt="A1Swiss" width="200"/>
<img src="https://user-images.githubusercontent.com/1732117/173170616-02932525-934e-40ce-ada8-7c635c9edc86.svg" alt="Hauptstrasse_Nummer_4" width="100"/>
<img src="https://user-images.githubusercontent.com/1732117/173170615-76774ea6-97d3-4e83-a230-97e7ee3ffcb6.svg" alt="CH-Hinweissignal-Nummerntafel_für_Hauptstrassen" width="125"/>

mlundblad:
It seems like Swiss motorway relations are now tagged as network=CH:motorway.

1ec5:
Switzerland currently has 30 occurrences of `network=motorway`, seven of `network=CH:motorway`, and two of `network=ch:motorway`. This seems far from settled, but maybe the discrepancy can start a discussion in the local community. I hope it goes better than in https://github.com/osm-americana/openstreetmap-americana/discussions/149.

We already support `network=ch:national` for the _Hauptstrassen_/_routes principales_/_strade principali_ (white on blue rectangle). There are currently five occurrences in the database.

https://github.com/osm-americana/openstreetmap-americana/blob/5752e3453f84744491d923771f8fd2992b156020/src/js/shield_defs.js#L4097-L4101

`network=CH:cantonal` is now quite common at 193 occurrences, mostly stemming from [widespread edits](https://www.openstreetmap.org/changeset/173858452) last year. However, I don’t know of any signage associated with cantonal routes, so we probably don’t need to do anything about them.

PR #852, opened 2023-04-08
add Swiss national route shields

claysmalley:
Related to #391. Some national routes have been tagged with `network=ch:national`, though motorways still have `network=motorway` which we are hesitant to implement without a country code.

[Screenshot from 2023-04-08 16-01-15](http://localhost:1776/#map=12.23/46.49841/7.27952)

### DIFF
--- a/doc-img/shield_map_world.svg
+++ b/doc-img/shield_map_world.svg
@@ -147,6 +147,7 @@ See the end of this file for a list of available jurisdictions and their codes. 
 .be,
 .bg,
 .by,
+.ch,
 .cz,
 .ee,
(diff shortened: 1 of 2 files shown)


1ec5 (approved):
Looks like this tag is being used for [_Hauptstrassen_](http://ra.osmsurround.org/searchRelation?name=&relationType=&route=road&ref=&network=ch%3Anational&operator=). I’ve updated [the Wikidata item](https://www.wikidata.org/wiki/Q1300787) with the `network=ch:national` tag so that it shows up in the legend.

PR #856, opened 2023-04-10
Railway halts should be styled like stations

claysmalley:
I overlooked this when reviewing #790, but `railway=halt` is a type of station, not a type of stop. In the US, the _de facto_ accepted practice is to use `railway=halt` for stations with flag-stop service. Other countries use different criteria to distinguish `railway=halt` from `railway=station`, but regardless, either can have one or many `railway=stop` (not rendered) associated with them.

Before/after:

[Screenshot from 2023-04-10 15-31-44 Screenshot from 2023-04-10 15-30-39](http://localhost:1776/#map=12.19/41.73421/-87.60683)

[Screenshot from 2023-04-10 15-41-20 Screenshot from 2023-04-10 15-41-49](http://localhost:1776/#map=15.2/41.740686/-87.600464)

### DIFF
--- a/src/layer/poi.js
+++ b/src/layer/poi.js
@@ -62,7 +62,7 @@ var iconDefs = {
   },
   railway_station: {
     classes: {
-      railway: ["station", "subway"],
+      railway: ["station", "halt", "subway"],
     },
     sprite: "poi_rail_circle",
     color: Color.poi.transport,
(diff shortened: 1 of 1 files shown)


Pengor (approved):
Fine by me, unfortunate this tag isn't used consistently in different places, but oh well 🤷‍♂️

Issue #105, opened 2022-01-27
Aboriginal land boundaries


1ec5:
Native American reservations and other tribal lands are an integral part of the system of boundaries in the United States. Plenty have already been mapped as [`boundary=aboriginal_lands`](https://wiki.openstreetmap.org/wiki/Tag:boundary%3Daboriginal_lands) relations. Many are large enough to be displayed at mid- and low zoom levels.

We should figure out a way to depict these boundaries that won’t be confused with non-indigenous state and local boundaries but still communicates their importance as delimiters of sovereignty. There was an extensive discussion in gravitystorm/openstreetmap-carto#3520 on this topic; some of the considerations may also apply to this style, although so far we’re rendering fewer things that constrain our options compared to openstreetmap-carto at the time.

For reference, most of the maps in my collection fill Indian reservations as they would fill parklands and military reservations, but this treatment doesn’t seem to be as intuitive as a boundary line, which is how Florida DOT and Best Western maps depict them:

<img src="https://user-images.githubusercontent.com/1231218/151273180-85622137-43da-4660-9ba3-cd52f3f2aa7e.jpg" width="250" alt="Florida DOT 2005"> <img src="https://user-images.githubusercontent.com/1231218/151273356-4ef0967b-de66-4206-af78-e90b7911ca93.jpg" width="250" alt="Nevada DOT"> <img src="https://user-images.githubusercontent.com/1231218/151273491-131aba7d-f715-4720-a47d-1eed876f1a5b.jpg" width="250" alt="Best Western Nevada">

[...]

PR #857, opened 2023-04-11
Add museum icon and brown "attraction" POI color

### LINKED ISSUE #820: Render museums
Museums are often important landmarks in big cities and small towns alike, and it would be fitting for them to be rendered fairly prominently. Would also make progress on rendering POIs in general (#692).

Museums present some difficulty in choosing a representative icon, as they can exhibit a wide variety of material, from art to archaeology to scientific demonstrations to items of local historical import and everything in between.

In https://github.com/ZeLonewolf/openstreetmap-americana/issues/794#issuecomment-1435095032, I suggested rendering museums as a columned neoclassical building, evoking the large institutions common in big cities. It could perhaps be similar to my original suggestion for town halls in #807, but without a flag. But a number of ideas could plausibly work. One thing I would like to shy away from is rendering museums as a building with an "M" in it, as this would violate our assumption of no default language.

Another problem arises when considering a color for the icon. The [contributing guidelines](https://github.com/ZeLonewolf/openstreetmap-americana/blob/34f541a978a6ffd3277b55bc192d908cf6349647/CONTRIBUTING.md) specify no color yet for "attractions", which is probably the best category for museums. Some shade of brown was suggested in https://github.com/ZeLonewolf/openstreetmap-americana/issues/128#issuecomment-1159801218. I think that would be appropriate.

wmisener:
Fixes #820, and makes progress on #692 (well, actually museums aren't explicitly listed in #692, but they probably should be). I've used a modified version of the columned, neoclassical building I originally proposed for town halls in #794, but without the flag:

poi_museum

This icon specifically evokes the large, established institutions of big cities, but I think the symbol is recognizable enough to represent museums of diverse sizes and settings.

This is the first POI in the "attraction" category, so the PR also adds and documents a new POI color, "attraction brown". Brown was suggested in #128, and it seems reasonable for attractions based on American usage, such as MUTCD signage. As @1ec5 pointed out on Slack, care may be needed when indigenous land boundaries are added (#105) to make sure the colors in the style clearly communicate their intended meaning. I've reused the brown color defined for highways, hex #693F23, but would be amenable to alterations if someone has a strong opinion.

Samples:
National Mall ([localhost](http://localhost:1776/#map=15.41/38.88993/-77.025717)):
Screen Shot 2023-04-11 at 8 25 55 AM

Woods Hole, MA ([localhost](http://localhost:1776/#map=16.75/41.523511/-70.667962)):
Screen Shot 2023-04-11 at 8 28 57 AM

San Francisco, CA ([localhost](http://localhost:1776/#map=16.36/37.779986/-122.417826)):
Screen Shot 2023-04-11 at 8 31 16 AM

Chicago, IL ([localhost](http://localhost:1776/#map=15.11/41.864862/-87.615266)):
Screen Shot 2023-04-11 at 8 42 40 AM
Note that the [Field Museum](https://www.openstreetmap.org/way/24825537) (`tourism=museum`) is rendered in this PR, but the [Shedd Aquarium](https://www.openstreetmap.org/way/24825568) (`tourism=aquarium` OMT class=aquarium, subclass=aquarium) and the [Adler Planetarium](https://www.openstreetmap.org/way/24825591) (`amenity=planetarium`, which doesn't appear to be in OMT, although I think this might be a tagging error) are not.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -376,7 +376,7 @@ For consistency, POI icons use the following color palette:
 | Infrastructure         | Pantone 294     | <img src="doc-img/pantone_294.svg" height=18 width=50 /> Blue               | 0 63 135    | #003f87     |
 | Consumer               | UTexas Orange   | <img src="doc-img/texas_orange.svg" height=18 width=50 /> Orange            | 191 87 0    | #bf5700     |
 | Outdoor                |                 | TBD (green?)                                                                |             |             |
-| Attraction             |                 | TBD (brown?)                                                                |             |             |
+| Attraction             | Pantone 469     | <img src="doc-img/pantone_469.svg" height=18 width=50 /> Brown              | 105 63 35   | #693f23     |
 | Airport                | Medium Purple C | <img src="doc-img/pantone_medium_purple_c.svg" height=18 width=50 /> Purple | 78 0 142    | #4e008e     |
 | Transport              | Pantone 234 C   | <img src="doc-img/pantone_234_c.svg" height=18 width=50 /> Mauve            | 162 0 103   | #a20067     |
 | Knockout               |                 | <img src="doc-img/background.svg" height=18 width=50 /> Lt Grayish Orange   | 249 245 240 | #f9f5f0     |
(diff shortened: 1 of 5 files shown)


ZeLonewolf (approved):
Visual inspection and code looks good to me.  I'm glad we could re-use the work on this.  Like to get someone on one of the fancier displays to make sure it looks good at all resolutions.

claysmalley (approved):
Looks good to me.

PR #859, opened 2023-04-13
hide bus and tram stop labels until very high zoom

claysmalley:
Cutting down on label clutter a little. North American bus stop names are typically unsigned and derived from the nearest two intersecting streets, so they're often quite long and add minimal semantic value.

test location: http://localhost:1776/#map=15.81/37.793137/-122.397591&language=mul

Screenshot from 2023-04-12 19-23-57
Screenshot from 2023-04-12 19-24-22
Screenshot from 2023-04-12 19-24-43


ZeLonewolf (approved):
Thanks, this is a definite improvement.  I would even advocate to take it one step further and not show bus stop icons until z16 (based solely on how this renders in NYC).

PR #864, opened 2023-05-09
Use "NSP" as the ref for Niagara Scenic Parkway.

ZLima12:
Currently, the road is tagged as being part of the `US:NY:Parkway` network, due to the similarity between its shield and other green parkway shields. While there still are differences, including the first letter of each word not being emphasized, the shield should still probably be rendered the same way as others in the network.

Should the road be retagged under its own network, this should be changed.

image

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -4285,6 +4285,7 @@ export function loadShields() {
     "Hutchinson River Parkway": "HRP",
     "Korean War Veterans Parkway": "KWVP",
     "Mosholu Parkway": "MP",
+    "Niagara Scenic Parkway": "NSP",
     "Pelham Parkway": "PP",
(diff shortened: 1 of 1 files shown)


ZLima12:
As can be seen [here](https://zelonewolf.github.io/openstreetmap-americana/#map=13.97/43.26119/-79.02163), the road is currently rendered with an empty shield.

ZeLonewolf:
Looks like it's different but similar to the rest of the NY series.  This seems like a reasonable approach to me and certainly better than a blank shield.

image

ZeLonewolf:
I've inserted a screen shot of the new rendering in the original PR comment, for easier review.

claysmalley (approved):
Fine by me.

ZeLonewolf:
Thanks!

PR #868, opened 2023-06-06
Fix failing github pages and view PR previews

### LINKED ISSUE #871: Legend fails to load
Errors when opening the legend.  Introduced in #699.

image

ZeLonewolf:
Fixes #871 

image

This fixes the github build scripts after the shield library PR, and enables a preview of a PR before deploying it.

This works by uploading the built deployment and transferring it via FTP to the tile server. It will post a comment with the preview links. I also removed (at least for now) the Chinese font stacks as it was taking way too long to run the build and upload. We can revisit that when we actually implement Han unification.

This also fixes the fact that legends were broken by #699, which is blocking other PRs :grimacing: 

This doesn't address other suggestions, such as using `tsc` to pre-compile typescript for better performance.

### DIFF
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -30,7 +30,7 @@ jobs:
           npm run style
           cp src/configs/config.aws.js src/config.js
       - name: Deploy via FTP
-        uses: genietim/ftp-action@releases/v4
+        uses: genietim/ftp-action@v4
         with:
           host: ${{ secrets.FTP_SERVER }}
           user: ${{ secrets.FTP_USERNAME }}
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -48,6 +48,7 @@ jobs:
             * [Map](https://preview.ourmap.us/pr/${{ github.event.pull_request.number }}/)
             * [Shield Test](https://preview.ourmap.us/pr/${{ github.event.pull_request.number }}/shieldtest.html)
             * [style.json](https://preview.ourmap.us/pr/${{ github.event.pull_request.number }}/style.json)
+            * [shields.json](https://preview.ourmap.us/pr/${{ github.event.pull_request.number }}/shields.json)
             * [taginfo.json](https://preview.ourmap.us/pr/${{ github.event.pull_request.number }}/taginfo.json)
 
             Sprite Sheets:
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -50,8 +50,8 @@ jobs:
             PR Preview:
             * [Map](https://preview.ourmap.us/pr/${{ github.event.pull_request.number }}/)
             * [Shield Test](https://preview.ourmap.us/pr/${{ github.event.pull_request.number }}/shieldtest.html)
-            * [Sprites 1x](https://preview.ourmap.us/pr/${{ github.event.pull_request.number }}/sprites/sprite.png)
-            * [Sprites 2x](https://preview.ourmap.us/pr/${{ github.event.pull_request.number }}/sprites/sprite@2x.png)
             * [style.json](https://preview.ourmap.us/pr/${{ github.event.pull_request.number }}/style.json)
             * [taginfo.json](https://preview.ourmap.us/pr/${{ github.event.pull_request.number }}/taginfo.json)
+            * ![Sprites 1x][https://preview.ourmap.us/pr/${{ github.event.pull_request.number }}/sprites/sprite.png]
+            * ![Sprites 2x][https://preview.ourmap.us/pr/${{ github.event.pull_request.number }}/sprites/sprite@2x.png]
           comment_tag: pr_preview
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -2,17 +2,19 @@ name: Deploy PR Preview
 on:
   pull_request:
     branches: [main]
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - closed
 permissions:
   contents: read
   pages: write
   pull-requests: write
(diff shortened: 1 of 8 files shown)


jleedev:
Huh, interesting.

Would be nice if the CI could produce a live preview.

Might also be good to make liberal use of [npm:debug](https://npmjs.com/package/debug) to make it extremely obvious what steps run and what outputs happen.

ZeLonewolf:
It looks like there are a few options do this from googling around.  I agree this would be a good idea.

ZeLonewolf:
This one seems the most promising:
https://github.com/marketplace/actions/deploy-pr-preview

.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,61 @@
...
+            PR Preview:
+            * [Map](https://preview.ourmap.us/pr/${{ github.event.pull_request.number }}/)
+            * [Shield Test](https://preview.ourmap.us/pr/${{ github.event.pull_request.number }}/shieldtest.html)
+            * [style.json](https://preview.ourmap.us/pr/${{ github.event.pull_request.number }}/style.json)
+            * [taginfo.json](https://preview.ourmap.us/pr/${{ github.event.pull_request.number }}/taginfo.json)
+
+            Sprite Sheets:
+
+            ![Sprites 1x](https://preview.ourmap.us/pr/${{ github.event.pull_request.number }}/sprites/sprite.png)
+            ![Sprites 2x](https://preview.ourmap.us/pr/${{ github.event.pull_request.number }}/sprites/sprite@2x.png)
```
1ec5, line 60: When using a 2× display, I’d find it more useful to view the 2× spritesheet at the same width as the 1× spritesheet. Otherwise, this just tells me at a glance that the icons can be twice as large in a spritesheet twice as large.

GitHub supports HTML `` syntax but lamentably sanitizes the `srcset` attribute, so we’d have to figure out a `width` to specify. Is it possible to know the width ahead of time?

scripts/fonts.json
```diff
@@ -31,13 +31,7 @@
     "Noto Sans Telugu": ["regular", "700"],
     "Noto Sans Thai": ["regular", "700"]
   },
-  "bundle-font-stacks": {
-    "Noto Sans HK": ["regular", "700"],
-    "Noto Sans JP": ["regular", "700"],
-    "Noto Sans KR": ["regular", "700"],
-    "Noto Sans SC": ["regular", "700"],
-    "Noto Sans TC": ["regular", "700"]
-  },
```
1ec5, line 40: I assume this was removed for performance reasons. We should ticket out any regression in rendering CJK.

1ec5 (commented):
Excited to get to a point of being able to visually review a PR without having to build it locally!

scripts/fonts.json
ZeLonewolf: Yes, and they're not actually used yet.

.github/workflows/deploy-preview.yml
ZeLonewolf: We could pick a width, but it'd be wrong on some set of displays. On mobile they are the same width.

scripts/fonts.json
ZeLonewolf: Ticketed in #872

.github/workflows/deploy-preview.yml
ZeLonewolf: Could we use a width percentage somehow?

ZeLonewolf: I tried some various CSS hacks to get this to happen and failed.

claysmalley:
`npm run status_map` seems to be broken:

shield_map_world

.github/workflows/deploy-preview.yml
1ec5: GitHub recently [added support for “responsive” images](https://github.blog/changelog/2022-08-15-specify-theme-context-for-images-in-markdown-ga/), but only based on light mode versus dark mode. If I try something like:

```html
<picture>
<source srcset="https://preview.ourmap.us/pr/868/sprites/sprite@2x.png" media="(min-resolution: 2dppx)">
<img src="https://preview.ourmap.us/pr/868/sprites/sprite.png" alt="">
</picture>
```

the `<source>` has no effect. There appears to be some Web Components magic going on that overrides the image with custom image switching that doesn’t understand resolutions.

It’s OK, not a blocker. I just open the image in a new tab and zoom out until it looks the same size.

ZeLonewolf: Feel free to edit the auto comment to experiment.

1ec5: Yeah, I threw everything I had at it, but HTML that should produce the right result does not, apparently because of some Web Components–based polyfill in JavaScript that overrides the correct image with the incorrect image based on the theme.

ZeLonewolf: Resolving conversation as CANTFIX :grinning:

ZeLonewolf:
Fixed in 44731dc

Is is worthwhile to have the status map in the preview auto-message?

claysmalley:
> Is is worthwhile to have the status map in the preview auto-message?

Probably unnecessary. Github's diff view has good SVG viewing features.

ZeLonewolf:
Well, FTP has clearly stopped working.

ZeLonewolf:
This is almost done from my end - I need to do some SSL and DNS updates and we'll be good to go.

ZeLonewolf:
This is ready to go :+1:

claysmalley (approved):
Looks good to me.

shieldlib/src/shield_renderer.ts
```diff
@@ -103,6 +111,17 @@ class AbstractShieldRenderer {
     return this;
   }
 
+  public onShieldDefLoad(
```
1ec5, line 114: This requires a new maplibre-shield-generator release on NPM.

PR #855, opened 2023-04-10
Add aerialway station icon

claysmalley:
Is it an enclosed gondola? Is it a chair lift? No, it's a delightfully ambiguous aerial tramway car!

Note that drag lifts also get this icon, as OpenMapTiles only differentiates aerialway subclasses on lines, not stations.

[Screenshot from 2023-04-10 11-02-55](http://localhost:1776/#map=14.23/39.48126/-106.05888)
[Screenshot from 2023-04-10 11-06-05](http://localhost:1776/#map=17.17/19.345307/-99.06362)
[Screenshot from 2023-04-10 11-11-35](http://localhost:1776/#map=15.66/-16.490002/-68.141044)
[Screenshot from 2023-04-10 11-12-59](http://localhost:1776/#map=15.15/47.150657/10.582845)

PR #869, opened 2023-06-06
Aerialway station in taginfo

wmisener:
Adds `aerialway=station` as a supported tagging to the project taginfo file. This was omitted in #855, which originally added rendering aerialway stations.


claysmalley (approved):
My bad :face_with_head_bandage: 

Thanks for catching this!

ZeLonewolf:
Need to resolve #868 before merging, otherwise map will be down :face_with_head_bandage:

PR #870, opened 2023-06-10
Distinct college/university icon

### LINKED ISSUE #694: Render universities/colleges distinctly from schools

wmisener:
Fixes #694. Changes the rendering for `amenity=college` and `amenity=university` (in OMT, `class=college` + `subclass=college`/`university`) to a mortarboard cap-style icon: poi_college. Previously, these were rendered as a generic school icon poi_school; this schoolhouse icon is now reserved for `amenity=school` and `amenity=kindergarten`.

I think the "higher ed"/"schools for adults" vs "schools for children" is a useful distinction for the map to make. Meanwhile, the distinction between an OSM `college` and `university` is a bit fuzzy in practice IMO, so this PR renders both tags the same. In #694, while there were multiple potential icons in various US map samples, the consensus was that a mortarboard cap was the most recognizable symbol of institutions of higher education. The new legend entry is "College or university", whereas the schoolhouse icon retains "School". I think this is not too confusing, but I'm open to suggestions.

The icon itself is a modified version of the maki symbol. I'm open to modifying the dimensions/length ratios/style/etc.

I've added rendering to begin at z14, one zoom level out compared to most other POIs, since colleges and universities can be quite important local destinations (I think that eventually starting zoom levels for POIs might need to be reconsidered more holistically, but that's out of scope for this PR). Rendering at z<14 is not currently possible, as non-transport POIs only appear in the tiles at z14 (https://github.com/openmaptiles/openmaptiles/issues/387). In the future, it might also be good to render these objects as areas as well, but I also leave that out for now.

Samples:
UCLA z14 ([localhost link](http://localhost:1776/#map=14/34.07262/-118.44771&language=)):
Screen Shot 2023-06-10 at 9 37 51 AM
Note that the Anderson School of Management [is tagged as](https://www.openstreetmap.org/way/184488683) `amenity=university`, which may or may not be a tagging error.

UCLA z15 ([localhost link](http://localhost:1776/#map=15/34.07052/-118.44835&language=)):
Screen Shot 2023-06-10 at 9 39 36 AM

Indiana University Bloomington ([localhost link](http://localhost:1776/#map=14.02/39.17449/-86.51178&language=)):
Screen Shot 2023-06-10 at 9 44 16 AM
The icon renders near the center of the university polygon. This may or may not be equivalent to the main activity center of the campus. This problem could be helped, though not really entirely solved, by rendering the area as a boundary.

City College of San Francisco ([localhost link](http://localhost:1776/#map=15.28/37.725185/-122.451305&language=)):
Screen Shot 2023-06-10 at 9 58 59 AM
This is a two-year community college, the predominant use of `amenity=college` I anecdotally observe in the US. I see these institutions are more close to a university than they are to e.g. a high school, so I think it makes more sense to render them together with universities.

Boston/Cambridge, MA ([localhost link](http://localhost:1776/#map=14.05/42.35303/-71.09176&language=)):
Screen Shot 2023-06-10 at 9 53 38 AM

Cambridge, UK ([localhost link](http://localhost:1776/#map=14.02/52.20467/0.1154&language=)):
Screen Shot 2023-06-10 at 9 54 48 AM
Again, these seem like tagging errors to me, though the UK university/college system is incredibly convoluted so I wouldn't go in and change the tags myself. But this seems at least an improvement over rendering them the same as elementary schools.


icons/poi_college.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="20" height="20" xmlns="http://www.w3.org/2000/svg">
+ <path d="M 10,4 2.5,7.5 4,8.166 C 4,10.143784 4,13.648377 4,16 H 5.5 V 8.9 L 10,11.000001 17.5,7.5 Z M 14.5,9.9 10,12.000001 C 9.1960909,11.604612 7.5793651,10.855791 6.5186269,10.367479 6.503761,10.408608 6.4919116,14.590274 6.4919116,14.590274 7.4564094,15.070546 8.7760134,15.516238 10,15.5 12.025778,15.47312 14.050277,14.429627 14.5,14 Z" style="stroke:#f9f5f0;stroke-opacity:1;paint-order:stroke fill markers;stroke-width:2;stroke-dasharray:none"/>
```
Pengor, line 2: This looks good to me, I'd just tweak it a bit:
* Change stroke color to white
* Height (and realign path due to that change)
* Remove unnecessary styling and cleanup what remains

```suggestion
<svg width="20" height="14" xmlns="http://www.w3.org/2000/svg">
 <path d="m10 1.1-7.5 3.5 1.5 0.666v7.734h1.5v-7l4.5 2.1 7.5-3.5zm4.5 5.9-4.5 2.1c-0.80391-0.39539-2.4206-1.1442-3.4814-1.6325-0.014866 0.041129-0.026715 4.2228-0.026715 4.2228 0.9645 0.48027 2.2841 0.92596 3.5081 0.90973 2.0258-0.02688 4.0503-1.0704 4.5-1.5z" stroke="#fff" stroke-width="2" paint-order="stroke fill"/>
```

scripts/taginfo_template.json
```diff
@@ -403,17 +403,17 @@
       "key": "amenity",
       "value": "college",
       "object_types": ["node", "area"],
-      "description": "Schools are marked by an icon representing a schoolhouse with a triangular flag raised above it.",
+      "description": "Colleges are marked by an icon representing a mortarboard cap.",
       "doc_url": "https://openmaptiles.org/schema/#poi",
-      "icon_url": "https://raw.githubusercontent.com/ZeLonewolf/openstreetmap-americana/main/icons/poi_school.svg"
+      "icon_url": "https://raw.githubusercontent.com/ZeLonewolf/openstreetmap-americana/main/icons/poi_mortarboard.svg"
```
Pengor, line 408: Icon name should match added icon
```suggestion
      "icon_url": "https://raw.githubusercontent.com/ZeLonewolf/openstreetmap-americana/main/icons/poi_college.svg"
```

scripts/taginfo_template.json
```diff
@@ -403,17 +403,17 @@
...
     },
     {
       "key": "amenity",
       "value": "university",
       "object_types": ["node", "area"],
-      "description": "Schools are marked by an icon representing a schoolhouse with a triangular flag raised above it.",
+      "description": "Colleges are marked by an icon representing a mortarboard cap.",
       "doc_url": "https://openmaptiles.org/schema/#poi",
-      "icon_url": "https://raw.githubusercontent.com/ZeLonewolf/openstreetmap-americana/main/icons/poi_school.svg"
+      "icon_url": "https://raw.githubusercontent.com/ZeLonewolf/openstreetmap-americana/main/icons/poi_mortarboard.svg"
```
Pengor, line 416: Icon name should match added icon
```suggestion
      "icon_url": "https://raw.githubusercontent.com/ZeLonewolf/openstreetmap-americana/main/icons/poi_college.svg"
```

Pengor (changes requested):
Nice, I think this looks great! I'd at least fix the paths in `taginfo_template.json` and the stroke color in the SVG file, but otherwise could be merged 👍

icons/poi_college.svg
wmisener: My read of the [contributing guidelines](https://github.com/ZeLonewolf/openstreetmap-americana/blob/main/CONTRIBUTING.md#color-scheme-1) was that the borders were supposed to be this shade of off-white, I guess to match the background?

Also not sure if the dimensions are supposed to be 20x20 even if the icon doesn't need that much room, or if the dimensions should be such that the icon just fits (or if the svg dimensions even matter).

ZeLonewolf:
I think the original filename of `poi_mortarboard` is better. As the number of icons grow, they'll be easier to understand in the code if they're closer in description to what's in the icon rather than what the icon represents.

icons/poi_college.svg
Pengor: As for dimensions, I believe most of our icons that don't fill the full 20x20 have reduced the SVG size to the fewest pixels necessary (there may be exceptions and we should probably clarify our position on that).

wmisener: Certainly none of the ones I've added have been (e.g. school, police badge, museum), clarification welcome.

Pengor (approved):

Pengor:
Sorry for the slowness on getting back to this. After the discussions on here and slack I don't think there's anything holding this design up from being merged

icons/poi_college.svg
wmisener: Marked this as resolved. If there's consensus for it, in the future the icon dimensions could easily be shrunk from 20x20 to just fitting.

ZeLonewolf:
Looks great, thanks!

PR #878, opened 2023-06-18
Migrate fonts to fontstack66 repo

ZeLonewolf:
This PR removes fontstack building from the style as it has been migrated to the [fontstack66](https://github.com/osm-americana/fontstack66) repository as of osm-americana/fontstack66#1.

This reduces Americana's build time to near-zero and drastically reduces the size of the build from 201mb to 3.4mb. This will improve the developer experience for most contributors at the expense of making things slightly less for convenient for a developer that wants to add a new font to the map.

I did a search for font-related issues in the repository, and they're all related to shield fonts, which is a separate issue.  Shield fonts are generated from the browser, while these fontstacks only apply to things like area, linestring and POI labels.


src/americana.js
```diff
@@ -32,7 +32,7 @@ export function buildStyle() {
   return Style.build(
     config.OPENMAPTILES_URL,
     `${baseUrl}/sprites/sprite`,
-    `${baseUrl}/fonts/{fontstack}/{range}.pbf`,
+    "https://osm-americana.github.io/fontstack66/{fontstack}/{range}.pbf",
```
claysmalley, line 35: This could benefit from having a config override to make it easier to test changes to the fontstack locally.

ZeLonewolf: Seems reasonable, just implemented that in https://github.com/ZeLonewolf/openstreetmap-americana/pull/878/commits/0568c400d0a46d6e42101085982b961efbe2406c and https://github.com/ZeLonewolf/openstreetmap-americana/pull/878/commits/e1c519e88731653691ef513f17ad158e5b048af1.

claysmalley (approved):

PR #879, opened 2023-06-19
Allow PR branches to run workflows with repo secrets

ZeLonewolf:
Unblocks #870 

This PR implements a deliberate security loosening as described in this [article](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/). It allows PRs from forks to run workflows with repo secrets, specifically the one that runs the PR preview code, which posts to AWS s3. Those secrets are account credentials to a limited-access account that only has permission to write to that specific s3 bucket and within a specific sub-folder. Given that we are reviewing PRs regularly and generally know our contributors, I assess the security risk of enabling the pull_request_target hook to be low.


1ec5:
By default, GitHub requires a maintainer to press a button to allow any hook, check, or action to run on a PR by a new contributor. That setting is still enabled, correct?

.github/workflows/deploy-preview.yml
```diff
@@ -1,6 +1,6 @@
 name: Deploy PR Preview
 on:
-  pull_request:
+  pull_request_target:
```
1ec5, line 3: [This article](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) seems to suggest that `pull_request_target` is inappropriate for workflows that build the PR branch’s contents, since it has access to the main repository and its secrets. (I don’t know what those secrets are; they could be trivial or they could leave you vulnerable.) The article suggests a different trigger, `workflow_run`, for building untrusted code using the `pull_request` trigger as a trampoline.

ZeLonewolf: Using the `workflow_run` trigger would not work in this case because it doesn't have access to the PR number, which is required for the preview upload.

ZeLonewolf: To answer the other question, these are the only secrets in the repo:

image

I have access controls implemented on the AWS side to restrict these credentials to only be able to write to a specific sub-folder in that one s3 button. The risk is a rogue PR that modifies the workflow to somehow log the secrets. But, we'd see the PR and what's in it, and the only thing they'd be able to do is (briefly) upload files to the bucket.

1ec5: The article is clear that, aside from those secrets, the action has access to this repository’s read-write token, which can be used to write to the main branch. We would at least need to set `persist-credentials` to `false`. But if the reason for this unsafe trigger is just to get the PR number, then the article provides a safer approach; see the `ReceivePR.yml` code snippet, which saves the PR number as an artifact.

1ec5 (commented):
I’d appreciate it if others who are more familiar with GitHub Actions or s3 could also take a look at this PR, just in case. I’ve seen enough inexplicable random PRs on the repositories that I maintain that I’m not inclined to prop the barn door ajar.

.github/workflows/deploy-preview.yml
ZeLonewolf: Passing the PR number should work in theory

ZeLonewolf: Ah, I see how this works.  The `workflow_run` job only runs in the main branch, not in the PR branch. So we will not able able to test this until merged unfortunately, but it will be secure.

ZeLonewolf:
In theory this transitions to the `workflow_run` trigger, but since that action only runs in the main branch, there's no way to actually test its functionality before merge.

.github/workflows/build-preview.yml
```diff
@@ -0,0 +1,36 @@
+name: Build PR Preview
+on:
+  pull_request:
+    branches: [main]
+    types:
+      - opened
+      - reopened
+      - synchronize
```
akadouri, line 8:
```suggestion
      - synchronize
    - workflow_dispatch
```
If you add `workflow_dispatch` here you'll get a little dropdown on the actions page which lets you run the action on command. If you wanted to test that way. https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow

You could also remove the`branches` filter, that should run it too.

akadouri: Commented on the first action but this applies to whatever you want to test.

ZeLonewolf:
Upon further research, it appears that the github variables I was using were not quite right.  [This article](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows) seems to have a more current syntax which I plan to try today.

PR #880, opened 2023-06-19
Add shield for US:NY:Truck

jleedev:
Five of these on the map. Appears to be the only such bannered routes.


jleedev:
Delightful!

NY 19 TRK near Brockport

ZeLonewolf (approved):
So much easier to review now!

image

PR #881, opened 2023-06-19
Move PR preview to Checks tab and add statistics

ZeLonewolf:
This PR adds statistics to the checks tab, which will show the change in layer size.
Additionally, this moves the PR Preview comment into the checks tab, to avoid email spamming.

PR #747, opened 2023-01-30
Add script for extracting style statistics

ZeLonewolf:
This PR adds a script that offers a set of utilities for analyzing layers. Additionally, I have renamed a number of layers in order to logically group them for the purpose of statistics generation (since the choice of layer id is essentially arbitrary).

Usage examples:
* `npm run stats -- -a -s`
* `npm run stats -- -c`
* `npm run stats -- -pp -pl background`
* `npm run stats -- -pg park`
* `npm run stats -- --help`

```
Usage: stats [options]

Options:
  -a, --all-layers                      summary layer stats
  -c, --layer-count                     count number of layers
  -s, --layer-size                      size of all layers
  -l, --layer <layer id>                stats about one layer
  -loc, --locales <locale1 locale2...>  language codes (default: ["mul"])
  -pp, --pretty                         pretty-print JSON output
  -pg, --print-group <group prefix>     print a list of the layers in a group
  -pl, --print-layer <layer id>         print the JSON of a layer
  -h, --help                            display help for command
```

PR #882, opened 2023-06-20
Add JSON output to stats script and use layer source

### LINKED ISSUE #777: Style statistics based on source layer
This is a follow-on from #747 to document additional work improve the stats generator.

> We could group the layers by source layer to get pretty similar statistics, modulo some layer organization weirdness in OpenMapTiles.

_Originally posted by @1ec5 in https://github.com/ZeLonewolf/openstreetmap-americana/pull/747#discussion_r1100926373_

ZeLonewolf:
Fixes #777

This PR overhauls the stats script.

First, it splits out all of the layer extraction commands into a separate "util" script since it's thematically different from stats generation.

Next, this changes the layer grouping to use the underlying source layer rather than thematic groups.  This will improve our ability to assess how we're using OpenMapTiles layers.

Lastly, it adds a JSON output option, which is a prerequisite to showing stats deltas as suggested in #881.


scripts/stats.js
```diff
@@ -62,32 +63,60 @@ for (let i = 0; i < layers.length; i++) {
...
+  } else {
+    group.forEach((lyr) => console.log(lyr));
+  }
 }
 
 if (opts.printLayer) {
   if (opts.pretty) {
-    console.log(JSON.stringify(layerMap.get(opts.printLayer), null, 2));
+    process.stdout.write(
+      JSON.stringify(layerMap.get(opts.printLayer), null, 2) + "\n"
```
1ec5, line 117: This output is more like JSON-LD. Anyone using this output programmatically will need to extract an arbitrary line to work with it. Let’s save these bits in a single variable scoped to the whole function and spit it out at the end.

ZeLonewolf:
`node scripts/stats -j -pp`
Returns:
```
{
  "layerCount": 349,
  "styleSize": 1005817,
  "layerGroup": {
    "background": {
      "size": 122,
      "count": 1
    },
    "landuse": {
      "size": 309,
      "count": 1
    },
    "park": {
      "size": 2365,
      "count": 3
    },
    "aeroway": {
      "size": 2059,
      "count": 7
    },
    "landcover": {
      "size": 361,
      "count": 2
    },
    "boundary": {
      "size": 13691,
      "count": 9
    },
    "water": {
      "size": 912,
      "count": 3
    },
    "waterway": {
      "size": 3543,
      "count": 3
    },
    "transportation": {
      "size": 920221,
      "count": 299
    },
    "building": {
      "size": 292,
      "count": 1
    },
    "transportation_name": {
      "size": 8838,
      "count": 3
    },
    "poi": {
      "size": 10094,
      "count": 2
    },
    "water_name": {
      "size": 4647,
      "count": 2
    },
    "aerodrome_label": {
      "size": 5951,
      "count": 4
    },
    "place": {
      "size": 32062,
      "count": 9
    }
  }
```

scripts/util.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+}
+
+let outputObj;
+
+if (opts.printGroup) {
+  let group = layerGroupMap.get(opts.printGroup);
+  let layers = [];
+  group.forEach((lyr) => layers.push(lyr));
+  outputObj = layers;
```
1ec5, line 50: This seems to be merely copying `group` to `layers`.

```suggestion
  outputObj = layerGroupMap.get(opts.printGroup);
```

scripts/util.js
```diff
@@ -0,0 +1,62 @@
...
+if (opts.printLayer) {
+  console.log(layerMap);
+  outputObj = layerMap.get(opts.printLayer) ?? {};
+}
+
+if (opts.pretty) {
+  process.stdout.write(JSON.stringify(outputObj, null, 2));
+} else {
+  process.stdout.write(JSON.stringify(outputObj));
+}
```
1ec5, line 62: Anything other than a string or number causes the third argument to be ignored:

```suggestion
process.stdout.write(JSON.stringify(outputObj, null, opts.pretty ? 2 : null));
```

scripts/util.js
1ec5, line 1: util.js isn’t a very descriptive name. Based on the description above, I guess this could be called something like extract_layer.js, at least until it gains more functionality.

scripts/util.js
```diff
@@ -0,0 +1,62 @@
...
+let outputObj;
+
+if (opts.printGroup) {
+  let group = layerGroupMap.get(opts.printGroup);
+  let layers = [];
+  group.forEach((lyr) => layers.push(lyr));
+  outputObj = layers;
+}
+
+if (opts.printLayer) {
```
1ec5, line 53: It isn’t obvious from the help text or documentation that this command only respects the last option you give it, ignoring any others.

CONTRIBUTING.md
```diff
@@ -194,6 +194,12 @@ There is a "stats" script that will generate various statistics about layer comp
 - `npm run stats -- -c` - total layer count
 - `npm run stats -- -h` - list all options
 
+There is a "util" script that will extract layer style data:
+
+- `node scripts/util -pl <layer>` - JSON contents of a specified layer
```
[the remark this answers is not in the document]
ZeLonewolf: I deliberately did not add this script to `package.json` because it produces JSON output, which is disrupted by the npm info logging that happens before the JSON is printed.

scripts/util.js
ZeLonewolf: Resolved by renaming and updating docs.

CONTRIBUTING.md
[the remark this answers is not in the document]
1ec5: It doesn’t need to be part of `npm run build`, just an extra optional run script like the existing `npm run style`:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/9eb5b36019ced9a49507e137bad3cb50220878eb/package.json#L32

scripts/stats.js
```diff
@@ -7,87 +7,81 @@ program
...
   let layer = layers[i];
   layerMap.set(layer.id, layers[i]);
   let layerSize = JSON.stringify(layer).length;
-  let layerGroup = layer.id.split("_", 1)[0];
-  layerSizeStats.set(layer.id, JSON.stringify(layer).length);
-  if (!layerGroupSizeStats.has(layerGroup)) {
-    layerGroupSizeStats.set(layerGroup, layerSize);
-    layerGroupCountStats.set(layerGroup, 1);
-    layerGroupMap.set(layerGroup, [layer.id]);
+  let layerGroup = layer["source-layer"] || layer.source || layer.type;
```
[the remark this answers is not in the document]
1ec5: I had misremembered that #747 was appending extra cruft onto each layer ID to indicate its group, but in most cases it was just replacing dashes with underscores. It still looks haphazard, but it isn’t a practical issue.

CONTRIBUTING.md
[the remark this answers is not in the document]
ZeLonewolf: Are you envisioning that the optional script would also output to a file?  I was thinking that the only reasonable use would be to pipe the output to "something else" and the info logging breaks that.

scripts/util.js
ZeLonewolf: I set the mutually exclusive options in Commander so you'll now get a console error if you try to combine commands that don't work together.

CONTRIBUTING.md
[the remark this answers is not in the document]
ZeLonewolf: Resolved by documenting -s option

CONTRIBUTING.md
```diff
@@ -190,15 +190,15 @@ When adding or changing style layer code, it can be helpful to assess the change
...
+- `npm run -s stats -- -h` - list all options
 
 There is an "extract_layers" script that will extract layer style data:
 
-- `node scripts/extract_layers -pl <layer>` - JSON contents of a specified layer
-- `node scripts/extract_layers -pg <source>` - list of layers from a specified source
-- `node scripts/extract_layers -h` - list all options
+- `npm run -s extract_layers -pl <layer>` - JSON contents of a specified layer
+- `npm run -s extract_layers -pg <source>` - list of layers from a specified source
+- `npm run -s extract_layers -h` - list all options
```
1ec5, line 201:
```suggestion
- `npm run -s extract_layer -pl <layer>` - JSON contents of a specified layer
- `npm run -s extract_layer -pg <source>` - list of layers from a specified source
- `npm run -s extract_layer -h` - list all options
```

1ec5 (approved):

PR #886, opened 2023-06-23
Refactor place icons

ZeLonewolf:
This PR does the following:

1. Removes two match blocks which discern national/state/city icons from town and village selectors, where they're necessary because those features are always dot icons. This will give is a small style size savings.
2. Renames the place icon file names to be consistent with our POI naming standards.


claysmalley:
Is the dot smaller on purpose? There appear to be two icons, `dot_city.svg` and `place_dot.svg`, that look similar but have different dimensions.

image

ZeLonewolf:
Ah no, that's a regression.

ZeLonewolf:
Restored the proper dot in feabf84:
image

I also found a few other cases where common styling could be extracted to a function. I think with a bit more work the whole city/town/village structure could even be merged into a single layer.

Pengor:
Whoops just saw my review was requested, I'll take a look later this week but I'm certainly fine with the goals of this PR

Pengor (approved):
Looks good 👍

PR #888, opened 2023-06-26
Don't autofocus the map within an iframe

### LINKED ISSUE #884: Viewport hogged when embedded in iframe
When OSM Americana is embedded in an iframe, it hogs the viewport by scrolling into view on load. The OSMUS website is having to work around this issue by scrolling back to the top of the page.

The culprit appears to be this code that focuses the `<canvas>` element just after loading the shield rendering plugin:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/09cf915acb8597359894b94ed20eeabbbca525b3/src/americana.js#L94

This code is an outgrowth of #61, which focused the canvas in order to ensure that GL JS’s keyboard shortcuts would work immediately on load. That behavior is reasonable, but it shouldn’t happen if the page is in a frame.

quincylvania:
Closes #884.

### DIFF
--- a/src/americana.js
+++ b/src/americana.js
@@ -91,7 +91,10 @@ function shieldDefLoad(shields) {
   map.addControl(legendControl, "bottom-left");
   map.addControl(sampleControl, "bottom-left");
 
-  map.getCanvas().focus();
+  if (window.top === window.self) {
+    // if not embedded in an iframe, autofocus canvas to enable keyboard shortcuts
+    map.getCanvas().focus();
+  }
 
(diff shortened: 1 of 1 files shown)


jleedev:
Curious if the `autofocus` attribute on the element would allow the browser to sort this out on its own. It has cross-origin implications FYI.

1ec5 (approved):
I think this is a pretty reasonable simple fix for a pretty visible bug. `autofocus` would be elegant, but I don’t know if it’s guaranteed to focus the element at the right time, after GL JS populates the original `<div>`.

PR #891, opened 2023-07-03
Add shield for Indiana’s John Hunt Morgan Heritage Trail

### LINKED ISSUE #811: John Hunt Morgan Heritage Trail shields

1ec5:
Added a shield for [Indiana’s John Hunt Morgan Heritage Trail](https://www.openstreetmap.org/relation/15638708). The state shape is based on the emblem in the center of the Indiana Toll Road shield (#228). The “JHM” initials are set in [Zilla Slab](https://github.com/mozilla/zilla-slab/). We could easily turn Zilla Slab into a [reverse-contrast](https://en.wikipedia.org/wiki/Reverse-contrast_typefaces) font by stretching it vertically, but that would make it illegible at this size, even on a 2× display.

[West Harrison, Indiana](https://zelonewolf.github.io/openstreetmap-americana/#map=15/39.25005/-84.82305):

[Harrison](https://zelonewolf.github.io/openstreetmap-americana/#map=15/39.25005/-84.82305)

Fixes #811.


ZeLonewolf (approved):
Not the most amazing readability at 1x, but the JHM on these shields is really more of a design flourish than text anyone is supposed to read.
image

PR #893, opened 2023-07-06
Add Ohio River Scenic Byway shield

### LINKED ISSUE #892: Add shield for Ohio River Scenic Byway
The Ohio River Scenic Byway is a National Scenic Byway that crosses Ohio, Indiana, and Illinois. Its [route superrelation](https://www.openstreetmap.org/relation/16058238) is tagged `US:ORSB`. It currently only has [a half-completed route relation in Ohio](https://www.openstreetmap.org/relation/12238016).

All three states post the same shield, depicting a blue Ohio River with green hills and a white sky in the background and a black silhouette of a typical river town skyline in the foreground. This shield is most likely copyrighted, but it should be possible to design a much simpler, copyright-free icon reminiscent of this shield.

[West Indiana State Road 56 and Ohio River Scenic Byway directional assembly](https://commons.wikimedia.org/wiki/File:Ohio_SB_19-10-16_403.jpg)

1ec5:
Added a heavily simplified shield for the Ohio River Scenic Byway. The icon’s composition remains similar enough to the original shield for recognizability, but the route name, ornamentation around the frame, and steamboat are all gone, and the remaining shapes are original. The skyline still has a pointy steeple but nothing else that’s discernible. Since there’s nothing much to put above the hills, this shield is a square instead of the taller shape seen in the field.

Like other scenic route shields, this shield departs from the standard color scheme for recognizability and better contrast. The river’s light blue is consistent with the sky in the 40-Mile Scenic Drive shield added in #576. It and the light green both come from [a California MUTCD sign](https://commons.wikimedia.org/wiki/File:MUTCD-CA_S32-3.svg) that uses these colors for bluejeans and tree leaves, respectively. The original shield’s black outlines are too bold at this size, so the hills are outlined in a standard darker green instead.

Fixes #892. Hold until [changeset 138,155,743](https://www.openstreetmap.org/changeset/138155743) makes it into the tiles.


ZeLonewolf:
Since we now have a second case for cyan, should we just standardize on one shade for the purpose of the palette?  Also note that the next tile update is pending a planetiler fix for some AWS pipeline issues as well as some script updates on my side to automate the generation of pmtiles.

1ec5:
> Since we now have a second case for cyan, should we just standardize on one shade for the purpose of the palette?

Yes, this cyan is a good go-to for picturesque scenic route shields. I would hesitate to use it for ordinary routes’ backgrounds or legends, which usually have more contrast by necessity.

The only other time we’ve used a light green was the lime green leaves in the Merritt Parkway shield (#758). Some online depictions of the Ohio River Scenic Byway shield use lime green, but the actual shields use a much more muted color that complements the sky blue river.

ZeLonewolf:
Looks good to me.

[image](https://preview.ourmap.us/pr/893/#map=11/39.2516/-81.6335)

ZeLonewolf (approved):

PR #895, opened 2023-07-09
Desuperlatize the readme

1ec5:
The readme’s [overly narrow superlative](https://tvtropes.org/pmwiki/pmwiki.php/Main/OverlyNarrowSuperlative) makes it sound like we’re scraping the bottom of the barrel for reasons to be notable enough for a Wikipedia article. This obscures the project’s actual accomplishments. I’ve probably left off a few. We may not be the first by some measures, but we’re proving that it can be done and it can even look good.

Much thanks to @pnorman for [pointing out](https://osmus.slack.com/archives/C01V02K52UX/p1688869438078429) this delicately worded claim to fame and @jleedev for giving me a word for one of my pet peeves.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,11 +10,18 @@ _A quintessentially American map style_ [🗺 View the map](https://zelonewolf.g
 
 The purpose of the Americana style is to:
 
-- Promote collaboration and common purpose in the American mapping community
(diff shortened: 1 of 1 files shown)


adamfranco (approved):
This reads much better now. 😃

PR #896, opened 2023-07-10
Replace images in readme with LaTeX

1ec5:
Replaced some swatch images in the readme with $\LaTeX$ code. This makes it a little more convenient to adjust the color palette in the future.


CONTRIBUTING.md
```diff
@@ -230,30 +230,30 @@ In addition, the following style guidelines are recommended:
...
-| <img src="doc-img/pantone_198.svg" height=18 width=50 /> Pink         | 198     | 223 70 97   | #df4661     |
-| <img src="doc-img/pantone_259.svg" height=18 width=50 /> Purple       | 259     | 109 32 119  | #6d2077     |
-| <img src="doc-img/pantone_187.svg" height=18 width=50 /> Red          | 187     | 191 32 51   | #bf2033     |
-| <img src="doc-img/pantone_116.svg" height=18 width=50 /> Yellow       | 116     | 255 205 0   | #ffcd00     |
-| <img src="doc-img/pantone_382.svg" height=18 width=50 /> Yellow-Green | 382     | 196 214 0   | #c4d600     |
-| <img src="doc-img/white.svg" height=18 width=50 /> White              | N/A     | 255 255 255 | #ffffff     |
-| <img src="doc-img/black.svg" height=18 width=50 /> Black              | N/A     | 0 0 0       | #000000     |
+| Color                                                              | Pantone | RGB         | Hex triplet |
+| ------------------------------------------------------------------ | ------- | ----------- | ----------- |
+| ${\color[RGB]{0,   63,  135}\rule[-4px]{50px}{18px}}$ Blue         | 294     | 0 63 135    | #003f87     |
```
1ec5, line 246: Oh that’s weird, now GitHub is unhappy with this syntax, but it was fine with it during preview:

Unable to render expression.

Blue

LaTeX is [supposed to be supported](https://github.blog/2022-05-19-math-support-in-markdown/) these days.

1ec5: Possibly related to orgs/community#45276.

ZeLonewolf: Amazing to see bit rot with such a short half-life...

1ec5: Wonder if simplifying the expression helps:

${\color[RGB]{0,63,135}\blacksquare}$

ZeLonewolf: I'm seeing a black square in Chrome.

1ec5: Yes, but if you’re using Chrome, you aren’t seeing the blue rectangle a few comments above.

I think this is simply a bug in either GitHub’s MathJax wrapper or a quirk of how they’ve configured MathJax. They understandably have a lot of code that tries to sanitize and otherwise restrict math expressions in Markdown. However, it would make no sense for these restrictions to apply to each browser differently (and not at all to Safari).

1ec5: GitHub is sanitizing the `\rule` macro pretty heavily. I was using it to approximate the dimensions of the images that were being replaced. I could still use `\LARGE` (not `\huge`), but I don’t think it’s really worth it.

1ec5: A few minutes later, the `\blacksquare` macros stopped working in Firefox, but they’re still working in Chrome and Safari. Sigh…

1ec5: Turns out these errors are only occurring in _my_ copy of Firefox, because I had previously activated [MathJax’s accessibility mode](https://docs.mathjax.org/en/latest/basic/accessibility.html) somewhere on GitHub. But if I view this page in a private browser, everything is fine.

ZeLonewolf (approved):
Looks normal on my setup (Ubuntu + Chrome) and I concur wit the switch to LaTeX and commons images. I'll leave it up to you to assess whether there's other browser compatibility things to check with the LaTeX code,

image

image

image

PR #899, opened 2023-07-22
Make code compatible with ES2019

### LINKED ISSUE #890: Make shield generator compatible with ES2019
The maplibre-shield-generator library is built for `ESNext` (which would be ECMAScript 2023 at the time of writing):

https://github.com/ZeLonewolf/openstreetmap-americana/blob/4ba85f903250b1747688d9729f0059d95dba0288/tsconfig.json#L3-L4

This build setting is so bleeding-edge that it could apparently open the door to [code that no released browser can parse at all](https://github.com/evanw/esbuild/releases/tag/v0.18.7) if we upgrade esbuild too soon. A more immediate issue is that even some of the recent ECMAScript versions would be incompatible with websites that might be interested in adopting the library.

For example, the OpenStreetMap Wiki recently integrated maplibre-shield-generator, loading it dynamically on [pages that embed OSM Americana](https://wiki.openstreetmap.org/wiki/Special:PermanentLink/2551555#New_templates_for_vector_maps). This makes it the first “core” OSM website to adopt client-side-rendered vector maps. Unfortunately, the mere presence of this library breaks all JavaScript for users of browsers such as Firefox 91 ESR (released in 2021). Since the library is compiled as a module, it becomes necessary to [dynamically import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import) it from another script, which leads to the following error:

> SyntaxError: import declarations may only appear at top level of a module

The OSM Wiki currently runs [MediaWiki 1.39](https://wiki.openstreetmap.org/wiki/Special:Version), which [provides “grade-A” support](https://www.mediawiki.org/wiki/Special:PermanentLink/6012138#Browser_support_matrix) for “modern browsers”, defined as the last three years of Chromium releases, two LTS releases of Firefox, and Safari 11 (2018). Scripts are loaded by [ResourceLoader](https://www.mediawiki.org/wiki/ResourceLoader/Core_modules#mw.loader.load), a built-in bundler and minifier. ResourceLoader bundles maplibre-shield-generator with other site scripts, such as the one that powers taginfo statistics and taglists.

The OSM Wiki isn’t required to support every configuration that MediaWiki does, but Americana is currently embedded on so few pages that limiting browser support based on it would be quite a tail-wagging-the-dog situation. I can also imagine other websites besides wikis facing similar hurdles from how an ESNext-targeted library might interact with other dependencies.

It would make a lot of sense for maplibre-shield-generator, as a MapLibre GL JS plugin, to support the same browsers that MapLibre GL JS does. As far as I know, MapLibre GL JS lacks a clearly articulated browser compatibility policy, but its build system is configured to target [ES2019](https://github.com/maplibre/maplibre-gl-js/blob/30c0e1d47f019f159edc6c6ef8c5e068f14a696e/tsconfig.json#L17).

ZeLonewolf:
Fixes #890

This PR changes the compatibility to ES2019 which is desired for compatibility reasons (see description in #890).  Mostly, this involved replacing top-level awaits with Promise code.

Open question: is this actually desirable in the main repo?  Should we make it so _only_the shield library is ES2019 instead?

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -1,5 +1,4 @@
 import { stat, copyFile, mkdir } from "node:fs/promises";
-import { pathToFileURL } from "node:url";
 
 import esbuild, { BuildContext, BuildOptions } from "esbuild";
 
@@ -83,4 +82,10 @@ export const build = (
   buildOptions: BuildOptions = {}
 ): Promise<BuildContext[]> => buildWith("build", buildOptions);
 
-await build();
+(async function () {
+  try {
+    await build();
+  } catch (error) {
+    console.error(error);
+  }
+})();
--- a/scripts/generate_style.js
+++ b/scripts/generate_style.js
@@ -4,7 +4,7 @@ import { Command } from "commander";
 import { validate } from "@maplibre/maplibre-gl-style-spec";
 
 import * as Style from "../src/js/style.js";
-import config from "../src/config.js";
+import { getConfig } from "../src/config.js";
 
 /**
  * Requires mapbox-gl-rtl-text:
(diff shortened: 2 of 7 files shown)


scripts/build.ts
```diff
@@ -1,5 +1,4 @@
 import { stat, copyFile, mkdir } from "node:fs/promises";
-import { pathToFileURL } from "node:url";
```
ZeLonewolf, line 2: Removed because it was unused, but not related to ES2019 compatibility.

PR #901, opened 2023-07-23
Package shield generator as an ordinary script

1ec5:
Package the shield generator library as an ordinary JavaScript script rather than an ECMAScript module. This enables webpage authors to install the library just as they would install MapLibre GL JS:

```html
<script type="text/javascript" src="maplibre-shield-generator.js"></script>
```

and access any exported symbol through the `window.mapLibreShieldGenerator` object. Unfortunately, Web applications written in modern ECMAScript won’t be able to fully encapsulate the shield generator library and hide it from the webpage context.

Fixes the problem described in https://github.com/ZeLonewolf/openstreetmap-americana/issues/890#issuecomment-1646717687.

### DIFF
--- a/shieldlib/README.md
+++ b/shieldlib/README.md
@@ -4,15 +4,21 @@ The Americana shield renderer is a library intended to draw highway shields on a
 
 ![Pictoral highway shields](https://wiki.openstreetmap.org/w/images/6/6d/Rendered_shields_americana.png)
 
-## Shield rendering workflow
+## Usage
 
-Rendering shields requires the following compoments:
+Rendering shields requires the following components:
 
 1. **Encode shield information in vector tiles**. First, your tiles must contain the information which tells the shield renderer what shields to draw. In OpenMapTiles, shield information is encoded in the [`transportation_name`](https://openmaptiles.org/schema/#transportation_name) vector tile layer with a series of attributes named `route_1`, `route_2`, etc. Each attribute contains a text string which contains all of the information needed to determine which graphic to display, including numeric route number if the shield is numbered. However, this library allows you to specify how the shield information has been encoded, and it's possible to stitch together data from multiple fields when encoding shield data.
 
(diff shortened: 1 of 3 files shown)


shieldlib/README.md
```diff
@@ -4,15 +4,21 @@ The Americana shield renderer is a library intended to draw highway shields on a
...
 
 2. **Expose shield information in a style layer**. Next, route information must be exposed in a maplibre expression using [image](https://maplibre.org/maplibre-gl-js-docs/style-spec/expressions/#types-image) in a structured string containing the route information. For example, you might encode Interstate 95 as an image named `shield|US:I=95`. Normally, the image expression is used to point to pre-designated sprites in a sprite sheet, but in this case, we're pointing to a sprite which doesn't exist called `shield|US:I=95`. This will trigger a `styleimagemissing` event which allows the shield renderer to create the required graphic on the fly. As an example of how to encode shield information, see OSM Americana's [`highway_shield`](https://github.com/ZeLonewolf/openstreetmap-americana/blob/main/src/layer/highway_shield.js) style layer.
 
-3. **Define a parser that describes how route information is encoded**. There are three parts to a route definition:
+3. Import the library:
+   
+   ```html
+   <script type="text/javascript" src="maplibre-shield-generator.js"></script>
```
ZeLonewolf, line 18: Should this link to an npm location or how does this work?

1ec5: This instruction is for projects that don’t use NPM – the more typical situation among Web developers. But yes, once we figure out how to make this build compatible with NPM, we should also include instructions for including it in that environment.

zekefarwell (commented):
Looks like this gets things working for a context that can't handle ES modules so that's good.  Unfortunately the demo site is currently importing the shield library as a module in a variety of places so more restructuring would be needed if it is to be compiled only in iife format.  `npm start` doesn't error out, but it shows a bunch of warnings (example below) and the map is blank.
```
 [WARNING] Import "ShieldRenderer" will always be undefined because the file "shieldlib/dist/maplibre-shield-generator.js" has no exports [import-is-undefined]
```
I've pushed up [another branch](https://github.com/ZeLonewolf/openstreetmap-americana/tree/zeke-shieldlib-iife-890) that seems to fix this issue and should allow the library to be used as an esmodule, common js module, or iife for older browser compatibility.  It configures the build to output several formats and specifies them all for different purposes in the package.json (based on the [esbuild recommendation here](https://esbuild.github.io/api/#main-fields)).  Feel free to merge into this PR if that seems like a good path forward.

The other option would be to restructure all the shield library imports to not use module syntax, but that might be a bit of a rabbit hole.

1ec5:
@zekefarwell that’s great – it makes a lot of sense to put out multiple flavors of the script depending on the environment. Would you mind opening a PR with your version?

PR #905, opened 2023-07-25
Zeke shieldlib iife 890

zekefarwell:
Building on #901, this PR updates the esbuild script to generate all three possible [output formats](https://esbuild.github.io/api/#format) and references them in the package.json [main, module, and browser fields](https://esbuild.github.io/api/#main-fields).  This should allow the library to be used in a variety of different contexts that may or may not supported ES Modules. I am opening this as a draft PR for now since it may make sense make include some more related changes before merging.  

[PR previews here](https://github.com/ZeLonewolf/openstreetmap-americana/blob/main/.github/workflows/s3-upload.yml#L78-L84) would let them be separately tested on a live URL.

The shield library is currently being built twice.  Once [in the library  build script](https://github.com/ZeLonewolf/openstreetmap-americana/blob/efff6cdecf68504a02b1343ddc760327f416616f/shieldlib/scripts/build.js#L9-L27), and once [in the main americana build script](https://github.com/ZeLonewolf/openstreetmap-americana/blob/efff6cdecf68504a02b1343ddc760327f416616f/scripts/build.ts#L52-L74).  I think the one in the main build script is vestigial from before the shield library got split out.  However, when I removed it, the [live reload feature](https://github.com/ZeLonewolf/openstreetmap-americana/blob/efff6cdecf68504a02b1343ddc760327f416616f/scripts/serve.ts#L8-L10C4) threw an error.  Looks like it somehow expects a `shieldLibContext` from the build script.  

I don't currently have time to figure out how either of these things work in order to make those changes.  If anyone else wants to add on to this PR, go right ahead.  Or if the consensus is just to move ahead without them that's fine too.

### DIFF
--- a/shieldlib/README.md
+++ b/shieldlib/README.md
@@ -4,15 +4,21 @@ The Americana shield renderer is a library intended to draw highway shields on a
 
 ![Pictoral highway shields](https://wiki.openstreetmap.org/w/images/6/6d/Rendered_shields_americana.png)
 
-## Shield rendering workflow
+## Usage
 
-Rendering shields requires the following compoments:
+Rendering shields requires the following components:
 
 1. **Encode shield information in vector tiles**. First, your tiles must contain the information which tells the shield renderer what shields to draw. In OpenMapTiles, shield information is encoded in the [`transportation_name`](https://openmaptiles.org/schema/#transportation_name) vector tile layer with a series of attributes named `route_1`, `route_2`, etc. Each attribute contains a text string which contains all of the information needed to determine which graphic to display, including numeric route number if the shield is numbered. However, this library allows you to specify how the shield information has been encoded, and it's possible to stitch together data from multiple fields when encoding shield data.
 
 2. **Expose shield information in a style layer**. Next, route information must be exposed in a maplibre expression using [image](https://maplibre.org/maplibre-gl-js-docs/style-spec/expressions/#types-image) in a structured string containing the route information. For example, you might encode Interstate 95 as an image named `shield|US:I=95`. Normally, the image expression is used to point to pre-designated sprites in a sprite sheet, but in this case, we're pointing to a sprite which doesn't exist called `shield|US:I=95`. This will trigger a `styleimagemissing` event which allows the shield renderer to create the required graphic on the fly. As an example of how to encode shield information, see OSM Americana's [`highway_shield`](https://github.com/ZeLonewolf/openstreetmap-americana/blob/main/src/layer/highway_shield.js) style layer.
 
-3. **Define a parser that describes how route information is encoded**. There are three parts to a route definition:
+3. Import the library:
+   
+   ```html
+   <script type="text/javascript" src="maplibre-shield-generator.js"></script>
+   ```
+
+4. **Define a parser that describes how route information is encoded**. There are three parts to a route definition:
 
    1. The `network` string, which defines a network with a common shield shape, graphic, and color
    1. The `ref` string, which defines a text sequence that should be drawn on top of the shield graphic
@@ -36,7 +42,7 @@ Rendering shields requires the following compoments:
    };
    ```
 
-4. **(Optional) Create predicates that define which shields will be handled**. For example, if all sprite IDs in your style that need a shield begin with the string `shield|`, this would look like:
+5. **(Optional) Create predicates that define which shields will be handled**. For example, if all sprite IDs in your style that need a shield begin with the string `shield|`, this would look like:
 
    ```typescript
    let shieldPredicate = (imageID: string) => imageID.startsWith("shield");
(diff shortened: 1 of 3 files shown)


zekefarwell:
The builds were failing originally due an import name resolution issue.  I worked around this with the same hack [already in use here](https://github.com/ZeLonewolf/openstreetmap-americana/blob/efff6cdecf68504a02b1343ddc760327f416616f/scripts/taginfo.js#L11-L18).  I imagine there is probably a better way, but I'm a bit out of my depth on this.  Unfortunately it the Ubuntu test build is still failing due to some other issue that I don't understand.  I'm actually getting the same failure when I run the tests locally on my Mac (Apple Silicon), but the MacOS build here is successfully.  🤷

1ec5:
I wonder if the import name resolution issue is caused by this use of a DOM type as an instance method’s return value. In the absence of an explicit import, it would be coming from the `window` when running in a browser.

https://github.com/ZeLonewolf/openstreetmap-americana/blob/efff6cdecf68504a02b1343ddc760327f416616f/shieldlib/src/document_graphics.ts#L13

ZeLonewolf:
I re-ran the test build several times in CI and it errored on the same test case each time.
I ran the test suite locally on my Ubuntu and it built normally.

The test failure is bizarre because it's erroring on one specific test case in the middle.

ZeLonewolf:
The issue is at:
```javascript
      expectGloss("en", "L’Aquila", "L'Aquila", "L’Aquila", "L'Aquila");
```

Is it possible that we're seeing some kind of difference in unicode diacritic processing between javascript versions that could be causing this?

1ec5:
> Is it possible that we're seeing some kind of difference in unicode diacritic processing between javascript versions that could be causing this?

That would be surprising to me, since `’` isn’t a diacritic at all. This is the only way I can explain the result of “L'Aquila”, but it suggests a possible standards compliance bug in the Node engine on that platform.

This test case was added in response to https://github.com/ZeLonewolf/openstreetmap-americana/pull/592#issuecomment-1332367624. Preferring OSM’s straight apostrophe over Wikidata’s curly apostrophe is suboptimal from a typographical standpoint but not incorrect linguistically. We shouldn’t insist on this behavior everywhere, but if it’s happening on some platforms, I guess it’s OK and we can replace this test case with something we feel more strongly about.

ZeLonewolf:
I'm exploring the root of this issue on the L'Aquila test case over in #908.  If I can isolate the issue, what I would plan to do is to restructure the test cases such that the expected test result matches the node-version-specific maplibre expression behavior.  Right now we are hard-coding the epected label and gloss, but we should be able to generate the expected result in _code_ and verify that the expression that we've built does the same thing as the javascript code (which in theory is using the same javascript functions to parse labels).

1ec5:
> Right now we are hard-coding the epected label and gloss

This is a good thing, in my opinion. We need to know when the expression yields the wrong end-user result in some configuration, regardless of whether it arrived at that result via the “right” JavaScript functions.

ZeLonewolf:
Does this PR conflict with #900 at all?

zekefarwell:
I don't think it should, but might as well merge it in and then update this branch to know for sure.

ZeLonewolf:
Dumb question, how do the three output artifacts get made available to end-users?  Does `npm publish` just take care of that or do we need to designate artifacts in the github CI?

zekefarwell:
I don't know the answer to this question and that's one of the reasons this PR is still a draft.  It looks like it is up to a library author to configure which files end up getting published or not.
https://stackoverflow.com/questions/31642477/how-can-i-publish-an-npm-package-with-distribution-files

ZeLonewolf:
I think for github CI artifacts, we can use https://github.com/actions/upload-pages-artifact and just add a section to the ubuntu builder that points to the output jar files.  Probably easiest if the output jar files post to a dedicated folder that we can point the upload-artifacts action to.

ZeLonewolf:
The addition of:
```json
"transpileOnly": true
```
Is what seems to have broken this build, based on a diff inspection.

ZeLonewolf:
This should build now, however, it's importing in the format below in a few spots:

```typescript
"@americana/maplibre-shield-generator/src/index"
```

Still need to figure out how to get those classes into the main exports.

controversial:
> how do the three output artifacts get made available to end-users

It’s conventional to publish all three versions in the same npm tarball, under different sub-paths, and then the “exports” conditions in the top level package.json will send importers to the right one.

for example
```json
{
  "exports": {
    ".": {
      "require": "./dist/cjs/index.cjs",
      "import": "./dist/esm/index.mjs",
    }
  }
}
```

ZeLonewolf:
@controversial, could you review this PR to see if it looks like it's doing the right things?

PR #908, opened 2023-07-30
Ubuntu CI test only

ZeLonewolf:
PR for the purpose of debugging PR #905 CI tests.

### DIFF
--- a/test/spec/label.js
+++ b/test/spec/label.js
@@ -46,7 +46,20 @@ function evaluatedLabelAndGloss(locales, properties) {
   return [label, gloss];
 }
 
+function diacriticSensitiveCaseInsensitiveMatch(locale, s1, s2) {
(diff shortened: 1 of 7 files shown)


test/spec/label.js
```diff
@@ -521,4 +643,23 @@ describe("label", function () {
...
     });
   });
+
+  describe("Regression checks", function () {
+    const locales = ["en"];
+    const properties = {
+      name: "L'Aquila",
+      "name:en": "L’Aquila",
+    };
+    it("L'Aquila special case 1", function () {
```
1ec5, line 653: Test case names should read like sentences describing the intended result.

ZeLonewolf:
Just to be clear, I'm primarily using this PR as a vehicle to identify the cause of the test case failure, since it only seems to be a problem on the CI system.

PR #909, opened 2023-07-31
Prevent all test-runners from using node v18.17.0

ZeLonewolf:
Node v18.17.0, released on July 18, 2023, introduced a bug which breaks one test case as shown in PR #905 and debugged in PR #908.

This PR forces node v18.16.0 in the Ubuntu CI runner.

I implemented this PR in two commits to demonstrate that the Ubuntu runner is in fact currently broken on node "18.x".

### DIFF
--- a/.github/workflows/build-preview.yml
+++ b/.github/workflows/build-preview.yml
@@ -19,10 +19,10 @@ jobs:
     steps:
       - name: Checkout 🛎️
         uses: actions/checkout@v3
-      - name: Use Node.js 18.x
+      - name: Use Node.js 18.16.1
         uses: actions/setup-node@v3
         with:
-          node-version: 18.x
+          node-version: 18.16.1 #18.17.0 is buggy
       - name: Save PR number
         env:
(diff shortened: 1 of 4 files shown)


zekefarwell:
Nice sleuthing!  I have Node 18.17 on my Mac and get the same test failure as the test-build-ubuntu workflow.  I installed Node 18.16.1 and can confirm that the test passes when run with that version.  I also installed Node 20.5.0 (released the same day as 18.17) and found that the test fails with the same error.

Holding the test workflow back to 18.16.1 seems like a good idea for now.  Seems like it would be good to do the same on the test-build-mac workflow too.  So far that one seems to still be using a cached version on 18.16.1, but I imagine it will bump up to 18.17 at some point soon and then the test will start failing there as well.

ZeLonewolf:
Updated the PR to change the fallback to 18.16.1 and apply it to all CI runners.

zekefarwell (approved):
Seems like a good solution for now.

PR #910, opened 2023-08-02
Add places of worship

### LINKED ISSUE #861: Places of worship
Places of worship are frequently important points of interest, both for potential attendees and for general wayfinding. I feel like they may be tricky to render well however. For one, there are a wide variety of major religions in the world, many of which have significant presence in the US. In addition to a generic place of worship symbol, individual symbols for major religions are typically rendered. For example, [Carto](https://wiki.openstreetmap.org/wiki/OpenStreetMap_Carto/Symbols#Religious_place) renders Christian, Jewish, Muslim, Taoist, Hindu, Buddhist, Shinto, and Sikh religious places with distinct icons. All of these religions seem suitable for distinct renderings in Americana; I could potentially imagine faiths in addition to this list, such as LDS/Mormon temples, also getting a distinct rendering (Carto currently renders them as generic places of worship).

A familiar approach in American maps is to use a square building topped by a religious symbol, most classically a Christian cross:
image

This works reasonably well for crosses, but I worry it may be difficult to generalize well to other symbols, and to a generic version, though it's probably doable. Alternatively, many other OSM renderers, including Carto, Temaki, and OSMAnd, use each religion's symbol itself. I'd love to see more samples of place of worship treatment in American maps, especially of non-Christian ones. Some that I've seen simply don't show them at all, or only show them when they are landmark buildings, with an icon similar to other major sights.

Another issue will be the color of the icons. In my opinion, places of worship do not fit neatly into any of the defined POI color categories. They are potentially closest to "attraction", which would entail a brown color. Many styles use black icons for religious sites, although Americana already uses black for places and roads.

wmisener:
Fixes #861.

This PR adds places of worship. The icon is based on the classic US atlas style of a square topped with a religious icon. The initial PR contains the places of worship of the 8 religions given an icon in Carto: Buddhist, Christian, Hindu, Jewish, Muslim, Sikh, Shinto, and Taoist. The symbols I chose are:

- Buddhist: wheel/dharmachakra poi_pow_buddhist
- Christian: cross poi_pow_christian
- Hindu: om  poi_pow_hindu
- Jewish: star of David poi_pow_jewish
- Muslim: star and crescent poi_pow_muslim
- Shinto: torii poi_pow_shinto
- Sikh: khanda poi_pow_sikh
- Taoist: yin-yang poi_pow_taoist

Places of worship do not necessarily fall neatly into any of the defined POI color categories. The initial PR uses attraction-brown, but there were suggestions in slack discussions for infrastructure-blue or perhaps its own color (maybe gray?). I'm open to suggestion on colors.

Other religions could be considered. Some other common `religion` values in the United States include `unitarian_universalist` (518 `place_of_worship`) and `bahai` (79). This PR also lacks distinct icons for LDS and Jehovah's Witness places of worship, which are rendered with crosses. Distinct renderings are blocked by OMT, which does not contain any `denomination` information that could distinguish these from other Christian POWs.

Samples:
Christian and Hindu ([localhost link](http://localhost:1776/#map=15.18/37.725193/-122.420873)):
Screen Shot 2023-08-01 at 10 53 33 PM

Buddhist, Christian, Jewish, and Shinto ([localhost link](http://localhost:1776/#map=15.18/37.787447/-122.433858)):
Screen Shot 2023-08-01 at 10 59 38 PM

Christian and Muslim ([localhost link](http://localhost:1776/#map=16.38/37.787833/-122.422837)):
Screen Shot 2023-08-01 at 10 57 28 PM

Buddhist, Christian, and Taoist ([localhost link](http://localhost:1776/#map=16.13/37.794934/-122.407639)):
Screen Shot 2023-08-01 at 11 03 09 PM

Christian and Sikh ([localhost link](http://localhost:1776/#map=17.02/37.960105/-122.316972)):
Screen Shot 2023-08-01 at 11 04 16 PM

Legend sample:
Screen Shot 2023-08-01 at 11 11 31 PM
I opted for the uniform "place of worship" rather than attempting to name each religion's.

Jerusalem Temple Mount ([localhost link](http://localhost:1776/#map=15.38/31.778644/35.232777)):
Screen Shot 2023-08-01 at 11 19 37 PM


ZeLonewolf:
Thanks for taking this on!

There was a [Slack discussion](https://osmus.slack.com/archives/C01V02K52UX/p1691041634613469) that discussed PoW icons and in summary I think we need to:
1. Use the "community anchor blue" color scheme for PoWs
2. The other-than-Christian symbols are too fine and the features should be thickened a bit so that they have the same effective visual weight

There was also some discussion about modifying the base rectangle size if needed for effective size contrast between the symbol and the base.

wmisener:
Updated the PR to color the symbols blue and add thicker lines for the Star of David. I'll continue iterating the symbol relative sizes via slack, though I can also post the test screenshots here if anyone desires

src/layer/poi.js
```diff
@@ -84,6 +84,70 @@ var iconDefs = {
...
+    color: Color.poi.infrastructure,
+    description: "Buddhist place of worship",
+  },
+  pow_christian: {
+    classes: {
+      place_of_worship: ["christian"],
+    },
+    sprite: "poi_pow_christian",
+    color: Color.poi.infrastructure,
+    description: "Christian place of worship",
```
1ec5, line 101: id-tagging-schema calls this tagging combination simply “[Christian church](https://github.com/openstreetmap/id-tagging-schema/blob/c0bfb2812c1d0ae94cc05fa0567002c787296551/data/presets/amenity/place_of_worship/christian.json#L46)”, and there are other religion-specific descriptions for other `religion` values.

wmisener: I'm open to this, and considered it for the initial PR, though I worried it gets into complicated terminology issues quickly. For instance, not all `amenity=place_of_worship` + `religion=christian` are churches (like LDS Temples or even [holy wells](https://wiki.openstreetmap.org/wiki/Tag%3Aplace_of_worship%3Dholy_well)), some sects of Judaism call their houses of worship 'temples' or 'shuls' instead of synagogues, what iD labels as "Sikh temples" are actually called [gurdwaras](https://en.wikipedia.org/wiki/Gurdwara), etc. But maybe these are small potatoes compared to the improvement in how understandable the legend would be.

1ec5: That’s true, your point probably matters more in the context of a public-facing map like Americana than something internal to OSM like iD. The diplomatic “place of worship” terminology isn’t inaccurate, just wordy, but we have enough room for it.

1ec5 (approved):

wmisener:
Ah thanks for going ahead and merging, I've been busy with work lately and haven't had much time to work on this. I think as some of the comments here and on Slack have noted, if it turns out "in the wild" that the religious symbols are too small, it could potentially be useful to make the squares slightly smaller and the icons slightly bigger. I'll also open separate issues to track other potential religions.

Issue #654, opened 2023-01-03
Spain’s “A-roads” do not necessarily have blue shields


1ec5:
Currently, this project assumes that every route relation tagged `network=ES:A-road` is an autopista or autovía that has a blue shield. Neither assumption is correct with respect to OSM. In reality, Spain’s route designation situation is almost as complex and inconsistent as in the U.S. Although the national government often applies a blue shield and “A-” or “AP-” prefix to autopistas and autovías, many have the prefix “R-” (for _radial_) or “AP-” (for _autopista de peaje_), making it less clear whether these routes should be considered “A-roads”.

Meanwhile, each of the autonomous communities has its own conflicting scheme for local autovías. For example, in Andalusía (which happens to start with an A), [Autovía A-92](https://www.openstreetmap.org/relation/9465880) has an orange shield, while [Autovía A-382](https://en.wikipedia.org/wiki/Autovía_A-382) has a green shield. In Castile and León, [Autovía A-231](https://en.wikipedia.org/wiki/Autovía_A-231) has a blue shield. In the Basque Country, [Autovía AP-1](https://www.openstreetmap.org/relation/404438) is neither nationally operated nor tolled but still has the “AP” prefix and blue shield.

Due to all these idiosyncrasies, the Spanish community has adopted the [`ref:colour`](https://wiki.openstreetmap.org/wiki/Key:ref:colour) key to indicate the shield color independently of the route network. It would be better to base the shield selection on this data than to maintain an override list ourselves, since the assigned colors apparently change regularly. The vector tiles would need to expose `ref:colour`, ideally coalescing similarly hexadecimal color references into more predictable color keywords. In the meantime, maybe we should remove the existing blue-only implementation to avoid confusion:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/e5fe35be2653a0cf96acec634e49fe3df829dce1/src/js/shield_defs.js#L3324-L3328

ZeLonewolf:
> The vector tiles would need to expose `ref:colour`, ideally coalescing similarly hexadecimal color references into more predictable color keywords.

Since it's tagged at the network level, this is feasible.  However, remember that network and ref are packed into a single tile attribute per route, so you have something like:

`route_1=ES:A-road=AP-1`

The `ref:color` key has 35K usages, so it's not unreasonable to use it. I would need to implement some scheme that packs the value into the string to make this work, perhaps by adding another `:<color>` on the end of the network portion, for example.

[...]

PR #912, opened 2023-08-03
Add shields for Nova Scotia scenic drives

### LINKED ISSUE #911: Nova Scotia scenic drive shields
Nova Scotia has [a number of well-signed scenic drives](https://en.wikipedia.org/wiki/List_of_Nova_Scotia_provincial_highways#Scenic_Routes) that tourists expect to see rendered on road maps. Some scenic routes are coterminous with numbered routes—like [NS Trunk 8/Kejimkujik Scenic Drive](https://en.wikipedia.org/wiki/Nova_Scotia_Trunk_8)—while most are more complex. The [Cabot Trail](https://en.wikipedia.org/wiki/Cabot_Trail) is a trunk route notable for being signed only as a scenic route, not by its route number, and so is especially important to get on the map.

Relations exist for all the routes, though the tagging and completeness are somewhat inconsistent and could use some maintenance.

#### List, linked to OSM

- [ ] [Bras d'Or Lakes Scenic Drive](https://www.openstreetmap.org/relation/11051558)
- [ ] [Cabot Trail](https://www.openstreetmap.org/relation/16153464)
- [ ] [Ceilidh Trail](https://www.openstreetmap.org/relation/10887030)
- [ ] [Digby Neck and Islands Scenic Drive](https://www.openstreetmap.org/relation/16153498)
- [ ] [Evangeline Trail](https://www.openstreetmap.org/relation/8592795)
- [ ] [Fleur-de-lis Trail](https://www.openstreetmap.org/relation/11147926)
- [ ] ~~[Fundy Shore Scenic Drive](https://www.openstreetmap.org/relation/10959709)~~ (apparently unsigned)
- [ ] [Glooscap Trail](https://www.openstreetmap.org/relation/10932493)
- [ ] [Kejimkujik Scenic Drive](https://www.openstreetmap.org/relation/8592777)
- [ ] [Lighthouse Route](https://www.openstreetmap.org/relation/8592936)
- [ ] [Marconi Trail](https://www.openstreetmap.org/relation/11115615)
- [ ] [Marine Drive](https://www.openstreetmap.org/relation/10983819)
- [ ] [Sunrise Trail](https://www.openstreetmap.org/relation/10975085)

#### Shields

[The shield designs](https://www.gribblenation.org/2018/01/nova-scotia-scenic-trails.html) are simple and consistent, and should be straightforward to adapt for Americana.

quincylvania:
Closes #911.

Creating a draft for sprite design feedback. Waiting for a tile refresh to review how these look on the map.


src/js/shield_defs.js
```diff
@@ -4201,6 +4204,33 @@ export function loadShields() {
 
   // Ref-specific cases. Each entry should be documented in CONTRIBUTE.md
 
+  shields["CA:NS:S"].overrideByRef = {
+    BdOLSD: {
```
1ec5, line 4208: Do people or publications normally refer to the routes by these initialisms? If so, let’s add an explanation to CONTRIBUTING.md, in the section where we justify other unsignposted refs.

icons/shield_ca_ns_s_md.svg
1ec5, line 1: I wonder if a few of these shields would be more legible if we don’t need to offset them to the left but can instead have them take up more space to the top and right.

src/js/shield_defs.js
quincylvania: I don't think so, I just made up the initialisms so we'd have something to key off of. I don't think we can key off the route name because I thiiiiink the name on the tile comes only from the way itself. The Merritt Parkway is the only one we do like that so far and it happens to have all way names match the parkway relation name.

This situation is kind of like the Great Lakes Circle Tour routes, but with a more items and less documentation. I imagine the problem will come up more often as we add additional networks of scenic routes.

I can add documentation for this.

icons/shield_ca_ns_s_md.svg
quincylvania: Sure, we could make the pictures slightly bigger by centering then and moving the water down. But personally I think I prefer sticking closer to the source designs. Though maybe seeing them on the map will affect this.

1ec5: Yeah, we’ll see how it turns out on the map. For the Ohio River Scenic Byway in #893, I made the shield 20&nbsp;pixels wide but taller than 20&nbsp;pixels, avoiding whitespace on either side, in order to maximize room for the shield’s details.

src/js/shield_defs.js
1ec5: For #451, the New Zealand community apparently decided to give each route a distinct `network` subvalue in order to avoid polluting `ref` with unsignposted initialisms. However, this fragmentation makes less sense when the shield designs all follow a common motif, as in this case.

I think inevitably we’ll need the tiles to contain something beyond the `network=ref` syntax in `route_*` properties. If we could go back and change the format, I would’ve suggested something like `route:2:network` or `route_network_2`, which would’ve also logically allowed for `route_name_2` or even `route_wikidata_2` for these tricky cases.

ZeLonewolf:
Tiles are updated!

quincylvania:
In terms of size, I think the original 16x20 looks okay, perhaps too small:

Screenshot 2023-08-07 at 2 48 17 PM
Screenshot 2023-08-07 at 2 49 28 PM

I tried 18x22 to keep even numbers and it seems a bit too big:

Screenshot 2023-08-07 at 2 53 26 PM
Screenshot 2023-08-07 at 2 53 38 PM

Ultimately I'd say 17x21 is a good compromise size as long as it doesn't cause rounding/aliasing issues.

Screenshot 2023-08-07 at 3 09 53 PM

ZeLonewolf:
Either the 17px or the 18px versions look pretty good to me in the screen shots.  I agree that the 16px one seems slightly undersized.

quincylvania:
I standardized on 17x21 px shields. I couldn't find any signage for the Fundy Shore Scenic Drive but otherwise this is complete and ready for review. FYI there was a `ref` mistake for the Sunrise Trail so that shield won't show up until a new tile update goes through but you can see it in the spritesheet.

src/js/shield_defs.js
1ec5: > This situation is kind of like the Great Lakes Circle Tour routes, but with a more items and less documentation.

The Great Lakes Circle Tour routes are known by the initialisms in `ref`, at least informally, so it isn’t a big deal if a non-shield-capable renderer displays the `ref` verbatim. I’d still reach out to the local community to get their opinion on this use of `ref`, in case there are more established initialisms for the routes.

1ec5 (commented):
These icons seem to hew to an 8-bit style, which happens to look great for these particular shields. I wonder if we should aim for a similar style in other shields, where we’re currently going for a very literal, realistic look and feel.

ZeLonewolf:
Regarding the choice of ref on bespoke tourist routes, don't we have `overrideByName` as an option?  Then we would be tied to the name, which should be beyond dispute. In any case, the choice of `ref` value on the route relation of an obscure tourist route seems like something we can just make up using reasonable conventions and change later if the community changes their mind down the line.  In short, I'd favor pushing this forward either as-is or with a name-based definition.

icons/shield_ca_ns_s_md.svg
1ec5: > Regarding the choice of ref on bespoke tourist routes, don't we have `overrideByName` as an option? Then we would be tied to the name, which should be beyond dispute. In any case, the choice of `ref` value on the route relation of an obscure tourist route seems like something we can just make up using reasonable conventions and change later if the community changes their mind down the line. In short, I'd favor pushing this forward either as-is or with a name-based definition.

@ZeLonewolf `overrideByName` was introduced for the Kentucky parkways. I agree that it would be an ideal solution… except that the `name` is just the way name, not the route relation’s name. For tourist routes, the road name isn’t reliably the route relation’s name, although it is in the case of the Cabot Trail.

src/js/shield_defs.js
1ec5: > Regarding the choice of ref on bespoke tourist routes, don't we have `overrideByName` as an option? Then we would be tied to the name, which should be beyond dispute. In any case, the choice of `ref` value on the route relation of an obscure tourist route seems like something we can just make up using reasonable conventions and change later if the community changes their mind down the line. In short, I'd favor pushing this forward either as-is or with a name-based definition.

@ZeLonewolf `overrideByName` was introduced for the Kentucky parkways. I agree that it would be an ideal solution… except that the `name` is just the way name, not the route relation’s name. For tourist routes, the road name isn’t reliably the route relation’s name, although it is in the case of the Cabot Trail.

quincylvania: Yep, here's an illustrative example. Names like "Lighthouse Route" don't always appear in the data, so we're left to rely on a `ref` at the moment.

Screenshot 2023-08-14 at 1 54 06 PM

ZeLonewolf: Yep, makes sense to me - thanks for refreshing how that works. I built the damn logic for those `route_X` attributes and I can't even keep it straight...

1ec5: Let’s ticket out something about a `route_#_name` scheme in OpenMapTiles. Something similar would also help with Spain’s per-route colors in #654.

quincylvania:
This is ready for review btw. The discussion kind of got lost in the code comments but I think everything's been addressed.

src/js/shield_defs.js
```diff
@@ -4208,6 +4211,45 @@ export function loadShields() {
...
+    },
+    DNISD: {
+      spriteBlank: "shield_ca_ns_s_dnisd",
+    },
+    ET: {
+      spriteBlank: "shield_ca_ns_s_et",
+    },
+    FdLt: {
+      spriteBlank: "shield_ca_ns_s_fdlt",
+    },
```
ZeLonewolf, line 4232: This seems to have a case difference with https://www.openstreetmap.org/relation/11147926

src/js/shield_defs.js
```diff
@@ -4208,6 +4211,45 @@ export function loadShields() {
...
+      spriteBlank: "shield_ca_ns_s_md",
+    },
+    MT: {
+      spriteBlank: "shield_ca_ns_s_mt",
+    },
+    ST: {
+      spriteBlank: "shield_ca_ns_s_st",
+    },
+  };
+
```
ZeLonewolf, line 4252: FSSD appears to be missing:
https://www.openstreetmap.org/relation/11147926

ZeLonewolf (changes requested):
Looks good pending the fix for FDLT. We can either add FSSD here or in a separate ticket.

src/js/shield_defs.js
quincylvania: The FSSD is documented on Wikipedia but I couldn't find any signage for it.

src/js/shield_defs.js
quincylvania: Thanks, fixed.

zekefarwell:
So if I understand correctly `ref=BdOLSD` on this route https://www.openstreetmap.org/relation/11051558 isn't a real identifier, but just made up to work around the fact that since no other field from a route relation is available in OpenMapTiles?  I wouldn't call myself a ground truth verifiability purist, but if that's the case this feels like a real stretch.  Seems to me like it would be better to work on improving OpenMapTiles with the desire for these type of scenic route shields as motivation.   I doubt these will be the last scenic routes folks will want to add.

quincylvania:
> So if I understand correctly ref=BdOLSD on this route https://www.openstreetmap.org/relation/11051558 isn't a real identifier, but just made up to work around the fact that no other field from a route relation is available in OpenMapTiles?

I wouldn't go so far as to say an initialism isn't a real identifier, but yes, these `ref` values are just unofficial workarounds (hacks). 

>  Seems to me like it would be better to work on improving OpenMapTiles with the desire for these type of scenic route shields as motivation. I doubt these will be the last scenic routes folks will want to add.

Agreed that something cleaner upstream would be nice, that there will be more cases in the future, and that this is the kind of issue Americana could push the envelope on.

But personally I think this is a "perfect is the enemy of good" scenario and would move forward as-is. We can improve the system later. But up to y'all.

zekefarwell:
> personally I think this is a "perfect is the enemy of good" scenario and would move forward as-is. We can improve the system later.

I also would consider it such a scenario if this was a fully self contained project.  However, we are just one part of the OSM ecosystem, and I don't think we want to be seen as a project that supports questionable tagging practices.  The stakes are quite low in this case, but I think it would be best to wait for a proper solution in OpenMapTiles.

ZeLonewolf:
I think openmaptiles/openmaptiles#1576 is not that far off, if the team is on board with that approach.  While testing it, I did find a pretty major bug in openmaptiles that's causing a massive reduction in shields from z11 and lower.  So my plan is to fix that issue and that push 1576 along.  Since it's been a very long time since the last openmaptiles release. I'm willing to be we're not that far off from a proper solution for ref-less routes.

ZeLonewolf:
Okay, the prerequisite is done.

I could use some design help in figuring out what the implementation for openmaptiles/openmaptiles#1576 should actually do to support these use cases.

ZeLonewolf:
Looks like the deploy preview is hit by actions/first-interaction#10.  I'll plan to do some research as to how that can be resolved.

ZeLonewolf:
The CI issues have been resolved.  The schema update remain unresolved.

icons/shield_ca_ns_s_md.svg
claysmalley: The current shield for `network=CA:NS:H` is only 20px tall because we made it before relaxing the height guidelines for narrow shields. I think the designs introduced in this PR are good for now. If we want to enlarge them later, along with the Arterial shield, let's put that in a new issue.

src/js/shield_defs.js
1ec5: As of #1032, each feature has a `route_#_name` property based on the route relation. The name gets plumbed through to the shield selection code, so you should be able to specify `overrideByName` without worrying about the roadway carrying a different name, and the relations no longer need to be tagged with initialisms as `ref`s.

route_1_network: CA:NS:R; route_1_ref: 322; route_1_name: Route 322; route_2_network: CA:NS:S; route_2_name: Lighthouse Route

1ec5 (commented):
From my perspective, the artwork in this PR is ready to ship, so the only thing that needs to change is migrating from `overrideByRef` to `overrideByName`.

quincylvania:
I've changed this to use `overrideByName` so I think it should be ready to go. I removed the made-up `ref` tags in this changeset: https://www.openstreetmap.org/changeset/150189553

claysmalley (approved):
[Looks good!](https://preview.ourmap.us/pr/912/#map=10.08/46.1729/-60.6214)

image

ZeLonewolf:
Thanks for sticking this one out!  Looks awesome!

quincylvania:
Thanks for running a project that improves tooling and sees things through even if it takes eight months!

PR #913, opened 2023-08-08
Update dependencies

jleedev:
Update sprites parameter (https://github.com/linz/basemaps/pull/2736)


jleedev:
Could probably use their command line API instead, I'm sure there's _a_ reason we scripted this, if not a _good_ reason.

ZeLonewolf (approved):
Looks good here and tested on Ubuntu.

PR #918, opened 2023-09-02
Add shields for US:NH:Turnpike

### LINKED ISSUE #343: NH Turnpike Shields
There are two turnpikes in New Hampshire that should get shields in this project. The Central Turnpike is called the [Everett Turnpike](https://www.openstreetmap.org/relation/13037911) and is signed with 
image
It is partially concurrent with US 3, I-293, and I-93, and it partially stands on its own. It also includes several spurs that are signed with the same shield on mile markers, though they aren't mapped in OSM yet.

The northern portion of the Eastern Turnpike is called the [Spaulding Turnpike](https://www.openstreetmap.org/relation/114529) and is signed with 
image
It's fully concurrent with NH 16, though if you call it Route 16 people will look at it you funny. 
(Images from Wikipedia)

The southern portion of the Eastern Turnpike is called the Blue Star Turnpike, but this name only exists on paper, so I don't think it should be rendered here. It's only signed as I-95. 

I currently have the 3 turnpikes tagged in the [one relation per-direction scheme](https://wiki.openstreetmap.org/wiki/Route_directions#One_relation_per_direction) scheme with the `network` tagged as "US:NH:Turnpike", the `ref` tag unset, and the `name` tag set to the common name (as it appears on the shield for the signed ones). There's no special tagging for the unsigned Blue Star Turnpike yet. Now that I think about it, it probably shouldn't be in OSM at all. Are there other named routes that are totally unsigned that we know about?

And what steps do we have to take to get the shields for the Spaulding and Everett Turnpikes included?

AntiCompositeNumber:
Uses the design by quincylvania in https://github.com/ZeLonewolf/openstreetmap-americana/issues/343#issuecomment-1416529985, but with the top corners squared off to more closely match the actual shield.

Closes #343


src/js/shield_defs.js
```diff
@@ -4285,6 +4288,17 @@ export function loadShields() {
     "Blue Route": bransonRouteShield(Color.shields.blue, Color.shields.white),
   };
 
+  shields["US:NH:Turnpike"].overrideByName = {
+    "Everett Turnpike": {
```
1ec5, line 4292: Just double-checking: can we be sure that each individual way along this route has this as the `name`? Unfortunately, OpenMapTiles only exposes the names of individual ways rather than the route relation as a whole.

1ec5 (commented):
~~Remember to run `npm run taginfo` to add the newly supported tag to the taginfo project file. Thanks!~~

1ec5:
> Remember to run `npm run taginfo` to add the newly supported tag to the taginfo project file. Thanks!

Never mind, this file is generated on the fly. It’s only the shield coverage map that requires a manual step, but this isn’t adding support for shields in a new country. Sorry for the confusion!

src/js/shield_defs.js
AntiCompositeNumber: Looks like the end of the Everett Turnpike needs to be fixed, as well as a toll booth on both turnpikes. I'll take care of that. Otherwise, there isn't a reason for the name to be changed.

AntiCompositeNumber: The way names are all correct now.

1ec5 (approved):
Looks good, thanks!

PR #474, opened 2022-07-04
Add shields for Chile

claysmalley:
Adds shields for national and regional routes of Chile.

In real life, national routes also have a blue variant shield used on motorways _(autopistas)._ Generally, all signage (including shields) on motorways is blue where signage elsewhere is green, similar to the situation in the UK (#106) where one route number may have different shields based on the importance of each segment in the overall hierarchy. Since we don't have a solution to this problem yet, the green shield is probably okay for all national routes.

Screenshot from 2022-07-04 16-04-51

Issue #421, opened 2022-06-15
Pan-American Highway shield


1ec5:
The [Pan-American Highway](https://en.wikipedia.org/wiki/Pan-American_Highway) is an intercontinental highway system analogous to the European E road system (#69) or Asian Highway system (#122). Unlike these other systems, the system itself is signposted with shields rather than the individual mainlines and spurs; no numbers have been assigned.

The overall system is mapped as [a superrelation](https://www.openstreetmap.org/relation/9396001) plus [a massive subrelation for the southern section](https://www.openstreetmap.org/relation/240861) south of the Darién Gap and [another subrelation for the northern section](https://www.openstreetmap.org/relation/1661488) called the Interamerican Highway. The superrelation includes unofficial portions that extend through North America. [Some spurs](https://www.openstreetmap.org/relation/9402626) have also been mapped as subrelations. All these relations lack a `network` tag, which would be a prerequisite for support by this renderer.

Most countries in Central and South America have posted a shield similar to a pre-1960s U.S. Route shield but defaced with a map of the highway running across the Americas. However, the specific shapes and colors [vary by country](https://commons.wikimedia.org/wiki/Category:Diagrams_of_route_signs_of_the_Pan-American_Highway). The unofficial extensions in North America are generally not signposted.

claysmalley:
Example of Venezuelan signage guidelines, thanks to @Wguayana:

![](https://user-images.githubusercontent.com/51529917/174329226-5be480d1-c816-4f51-91df-7a05c930293a.jpg)

Issue #357, opened 2022-06-02
Colored Banners


ahallification:
**Colored Banners**
In Vermont, banners match the text color of the shield they modify. From what I can tell, americana has made no such distinction on this so far, and I'm not sure if it has been discussed. Dependent on the color of the shields and the modifier (which doesn't get a background), I think it would be cool to match the texts. Some colors might not work visually.

Action Item: Can we modify the banneredShield text color for banners in Vermont to match the shield color?
I did a quick review of other states and the only one that jumped out to me that could also be a candidate was South Carolina, which matches their banners to the blue shields. Most other banners seem to be paired with black shields.
![Capture](https://user-images.githubusercontent.com/5446730/171685304-c43ae7f3-cea9-459a-a3a0-d1b560954b93.PNG)
![Capture](https://user-images.githubusercontent.com/5446730/171685656-74c40719-9598-4580-894f-4c2f9f05f24c.PNG)

Outside of state shields, it looks like US:I could be explored as well, although I don't know how it would look with the multiple colored shield.

claysmalley:
Currently, we do this with business state routes in Maryland, by using `colorLighten` on a black-and-white design.

The text halo looks a little more dramatic than I remember when I first added these (#295). @ZeLonewolf changed the text halo in #337.

![Screenshot from 2022-06-02 13-18-29](https://user-images.githubusercontent.com/1732117/171687168-c0867657-c285-4c70-b17d-036a193c4428.png)

claysmalley:
Yeah, it's getting chopped off on the sides and top. Here's 200% zoom:

![Screenshot from 2022-06-02 13-37-04](https://user-images.githubusercontent.com/1732117/171691149-3f581fa0-0bbd-4e21-9974-80468002b18a.png)

ZeLonewolf:
Yuck okay.

So my _intent_ in fixing the halo code was for all banners to be standard black with a white/light halo.  I didn't anticipate introducing the behavior which made banners render in the colorLighten color.  So we need to decide:

1. Do we want colored text banners or all black?
2. If they're colored, how do we handle the halo?  Can it always be white or does it need to differ based on the text color?

Separate from all these, the clipping needs to be addressed, at least on the top.

claysmalley:
> I didn't anticipate introducing the behavior which made banners render in the colorLighten color.

You're good; it did that previously. What I was primarily talking about wrt #337 was the wonky halo.

[...]

PR #922, opened 2023-09-10
Add vintage U.S. Route shield for historic routes

### LINKED ISSUE #921: Historic U.S. Route shield is too subtle
The shield for Historic U.S. Routes is identical to the usual U.S. Route shield but recolored from black to brown. This is too subtle against an off-white background, especially in some color profiles.

[County Route 10 and Historic U.S. Route 66](https://zelonewolf.github.io/openstreetmap-americana/#map=13/35.1172/-114.14085)

We’ve been getting feedback in [OSMUS Slack](https://osmus.slack.com/archives/C01V02K52UX/p1694307360310139) about this shield causing confusion:

> Another one is that US-80 has been decommissioned for decades now. There are a number of relations for the historical route US-80 which apparently are being rendered as a still commissioned route. Would it be possible to have a historical route shield for the historical routes?

1ec5:
Added a vaguely 1948-era U.S. Route shield for Historic U.S. Routes. This shield is the same as the one we’ve been using (just a modern U.S. Route shield recolored from black to brown), but with a crossbar up top, similar to what we’ve done with the Maryland state route shield.

Here’s a before and after:

[Before: Arizona State Route 66 and Historic U.S. Route 66 at Interstate 40 and U.S. Route 93](https://zelonewolf.github.io/openstreetmap-americana/#map=13/35.22067/-114.00642) [After: Arizona State Route 66 and Historic U.S. Route 66 at Interstate 40 and U.S. Route 93](https://zelonewolf.github.io/openstreetmap-americana/#map=13/35.22067/-114.00642)

This treatment is consistent with many states’ signs for these historic routes. Some states go further back to the 1926 style with a blocky typeface, but we haven’t been tailoring shields to each country’s typefaces anyways. Originally, I had considered adding a “HIST” banner at the top, but I think the crossbar is distinctive enough.

[Historic U.S. 165 in Louisiana](https://commons.wikimedia.org/wiki/File:US_165_Historic_Route.JPG)

[Historic U.S. 66 in Arizona](https://commons.wikimedia.org/wiki/File:09_Seligman_2019.jpg)

Fixes #921.


zekefarwell:
Looks great!  

> Originally, I had considered adding a “HIST” banner at the top, but I think the crossbar is distinctive enough.

I think this is a good idea.  For those of us in the know, the brown and the crossbar is distinctive enough, but for those who spend less time thinking about highway shields it's probably not clear what the different style means.  Perhaps the legend is enough to answer that question for folks, but a HIST banner would make it clearer and would match the "historic" wording that is generally on the signs.

ZeLonewolf:
I think the retro shield looks great but it doesn't solve the problem of the brown being too subtle. So I'd say we either need a banner or some way to make the brown be more obvious.

jleedev:
White text on a brown shield? It would not be entirely unrealistic.

1ec5:
> I think this is a good idea. For those of us in the know, the brown and the crossbar is distinctive enough, but for those who spend less time thinking about highway shields it's probably not clear what the different style means. Perhaps the legend is enough to answer that question for folks, but a HIST banner would make it clearer and would match the "historic" wording that is generally on the signs.

I think we’re basically at the point of the difference between an Interstate and an Interstate Business Loop/Spur, where the user will at least notice there’s something unusual about the shield, cuing them to consult the legend. That said, I’m still open to adding a “HIST” banner. It’s just that we only support drawing the banner in black at the moment, which is kind of awkward.

[HIST banner: Arizona State Route 66 and Historic U.S. Route 66 at Interstate 40 and U.S. Route 93](https://zelonewolf.github.io/openstreetmap-americana/#map=13/35.22067/-114.00642)

> White text on a brown shield? It would not be entirely unrealistic.

I like how this would look visually, but it does take some liberties with the shield, which is white in every state.

[White on brown: Arizona State Route 66 and Historic U.S. Route 66 at Interstate 40 and U.S. Route 93](https://zelonewolf.github.io/openstreetmap-americana/#map=13/35.22067/-114.00642)

Note that the former U.S. Route 412 Scenic in Oklahoma was also set against a brown sign back. Idaho also used to signpost the Thousand Islands Scenic Byway [concurrency](https://www.openstreetmap.org/way/142994987) with U.S. 30 (not U.S. 30 Scenic) as [a white-on-brown shield](http://www.aaroads.com/shields/show.php?image=ID19790843&view=4), though ITD has since [replaced it](https://www.bing.com/maps?osid=b997cc64-0290-4f2b-bd7b-5a8e62a47440&cp=42.934467~-114.988675&lvl=17&dir=174.47498&pi=0.16404386&style=x&v=2&sV=2&form=S00027) with a standard U.S. Route shield.

ZeLonewolf:
> I like how this would look visually, but it does take some liberties with the shield, which is white in every state.

While the white on brown may not be perfectly accurate, it very nicely communicates the familiar feel of the historic shield.

zekefarwell:
White on brown looks nice (reminds me of the USFS shields), but given how closely this style replicates real world shields the brown on white seems the better choice to me.  

> I’m still open to adding a “HIST” banner. It’s just that we only support drawing the banner in black at the moment, which is kind of awkward.

Good point.  A brown banner would look better.  Perhaps adding color support for banners can be a future improvement.  There are other shields where the banner matching the primary color would look better and be more accurate. Green on Vermont shields for example:

Screenshot 2023-09-10 at 3 10 05 PM

Screenshot 2023-09-10 at 3 12 23 PM

ZeLonewolf:
> Good point. A brown banner would look better. Perhaps adding color support for banners can be a future improvement. There are other shields where the banner matching the primary color would look better and be more accurate. Green on Vermont shields for example:

Agreed.  This would be a welcome feature for the shield library.  Currently it only accepts a global default color, but what's needed is a per-network override option.  I suspect it would be relatively easy to implement.

1ec5:
> A brown banner would look better. Perhaps adding color support for banners can be a future improvement. There are other shields where the banner matching the primary color would look better and be more accurate. Green on Vermont shields for example:

#357 tracks drawing banners in colors other than black. In the meantime, I’ve added the “HIST” banner in black.

icons/shield_us_us_historic_2.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="20" height="20" xmlns="http://www.w3.org/2000/svg">
```
claysmalley, line 1: These icons might have a generic use in some Latin American countries. Consider a more generic filename like `badge_crossbar`.

1ec5: Done. shield_badge_crossbar_{2,3}.svg is now black on white like shield_badge_{2,3}.svg, and this network’s definition recolors it to brown.

When we get around to implementing Bolivia’s federal routes and Paraguay’s primary routes, we’ll be able to recolor these images to white on green. Some other countries have similar-looking shields that aren’t similar enough to reuse these images:

* Chile’s national routes are shaped differently at the top: https://github.com/ZeLonewolf/openstreetmap-americana/pull/474#discussion_r913828239.
* The Pan-American Highway shield is inspired by the old U.S. Route shield, but it’s defaced with a map of the Americas: https://github.com/ZeLonewolf/openstreetmap-americana/issues/421#issuecomment-1159569738.
* Peru’s shields are stretched vertically.

claysmalley (approved):
I was on the fence about including both the crossbar and the "HIST" banner, but seeing it in context really hammers it home. Looks good to me.

Example location ([production](https://americanamap.org/#map=10.2/35.2184/-101.7296), [localhost](http://localhost:1776/#map=10.2/35.2184/-101.7296)):

image

PR #925, opened 2023-09-17
Don't render one-way arrows on highway features that aren't rendered

dschep:
I noticed that the style is rendering oneway arrows on paths. This PR removes those (and on tracks, as they aren't rendered either)

on `main`:
image

on this branch:
image


src/layer/oneway.js
```diff
@@ -10,6 +10,7 @@ export const surface = {
     ["==", ["get", "oneway"], 1],
     ["!=", ["get", "ramp"], 1],
     ["!", ["in", ["get", "brunnel"], ["literal", ["bridge", "tunnel"]]]],
+    ["!", ["in", ["get", "class"], ["literal", ["path", "track"]]]],
```
1ec5, line 13: Nit: I find this somewhat more readable, though I don’t think it would affect performance:

```suggestion
    ["match", ["get", "brunnel"], ["bridge", "tunnel"], false, true],
    ["match", ["get", "class"], ["path", "track"], false, true],
```

1ec5 (approved):
This looks good to me. Let me know if you agree with the suggested changes; I think it’s fine to merge regardless.

dschep:
I felt that using the same type of expression as directly above (for brunnel) was more readable.

ZeLonewolf:
I think this version is fine, especially since we'll be presumably throwing it away once tracks and paths actually get implemented anyways. Based on @adamfranco's comments on #717, we're not concerned about a conflict with ongoing work.  Therefore, I agree that this is good to merge.  Thanks for the PR!

dschep:
Yep! I'm also hoping this is temporary seeing as I ❤️ trails

PR #934, opened 2023-09-24
Add bridge rendering

### LINKED ISSUE #169: Distinguish highway bridges from non-bridges

claysmalley:
Fixes #169

Bridges are given a fill extrusion just like buildings, but with a red tinge and different heights for different values of `layer`. The same red tinge is applied to an extra casing around linear road and rail bridge features.

## Example locations
[production](http://zelonewolf.github.io/openstreetmap-americana/#map=16.07/41.887307/-87.639202), [localhost](http://localhost:1776/#map=16.07/41.887307/-87.639202)
image

[production](http://zelonewolf.github.io/openstreetmap-americana/#map=16.07/41.887307/-87.639202), [localhost](http://localhost:1776/#map=16.07/41.887307/-87.639202)
image

[production](http://zelonewolf.github.io/openstreetmap-americana/#map=17/40.436284/-79.973147), [localhost](http://localhost:1776/#map=17/40.436284/-79.973147)
image

[production](http://zelonewolf.github.io/openstreetmap-americana/#map=13.53/29.73822/-95.37227), [localhost](http://localhost:1776/#map=13.53/29.73822/-95.37227)
image

### DIFF
--- a/src/layer/bridge.js
+++ b/src/layer/bridge.js
@@ -1,0 +1,123 @@
+"use strict";
+
+import * as Color from "../constants/color.js";
+import * as Util from "../js/util.js";
+
+// Bridge areas
+export const bridge = {
+  type: "fill-extrusion",
+  source: "openmaptiles",
+  "source-layer": "transportation",
+  id: "bridge",
+  minzoom: 13,
+  layout: {
+    visibility: "visible",
+  },
+  paint: {
+    "fill-extrusion-color": Color.bridgeFill,
+    "fill-extrusion-height": ["+", 3, ["coalesce", ["get", "layer"], 0]],
+    "fill-extrusion-opacity": 0.6,
+  },
+  filter: ["all", ["==", ["get", "class"], "bridge"]],
+};
+
+// Bridge casing for highways and railways
+export const bridgeCasing = {
+  type: "line",
+  source: "openmaptiles",
+  "source-layer": "transportation",
+  id: "bridge_casing",
+  minzoom: 13,
+  layout: {
+    "line-cap": "butt",
+    "line-join": "bevel",
+    visibility: "visible",
+  },
+  paint: {
+    "line-color": Color.bridgeFill,
+    "line-opacity": 0.8,
+    "line-width": Util.zoomInterpolate([
+      "match",
+      ["get", "class"],
+      ["rail", "transit"],
+      ["match", ["get", "service"], ["siding", "spur", "yard"], 10, 14],
+      "motorway",
+      ["match", ["get", "ramp"], 1, 18, 28],
+      "trunk",
+      [
+        "match",
+        ["get", "expressway"],
+        1,
+        34,
+        ["match", ["get", "ramp"], 1, 16, 28],
+      ],
+      "primary",
+      [
+        "match",
+        ["get", "expressway"],
+        1,
+        32,
+        ["match", ["get", "ramp"], 1, 15, 24],
+      ],
+      "secondary",
+      [
+        "match",
+        ["get", "expressway"],
+        1,
+        26,
+        ["match", ["get", "ramp"], 1, 14, 20],
+      ],
+      ["tertiary", "busway", "bus_guideway"],
+      [
+        "match",
+        ["get", "expressway"],
+        1,
+        20,
+        ["match", ["get", "ramp"], 1, 13, 18],
+      ],
+      "minor",
+      14,
+      "service",
+      [
+        "match",
+        ["get", "service"],
+        ["alley", "driveway", "drive-through", "parking_aisle"],
+        9,
+        11,
+      ],
+      20,
+    ]),
+  },
+  filter: [
+    "all",
+    ["==", ["get", "brunnel"], "bridge"],
+    [
+      "in",
+      ["get", "class"],
+      [
+        "literal",
+        [
+          "motorway",
+          "trunk",
+          "primary",
+          "secondary",
+          "tertiary",
+          "busway",
+          "bus_guideway",
+          "minor",
+          "service",
+          "rail",
+          "transit",
+        ],
+      ],
+    ],
+  ],
+};
+
+export const legendEntries = [
+  {
+    description: "Bridge",
+    layers: [bridge.id],
+    filter: ["==", ["get", "class"], "bridge"],
+  },
+];
--- a/src/constants/color.js
+++ b/src/constants/color.js
@@ -1,6 +1,8 @@
 export const backgroundFill = `hsl(30, 44%, 96%)`;
 export const backgroundFillTranslucent = `hsla(30, 44%, 96%, 0.8)`;
 
+export const bridgeFill = "hsl(0, 20%, 80%)";
+
 export const waterFill = "hsl(211, 50%, 85%)";
 export const waterFillTranslucent = "hsla(211, 50%, 85%, 0.5)";
 export const waterIntermittentFill = "hsla(211, 60%, 85%, 0.3)";
--- a/src/js/util.js
+++ b/src/js/util.js
@@ -38,3 +38,35 @@ export function zoomMultiply(arr, multiplier) {
   }
   return transformedArray;
 }
+
+//Create a zoom interpolation expression, given width at zoom 20
+export function zoomInterpolate(widthZ20) {
+  return [
+    "interpolate",
+    ["exponential", 1.2],
+    ["zoom"],
+    8,
+    multiplyMatchExpression(widthZ20, 1 / 16),
+    12,
+    multiplyMatchExpression(widthZ20, 1 / 4),
+    20,
+    widthZ20,
+  ];
+}
+
+export function multiplyMatchExpression(value, factor) {
+  if (Array.isArray(value)) {
+    var result = [value[0], value[1]];
+    for (let i = 2; i < value.length - 1; i++) {
+      if (i % 2 == 0) {
+        result.push(value[i]);
+      } else {
+        result.push(multiplyMatchExpression(value[i], factor));
+      }
+    }
+    result.push(multiplyMatchExpression(value[value.length - 1], factor));
+    return result;
+  } else {
+    return value * factor;
+  }
+}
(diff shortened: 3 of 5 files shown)


wmisener:
Cool addition. Just to clarify, this is for rendering `man_made=bridge` areas only, not a halo around ways tagged `bridge=yes` right? Or is it both?

src/layer/bridge.js
```diff
@@ -0,0 +1,123 @@
...
+
+// Bridge casing for highways and railways
+export const bridgeCasing = {
+  type: "line",
+  source: "openmaptiles",
+  "source-layer": "transportation",
+  id: "bridge_casing",
+  minzoom: 13,
+  layout: {
+    "line-cap": "butt",
```
1ec5, line 32: This layer and the main road bridge layer need a line cap of `square`. Otherwise, the bridge area seems to extend beyond the roadway, obscuring a small segment of the roadway past either end of the bridge.

1ec5 (commented):
This is a nifty effect, but I’m not sure yet that it fits very well with the general aesthetic of this style. It moves us away from schematic roads toward something more skeuomorphic. I appreciate that the translucent bridge overlays are pretty much only visible at the high zoom levels where we depict roadways as 1.5D features and we want to eventually build out street-level detail. Maybe skeuomorphism won’t look so out of place eventually. However, at these zoom levels, there’s something awkwardly redundant about giving the roadway both a dark casing and the translucent fill around that. It’s almost as if a bridge sits on top of another bridge. Is there a way to better integrate the two treatments?

Conceptually, I think the bridge areas can be very useful for masking out conflicts between roadways/railroad tracks at different levels: https://github.com/ZeLonewolf/openstreetmap-americana/issues/169#issuecomment-1733025912. However, I don’t think this necessarily entails depicting bridges more or less as buildings. Users are adept at inferring the physical layering across gaps, so I don’t think we really have to make it explicit by extruding the bridges.

1ec5:
> Cool addition. Just to clarify, this is for rendering `man_made=bridge` areas only, not a halo around ways tagged `bridge=yes` right? Or is it both?

It’s both. At low zoom levels, only the casing layer tends to be visible, but at high zoom levels, the casing is still subtly visible within the bridge area.

wmisener:
Just wanted to note here that @adamfranco's track PR implements a "guardrail"/casing approach to track bridges: https://github.com/ZeLonewolf/openstreetmap-americana/pull/717#issuecomment-1416502830. The considerations are slightly different since tracks wouldn't have a casing, but obviously I think we'd want to unify the approaches coherently.

src/layer/bridge.js
```diff
@@ -0,0 +1,123 @@
+"use strict";
+
+import * as Color from "../constants/color.js";
+import * as Util from "../js/util.js";
+
+// Bridge areas
+export const bridge = {
+  type: "fill-extrusion",
```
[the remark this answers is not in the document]
claysmalley: Without a fill extrusion, bridges over buildings looked wonky. I'll see if the guardrail approach improves this.

image

src/layer/bridge.js
claysmalley: I tried that and it looks weirder to me:

`"line-cap": "butt",`
image

`"line-cap": "square",`
image

1ec5: Hm, maybe just the normal bridge roadway layer then?

501Ghost:
> with a red tinge

Will that become yellow for toll roads?

claysmalley:
> > with a red tinge
> 
> Will that become yellow for toll roads?

No. My intent was to make bridges look like buildings, but slightly different. I wasn't considering changing the color of the bridge based on the tagging.

Besides, I'm leaning towards replacing this with a knockout/"guardrail" effect as above, so the color would be a moot point.

claysmalley:
From [#169](https://github.com/ZeLonewolf/openstreetmap-americana/issues/169#issuecomment-1733045248):

> A solution to this would be to compose two different knockouts per bridge: one based on the bridge area and another based on the roadway that’s always wider than the roadway. In turn, there can be two guardrail layers based on the same features. To keep the bridge area–based guardrail layer from looking like an outline, the roadway-based knockout layer can cover up the bridge area–based guardrail layer at the abutments. To avoid redundant guardrails, both guardrail layers would appear beneath both knockout layers.

I much prefer this idea. I've gone ahead reimplemented bridges to be in a knockout/guardrail style. It conflicts a little with my desire for bridges to be slightly transparent at high zoom, but I managed to fudge the guardrail opacity so that it's barely noticeable under a bridge.

## Example locations
[production](http://zelonewolf.github.io/openstreetmap-americana/#map=15.74/41.875563/-87.645814), [localhost](http://localhost:1776/#map=15.74/41.875563/-87.645814)
image

[production](http://zelonewolf.github.io/openstreetmap-americana/#map=16.07/41.887307/-87.639202), [localhost](http://localhost:1776/#map=16.07/41.887307/-87.639202)
image

[production](http://zelonewolf.github.io/openstreetmap-americana/#map=17/40.436284/-79.973147), [localhost](http://localhost:1776/#map=17/40.436284/-79.973147)
image

[production](http://zelonewolf.github.io/openstreetmap-americana/#map=13.53/29.73822/-95.37227), [localhost](http://localhost:1776/#map=13.53/29.73822/-95.37227)
image

wmisener:
I like this effect much better. One minor note, it looks like something odd happens with the overlapping knockouts when the road/bridge curves: https://preview.ourmap.us/pr/934/#map=17.74/34.027658/-118.453153.
Screen Shot 2023-10-01 at 10 02 37 PM
In [OSM](https://www.openstreetmap.org/way/30968236), the bridge ways are glued to the bridge area. I'm guessing this is tricky, and certainly not a blocking issue.

wmisener:
Also, poking around in this PR has already provided some useful layering QA opportunities 😎 . I don't know if any other style correctly layers bridge areas, nice!
roadway wrongly at `layer=2`, bridge area at `layer=1`
Screen Shot 2023-10-01 at 10 09 48 PM

bridge area at `layer=1`, roadway wrongly missing `layer`
Screen Shot 2023-10-01 at 10 10 10 PM

claysmalley:
> One minor note, it looks like something odd happens with the overlapping knockouts when the road/bridge curves: 
> 
> ... 
> 
> In [OSM](https://www.openstreetmap.org/way/30968236), the bridge ways are glued to the bridge area. I'm guessing this is tricky, and certainly not a blocking issue.

It is indeed tricky. I don't really have a solution for it.

1ec5:
> [production](http://zelonewolf.github.io/openstreetmap-americana/#map=17/40.436284/-79.973147), [preview](https://preview.ourmap.us/pr/934/#map=17/40.436284/-79.973147), [localhost](http://localhost:1776/#map=17/40.436284/-79.973147)
> image

The road casing is redundant to the bridge area. Can we deemphasize the road casing at this zoom level? In localities without separately mapped bridge areas, the bridges would be less prominent, but this could be an incentive to map bridge areas.

claysmalley:
> The road casing is redundant to the bridge area. Can we deemphasize the road casing at this zoom level? In localities without separately mapped bridge areas, the bridges would be less prominent, but this could be an incentive to map bridge areas.

Here's my attempt. I didn't touch the opacity of expressway casing because it doesn't look good with link roads.

image

image

1ec5:
Thanks for trying that. Unfortunately it didn’t quite turn out as clear as I had hoped. I’m starting to wonder if the knockout approach is really appropriate at zoom levels where roads are filled. Looking back at https://github.com/ZeLonewolf/openstreetmap-americana/issues/169#issuecomment-1043669724, I do see a pattern where maps only bother with a knockout when the road is stroked.

On this street-scale map of Chicago, only the overpasses carrying minor streets get a knockout with guardrails. Even railroad tracks (in gold on this map) don’t get a knockout, presumably because they already contrast well enough against the green freeways:

Eisenhower Expressway

This Minnesota Department of Transportation map treats a freeway as a pair of independent stroked roads (similar to what aaroads-wiki#2 does with expressways). The freeways are red but have black guardrails:

MNDOT 2019

This black-and-white map of Springfield, Ohio, does combine road casings with bridge knockouts and guardrails. It’s exquisite linework. I think the only thing keeping the casings from looking redundant is that the guardrails are guardrails – they don’t form a closed polygon:

I-70

This map of New York City is possibly the only one in my collection that tries to depict the shape of each overpass – but still not as closed polygons. It would’ve been a lot more confusing with polygons:

New York City & Long Island Street Map & Visitors Guide by American Map Corporation

ZeLonewolf:
This is a tough one, and it scratches at whether we even render these polygons.  I do think we should at the highest zooms at least.  I wonder if perhaps a color treatment more reminiscent of pavement might better satisfy what we're trying to show.

ZeLonewolf:
I achieved the following effect with:

```
export const bridgeFill = "hsl(248, 25%, 85%)";
export const bridgeBackgroundFill = "hsl(248, 25%, 36%)";
```

image

The slightly bluer effect seemed more pavement-y.  Experimenting, I think we probably don't want this to show up until z17, otherwise it feels like too much clutter.

I'm not really a fan of how the background fill creates breaks when two adjacent bridge areas have separate levels.  I tried a straight knockout effect, which looks decent on the sides but has an even more pronounced effect on the bridge joins:

image

claysmalley:
Bridge areas are now hidden at z<15, and the guardrail is now a knockout at z>15.

image

image

PR #942, opened 2023-10-05
Script for generating render samples

ZeLonewolf:
This PR adds a script for generating render samples, using a configuration file to specify the lat/lon/zoom location of the sample and the clipping window to use.

This script presumes the server to be running locally on port 1776, which is the default.

Usage: `npm run generate_samples -- list_of_map_locations.json`

This is a first step that will enable CI integration of map render samples. Ultimately I would like the CI to show images of places in the map that have changed for a PR by generating a folder of images for both main and the PR branch. Then, whichever images are different can be loaded into a GitHub check so we can see visually what changes and we can start a practice of adding new locations to the samples list when we make changes. However, for any of this to work, we first need the ability for main to generate sample images.

Below are the three sample images presently provided in the config file as a starting point:

full_window
downingtown_alt_trk_bus
historic_us66_oklahoma


scripts/generate_samples.ts
```diff
@@ -0,0 +1,88 @@
+import fs from "fs";
```
jleedev, line 1:
```suggestion
import fs from "node:fs";
```

Becoming the preferred style for imports that are part of nodejs and not coming from npm, for some reason (something about interoperability).

scripts/generate_samples.ts
```diff
@@ -0,0 +1,88 @@
...
+
+type LocationClip = {
+  x: number;
+  y: number;
+  width: number;
+  height: number;
+};
+
+type SampleSpecification = {
+  location: string;
```
jleedev, line 21: Document these? location is `z/lon/lat`, for instance. They're all apparently required, are there defaults?

The web page has a bunch of stuff on it, I think clip is intended to select a rectangle from the middle but it might be nice to have a web page without the stuff on it so that clip can be "the whole viewport".

scripts/generate_samples.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  await createImage(screenshot);
+}
+
+async function createImage(screenshot: SampleSpecification) {
+  await page.goto(`http://localhost:1776/#map=${screenshot.location}`);
+
+  // Wait for map to load, then wait two more seconds for images, etc. to load.
+  try {
+    await page.waitForFunction(() => window.map && window.map.loaded(), {
```
jleedev, line 64:
```suggestion
    await page.waitForFunction(() => window.map?.loaded(), {
```

scripts/generate_samples.ts
```diff
@@ -0,0 +1,88 @@
+import fs from "fs";
+import { chromium } from "@playwright/test";
+
+// Declare a global augmentation for the Window interface
+declare global {
+  interface Window {
+    map: {
```
jleedev, line 7:
```suggestion
    map?: {
```

Based on usage below this could be undefined.

scripts/generate_samples.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  try {
+    await page.screenshot({
+      path: `${sampleFolder}/${screenshot.name}.png`,
+      type: "png",
+      clip: screenshot.clip,
+    });
+    console.log(`Created ${sampleFolder}/${screenshot.name}.png`);
+  } catch (err) {
+    console.log(err);
```
jleedev, line 84:
```suggestion
    console.error(err);
```

scripts/generate_samples.ts
```diff
@@ -0,0 +1,88 @@
...
+
+fs.mkdirSync(sampleFolder, { recursive: true });
+
+const browser = await chromium.launch({ headless: true });
+const context = await browser.newContext();
+
+const page = await context.newPage();
+
+for (const screenshot of screenshots) {
+  // Set viewport and double deviceScaleFactor to get a closer shot of the map
```
jleedev, line 53: I don't see deviceScaleFactor.

package.json
```diff
@@ -24,6 +24,7 @@
     "code_format:prettier": "prettier --write --list-different .",
     "code_format:svgo": "svgo -q -f icons/",
     "extract_layer": "node scripts/extract_layer",
+    "generate_samples": "npx playwright install && ts-node scripts/generate_samples.ts",
```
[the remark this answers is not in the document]
ZeLonewolf: I did an npm install of playwright into dev and when I ran playwright, it told me to do this npx install command which seems to download and install the web browser? So not sure how that's supposed to work?

scripts/generate_samples.ts
ZeLonewolf: Should we just have a separate web page without the controls that we can use as an option and that drop the clip entirely?  That would certainly simplify this config.

package.json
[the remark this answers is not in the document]
ZeLonewolf: Specifically, I did something like `npm install --save-dev @playwright/test` and when I tried to run playwright it gave me the message telling me to run that npx command.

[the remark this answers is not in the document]
jleedev: Ohhhhh ok I played with it a little. It seems that you can either depend on `playwright` or `@playwright/test`, the latter includes a whole test framework, but both supply a bin named `playwright`.

I'd suggest not using `npx` or `npm exec` inside of the `scripts` section as it's basically implied -- it already sets PATH to the path to installed things. `npx` can be used to install things globally, which is how this read to me at first. 

But also, `playwright install` downloads a whole gigabyte of browsers that you don't need and for the wrong architecture.

```suggestion
    "generate_samples": "ts-node scripts/generate_samples.ts",
```

and more documentation, but also it gives a nice error asking you to install.

dev/README.md
```diff
@@ -11,3 +11,11 @@ The highway shield [color palette](americana.gpl) can be imported into Inkscape
...
 4. Restart Inkscape.
 5. Click the ◀ button to the right of the color palette strip at the bottom of the window (or in the top-right corner of the Color Palette panel), then choose americana.gpl from the menu.
+
+## Map sample images
+
+Map sample images can be generated with a script. See [sample_locations.json](test/sample_locations.json) for the format.
+
+1. Create a JSON file with the map location and clipping rectangles
+2. Start the server in the background with `npm start &`
+3. Run the generate_samples script and pass the JSON file location: `npm run generate_samples -- test/sample_locations.json``
```
jleedev, line 21:
```suggestion
3. Configure playwright: either `npx playwright install chromium` or `export CHROME_BIN=/usr/bin/chromium`
4. Run the generate_samples script and pass the JSON file location: `npm run generate_samples -- test/sample_locations.json``
```

scripts/generate_samples.ts
```diff
@@ -0,0 +1,88 @@
...
+const jsonSampleLocations = process.argv[2] ?? "test/sample_locations.json";
+
+console.log(`Loading sample locations from ${jsonSampleLocations}`);
+
+const screenshots: SampleSpecification[] =
+  loadSampleLocations(jsonSampleLocations);
+
+fs.mkdirSync(sampleFolder, { recursive: true });
+
+const browser = await chromium.launch({ headless: true });
```
jleedev, line 47:
```suggestion
const browser = await chromium.launch({
  headless: true,
  executablePath: process.env.CHROME_BIN,
});
```

jleedev (commented):
Taking `playwright install` out of the script.

scripts/generate_samples.ts
ZeLonewolf: Will this work for both documented methods of `npx playwright install chromium` and `export CHROME_BIN=/usr/bin/chromium` ?

jleedev: Yes, I tested it; we could omit the variable if it's undefined but that seems unnecessary.

jleedev (approved):
Looks good, we can streamline the html page and json formats later.

PR #945, opened 2023-10-07
Improved render sample config

### LINKED ISSUE #944: Simplify format for sample generation script
Should we just have a separate web page without the controls that we can use as an option and that drop the clip entirely?  That would certainly simplify this config.

_Originally posted by @ZeLonewolf in https://github.com/ZeLonewolf/openstreetmap-americana/pull/942#discussion_r1347439745_

ZeLonewolf:
Fixes #944

In #942, we added the ability to generate render samples.  This PR improves the render sample functionality by replacing the "clip" function with a boolean that lets you configure whether or not to include the Americana controls in the sample clip.

This required the `americana.js` to be refactored in order to create "bare" and "fully featured" html versions of the map.  Then, when a user specifies whether to use the controls, the screenshot generator simply invokes either the "bare" or "fully featured" html pages as needed.

I created a "bare" version of `americana.js` and `index.html`, and refactored out common code into `src/js/map_builder.ts`.

One additional improvements to render samples not included in this PR that I would like to explore is an event-driven way to check that the map has loaded, rather than a fixed 500ms delay, which has a timeout risk.

The following files are samples generate with the new version of the script:

full_window
downingtown_alt_trk_bus
historic_us66_oklahoma

### DIFF
--- a/src/bare_americana.js
+++ b/src/bare_americana.js
@@ -2,7 +2,7 @@
 
 import "maplibre-gl/dist/maplibre-gl.css";
 
-import { createMap } from "./js/map_builder.js";
+import { createMap, loadRTLPlugin } from "./js/map_builder.js";
 
 loadRTLPlugin();
 
--- a/src/bare_americana.js
+++ b/src/bare_americana.js
@@ -2,16 +2,20 @@
 
 import "maplibre-gl/dist/maplibre-gl.css";
 
-import { createMap, loadRTLPlugin } from "./js/map_builder.js";
+import { createMap, loadRTLPlugin, buildStyle } from "./js/map_builder.js";
 
 loadRTLPlugin();
 
-export const map = createMap(
-  window,
-  (shields) => shieldDefLoad(),
-  [-94, 40.5],
-  4
-);
+export const map = createMap(window, (shields) => shieldDefLoad(), {
+  container: "map", // container id
+  hash: "map",
+  antialias: true,
+  style: buildStyle(),
+  center: [-94, 40.5],
+  zoom: 4,
+  fadeDuration: 0,
+  attributionControl: false,
+});
 
 function shieldDefLoad() {
   if (window.top === window.self) {
--- a/src/bare_americana.js
+++ b/src/bare_americana.js
@@ -1,0 +1,27 @@
+"use strict";
+
+import "maplibre-gl/dist/maplibre-gl.css";
+
+import { createMap } from "./js/map_builder.js";
+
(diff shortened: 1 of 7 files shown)


jleedev (approved):
I like the new entry point. I have some concerns about getting this to load, timeouts, etc. but they are unrelated. The timeouts can probably be removed.

ZeLonewolf:
> I like the new entry point. I have some concerns about getting this to load, timeouts, etc. but they are unrelated. The timeouts can probably be removed.

I tried removing the timeouts, but there seems to be some fade-in effect that's happening with the shields:

downingtown_alt_trk_bus

It feels like maybe there's some animation setting that can be disabled but I'm not sure what.

jleedev:
Yes:

```js
new maplibre.Map({
  fadeDuration: 0,
})
```

Rather than `center` and `zoom`, map_builder should just accept the `maplibre.MapOptions`.

Perhaps `page.evaluate('map.once("idle")')` would work from a "things fading in side", but the page hung and consumed a gig of memory when I tried.

ZeLonewolf:
I was at least able to turn off the 500ms delay, that seems to work without a hitch.

src/bare_americana.js
```diff
@@ -2,16 +2,20 @@
...
-  4
-);
+export const map = createMap(window, (shields) => shieldDefLoad(), {
+  container: "map", // container id
+  hash: "map",
+  antialias: true,
+  style: buildStyle(),
+  center: [-94, 40.5],
+  zoom: 4,
+  fadeDuration: 0,
```
1ec5, line 16: The default crossfade is nice when switching languages. Label placement sometimes differs dramatically from one zoom level to another; the fade duration mitigates that effect. We should get to the bottom of why you were seeing the hung crossfade so we can reenable that functionality.

ZeLonewolf: The cross fade isn't hung, it just takes up unnecessary time in a headless utility use case. The fade is only disabled for the purpose of generating render samples.

jleedev: The docs say

> fade-in/fade-out animation for label collisions **after** initial map load

I imagine it's load, 300ms, idle.

If the "bare" map page is useful for something other than driving the screenshot script then a more flexible setup would be needed.

1ec5: Ah, I misunderstood. This is perfectly reasonable for a headless setup. That said, we probably should still add the attribution no matter what. It can be smaller if necessary.

PR #955, opened 2023-10-07
[demo] use Oswald web font on shields

claysmalley:
PR for preview purposes only; see discussion [here](https://github.com/ZeLonewolf/openstreetmap-americana/discussions/954)

### DIFF
--- a/shieldlib/src/screen_gfx.ts
+++ b/shieldlib/src/screen_gfx.ts
@@ -4,7 +4,7 @@ import rgba from "color-rgba";
 
 const defaultFontFamily = '"sans-serif-condensed", "Arial Narrow", sans-serif';
 export const shieldFont = (size: number, fontFamily: string) =>
-  `bold ${size}px ${fontFamily || defaultFontFamily}`;
+  `500 ${size}px ${fontFamily || defaultFontFamily}`;
 export const fontSizeThreshold = 12;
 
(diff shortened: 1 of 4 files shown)


1ec5:
There seems to be a race condition in which the font loads after some of the shields have been rendered and cached in the sprite sheet. This randomly affects different shields each time I reload the page in Firefox:

U.S. Route 35 in a lighter font
U.S. Route 35 in a heavier font

Each browser engine has slightly different behavior around Web fonts. If I remember correctly, WebKit generally draws blank text until the font loads, which would be worse in a case like this.

We might want to [listen for a font loading event](https://developer.mozilla.org/en-US/docs/Web/API/FontFaceSet/loadingdone_event) and trash and repopulate the shields in the sprite sheet at that time.

zekefarwell:
I think the issue may be that different font weights are being used randomly.  Four weights are loaded from google fonts:
https://github.com/ZeLonewolf/openstreetmap-americana/blob/a70c4c7bb3564e3c79eabbb77ffebb333566cc74/src/index.html#L9
I'm not sure why this would be happening, but the number shapes all look the same to me just randomly varying by weight with each page load.

zekefarwell:
> the number shapes all look the same to me

Scratch that.  After some testing I've seen varying number shapes too.  Like these different 9s here:
Screenshot 2023-10-08 at 12 08 33 PM

PR #956, opened 2023-10-08
Show before-and-after sample images in GitHub workflow

### LINKED ISSUE #958: Generate side-by-side sample clips of PR changes
**Proposal**:

GitHub actions should generate side-by-side images of PR changes, for example:

| Sample Name            | Before                                                    | After                                                   |
| ---------------------- | --------------------------------------------------------- | ------------------------------------------------------- |
| historic_us66_oklahoma | historic_us66_oklahoma_before | historic_us66_oklahoma_after |

For demonstration purposes, the banner on the Historic US 66 shield has been changed to a nonsense value. Also note for the real implementation, I intend to include the attribution notice in the sample clip as there's a strong likelihood that they get copied around.

**Design**:

The repository will hold a configuration file which contains a list of "interesting" locations which will be used by a GitHub action to generate screenshots of that location on the map. The action will generate screenshots for both the main branch and the PR branch and collect a set of images which have differences between the two branches. The action will use the PR version of the screenshot configuration so that it's the same one-to-one comparison.

Then, only those locations which have differences will be added to the "PR Preview" check after the existing content:

image

**Process Changes**:

Once we have this functionality in place, we should encourage PR creators to add sample locations to the file when new features are added.  This will help us visually catch regressions in future PRs and simplify the process of showing changes.  I anticipate adding instructions to this effect in our contributor guide.

**Implementation**:

#942 and #945 added the ability to render samples from a script, at the command line. I intend to use this functionality along with additional scripting to run the comparisons and update the appropriate GitHub action. Since this will require changes to the "privileged" GitHub action, a portion of the implementation will need to include un-previewable changes to main that I will commit separate from the in-progress PR #956 that's implementing this capability.

I'm looking for concurrence on this approach given that I'll need to push a change to main to enable the changes to the privileged action script.

ZeLonewolf:
Fixes #958 

This PR add before and after image samples to the "PR Preview" GitHub check tab.  It only shows locations where there is a change in rendering.  The sample locations are defined in tests/sample_locations.json. The before and after images are generated in userland, uploaded to s3, while a markdown file with the change is pushed over the the privileged CI runner.

I had to change the GitHub action to run the before and after builds in separate folders so they wouldn't stomp on each other.  I also use the PR branch's render test list and apply it to both the main and PR branch.  This ensures that we have apples to apples comparisons. For this reason, the comparison and clipping scripts are also run out of PR branch code.

I also tweaked some of the sample location sizes and added a dense area of NYC and a recently located 7-way concurrency because Americana :)

I've also made a change the Historic US-66 to demonstrate the change in the checks tab, however, I will remove that change prior to merging this PR.

Below is what it looks like in the checks tab, and you can click down in the checks to view it. There are links to the location in the current demo map as well as the PR preview map.

image

If we think it's useful, a future PR could also generate an image diff to show the pixel differences, perhaps using [pixelmatch](https://www.npmjs.com/package/pixelmatch).


.github/workflows/build-preview.yml
```diff
@@ -39,10 +39,10 @@ jobs:
       - name: Capture main branch sample clips
         id: main-samples
         run: |
-          npm start & command_pid=$!
-          sleep 15
+          npm start &
+          while ! nc -z localhost 1776; do sleep 1; done
           npm run generate_samples
-          kill $command_pid
+          kill $(lsof -t -i:1776) 2>/dev/null || echo "No process found listening on port 1776."
```
jleedev: man bash
/jobspec

`kill %`

.github/workflows/build-preview.yml
```diff
@@ -39,10 +39,10 @@ jobs:
       - name: Capture main branch sample clips
         id: main-samples
         run: |
-          npm start & command_pid=$!
-          sleep 15
+          npm start &
+          while ! nc -z localhost 1776; do sleep 1; done
```
jleedev, line 43: until

.github/workflows/build-preview.yml
ZeLonewolf: I found that the PID of `npm start` is not actually the PID of the :1776 web server, which is why I changed it to this lsof business.

.github/workflows/s3-cleanup.yml
```diff
@@ -11,9 +11,6 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v2
-
```
ZeLonewolf, line 16: Removed because checking out code is not needed in this cleanup workflow.

src/js/shield_defs.js
```diff
@@ -500,7 +500,7 @@ export function loadShields() {
       textColor: Color.shields.brown,
       colorLighten: Color.shields.brown,
     },
-    ["HIST"],
+    ["XYZA"], //Change to be removed before merging
```
ZeLonewolf, line 503: I will remove this change before merging.  Leaving it in for now so we can see an example of a map change.

claysmalley:
Looks like the PR preview isn't pulling in the web font. The Interstate 40 shield has the wrong width.

scripts/generate_samples.ts
```diff
@@ -92,7 +91,7 @@ async function createImage(screenshot: SampleSpecification) {
 }
 
 function delay(ms: number): Promise<void> {
```
jleedev, line 93:
```suggestion
import { setTimeout as delay } from "node:timers/promises";
```

ZeLonewolf:
This PR incorporates changes from #991.

src/js/shield_defs.js
```diff
@@ -500,7 +500,7 @@ export function loadShields() {
       textColor: Color.shields.brown,
       colorLighten: Color.shields.brown,
     },
-    ["HIST"],
+    ["XYZA"],
```
ZeLonewolf, line 503: This change will be reverted prior to merging. I changed it in order to demonstrate what a before-and-after sample would look like in the build actions.

test/sample_locations.json
```diff
@@ -12,16 +12,24 @@
...
+      "width": 400,
+      "height": 400
+    }
+  },
+  {
+    "location": "15/33.92208/-83.37654",
+    "name": "athens_ga_7_way_concurrency",
+    "viewport": {
+      "width": 400,
+      "height": 400
```
ZeLonewolf, line 32: I added this because it's a good sample location for multiple-concurrency shields and shield coloration.

akadouri:
Is `https://github.com/ZeLonewolf/openstreetmap-americana/pull/956/checks?check_run_id=19668104694` supposed to have a preview of all the sample locations?

ZeLonewolf:
The preview is intended to only show sample locations for places where the PR changes the rendering from what's in the main branch.

ZeLonewolf:
Thanks @akadouri for your review, are you good with merging this PR?

PR #962, opened 2023-10-12
Workspace tweaks

jleedev:
Miscellaneous fixes:

- Don't declare a dependency on a `file:…` version of something. I think this sort of thing should never be checked in, ever, but I'm not sure. As long as the declared dependency is semver-compatible with the version present in the workspace, npm install will symlink the dependency from node_modules to the workspace. Depending on `*` will guarantee this, although other techniques exist in the wild. (We're not publishing much to npm so I'm not really concerned).
- Remove npx, exec, and npm exec from scripts. These either do nothing and are redundant or allow the possibility of wrong things, such as finding globally installed packages; exec also confuses yarn's shell implementation.
- Instead of a literal `npm run` in a script, delegate this to `npm-run-all` via `run-s`, which is clever enough to spawn a yarn process.
- Add missing (peer-)dependencies on mocha and @types/node as needed to hush more warnings.
- Remove explicitly declared dependency on two conflicting versions of glob, which I'm honestly amazed that npm is willing to put up with at all.

(What follows is mostly ranting.)

This improves compatibility with yarn and pnpm. These workflows should work:

1. \>pnpm-workspace.yaml; pnpm install; npm run build
2. yarn config set nodeLinker pnpm; yarn install; yarn run build
3. yarn config set nodeLinker node-modules; yarn install; yarn run build

And they are _much_ faster than regular npm, and produce correct results instead of incorrect results.

This might work someday but it is experimental:

4. npm config -L project set install-strategy=linked; npm install; npm run build

(In particular, it seems to run install scripts _multiple times_ for the _exact same version_. And then fails to actually link the workspace dependency, so nothing works. Maybe someday. The other documented install-strategies of nested and shallow don't seem to do anything at all, for some reason.)

Yarn Plug'n'Play doesn't work, and it's weird, and slow:

5. yarn config set nodeLinker pnp; yarn install; yarn run build

Yarn Plug'n'Play has some interesting ideas, but it monkeypatches node's fs module and breaks when we need to use a module loader with ts-node. What's more, yarn, pnpm, and esbuild don't seem to quite do the same thing in PnP mode, leading to breakage.

Incremental use of npm install can leave things in a potentially inconsistent state. npm ci generally works, but deleting and recreating node_modules is not fantastic, especially when binary dependencies need to be rebuilt every time. Both pnpm and yarn are capable of converging node_modules to a consistent state without having to destroy and rebuild the entire thing. pnpm even seems to cache the built binary dependencies external to the project's node_modules.


ZeLonewolf (approved):
Everything seems to be in order and works great here.  The only thing I don't know how to verify is whether we're actually pulling in the local shieldlib or if it's pulling from npm.

jleedev:
> Everything seems to be in order and works great here. The only thing I don't know how to verify is whether we're actually pulling in the local shieldlib or if it's pulling from npm.

```sh
~/pkg/openstreetmap-americana ¶
[ ../../shieldlib = "$(readlink ./node_modules/@americana/maplibre-shield-generator)" ]
```

If it's a directory, or a symlink pointing anywhere else, then it was installed from npm (or yarn/pnpm installed the file: dependency into node_modules, which is annoying as it will not have been built yet).

PR #965, opened 2023-10-18
Change shield font to Noto Sans Condensed

### LINKED ISSUE #834: Two-digit Interstate shields are too wide

claysmalley:
Fixes #834

Based on discussion in #954, Noto Sans Condensed satisfies all our needs, and matches the existing style. This PR changes the shield font to Noto Sans Condensed, and updates various shield definitions to accommodate the change.

I've set up a repository for web fonts at [osm-americana/web-fonts](https://github.com/osm-americana/web-fonts). If we want to support route networks that incorporate other scripts like Georgian and Arabic, we'll need to add their corresponding fonts.


icons/shield_ca_mb_winnipeg.svg
claysmalley, line 1: Replaced with a JS-drawn trapezoid. I probably should have done this in #585, honestly.

icons/shield_ca_ns_h.svg
claysmalley, line 1: Replaced with a JS-drawn escutcheon. It was either that, or vertically stretch the SVG.

icons/shield_ca_tch_2.svg
claysmalley, line 1: Visually identical to the 'deleted' `shield_ca_tch.svg`. Drawn in a different way to make widening easier.

src/fonts.css
```diff
@@ -0,0 +1,18 @@
+@font-face {
+  font-family: "Noto Sans Condensed";
+  src: url("https://osm-americana.github.io/web-fonts/noto/NotoSans-CondensedMedium-PropNums.woff2")
```
1ec5, line 3: I opened osm-americana/web-fonts#1 as tail work for automating the creation of these files.

src/index.html
```diff
@@ -3,6 +3,8 @@
   <head>
     <meta charset="utf-8" />
     <title>OpenStreetMap Americana</title>
+    <link rel="preconnect" href="https://osm-americana.github.io" />
+    <link href="fonts.css" rel="stylesheet" />
```
1ec5, line 7: Is the https://github.com/ZeLonewolf/openstreetmap-americana/pull/955#issuecomment-1751888674 still possible, but less likely? Or should we continue to investigate a JavaScript-based alternative to the CSS stylesheet that can respond to font loading events?

1ec5 (approved):
The shield definition changes look great to me at a glance. Thanks for facilitating the discussion around choosing a font. I haven’t heard much enthusiasm for continuing to rely on system fonts, so at least we can standardize on this font while we pursue further improvements either to the font selection or to metrics in #768.

src/fonts.css
```diff
@@ -0,0 +1,18 @@
...
+      format("woff2"),
+    url("https://osm-americana.github.io/web-fonts/noto/NotoSans-CondensedMedium-PropNums.woff")
+      format("woff");
+  font-weight: 500;
+  font-style: normal;
+}
+@font-face {
+  font-family: "Noto Sans Armenian Condensed";
+  src: url("https://github.com/osm-americana/web-fonts/raw/main/noto/NotoSansArmenian-CondensedMedium-PropNums.woff2")
```
ZeLonewolf, line 12: Rather than pointing to the raw file, we should deploy to github.io and point to that location.

icons/shield_kr_expressway_2.svg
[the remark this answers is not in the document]
claysmalley: The Korean expressway shield bulges out in a different way. I think our logo is fine :sweat_smile:

src/shieldtest.js
```diff
@@ -151,10 +150,10 @@ let networks = [
 
   // Fancy Rectangles
   "US:AK",
+  "US:ID",
   "US:KY:Parkway",
```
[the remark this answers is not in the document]
zekefarwell: Oh that is not so great for this use case.  Looks like an alternate J is available as an open type feature.  We can probably freeze it into the web font just like the proportional numerals.
https://github.com/notofonts/latin-greek-cyrillic/issues/377

src/index.html
```diff
@@ -96,6 +98,15 @@
...
   <body>
+    <p
+      style="
+        font-family: 'Noto Sans Condensed';
+        font-weight: 500;
+        visibility: hidden;
+      "
+    >
+      Invisible text so the font will load early
+    </p>
```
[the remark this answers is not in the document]
zekefarwell: I'm sure there must be a way to ensure the font is loaded before the first shield is rendered without relying on a hidden text hack.

src/index.html
zekefarwell: Yes still possible.  Try shield test with browser cache disabled and you'll see a fallback font for the first few rows.

src/fonts.css
zekefarwell: Shouldn't the font files be pulled in by the build script so they are served from the same domain as the rest of the style?

src/shieldtest.js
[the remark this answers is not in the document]
claysmalley: Issue tracked here: osm-americana/web-fonts#3

PR #972, opened 2023-10-21
limit oneway arrows to currently supported highway features

claysmalley:
Currently, oneway arrows appear on railroads in Americana. `oneway=yes` on `railway=*` is a tagging error mistakenly exposed on OpenMapTiles, and `railway:preferred_direction=forward|backward` is the correct tag for a track on which trains only travel in one direction in most circumstances. This PR limits oneway arrows to highway features, preventing them from showing up on railroads.

[production](http://zelonewolf.github.io/openstreetmap-americana/#map=15.56/35.691993/139.700458), [preview](https://preview.ourmap.us/pr/972/#map=15.56/35.691993/139.700458), [localhost](http://localhost:1776/#map=15.56/35.691993/139.700458)

before:
image

after:
image


ZeLonewolf:
Is there a corresponding ticket in OMT that should be opened?

ZeLonewolf (approved):
Looks good.  I agree with being explicit about the classes with oneway treatment.  Once we have finer control over the tile schema, we might de-optimize that check and handle this on the tile side.  The layer size increase is not significant (~400 bytes).

PR #973, opened 2023-10-21
add shields for Albany County, Wyoming

### LINKED ISSUE #968: Albany County Wyoming Shields
County roads in Albany County, WY (proposed network US:WY:Albany) are signed with a mix of either the standard US CR pentagon or on the streetblades themselves (green square with white border). There doesn't appear to be one dominant way over the other from my knowledge (lived in Laramie for two years). I would recommend using the pentagons to display all of them, but if you want me to break them out into two separate networks I can do that as well (US:WY:Albany1 and US:WY:Albany2 for example).

claysmalley:
Fixes #968


ZeLonewolf:
Albany county: https://www.openstreetmap.org/relation/1411398

ZeLonewolf:
I've added three route networks:
https://overpass-turbo.eu/?w=%22network%22%3D%22US%3AWY%3AAlbany%22+global&R#

ZeLonewolf:
For some reason, the route data is not showing up in the tiles, even though it looks like the tiles have updated.
Location: https://tile.ourmap.us/inspect.html#14/41.27212/-105.57734
These roads are tertiary, which should be fine for showing shields.

ZeLonewolf (approved):
Looks good!

PR #975, opened 2023-10-23
Remove PR preview files when PRs are closed

ZeLonewolf:
This PR removes preview files from s3 once a PR is merged.  Images in the checks tab will still remain, since those are archived by GitHub's content system, including the image diffs which will be provided by #956. However, links to live previews of merged PRs will become dead.

While s3 is cheap, there's really no need to keep old previews around once a PR has been merged.

### DIFF
--- a/.github/workflows/deploy-pr-checks.yml
+++ b/.github/workflows/deploy-pr-checks.yml
@@ -98,12 +98,14 @@ jobs:
           interval: 500
       - name: Generate Preview text
         run: |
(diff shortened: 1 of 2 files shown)


claysmalley (approved):
I don't see why not.

ZeLonewolf:
Check tabs imagery behavior verified in https://github.com/ZeLonewolf/openstreetmap-americana/pull/974/checks?check_run_id=17938611920

PR #982, opened 2023-10-28
add South Dakota secondary shields

### LINKED ISSUE #970: South Dakota "secondary" route shields
So South Dakota doesn't have an official secondary route system, but some minor highways (namely 251, 258, and 473 from my memory) are signed using a small white rectangle with a black outline. Is this worth splitting from the US:SD system (proposed network would be US:SD:Secondary if this moves forward)?

claysmalley:
Fixes #970

[example location](https://preview.ourmap.us/pr/982/#map=13.21/43.20364/-99.43508)


ZeLonewolf (approved):
Looks good to me!

PR #984, opened 2023-10-28
Guard against empty search results

jleedev:
When pressing Enter in the search box, if there are no results (whether still loading or if there are no results) then unwanted things happen: The form submits and the page navigates, or the text box is sadly cleared without any other feedback (and an exception is thrown).

If there are no results, do nothing when Enter is pressed (leaving the typed entry intact).


ZeLonewolf (approved):
Looks good, thanks!

PR #987, opened 2023-10-30
Load map after font load

ZeLonewolf:
This PR waits to load maplibre until web fonts have been downloaded.  The allows the shield library to generate shield sprites only after the necessary fonts are in place.

### DIFF
--- a/src/americana.js
+++ b/src/americana.js
@@ -41,18 +41,7 @@ const initializeMap = async () => {
   });
 };
 
-let map;
-
-if ("fonts" in document) {
-  document.fonts.ready.then(() => {
-    map = initializeMap();
-  });
-} else {
-  // Font Loading API not supported, run the code immediately
-  map = initializeMap();
-}
-
-export { map };
+export const map = await initializeMap();
 
 let options = {};
 
--- a/src/americana.js
+++ b/src/americana.js
@@ -28,15 +28,30 @@ upgradeLegacyHash();
 
 loadRTLPlugin();
 
-export const map = createMap(window, (shields) => shieldDefLoad(shields), {
-  container: "map", // container id
-  hash: "map",
-  antialias: true,
-  style: buildStyle(),
-  center: [-94, 40.5],
-  zoom: 4,
-  attributionControl: false,
-});
+const initializeMap = () => {
+  return createMap(window, (shields) => shieldDefLoad(shields), {
+    container: "map", // container id
+    hash: "map",
+    antialias: true,
+    style: buildStyle(),
+    center: [-94, 40.5],
+    zoom: 4,
+    attributionControl: false,
+  });
+};
+
+let map;
+
+if ("fonts" in document) {
+  document.fonts.ready.then(() => {
+    map = initializeMap();
+  });
+} else {
+  // Font Loading API not supported, run the code immediately
+  map = initializeMap();
+}
+
+export { map };
 
 let options = {};
 
--- a/src/americana.js
+++ b/src/americana.js
@@ -28,7 +28,8 @@ upgradeLegacyHash();
 
 loadRTLPlugin();
 
-const initializeMap = () => {
+const initializeMap = async () => {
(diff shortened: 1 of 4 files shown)


src/americana.js
```diff
@@ -28,15 +28,30 @@ upgradeLegacyHash();
...
-  container: "map", // container id
-  hash: "map",
-  antialias: true,
-  style: buildStyle(),
-  center: [-94, 40.5],
-  zoom: 4,
-  attributionControl: false,
-});
+const initializeMap = () => {
+  return createMap(window, (shields) => shieldDefLoad(shields), {
```
jleedev, line 32:
```suggestion
const initializeMap = async () => {
  await document.fonts?.ready;
  return createMap(window, (shields) => shieldDefLoad(shields), {
```

src/americana.js
```diff
@@ -28,15 +28,30 @@ upgradeLegacyHash();
...
+if ("fonts" in document) {
+  document.fonts.ready.then(() => {
+    map = initializeMap();
+  });
+} else {
+  // Font Loading API not supported, run the code immediately
+  map = initializeMap();
+}
+
+export { map };
```
jleedev, line 54:
```suggestion
export const map = await initializeMap();
```

src/bare_americana.js
```diff
@@ -6,16 +6,31 @@ import { createMap, loadRTLPlugin, buildStyle } from "./js/map_builder.js";
...
-  hash: "map",
-  antialias: true,
-  style: buildStyle(),
-  center: [-94, 40.5],
-  zoom: 4,
-  fadeDuration: 0,
-  attributionControl: false,
-});
+const initializeMap = () => {
+  return createMap(window, (shields) => shieldDefLoad(), {
```
jleedev, line 10:
```suggestion
const initializeMap = async () => {
  await document.fonts?.ready;
  return createMap(window, (shields) => shieldDefLoad(shields), {
```

src/bare_americana.js
```diff
@@ -6,16 +6,31 @@ import { createMap, loadRTLPlugin, buildStyle } from "./js/map_builder.js";
...
+if ("fonts" in document) {
+  document.fonts.ready.then(() => {
+    map = initializeMap();
+  });
+} else {
+  // Font Loading API not supported, run the code immediately
+  map = initializeMap();
+}
+
+export { map };
```
jleedev, line 33:
```suggestion
export const map = await initializeMap();
```

jleedev (changes requested):

ZeLonewolf:
The simple `await` version doesn't seem to work here:

image

I can't explain why, or if it was just a fluke that it worked the last time.

src/americana.js
```diff
@@ -28,15 +28,20 @@ upgradeLegacyHash();
...
-  container: "map", // container id
-  hash: "map",
-  antialias: true,
-  style: buildStyle(),
-  center: [-94, 40.5],
-  zoom: 4,
-  attributionControl: false,
-});
+const initializeMap = async () => {
+  await document.fonts?.ready;
```
1ec5, line 32: I don’t think the map should be blocked on loading the Web font, just as it shouldn’t be blocked on loading the fontstack or sprites. If something causes the Web font to fail to load, that isn’t such a critical problem that the user should be left with a blank page. This is already how WebKit handles Web fonts for native text, thanks to Apple’s perfectionism, and it causes all sorts of problems like https://github.com/ZeLonewolf/openstreetmap-americana/pull/955#issuecomment-1751888674.

Either this behavior should be limited to headless mode, or we should continue to load the map right away but purge the shield-related sprites as soon as the Web font loads. I’d prefer the latter because it would address the issue where some Safari users are seeing blank shields.

jleedev: While it's not possible to make styleimagemissing async, purging the shields when the font loads should work.

And probably will avoid any visual glitches.

ZeLonewolf:
Closing to re-implement as a purge and re-load.

PR #990, opened 2023-10-30
Add VT Truck route support

zekefarwell:
I added the first `network=US:VT:Truck` route.  Here's the definition to support shield rendering for it.  
https://www.openstreetmap.org/relation/16617445
https://www.mapillary.com/app/?pKey=905552603342258

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -2490,6 +2490,11 @@ export function loadShields() {
     ["ALT"],
     Color.shields.green
   );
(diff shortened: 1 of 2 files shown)


zekefarwell:
Looks like the new route is in the tiles now and working as expected.
https://preview.ourmap.us/pr/990/#map=12.09/44.90284/-72.2528
Screenshot 2023-10-31 at 1 02 05 PM

claysmalley (approved):

PR #991, opened 2023-10-31
Re-render shields after fonts load

ZeLonewolf:
This PR updates the shield renderer to re-generate all shields that were previous rendered when fonts loads. This ensures that a font update is replicated into prior generated shields but doesn't block the loading of the map.

I tested this by inserting delays after the font load and it seems to update the shields without glitching.

Follow-on from #987


shieldlib/src/shield_renderer.ts
```diff
@@ -1,4 +1,8 @@
-import { Map, MapStyleImageMissingEvent, StyleImage } from "maplibre-gl";
+import {
+  Map as MapLibre,
```
ZeLonewolf, line 2: I did this in order to use the built-in Map class.

shieldlib/src/shield_renderer.ts
```diff
@@ -123,9 +131,25 @@ export class AbstractShieldRenderer {
...
+    document.fonts.ready.then(() => {
+      this._fontsLoaded = true;
+      if (this._preFontImageCache.size == 0) {
+        return;
+      }
+      console.log("Re-processing shields with loaded fonts");
+
+      // Loop through each previously-loaded shield and re-render it
+      for (let [id, routeDef] of this._preFontImageCache.entries()) {
+        map.removeImage(id);
```
1ec5, line 146: Calling `removeImage()` and `addImage()` doesn’t seem to work. If I disable my browser’s cache, go directly to [Mechanisburg](https://preview.ourmap.us/pr/991/#map=14/40.07107/-83.5519), and zoom out and pan over the [Springfield](https://preview.ourmap.us/pr/991/#map=10/39.9396/-83.8312), I see the State Route 4 shield with the fallback font but the U.S. Route 68 and Interstate 70 shields with the downloaded Web font.

Springfield

[`updateImage()`](https://maplibre.org/maplibre-gl-js/docs/API/classes/maplibregl.Map/#updateimage) is specifically designed for this purpose. Calling it would require refactoring `SpriteRepository.putSprite()`.

ZeLonewolf: Good catch.  I have a good reproducer as well.

ZeLonewolf: I'm actually not having expected results with updateImage() either:

image

Is there something I can do to forcibly re-render the old shields beyond the sprite sheet update?

1ec5: As of maplibre/maplibre-gl-js#1805, the map is able to juggle multiple spritesheets (or “sprites” as that PR confusingly calls them). Could we add the shield sprites to one spritesheet, then dump that spritesheet once the fonts load? I wonder if that would clear the cache and convince the map to reevaluate the image expressions.

shieldlib/src/shield_renderer.ts
```diff
@@ -123,9 +140,24 @@ export class AbstractShieldRenderer {
...
+    document.fonts.ready.then(() => {
+      this._fontsLoaded = true;
+      if (this._preFontImageCache.size == 0) {
+        return;
+      }
+      console.log("Re-processing shields with loaded fonts");
+
+      // Loop through each previously-loaded shield and re-render it
+      for (let [id, routeDef] of this._preFontImageCache.entries()) {
+        missingIconLoader(this._renderContext, routeDef, id);
```
jleedev, line 155: Need to add a `true` argument here for update, and thread the argument through in shield.js.

shieldlib/src/shield.d.ts
```diff
@@ -15,7 +15,8 @@ export function storeNoShield(
 export function missingIconLoader(
   renderContext: ShieldRenderingContext,
   routeDef: RouteDefinition,
-  spriteID: string
+  spriteID: string,
+  update?: boolean
```
jleedev, line 19: Did you perhaps forget to commit shield.js

ZeLonewolf: What is missing?

ZeLonewolf:
I could not avoid some kind of font pre-load in order to make this all work.  At least this version seems slightly less kludgy (or at least centralizes the kludge into one spot in the code...)

src/index.html
```diff
@@ -5,6 +5,7 @@
     <title>OpenStreetMap Americana</title>
     <link rel="preconnect" href="https://osm-americana.github.io" />
     <link href="fonts.css" rel="stylesheet" />
+    <script type="module" src="js/load_fonts.js"></script>
```
1ec5, line 8: Do we need to document this additional script as an installation step somewhere?

src/js/load_fonts.js
```diff
@@ -0,0 +1,26 @@
...
+  // Make the div invisible
+  div.style.position = "absolute";
+  div.style.opacity = "0";
+  div.style.pointerEvents = "none";
+  div.style.zIndex = "-1";
+  div.style.left = "-1000px";
+  div.style.top = "-1000px";
+
+  // Append the div to the body
+  document.body.appendChild(div);
```
1ec5, line 20: Does building this placeholder in code pull it into the NPM package? The host webpage’s code may not be anticipating a placeholder div as a top-level child of the `<body>`, regardless of how it’s styled. Maybe it would be safer to insert this placeholder as a child of the map container?

jleedev:
Oh, this is less than ideal:

> Error: The width and height of the updated image must be that same as the previous version of the image

jleedev:
Try adding this in setShields():

```js
    const fontSpec = '1em ' + shieldSpec.options.shieldFont;
    if (!document.fonts.check(fontSpec)) {
      this._fontsLoaded = false;
      document.fonts.load(fontSpec).then(() => this.onFontsLoaded());
    } else {
      this._fontsLoaded = true;
    }
```

I think that's the font loading entrypoint we need.

This will avoid depending on document.fonts.ready as a sort of global state, making this less brittle when embedded. And avoid depending on _all_ of the document's fonts being ready, in case the document wants other fonts unrelated to shield sprites.

Broad overview of what this means: The CSS stylesheet installs the fonts into `document.fonts` in a "pending" state. They aren't loaded until used. The `check()` method of the `FontFaceSet` will synchronously return a boolean whether the given CSS `font: …` rule has an available web font, or not. If not, we use the `load()` method to trigger it.

Alternative to CSS, the `new FontFace()` constructor could be used, which is effectively the same thing to CSS except a weird API that nobody uses. It just adds the fonts to document.fonts but in a pending state, only loaded when needed.

Although it also supports adding the fonts with an ArrayBuffer, which sort of allows more precise control over loading. But I think loading the fonts asynchronously and recovering the fallback sprites is acceptable (and allows the font to safely fail for any reason).

The thing about triggering font loading with a hidden text is that (I don't think) there's no way to ask specifically, which web fonts did we use for _this_ DOM element. So you can only use `document.fonts` as a whole.

Another way could be this:

```js
Promise.all(Array.from(document.fonts, f => f.load())).then(() => … fonts loaded);
```

But I think asking specifically for the value of `shieldFont` is a good idea.

It looks like this:

```js
> shieldFont=`"Noto Sans Condensed", "Noto Sans Armenian Condensed", sans-serif-condensed, "Arial Narrow", sans-serif`
> await document.fonts.load('1em ' + shieldFont)
(2) [FontFace, FontFace]
0: FontFace {family: 'Noto Sans Condensed', status: "loaded"}
1: FontFace {family: 'Noto Sans Armenian Condensed', status: "loaded"}
```

So it gives us all the fonts we asked for.

ZeLonewolf:
This works delightfully.  I had to do some gymnastics because setShields can be invoked asynchronously (from URLShieldLoader), so it's not guaranteed that the Map is available before the font name.

ZeLonewolf:
Hmm I wonder if this works okay with the taginfo shield generator.

jleedev:
Does taginfo not run on pull requests?

```js
    if (!document.fonts.check(fontSpec)) {
         ^
ReferenceError: document is not defined
```

Ah, I see. I suppose this needs to be encapsulated into headless_graphics and document_graphics.

jleedev:
```js
class DOMGraphicsFactory implements GraphicsFactory {
  static loadFonts(shieldFont: string): Promise<void> {
    …
  }
}

**Comment by jleedev** [comment]
> so it's not guaranteed that the Map is available before the font name.

I don't think that should matter, modulo having to supply a certain argument to a certain function.

If the shield renderer has a font name and a document, it can proceed to ask the document to load that font. When the font is ready, it needs to ask the map to replace images and redraw itself, but that won't happen if there isn't a map yet, because the map is the driver of asking for sprites in the first place.

**Comment by ZeLonewolf** [comment]
Taginfo graphics appear to be working normally, e.g.:
https://preview.ourmap.us/pr/991/shield-sample/shield_24.svg

**Comment by ZeLonewolf** [comment]
To make a long story short, taginfo generation doesn't have a problem because the taginfo script only draws shapes without text.  Also, the test in `reloadShieldsOnFontLoad()` will short-circuit on the `_fontsLoaded()` check.  So I think this implementation should work for all cases.

File: shieldlib/src/shield_renderer.ts [hunk 7]
```diff
@@ -52,15 +56,27 @@ export type ShapeDrawFunction = (
...
-  putSprite(spriteID: string, image: ImageData, pixelRatio: number): void {
-    this.map.addImage(spriteID, image, { pixelRatio: pixelRatio });
+  putSprite(
+    spriteID: string,
+    image: ImageData,
+    pixelRatio: number,
+    update: boolean
+  ): void {
+    if (update) {
+      console.log(`update ${spriteID}`);
```
1ec5, line 73: Do we need all this logging anymore? At least in Firefox’s Web inspector, the debugger lets you configure a breakpoint to log a message.

shieldlib/src/shield_renderer.ts
```diff
@@ -52,15 +56,27 @@ export type ShapeDrawFunction = (
...
+      this.map.removeImage(spriteID);
+      this.map.addImage(spriteID, image);
+    } else {
+      console.log(`add ${spriteID}`);
+      this.map.addImage(spriteID, image, { pixelRatio: pixelRatio });
```
1ec5, line 78: The last parameter to `addImage()` accepts [some other options](https://maplibre.org/maplibre-gl-js/docs/API/types/maplibregl.StyleImageMetadata/), of which `sdf` may be the most relevant to a client of this library. Should we plumb through an entire options object?

1ec5 (approved):

Issue #779, opened 2023-02-09
Unnecessary components to road line-sort-key


ZeLonewolf:
The below code is the definition for the `line-sort-key` property in the roads layer. It provides overlap rules for overlapping features within a style layer. This helps solve the problem of multiple adjacent style lines which overlap.  For example, at this location in Houston, we want the toll road always drawn above the non-toll roads.  In this case, the roads have the same `layer` value, so this code ensures that tolls are drawn on top.

![image](https://user-images.githubusercontent.com/3254090/217714963-2fd7d23c-1fa0-4ad6-9e75-c954166f84e6.png)

https://github.com/ZeLonewolf/openstreetmap-americana/blob/4d13fa2be2336e018f930e9344586f6b1088e927/src/layer/road.js#L94-L119

However, these distinctions are _already_ sorted, based on layer ordering as shown below in the bridge. It appears that much of this line sort expression can be removed or be made more specific.  For example, all motorways and trunks can be sorted based on toll/no-toll and ramp/no-ramp attributes. Lesser classes can be ramp/no-ramp only.  There's no need for class-specific distinctions because they won't be honored in separate layers.

https://github.com/ZeLonewolf/openstreetmap-americana/blob/4d13fa2be2336e018f930e9344586f6b1088e927/src/layer/index.js#L146-L179

PR #995, opened 2023-10-31
Rewrite road layer

### LINKED ISSUE #746: Exits to nowhere look weird
On/off ramps show up at zoom 9, but many of the roads they link to aren't visible yet. This leads to a "barbed wire" effect that doesn't look great.

Screenshot 2023-01-30 at 10 02 33 AM

A few options:

- Don't show any links until higher zooms when more roads show up (would hide important motorway junctions)
- Don't show links that go to an invisible road (could be a difficult data issue but idk)
- Use little shape icons to indicate highway junctions at low/mid zooms. Some paper atlases do this to create a node and edge network effect

### LINKED ISSUE #992: Offset one-way arrows away from roads
This style currently tries very hard to superimpose one-way arrows on roads. While we should make use of the space within a filled road, we’re running into a lot of problems, especially when the road is only nominally filled.

For example, the arrows at zoom level 15 on local roads and especially alleys gets almost buried in the road casing, making it more difficult to discern the arrow’s direction: https://github.com/ZeLonewolf/openstreetmap-americana/pull/512#issuecomment-1597629098. Meanwhile, we’ve had to invert the arrow’s color on some darker filled roads, but neither black nor white will yield much contrast against a red road or a road that happens to intersect or pass under: https://github.com/ZeLonewolf/openstreetmap-americana/pull/512#issuecomment-1193495684 https://github.com/ZeLonewolf/openstreetmap-americana/pull/512#issuecomment-1200311309. We tried to mitigate this issue by removing one-way arrows from ramps, but sometimes ramps do benefit from the arrows: https://github.com/ZeLonewolf/openstreetmap-americana/pull/512#issuecomment-1273759489.

Many print maps avoid filling roads even at high zoom levels, so they have to account for collisions between stroked roads and one-way arrows. They typically deconflict the arrows in the same manner as road name labels, by offsetting them to one side of the roadway. Though this treatment takes up more space, it ensures the arrows’ legibility and allows for them to be styled more consistently.

MNDOT 2019 American Map New York City

I implemented offset one-way arrows in the AARoads Wiki fork in https://github.com/aaroads-wiki/openstreetmap-americana/pull/2#issuecomment-1732549849. Overall, I think it reinforces the aesthetic that we’re going for with the road name labels. One downside is that it highlights the idiosyncratic arrow placement along these roadway lines, but I think that’s partly because the tiles aren’t merging roadways aggressively enough (onthegomap/planetiler#298) and don’t contain feature IDs that MapLibre can use to join features across tile boundaries (onthegomap/planetiler#377).

[OSM Americana](https://zelonewolf.github.io/openstreetmap-americana/#map=15/44.96595/-93.27599) | [AARoads Wiki](https://aaroads-wiki.github.io/openstreetmap-americana/#map=15/44.96595/-93.27599)
----|----
[Interior arrows](https://zelonewolf.github.io/openstreetmap-americana/#map=15/44.96595/-93.27599) | [Exterior arrows](https://aaroads-wiki.github.io/openstreetmap-americana/#map=15/44.96595/-93.27599)

### LINKED ISSUE #993: Trunk roads are brighter than freeways
I find it jarring that trunk roads are brighter – and thus more prominent – than freeways:

[U.S. Route 50](https://zelonewolf.github.io/openstreetmap-americana/#map=13/39.11464/-84.42119)

This might’ve happened because print maps often stroke U.S. Routes in bright red, but they also don’t tend to make any surface street as wide as this style does. The freeway color seems to be inspired by [AAA](https://www.aaa.com/aaa/common/mapgallery/ca/sanfranciscoarea.pdf) and other publishers that apply the same tint of red to both U.S. Routes and freeways. The only distinction between these two road classifications is the black casing, to which some other map publishers add a black line down the median.

[U.S. Route 101 and State Route 1](https://www.aaa.com/aaa/common/mapgallery/ca/sanfranciscoarea.pdf) ODOT

### LINKED ISSUE #1029: Minor road tunnels
Example: https://zelonewolf.github.io/openstreetmap-americana/#map=14.59/40.77207/-73.97353

On low zoom, these roads are all black instead of some colour like other major road types (freeway, etc.)
On high zoom, these roads are all white.
Tunnels, on the other hand, are white also on low zoom, creating a strong contrast between black roads and white tunnels.
This also makes the road type somewhat ambiguous.
Major road tunnels are subtly coloured to avoid this.

image

1ec5:
This is a full rewrite of the road layer code based on aaroads-wiki#2.

Tasks:

* [x] Merge and rebase changes to roads layers from AARoads fork
* [ ] Restore road widths and colors based on functional classification
* [ ] Restore space-filling streets at high zoom levels, particularly #1183
* [ ] Restore casing on freeways
* [ ] Check whether #779 is fixed
* [ ] Eliminate bridge casing and always sort filled road lines over stroked road lines at lower zoom levels (possible tail work)
* [ ] Experiment with single offset lines for freeway and expressway casing instead of line gap (possible tail work)

Fixes #746, fixes #992, fixes #993.


ZeLonewolf:
How are you addressing #746 (exits to nowhere)?

1ec5:
> How are you addressing #746 (exits to nowhere)?

This design reduces freeway ramps to thin black strokes, even when they connect freeways to freeways, but tries to keep them visible at any zoom level where freeways are visible (OpenMapTiles is a limiting factor here). See the before-and-after screenshots in https://github.com/aaroads-wiki/openstreetmap-americana/pull/2#issuecomment-1730544721 https://github.com/aaroads-wiki/openstreetmap-americana/pull/2#issuecomment-1742491512 where I first implemented this approach, and also [this discussion in Slack](https://osmus.slack.com/archives/C01V02K52UX/p1695313056053639). I think there will be more continuity between freeways and their ramps once we restore the black casing to freeways.

The treatment of freeway-to-freeway ramps may come as a surprise to anyone who’s been focused on highway connectivity as part of the U.S. mapping community’s reclassification effort, with its emphasis on identifying a coherent, well-connected network of freeways and trunk roads. However, the effect of “pinching” the lines at major freeway interchanges has plenty of precedent among American maps at scales where interchanges are more than mere diamonds but roads are still stroked. At least this treatment is more prevalent on these maps than the “exits to nowhere” that are typical of OSM-based maps.

MNDOT 2019 Compass Maps San Mateo County 2006 Universal Maps Cincinnati Area Rand McNally Dallas Metro

Some maps do keep freeway–freeway ramps as filled lines that look identical to the freeway, but they’re in the minority. (If anything, these maps would be arguments for tagging these connections as `highway=motorway` rather than `highway=motorway_link`.)

INDOT Universal Map Chicago

Anecdotally, I’ve found that deemphasizing ramps highlights many misclassified ramps that would potentially affect turn-by-turn navigation. I think this benefit would outweigh the downside of tempting mappers to reclassify freeway–freeway ramps as freeways.

src/layer/oneway.js
```diff
@@ -79,6 +67,8 @@ export const surface = {
     ],
     "symbol-placement": "line",
     "icon-rotation-alignment": "map",
+    // Assumes driving on the right.
+    "icon-offset": [0, 10],
```
claysmalley, line 71: I like the offset effect, but it looks really clunky in left-hand-drive localities, including the US Virgin Islands. Perhaps an asymmetric oneway icon could make it look a little better.

1ec5: I was thinking this might fall under the notion that we’re primarily optimizing for North America, but a regression in left-hand driving countries would be far from ideal. We could bundle a GeoJSON of left-hand driving countries and reference it within a `within` expression, similar to #749. Performance might be an issue, however.

1ec5: Really the effect I’m going for is how printed maps traditionally would append an arrow to the (offset) label of each one-way street, rather than treating it as a separately placed repeating element. That would have the benefit of less repetition. But I don’t think that’s feasible here, because the `oneway` field is only present on the `transportation` layer.

1ec5: I guess [25.7&nbsp;kilobytes](https://github.com/openstreetmap/iD/blob/3f35a35dd14dfed6b4dbe62eee64e1c974247332^/data/drive-left.json) or [57.2&nbsp;kB](https://github.com/Project-OSRM/osrm-backend/blob/367933fc1a9705e8ea852b8470f9db2d0f6f73fb/data/driving_side.geojson) isn’t too bad, though it sure would be nice if every road just had its country code on it.

adamfranco:
While I'm not completely sold on the blur for unpaved roads, there does seem to be an error with unpaved tertiary roads coming in full weight, then dropping to blurred soon after.

See https://preview.americanamap.org/pr/995/#map=11.99/44.07183/-73.01786 versus https://preview.americanamap.org/pr/995/#map=12.01/44.0718/-73.01774

1ec5:
Oh yeah, the blur is just a temporary workaround. A dashed line makes more sense for this kind of style. Restoring dashes will require multiple layers for roads. We also need to bring back road casing.

PR #996, opened 2023-11-07
Replaces old dot_city icon reference with current place_dot

cboone:
Visually this is a no-op, as the extant `dot_city` references are set to `opacity: 0` and just serve as bridge label buffers. But the missing image warnings in the console were bugging me, so this fixes them.


ZeLonewolf (approved):
Looks good, thanks for catching that and for the contribution!

cboone:
Thanks for an awesome project!

PR #999, opened 2023-12-02
Add gas stations

whubsch:
This PR adds an icon for gas stations. I feel like this is an important one to have on an America-centric map. The icon itself is fairly standard, I think, for what Americans would expect to see representing a gas station and falls into the "consumer" color category. I chose to select the `fuel` subclass because I saw discussion in openmaptiles/openmaptiles#1544 to add `charging_station` to the `fuel` class. Happy to discuss and/or alter any of this.

Screenshot 2023-12-02 at 2 30 41 PM

[Link on localhost](http://localhost:1776/#map=16.43/38.871954/-77.153927)
Screenshot 2023-12-02 at 2 30 55 PM


src/layer/poi.js
```diff
@@ -297,7 +309,11 @@ export const poi = {
         "tram_stop",
       ],
       15,
-      [...getSubclasses(iconDefs.bar), ...getSubclasses(iconDefs.coffee)],
+      [
+        ...getSubclasses(iconDefs.fuel),
```
claysmalley, line 313: Gas stations are present in the tiles at zoom 14, and I think they're worth highlighting at that zoom level. If the labels look too cluttered, we can give them a higher minimum zoom than the icons, just like we do with bus stops.

icons/poi_fuel.svg
claysmalley, line 1: I like the icon; it definitely fits the Americana theme. The curved top makes me think of old-timey gas pumps, but it still looks recognizable and modern.

image

A lot of straight lines aren't aligned with pixel boundaries. This makes lines look less sharp at icon size. My suggestions:
* change icon height to 20 pixels
* make the gas hose, and the gaps between, 1 pixel wide
* move remaining points to whole-number pixel boundaries
* make the gas pump display a solid white rectangle (otherwise it looks as if the pump has a window)

whubsch: Thanks! I hadn't even thought about lining the icon up to pixels, and agree it looks a lot sharper now 😄 I think at a height of 20 pixels, it's on the verge of being a little too big, but let me know what you think.

commit 113a3e2c283c90cb30c3f43dfd8476e3bbbaa746

Screenshot 2023-12-03 at 7 44 44 AM

src/layer/poi.js
whubsch: Done in commit e8c98f3acb8b1571478bb2c6ece6f061f695161c

claysmalley: Looks a little better. Though this confirms my worry that the labels are cluttered at low zoom. Here's z15 and z16, respectively:

image
image

The layer's `text-field` is set to an empty string if certain conditions are met, namely if the POI is a `bus_stop` or `tram_stop` and the zoom level is too low. Let's try adding `fuel` to the list on line 338.

wmisener:
Looks good, thanks for adding this! Regarding the color, I wrote most of this on Slack, but reposting here for completeness:

Gas stations are certainly the perfect overlap of “consumer”, “infrastructure”, and “transport”, so the color they should have is definitely ambiguous under the current guidance. IMO down the line we might want separate colors for food and shops, rather than one “consumer” color. Maybe gas could be this future shop color? Until then, I think your choice of "consumer" orange is fine, though I don't think blue or mauve would be egregiously wrong either.

claysmalley (approved):
Looks good! Well done.

ZeLonewolf:
Thank you for your contribution!

PR #1003, opened 2023-12-13
Replace ts-node with tsx

jleedev:
Fix a module hook incompatibility. And hey, it's no longer "experimental" so this (facet of the house of cards) should be stable now.

Should now be able to run on any version of Node.js 18 or later.

Tested on:   18.18.2,  18.19.0,  21.4.0


ZeLonewolf:
Need to replace ts-node calls in package.json and one ts-node call in one of the github workflows.

jleedev:
Well, "Build PR Preview / deploy-preview" is failing at the "Install and Build Main Branch", because it is broken.

ZeLonewolf:
Looks like there is still a `ts-node scripts/build` in that action's output

ZeLonewolf:
This action:

```
      - name: Use Node.js 18.x
        uses: actions/setup-node@v3
        with:
          node-version: 18.x
```
Needs to move up higher, before the first invocation of npm.

ZeLonewolf:
There are also two actions labelled "deploy-preview", so one of them should probably be renamed for sanity sake when looking at output runs.

jleedev:
> Looks like there is still a ts-node scripts/build in that action's output

Yes, because it switches to a different branch.

> Needs to move up higher, before the first invocation of npm.

I don't disagree, but that's not directly related to this change.

ZeLonewolf (approved):

ZeLonewolf:
I believe this is safe to merge, and then main branch will be sane again

PR #1010, opened 2023-12-19
Cross-browser shield text

### LINKED ISSUE #625: Shield labels don't fit properly with Safari

ZeLonewolf:
Fixes #625 

This PR fixes differences between chrome and webkit by using fudge factors.

It also fixes several "I don't know how this even worked before" bugs in the banner layout code, and does some related typescript cleanup in the affected routines.

They are not pixel-perfect between browsers but they're very close.  Tested on Ubuntu.

**Webkit**
image

**Chrome**
image

**Firefox**
image


shieldlib/src/shield_banner.ts
```diff
@@ -133,7 +133,7 @@ function drawBannerTextComponent(
   textComponent: boolean
 ): void {
   const bannerPadding = {
-    top: r.options.bannerPadding,
+    top: 0,
```
ZeLonewolf, line 136: Removed because the padding is accomplished by placing a gap between successive layouts, not by buffering the layout itself.

shieldlib/src/shield_banner.ts
```diff
@@ -161,7 +161,7 @@ function drawBannerTextComponent(
       text,
       textLayout.xBaseline,
       textLayout.yBaseline +
-        bannerIndex * r.px(r.options.bannerHeight - r.options.bannerPadding)
+        bannerIndex * r.px(r.options.bannerHeight + r.options.bannerPadding)
```
ZeLonewolf, line 164: I'm not even sure how this code worked with this bug.

shieldlib/src/shield_text.ts
```diff
@@ -157,34 +165,49 @@ export function layoutShieldText(
...
-  var xBaseline = padLeft + availWidth / 2;
+  //Adjust for excess text height measured in Webkit engine specifically
+  if (isRunningInWebKit()) {
+    textHeight *= 0.54;
+  }
+
+  let availHeight: number = bounds.height - padTop - padBot;
+  let availWidth: number = bounds.width - padLeft - padRight;
+
+  let xBaseline: number = padLeft + availWidth / 2;
```
ZeLonewolf, line 201: This is the main fix.  We check both ascent/descent and left/right to get height/width, and then we make empirical adjustments to account for how the different browsers actually handle these fonts.

shieldlib/src/shield_text.ts
```diff
@@ -157,34 +165,49 @@ export function layoutShieldText(
...
 
-  let textConstraint = textLayoutFunc(
-    { height: availHeight, width: availWidth },
-    { height: textHeight, width: textWidth },
+  let spaceAvail: Dimension = { height: availHeight, width: availWidth };
+  let measuredTextBounds: Dimension = { height: textHeight, width: textWidth };
+
+  let textConstraint: TextTransform = textLayoutFunc(
+    spaceAvail,
+    measuredTextBounds,
```
ZeLonewolf, line 210: Just a nicety to add typescript typing and make it clear what's happening.

shieldlib/src/shield_text.ts
```diff
@@ -195,20 +218,23 @@ export function layoutShieldText(
   );
 
   ctx.font = Gfx.shieldFont(fontSize, r.options.shieldFont);
-  ctx.textAlign = "center";
+  ctx.textAlign = "left";
   ctx.textBaseline = "top";
```
ZeLonewolf, line 222: top/left makes the positioning math a bit easier and more consistent.

zekefarwell:
I tested this out on Mac OS 13, Windows 11, iOS 17, and Android 14.  Looks quite consistent across all the browsers I tested!  Examples on the interstate shield only for a focused comparison:

Windows-Chrome
Chrome on Windows

Mac-Chrome
Chrome on Mac

Android-Chrome
Chrome on Android

Mac-Safari
Safari on Mac

iOS-Safari
Safari on iOS

Windows-FF
Firefox on Windows

Mac-FF
Firefox on Mac

Android-FF
Firefox on Android

Definitely some vertical padding adjust needed for this and other shields, but that can probably be a follow up PR.  Another issue I'm seeing is Chrome on Windows getting the wide shield blank for 2 digit refs.  On the plus side, Safari and Firefox on Windows get the narrow shield blank for 111.

ZeLonewolf:
> Definitely some vertical padding adjustment needed for this and other shields, but that can probably be a follow up PR. Another issue I'm seeing is Chrome on Windows getting the wide shield blank for 2 digit refs. On the plus side, Safari and Firefox on Windows get the narrow shield blank for 111.

It looks like that only happens for the slightly wider 2-digit numbers, so I suspect I can just micro-adjust that threshold and achieve the intended effect.

ZeLonewolf:
Reducing the threshold from 12 to 11.8 seems to be enough for the right 2-digit to 3-digit transition on Windows/Chrome.

zekefarwell:
> Reducing the threshold from 12 to 11.8 seems to be enough for the right 2-digit to 3-digit transition on Windows/Chrome

Looks like it.  I can no longer reproduce.  👍

I noticed a couple of regressions that it may may be nice to resolve before merging this.  The banner text is larger on this branch than on main.  Also in some cases the banner text size changes between the narrow and wide variants of a certain shield.

banner-text-PR-1010

This PR


banner-text-main

Main branch


Some shields have a thicker top border on this branch that isn't present on main.  I'm only seeing this issue on (some) bannered shields so it may be related to the banner text somehow.

thick-border-PR-1010

This PR


thick-border-main

Main Branch

claysmalley:
> Some shields have a thicker top border on this branch that isn't present on main. I'm only seeing this issue on (some) bannered shields so it may be related to the banner text somehow.

This appears to be affecting shields generated by draw functions. SVG shields look okay.

ZeLonewolf:
[755117b](https://github.com/ZeLonewolf/openstreetmap-americana/pull/1010/commits/755117b981b481da10895803230ac55b5db76ca9) fixes the weirdness we were seeing with some shields.  I simplified the code that draws shields when banners are present and removed an extra draw that wasn't supposed to be there.

shieldlib/src/shield_text.ts
1ec5: This makes me wonder if WebKit is reporting the full line height as the computer text height.

1ec5 (approved):
Looks better than before in MobileSafari:

Safari

PR #1012, opened 2023-12-20
Add shields for Aruban road network

Blijbol:
This adds black-on-yellow shields for road routes on Aruba. The A- or B-suffix depends on the direction of travel.

Sample road signs:
* https://commons.wikimedia.org/wiki/File:Road_infrastructure_in_Aruba.jpeg
* https://commons.wikimedia.org/wiki/File:Road_infrastructure_in_Aruba_16_59_50_641000.jpeg

(Older road signs use white-on-blue instead of black-on-yellow.)

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -3392,10 +3392,10 @@ export function loadShields() {
     34
   );
 
-  // Netherlands
+  // Netherlands, Kingdom of the (European Netherlands, Aruba, and Curacao)
   // https://wiki.openstreetmap.org/wiki/The_Netherlands_road_network
   shields["NL:A"] = roundedRectShield(Color.shields.red, Color.shields.white);
-  shields["NL:N"] = roundedRectShield(
+  shields["NL:N"] = shields["AW:route"] = roundedRectShield(
     Color.shields.yellow,
     Color.shields.black
   );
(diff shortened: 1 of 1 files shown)


ZeLonewolf:
Preview location (basically the whole country):
https://preview.ourmap.us/pr/1012/#map=12.41/12.51651/-69.98641

Blijbol:
> Are shields in Aruba always fixed-width rectangles?

We might need to wait for route 10 to find out? Currently they are all one digit and one letter.

ZeLonewolf:
There are already both 1-digit and 2-digit routes (thanks to letter suffixes).  If they are both using a fixed wide width as shown in the linked pictures then I agree with a fixed wide width in the style.  The examples linked are all 2-digit so I think we'd need to look at what the 1-digit version looks like.

Blijbol:
Unfortunately, all currently existing shields are one numerical digit (1/2/3/4/5/8/9) plus one letter (A/B).

ZeLonewolf:
In the code, the Aruba shields mirror the Netherlands shields.  How is with handled in mainland Dutch shields?

Blijbol:
Netherlands mainland shields are variable-width.

ZeLonewolf:
Given the political association between The Netherlands (mainland) and Aruba, it makes sense to me that we would model the Aruba shields the same way as is presently implemented in this PR.

claysmalley (approved):
Fine by me.

PR #1018, opened 2023-12-23
Limit park rendering to nature reserve, national park, and protected area

ZeLonewolf:
As noted in #105, this PR whitelists the allowable `class` values to render in the park layer only to those features that should be rendered like parks.


ZeLonewolf:
Wrong branch base

PR #1019, opened 2023-12-23
Omit indigenous lands from park rendering

ZeLonewolf:
As noted in #105, this PR whitelists the allowable class values to render in the park layer only to those features that should be rendered like parks.

### DIFF
--- a/src/layer/park.js
+++ b/src/layer/park.js
@@ -3,8 +3,15 @@
 import * as Label from "../constants/label.js";
 import * as Color from "../constants/color.js";
 
+const parkLayerFilter = [
+  "in",
+  ["get", "class"],
+  ["literal", ["nature_reserve", "national_park", "protected_area"]],
+];
+
 export const fill = {
   id: "protected-area_fill",
+  filter: parkLayerFilter,
   type: "fill",
   paint: {
     "fill-color": Color.parkFill,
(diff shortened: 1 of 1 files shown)


src/layer/park.js
```diff
@@ -3,8 +3,15 @@
 import * as Label from "../constants/label.js";
 import * as Color from "../constants/color.js";
 
+const parkLayerFilter = [
+  "in",
+  ["get", "class"],
+  ["literal", ["nature_reserve", "national_park", "protected_area"]],
```
1ec5, line 9: [`class`](https://openmaptiles.org/schema/#park) can potentially be anything, based on `protection_title` taggified. Do we need to include other common values like [`protected_site`](https://openmaptiles.org/schema/#park) and [`přírodní_památka_(pp)`](https://taginfo.openstreetmap.org/tags/protection_title=P%C5%99%C3%ADrodn%C3%AD%20pam%C3%A1tka%20(PP)#overview)?

ZeLonewolf: Oh, I thought I fixed that but I see that I never finished openmaptiles/openmaptiles#1491, so I'll change that to exclusion

1ec5 (approved):

src/layer/park.js
ZeLonewolf: Changed in 9642b23.
I also see now that aboriginal_lands is not differentiated at the zoom so I've also removed park layer rendering from z4 until we can get a fix.

PR #1022, opened 2023-12-24
Universities from z10

wmisener:
With the implementation of https://github.com/phanecak-maptiler/planetiler-openmaptiles/pull/8 in the tiles, universities/college POIs can appear in the tiles as early as z10, with different ones appearing at different times based on their size (larger size=earlier appearance, down to z14 for the smallest ones). This PR changes the earliest possible zoom level for colleges and universities from z14 to z10, allowing them to render whenever they appear in the tiles.

Screenshots below, though note that the icon rendering is currently broken due to universities appearing in the tiles as `class=office` rather than `class=college`, which appears to be due to https://github.com/phanecak-maptiler/planetiler-openmaptiles/pull/29. This wouldn't be hard to patch in Americana, but I don't know if it'd be good practice since it's clearly an OMT/planetiler bug.

Examples:
Michigan State at z11 ([localhost link](http://localhost:1776/#map=11.04/42.716/-84.4694)):
Screen Shot 2023-12-24 at 11 50 21 AM

Cornell at z12 ([localhost link](http://localhost:1776/#map=12.13/42.4371/-76.47602)):
Screen Shot 2023-12-24 at 11 54 08 AM

University of Arizona at z13 ([localhost link](http://localhost:1776/#map=13.08/32.23117/-110.9603)):
Screen Shot 2023-12-24 at 11 55 59 AM

Berry College at z13 ([localhost link](http://localhost:1776/#map=13.06/34.29477/-85.18781)):
Screen Shot 2023-12-24 at 12 12 21 PM


ZeLonewolf:
So the "office" bug is definitely on the openmaptiles side as I've been able to replicate it in an openmaptiles build:

image

ZeLonewolf:
I've just set up:
https://d1zqyi8v6vm8p9.cloudfront.net/rhode-island.json

This has the openmaptiles fix in openmaptiles/openmaptiles#1607 which can be used for testing/render samples.

ZeLonewolf:
No University in Rhode Island is large enough to trip the zoom 10 criterion, so I'm currently rendering an extract of Michigan that we can use for testing.

We do have zoom 11 and 12 examples however:

Zoom 11: University of Rhode Island, Alton Jones Campus ([localhost link](http://localhost:1776/#map=12/41.61466/-71.7221))
image

Zoom 12: University of Rhode Island main Campus ([localhost link](http://localhost:1776/#map=12/41.48618/-71.51562))
image

The zoom 11 example is interesting because it's a massive rural facility that "includes a 75-acre lake, ponds, streams, trails, and a 19th-century farm". Not really a college campus but used by the university as a conference center and presumably for some kind of woodland research type usages.  I'll have to think about this some more because I'm not sure if this is really an `amenity=university` or some other type of facility that happens to be owned by the University.

The z12 icon makes total sense in context with this example, but a university at zoom 11 is _incredibly_ prominent on the map to the point where I wonder if it really belongs.  It's at a peer level with cities.  But if the Alton Jones case is an outlier that "isn't really a university in the sense of what that tag means and just a conference and retreat center owned by the university".  So I don't want to put a lot of stake in this singular Rhode Island example.

I think for some of these huge midwestern universities, calling them "on par with cities in towns" is probably correct parity for universities that boast (in the example of Michigan State) a student population of 40,000.

So In short I'd like to see some render samples and get concurrence that such low zoom treatment is the right call, especially for the z10/11 level where there really aren't much in the way of other POIs shown.

ZeLonewolf:
Here is the Michigan State University example ([localhost link](http://localhost:1776/#map=10/42.7023/-84.4775)) using a Michigan tile render (same URL as the RI example, just change rhode-island to michigan):
image

I think this works quite nicely.

ZeLonewolf:
I'm not seeing Michigan State at z10 on the [current PR preview](https://preview.ourmap.us/pr/1022/#map=10/42.6912/-84.4579)

We'll need to investigate why that is (presumably tile-side)

wmisener:
Indeed, MSU now renders from z11 onward. In fact, I struggle to find any other university that renders at z11 now, not to mention at z10.

It's hard to tell, but I think at some point in the 3.15 upgrade things in maptiler switched from deciding how to render based on the fraction of the world's area a university takes up to being a fraction (10%) of a tile's area (relevant code [here](https://github.com/phanecak-maptiler/planetiler-openmaptiles/blob/c285d5d3da8b0acb31ce5c784f6e62bc25656875/src/main/java/org/openmaptiles/layers/Poi.java#L308) I think). Maybe that somewhat shifted the levels at which things appear?

Although actually, all of my original examples from the original PR still appear at the zoom levels I specified then (but now with the correct icon). All the ones that you posted from your state-level extracts seem to render a zoom level later: the Alton Jones campus appears at z12, and the main URI campus appears at z13 for me now. Maybe there was some weird thing where using the extract calculated the tile/planet size strangely, or a maptiler versioning thing on your local machine?

ZeLonewolf (approved):
Thanks for jockeying this, I'm comfortable that this does the job and other efforts to bring up universities from z11 to z10 can happen on the tile side.

PR #1027, opened 2024-01-11
Adding vehicle POI icons

whubsch:
Adding an icon to render icons for `shop=car`, `shop=car_repair`, and `amenity=taxi`. I wanted to handle them all in one go because they share the same base car icon, although in the end I tweaked the `car_repair` car to be a little shorter to accomodate the wrench above it. I ended up filling in the windshield after some discussion on Slack, as that seems most consistent with the bus and fuel icons. I had also wanted to create an icon for `car_wash` while I was at it, but OpenMapTiles does not support that yet. This PR makes progress on #692. Open to any feedback, discussion, or suggestions!

[localhost link](http://localhost:1776/#map=17.52/40.750319/-73.992944)
Screenshot 2024-01-11 at 6 22 02 AM
[localhost link](http://localhost:1776/#map=16.42/38.897058/-77.108183)
Screenshot 2024-01-11 at 6 21 41 AM


wmisener:
I think these look nice! Can you make sure you update the taginfo file as well?

whubsch:
Done.

ZeLonewolf (approved):
These look awesome, thanks!

PR #1028, opened 2024-01-13
Developer script for overlaying grids on icons

ZeLonewolf:
This PR adds a script for generating pixel grid previews for an SVG, as well as some documentation describing its use.  This gives a developer or reviewer an additional tool to verify pixel alignment of new icons.

Follow-on work from this PR can potentially integrate this into CI, so that we can see a preview of the pixel grid alignment of new or modified icons in the map. The coarse grid shows the 1x pixel alignment, while the fine grid shows the 2x pixel alignment.

poi_bus_preview
poi_fuel_preview
poi_health_cross_preview
poi_martini_glass_preview
shield_us_interstate_2_preview
shield_us_nj_njtp_noref_preview
shield_us_wi_rustic_preview


scripts/icon_grid.ts
```diff
@@ -0,0 +1,73 @@
...
+        { input: resizedSvgBuffer, blend: "over" },
+        { input: gridPattern, tile: true, blend: "over" },
+      ])
+      .toFile(outputFilename)
+      .then(() => console.log(`Wrote ${outputFilename}`))
+      .catch((err) => console.error(err));
+  } catch (error) {
+    console.error("Error:", error);
+  }
+}
```
jleedev, line 66: It's not really an issue in the context of a script, but swallowing exceptions is bad, and swallowing exceptions from a promise is also bad.

As a general rule, since this is an `async` function, it should `await` on every promise and not just discard them, and it should only use `.then()` and `.catch()` for the purpose of chaining together asynchronous operations, not for logging errors.

The top-level code can have the try/catch instead.

```suggestion
async function convertAndScaleSVG(svgFilename: string): Promise<void> {
  const svgBuffer = fs.readFileSync(svgFilename);

  // Get dimensions of the original SVG
  const metadata = await sharp(svgBuffer).metadata();
  const width = metadata.width! * scale;
  const height = metadata.height! * scale;

  // Resize the SVG
  const resizedSvgBuffer = await sharp(svgBuffer, {
    density: 72 * scale,
  })
    .resize(width, height)
    .toBuffer();

  const xOffset: number = width % 2 == 0 ? scale / 2 : 0;
  const yOffset: number = height % 2 == 0 ? scale / 2 : 0;

  // Generate a pixel grid pattern
  const gridPattern = generateGridPattern(xOffset, yOffset);

  // Composite the scaled image over the grid
  await sharp({
    create: {
      width,
      height,
      channels: 4,
      background: { r: 255, g: 255, b: 255, alpha: 0 },
    },
  })
    .composite([
      { input: resizedSvgBuffer, blend: "over" },
      { input: gridPattern, tile: true, blend: "over" },
    ])
    .toFile(outputFilename);
  console.log(`Wrote ${outputFilename}`);
}
```

scripts/icon_grid.ts
```diff
@@ -0,0 +1,73 @@
...
+
+if (!svgFilename) {
+  console.error("Please provide an SVG filename.");
+  process.exit(1);
+}
+
+convertAndScaleSVG(svgFilename);
```
jleedev, line 73:
```suggestion
try {
  await convertAndScaleSVG(svgFilename);
} catch (error) {
  console.error("Error: ", error);
  process.exit(1);
}
```

jleedev:
This is interesting, two feature requests:

- Render the icon at actual size (well, 1x and 2x), and scale up with nearest neighbor, then apply the overlay grid. To see both how the scalable curves fit the grid and how they actually land on the grid.
- Change the background, the POI icons have a white border that's difficult to see.

ZeLonewolf:
Differential background is definitely a clear improvement.  Though I may set it to something not on grayscale to handle the scaling that you're suggesting.
poi_fuel_preview

claysmalley (approved):

PR #1031, opened 2024-01-26
Adding hotel and hostel POI icons

whubsch:
This PR adds two icons, one for `hotel`, `motel`, and `guest_house`, and a bunk-bed style one for `hostel`. I previously had included dormitory with the `hostel` icon, but on the Slack we determined that because dormitories are not open or available for use by the public, they should not be included in this PR. I have the figures in the hostel icon sleeping in different directions primarily to avoid users mistaking the icon for two hotels right on top of each other, even though I know that's not how most people sleep 😄 

This PR makes progress on #692. Open to any feedback or suggestions!

[Localhost link](http://localhost:1776/#map=16.1/38.906341/-77.025889)
Screenshot 2024-01-26 at 6 17 57 AM

Screenshot 2024-01-26 at 6 17 44 AM


ZeLonewolf:
I think these icons look great!  Let's add a test location as well to the samples list in test/.  The spot you linked above is fine, as well as [this](https://preview.ourmap.us/pr/1031/#map=16/40.755264/-73.987405) spot in Manhattan.

ZeLonewolf (approved):
These look great, thanks for contributing!

PR #1032, opened 2024-01-27
Rename shield image names based on structured route properties

1ec5:
This is a ~~proof of concept for~~ refactor integrating openmaptiles/openmaptiles#1620, which restructures the `transportation` layer’s feature properties to put route relation tags in structured properties instead of the inflexible `network=ref` mini-language. This PR also plumbs route names (as opposed to road names) and route colors through to the shield selection code, though support for route colors will require an additional change to the shield library not included here.

### DIFF
--- a/src/configs/config.aws.js
+++ b/src/configs/config.aws.js
@@ -3,7 +3,7 @@
 /*
 Planetiler tile server, hosted at AWS
 */
-const OPENMAPTILES_URL = "https://tile.ourmap.us/data/v3.json";
+const OPENMAPTILES_URL = "https://tile.ourmap.us/data/omt_3_15.json";
 
 /*
 The following two variables override the color of the bounding box and halo of
--- a/src/js/shield_format.ts
+++ b/src/js/shield_format.ts
@@ -3,6 +3,8 @@ import type {
   RouteParser,
 } from "@americana/maplibre-shield-generator";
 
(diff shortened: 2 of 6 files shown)


ZeLonewolf:
Now that openmaptiles/openmaptiles#1620 has been merged, this PR can proceed.  We'll hold off on the final merge until the tiles are updated.  Alternately, if the tiles are updated before the PR is ready, I think that means this PR will become a critical fix for Americana as well as other downstream styles such as AARoads or Gnome?

1ec5:
GNOME Maps doesn’t use Americana tiles, but Overpass Ultra does.

ZeLonewolf:
We should be able to use:
https://tile.ourmap.us/data/omt_3_15.json

As the current live-updating bleeding-edge planet.  Also we should get all existing shield clients to make this change and switch to that endpoint.  That should allow for a clean transition.

Inspect link here:
https://tile.ourmap.us/inspect-omt_3_15.html

(Haven't been able to test this for reasons but above is the intent)

1ec5:
[Interstate 695 is interrupted by the Patapsco River where the Francis Scott Key Bridge once stood.](https://zelonewolf.github.io/openstreetmap-americana/#map=12/39.22091/-76.52774)

ZeLonewolf (approved):
LGTM!

PR #1014, opened 2023-12-21
Add library icon

wmisener:
Fixes #821, and makes progress on #435 and #692. Adds an icon for libraries: poi_library
Based on feedback in #821, I riffed off the MUTCD I4-1 sign, which depicts a person reading and should be highly recognizable to an American audience.

I'd be open to an approach that uses a lone book symbol as well.

Went with blue coloring, based on discussions [like this](https://osmus.slack.com/archives/C01V02K52UX/p1691089744839109?thread_ts=1691041634.613469&cid=C01V02K52UX) and others. I think there's an argument to be made that libraries also fit the definition of "Attraction: places where people go for entertainment, leisure, or curiosity", which would be brown, but "community anchor"/"infrastructure" is probably more in line with current style symbology.

Previews:
UCLA ([localhost link](http://localhost:1776/#map=15.07/34.07068/-118.44458))
Screen Shot 2023-12-21 at 7 05 24 AM

Gallup, NM ([localhost link](http://localhost:1776/#map=15.13/35.526549/-108.740541))
Screen Shot 2023-12-21 at 7 29 51 AM

New York, NY ([localhost link](http://localhost:1776/#map=16.36/40.752812/-73.982632))
Screen Shot 2023-12-21 at 7 33 56 AM

Legend:
Screen Shot 2023-12-21 at 7 44 16 AM

PR #1034, opened 2024-02-18
Add Bookstores

wmisener:
Based on the discussion in #1014, I'm reusing the upright book icon added for libraries but in "consumer orange" for bookstores (`shop=books` in OSM, class `library` + subclass `books` in OMT). They appear at z16, the same as similar consumer POIs like coffee shops, bars, car repair shops, etc.

Screen Shot 2024-02-17 at 10 02 48 PM

Screen Shot 2024-02-17 at 10 10 34 PM

Screen Shot 2024-02-17 at 10 12 16 PM

Screen Shot 2024-02-17 at 10 13 19 PM


ZeLonewolf:
Did you intend to modify icons/shield_us_az_scenic.svg with this PR?

wmisener:
> Did you intend to modify icons/shield_us_az_scenic.svg with this PR?

Oops, no. I think running code_format might've picked it up? It doesn't look like a material change to the icon

adamfranco:
I think this looks good, but I wonder if we should worry about accessibility for various types of color blindness? These colors might not be an issue, or maybe the presence of labels is enough, but just thought I should flag this before we make too many identical icons in different categories.

ZeLonewolf:
> accessibility for various types of color blindness?

It strikes me that one of the things we could implement (since this a vector style, after all) is the ability to toggle various color blindness settings that could change the colors en masse on the fly (in a similar vein to what we do with label language.

If different colors that mean different things aren't color-blind discernible, we probably shouldn't rely on differentiated icons to paper over that.

wmisener:
> I think this looks good, but I wonder if we should worry about accessibility for various types of color blindness? These colors might not be an issue, or maybe the presence of labels is enough, but just thought I should flag this before we make too many identical icons in different categories.

Thanks for raising this. Using the nifty tool here: https://davidmathlogic.com/colorblind/#%23F38F00-%23003F87-%23693F23-%23A20067-%234E008E-%23006747 with our current POI colors: 
Screen Shot 2024-02-18 at 11 11 32 AM

This seems to indicate that orange and blue are essentially always distinguishable. The purple (used for airport features) tends to look blue, while mauve (used for transport features) can look similar to either brown or blue depending on the type of color-blindness (the first two are more common than the third).

Based on this, it might be worth considering changing the shade of blue and/or purple we use. If we implement green icons, care will be needed to make sure they don't look identical to our brown icons. Otherwise, as Brian suggests, we could implement on-the-fly color switching, or clickable icons that display more information (I could imagine a subtitle under the name with the legend entry, for example). But these accessibility considerations might merit their own issue for further discussion.

I still think with some care, we can make distinguishable colors, and the benefits that reusing icons has for style coherence outweigh the downsides of potential confusion.

claysmalley (approved):
Looks good to me.

Let's put tail work about colorblind analysis into a new issue.

ZeLonewolf:
Thanks for working this, forked off the CVD work to #1038.

Issue #809, opened 2023-02-26
Brand logos


1ec5:
To complement the usual POI icons, we could special-case certain chain stores and mark their locations by the brand’s logo.

## Precedent

There’s a long history of print maps highlighting specific chains, especially in tourist-oriented maps. Even travel-oriented maps will often call out a specific chain. This used to be how chains published store locators before the Google Maps API came along. Consider it America’s response to the Japanese maps that unfailingly mark every 7-Eleven and Lawson location with those chains respective logos.

![Best Western San Francisco Bay Area](https://user-images.githubusercontent.com/1231218/221390138-21c5eadb-bfcd-4d96-ac31-fd301c3909d3.jpg)
_Best Western may not be the most neutral cartographer out there._

![Huff Hamilton County](https://user-images.githubusercontent.com/1231218/221390171-378ccc01-bfa1-46f7-8c1f-b04a87aeab06.jpg)
_The most important thing to know about Cincinnati is that Huff Realty is ready to sell you property at their convenient location out in Green Township._

![Ontario Ministry of Transportation](https://user-images.githubusercontent.com/1231218/221390217-50b72dfa-47a5-4908-b13f-83da05fb68fe.jpg)
_The Ontario Ministry of Transportation doesn’t care for commercialism. Instead, their icon for parks is the park’s operator’s logo._

## Implementation

With the help of name-suggestion-index, OSM has been annotating individual store locations with the Wikidata QID of the brand or operator in `brand:wikidata` and `operator:wikidata`, respectively. Unfortunately, these keys aren’t present in the tiles. With these QIDs, we’d be able to easily match a select few to prebundled icons, or we could do something more dynamic and grab the icon from Wikimedia Commons, as other name-suggestion-index consumers do.

## Design

Some logos work better as icons than others. Wikidata’s [small logo or icon (P8972)](https://www.wikidata.org/wiki/Property:P8972) or [icon (P2910)](https://www.wikidata.org/wiki/Property:P2910) property would be much more reliable than its [logo image (P154)](https://www.wikidata.org/wiki/Property:P154) property. Some brands may need overrides specific to a map, because their usual logos lack contrast against a map background.

PR #1035, opened 2024-02-19
Add post offices

wmisener:
Makes progress on #435 and #692. Renders post offices (`amenity=post_office`, aka in OMT `class=post` + `subclass=post_office`) with an envelope icon. This is quite similar to the [Carto](https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/amenity/post_office.svg) and [Maki](https://github.com/mapbox/maki/blob/main/icons/post.svg) approaches.
Screen Shot 2024-02-18 at 6 07 04 PM

I've sometimes seen American maps render an envelope with the bottom folds as well, making more of an 'x' shape. See e.g. [the NPS](https://www.nps.gov/npgallery/GetAsset/cd253093-6813-471e-9d16-042c2a373a97) or [the MUTCD](https://wiki.openstreetmap.org/wiki/File:MUTCD_RS-026.svg). I'm open to this style of envelope, but in the end I didn't think that the extra visual clutter made it that much more identifiable.

The icon comes in at z15, the same as other 'community anchor' POIs. I went with blue color, since I see post offices as more of a community service provider than a consumer destination.
Screen Shot 2024-02-18 at 9 13 48 AM

Unfortunately, in my opinion, there is no information in the tiles we can reliably use to distinguish public post offices (like the USPS) from private ones (UPS Stores, etc), as mentioned in https://github.com/ZeLonewolf/openstreetmap-americana/issues/435#issue-1275881342. So this PR renders them both identically, although I see distinguishing them in the future as desirable. This would be a good use case for using `operator:type` or `operator:wikidata` were those tags to be exposed in OMT, similar to what's been discussed more broadly for supporting brand logos (#809).
Screen Shot 2024-02-18 at 9 14 19 AM

### DIFF
--- a/icons/poi_envelope.svg
+++ b/icons/poi_envelope.svg
@@ -1,0 +1,3 @@
+<svg width="20" height="20" xmlns="http://www.w3.org/2000/svg">
+ <path d="m 2,4 v 1 l 8,5 8,-5 V 4 Z m 0,3 v 9 H 18 V 7 l -8,5 z" style="stroke:#fff;stroke-width:2;stroke-dasharray:none;stroke-opacity:1;paint-order:stroke fill markers;stroke-linejoin:miter;stroke-linecap:round"/>
+</svg>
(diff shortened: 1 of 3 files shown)


ZeLonewolf (approved):
Looks great, thanks!  Sorry for the delay in reviewing.

PR #1042, opened 2024-04-18
change shields of Lincoln and Rock Counties, Minnesota

### LINKED ISSUE #1041: Rock County MN doesn't distinguish between CR's and CSAH's
So per https://www.dot.state.mn.us/maps/gdma/data/maps/county/rock.pdf, there's a nice, defined difference between the CSAH system (normal pentagon) and the CR system (white square). However, in the field, Rock County doesn't distinguish between them (https://www.bing.com/maps?osid=fac4b26f-6bf9-453f-ad51-bf949a2a7d76&cp=43.615358~-96.093568&lvl=15.814454&dir=26.544998&pi=-9.240571&style=x&v=2&sV=2&form=S00027, CSAH to the left, CR to the right).

As such, the US:MN:Rock:CSAH shield should be changed to the white square to match what's in the field. 

This might not be the only county in MN that does this (based on what I remember driving through rural MN), but I don't have a definite answer right now.

claysmalley:
Fixes #1041


ZeLonewolf:
Indeed we have the desired white squares:

image

ZeLonewolf (approved):
LGTM!

PR #1045, opened 2024-04-20
add Mississippi Scenic Route shields

claysmalley:
Scenic Routes of Mississippi have shields based on the State Route design, but blue instead of black, and with a "Scenic" banner.

The "SCEN" banner chosen here could potentially be omitted, but I don't think the blue and black contrast well enough to make the distinction clear without it.

[example](https://preview.ourmap.us/pr/1045/#map=12.09/34.39986/-89.82934)

image

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -1424,6 +1424,11 @@ export function loadShields() {
 
   // Mississippi
   shields["US:MS"] = ovalShield(Color.shields.white, Color.shields.black);
+  shields["US:MS:Scenic"] = banneredShield(
+    ovalShield(Color.shields.white, Color.shields.blue),
+    ["SCEN"],
+    Color.shields.blue
+  );
   [
     "Alcorn",
     "Calhoun",


src/js/shield_defs.js
```diff
@@ -1424,6 +1424,11 @@ export function loadShields() {
 
   // Mississippi
   shields["US:MS"] = ovalShield(Color.shields.white, Color.shields.black);
+  shields["US:MS:Scenic"] = banneredShield(
+    ovalShield(Color.shields.white, Color.shields.blue),
+    ["SCEN"],
```
1ec5, line 1429: The banner is also appropriate because there’s always a “SCENIC ROUTE” banner over the shield in the real world.

“SCEN” isn’t the most intuitive abbreviation. Maybe separately we could try to accommodate longer banner text by reducing the pitch or squishing the text horizontally. That could let us spell out “TRUCK” too.

1ec5 (approved):

PR #1046, opened 2024-04-20
assign Minnesota counties correct shields

### LINKED ISSUE #1043: Identify all counties of Minnesota whose shields don't distinguish CSAH and CR
Tail work from #1041.

All Minnesota counties have a distinction between County State Aid Highways (CSAH; tagged `network=US:MN:<county>:CSAH`) and County Roads (CR; tagged `network=US:MN:<county>:CR`). Typically, CSAHs have blue pentagonal shields and CRs have white square shields. But some counties opt to sign CSAHs and CRs the same, using one of these two designs.

We need to compile a list of counties in Minnesota that have chosen to use solely one shield, and which one. Likewise, if any counties deviate from the typical Minnesota county shield design (e.g. yellow text instead of white), this should be documented as well.

cc @the-spui-ninja

claysmalley:
Fixes #1043 

Also adds support for `network=US:MN:Intercounty`.


ZeLonewolf (approved):
Code looks good by me.  I panned around and did see a few outliers that probably need tagging updates across the state, e.g. [this one](https://www.openstreetmap.org/relation/4446442).

PR #1049, opened 2024-04-24
Add Halifax bridge shields

quincylvania:
Followup from #912. Halifax has these [little signs](https://c8.alamy.com/comp/2R5GM9R/road-signs-for-the-two-halifax-harbour-bridges-in-halifax-nova-scotia-canada-bayers-road-2R5GM9R.jpg) for its two major bridges that match the design of the scenic route signs of the province.

https://www.openstreetmap.org/relation/9240721
https://www.openstreetmap.org/relation/17510849 ~~(waiting to propagate to tiles)~~

Screenshot 2024-04-24 at 10 41 14 AM


quincylvania:
https://www.openstreetmap.org/relation/17510849 is now propagated to the tiles but the shield still isn't showing up. Not sure why.

https://preview.ourmap.us/pr/1049/#map=15.76/44.677366/-63.611834

ZeLonewolf:
image

quincylvania:
This is now ready for review.

Screenshot 2024-04-30 at 10 39 52 PM

claysmalley (approved):
Looks good to me!

It's a bummer we can't just use one icon and recolor it, though.

quincylvania:
> It's a bummer we can't just use one icon and recolor it, though.

Maybe in the future we can achieve this through some sort of composite. Having the icon be separate from the generic white rectangle background.

PR #1057, opened 2024-05-08
Add various missing bannered shields in the US

claysmalley:
This adds support for 6 additional bannered shields in the US with at least one instance on OSM, confirmed by street-level imagery.


ZeLonewolf (approved):
LGTM

PR #1067, opened 2024-05-16
Allow specifying shield networks using a prefix and wildcard

ZeLonewolf:
WIP


ZeLonewolf:
Based on discussion in Slack, this approach, while interesting, likely introduces other undesired side effects, most seriously being the unintended styling of route networks that happen to meet the filter.

PR #1068, opened 2024-05-17
Add shield for Inner Loop, Rochester, NY

claysmalley:
The [Inner Loop](https://en.wikipedia.org/wiki/Inner_Loop_(Rochester)) of Rochester, New York has an idiosyncratic shield: a [Bus Carrell-style](https://www.atlasobscura.com/articles/forest-service-signs-history) white trapezoid with rounded corners, with the black text "INNER LOOP" ([example 1](https://www.bing.com/maps?cp=43.160874%7E-77.612446&lvl=22.0&mo=om.1&pi=34.2&style=x&dir=48.8), [example 2](https://www.bing.com/maps?cp=43.161687%7E-77.611031&lvl=22.0&mo=om.1&pi=1.7&style=x&dir=302.3)).

For the shield, I chose to omit the "INNER" and use just the text "LOOP", as there is no other loop route with similar signage in Rochester (the Outer Loop is instead signed as a sequence of Interstate and State routes). I considered "IL" as well, but I would expect that to be less intuitive.

Untitled

[image](preview.ourmap.us/pr/1068/#map=13.15/43.15587/-77.61005)
[](https://github.com/ZeLonewolf/openstreetmap-americana/assets/1732117/1a4cbf90-626d-4f08-b9c0-f890fc2984f4)[](https://github.com/ZeLonewolf/openstreetmap-americana/assets/1732117/ba4c7b9b-91fa-4ef4-a202-6dcc29fb5abc)[](https://github.com/ZeLonewolf/openstreetmap-americana/assets/1732117/a01db64c-39ff-4745-8412-e1eaa0a05238)[](https://github.com/ZeLonewolf/openstreetmap-americana/assets/1732117/f7c771bf-5c4e-42cb-aae9-e39b2de65d72)[](https://github.com/ZeLonewolf/openstreetmap-americana/assets/1732117/e0e330e6-a239-4cbe-8ea4-9791f69a1b25)[](https://github.com/ZeLonewolf/openstreetmap-americana/assets/1732117/48bad46e-01ae-44cf-b235-0a16742c9fc0)[](https://github.com/ZeLonewolf/openstreetmap-americana/assets/1732117/c6913d55-6ab6-4d00-bd96-7642bf3f0e41)[](https://github.com/ZeLonewolf/openstreetmap-americana/assets/1732117/46d6823c-14ff-4c4d-8375-87ea231038ed)[](https://github.com/ZeLonewolf/openstreetmap-americana/assets/1732117/2a3c174f-4d0f-4e09-b05b-5f6463e8d665)[](https://github.com/ZeLonewolf/openstreetmap-americana/assets/1732117/b37e2d5c-5e19-42b0-946d-00797ae1d11c)[](https://github.com/ZeLonewolf/openstreetmap-americana/assets/1732117/aa080f00-7668-41d4-84fa-37d0f5317069)


jleedev:
I think the orange color is more proper; while it's sometimes in white I see that as akin to the Pennsylvania Turnpike signs sometimes being in white simply to contrast with a green background (when on guide signs vs standalone). It's also careening towards being a historical object so I expect the signage in its current state is pretty much "what you see is what you get".

claysmalley:
> I think the orange color is more proper; while it's sometimes in white I see that as akin to the Pennsylvania Turnpike signs sometimes being in white simply to contrast with a green background (when on guide signs vs standalone).

While it may have been historically signed in orange, I couldn't find any extant examples of this color scheme on street-level imagery. Even the standalone signs are black-and-white, despite the lack of a green background.

ZeLonewolf:
Wondering if it would be a little nicer to vertically center the LOOP text, perhaps with a shallower angle on the sides?

src/js/shield_defs.js
```diff
@@ -3856,6 +3863,9 @@ export function loadShields() {
     },
   };
 
+  shields["US:NY:Inner_Loop"].refsByName = {
+    "Inner Loop": "LOOP",
+  };
```
ZeLonewolf, line 3868: This works, but I wonder now if we should add a function to hard-code the ref based on network for cases like this.

jleedev:
Street View at the (checks notes) Southern terminus, East Main Street.

Sep 2014: Orange https://maps.app.goo.gl/rsUnk4REQr6Mci9b8
Nov 2015: White https://maps.app.goo.gl/V8BgvAg7VWvMLrrZ8

White shield SGTM.

src/js/shield_defs.js
1ec5: There’s only ever going to be one route in the `US:NY:Inner_Loop` network, right? It doesn’t need to be a function; we could add support for setting a static `ref` property in the shield definition and using that instead of `refsByName`.

ZeLonewolf: #1070 adds this support

claysmalley: New functionality integrated in 1b3799d

ZeLonewolf:
Just for kicks, I tried evening out the top and bottom padding for trapezoidDownShield to 3px each instead of 2px/4px:

image

claysmalley:
Padding values adjusted to vertically center the text.

ZeLonewolf (approved):
LGTM!
image

Issue #1062, opened 2024-05-13
Allow re-usable shield definitions


ZeLonewolf:
Chipping away at #889.

We should be able to define a shield definition by name, and then reference that name in other parts of the ShieldJSON.

The current structure is:
```json
{
  "networks": {
    "default": {
      "key": "value"
    },
    "network1": {
      "key": "value"
    },
    "network2": {
      "key": "value"
    }
  }
}
```

A reorganized format which eliminates repetition might look something like this

```json
{
  "defaults": {
    "key": "value"
  },
  "definitions": {
    "white_circle": {
      "mapping": {
        [
          {
            "networks": ["US:NJ", "US:DE"],
          },
          {
            "networks": ["US:NJ:Truck", "US:DE:Truck"],
            "banners": ["TRK"]
          }
        ]
      }
      "key": "value"
    },

  }
}
```
In that form, the "key":"value" represents all the elements of the current definition language that define the color, shape, etc., of the shields.

This inverts the current structure and would necessitate a full re-write or replacement of the shield_defs.js file that currently generates the ShieldJSON.  Suggestions welcome on the best way to structure this.

1ec5:
Even an intermediate refactor would enable the definitions to be reused more broadly. For example, this definition:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/9df8b999d544ab1cfec6f43d4d99bbb42e9efe5b/src/js/shield_defs.js#L169-L175

which calls this function directly:

https://github.com/ZeLonewolf/openstreetmap-americana/blob/9df8b999d544ab1cfec6f43d4d99bbb42e9efe5b/shieldlib/src/shield_helper.d.ts#L50-L57

could be rewritten as:

```js
{
  "CA:MB:Winnipeg": {
    "type": "trapezoidDownShield",
    "sideAngle": 10,
    "fillColor": "white",
    "strokeColor": "black",
    "textColor": "black",
    "radius": 2
  }
}
```

It would be more verbose than what we have now, but it would be easier to validate and transform to various formats, and we could more naturally split it up into multiple files, similar to how name-suggestion-index and id-tagging-schema are structured. However, if we prefer the current format for less redundancy, we could have it generate this JSON instead.

ZeLonewolf:
[...]

PR #1069, opened 2024-05-18
Make dedicated pill shape function

ZeLonewolf:
This PR adds "pill" as a first-class shape that a shield style can specify.  Behind the scenes in the shield library, it still just invoking roundedRectangle, but this allows the shield style definition to not have to do the hack of specifying a pill by declaring a rounded rectangle with a large radius. This removes the hack mentioned in #1062 for taginfo and saves a few bytes of ShieldJSON.

I anticipate doing the same job for circle and homePlate shapes.


shieldlib/src/shield_canvas_draw.ts
```diff
@@ -97,6 +97,16 @@ function ellipse(
   return width;
 }
 
+export function pill(
+  r: ShieldRenderingContext,
+  ctx: CanvasRenderingContext2D,
+  params: ShapeBlankParams,
+  ref?: string
+): number {
+  params.radius = r.shieldSize() / 2;
```
claysmalley, line 106: This doesn't seem to be working right for 200% zoom:

[image](https://preview.ourmap.us/pr/1069/#map=14.2/38.19356/-81.49928)

claysmalley (commented):
> I anticipate doing the same job for circle and homePlate shapes.

How come? Because they have different `prettyShapeName`s?

ZeLonewolf:
> How come? Because they have different `prettyShapeName`s?

I think it's generally useful and more readable in the ShieldJSON to have the shield definition be able to specify common shapes by name, particularly when then can consume defaults (in this case, corner radii). I confess that I don't fully understand @1ec5's vision for shield definition re-usability, but I see value in a slightly richer dictionary of shield shapes being more extendable down the line.

shieldlib/src/shield_canvas_draw.ts
ZeLonewolf: Fixed in dd2192e. I also added a note explaining the logic at that spot in the code.

claysmalley (approved):

PR #1074, opened 2024-05-19
Group GLCT definitions together

ZeLonewolf:
This is a minor refactor to put all the GLCT (Great Lakes Circle Tour) shield definition code together in the same place in shield_defs.json.  They were thousands of lines apart which made it confusing to understand what was going on.

image


claysmalley:
All of the `overrideByRef`, `overrideByName` and `refsByName` cases are at the end of the file. Either we should keep all of them together, or get rid of this section and put all of the special cases inline.

quincylvania:
I could see the number of "special cases" growing to the point where they're not so special. I think it makes more sense just to organize everything consistently.

ZeLonewolf:
Superceded by #1081

PR #1076, opened 2024-05-19
Move text on US interstate shields down 1 pixel

ZeLonewolf:
The numbers on Interstate shields bump into the crown, I think they look a little nicer bumped down a pixel:

image

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -451,7 +451,7 @@ export function loadShields() {
       left: 4,
       right: 4,
       top: 6,
-      bottom: 4,
+      bottom: 5,
     },
   };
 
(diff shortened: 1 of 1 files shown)


src/js/shield_defs.js
```diff
@@ -450,8 +450,8 @@ export function loadShields() {
     padding: {
       left: 4,
       right: 4,
-      top: 5,
-      bottom: 5,
+      top: 6,
+      bottom: 4,
```
claysmalley, line 454:
```suggestion
      bottom: 5,
```

There's enough room to squash this vertically by one pixel without affecting icon selection. I think this looks a little better for 3+ digits. Before/after:

image

image

ZeLonewolf: LGTM

PR #1081, opened 2024-05-30
Refactor all special case shield definitions inline

claysmalley:
Corollary to #1074. This PR removes the special case section at the bottom of `shield_defs.js`, and collates all the instances of `overrideByRef`, `overrideByName` and `refsByName` inline with their geographically sorted definitions.


ZeLonewolf:
Do we know why the ShieldJSON got smaller in the stats?  Just want to make sure we're not losing anything in the refactor.

claysmalley:
I changed `US:NH:Turnpike` to use the color constants instead of hardcoded hex values, and `US:TX:Harris:HCTRA` to be variable-width. Would either of those changes mess with the ShieldJSON size?

ZeLonewolf:
I actually misspoke, the stats say the ShieldJSON is 1,484 bytes larger.  That sounds like there's somehow an extra shield entry in this.  I went shield by shield and couldn't find it though.

ZeLonewolf:
I ran a script to compare the two shield json files between this PR and main, and the attached report shows the differences (in theory).
[compare_report.txt](https://github.com/user-attachments/files/15523969/compare_report.txt)

src/js/shield_defs.js
```diff
@@ -105,12 +105,33 @@ export function loadShields() {
...
+      },
+      LSCT: {
+        spriteBlank: "shield_glct_lsct",
+      },
+    },
+  };
+  shields["GLCT:Loop"] = {
+    ...shields["GLCT"],
+    banners: ["LOOP"],
+    bannerColor: Color.shields.brown,
```
ZeLonewolf, line 127: I think this is supposed to be bannerTextColor, let's check the map and see how it's rendering.

src/js/shield_defs.js
```diff
@@ -446,6 +518,13 @@ export function loadShields() {
 
   // Yukon
   shields["CA:YT"] = roundedRectShield(Color.shields.white, Color.shields.red);
+  shields["CA:YT"].overrideByRef = {
+    2: roundedRectShield(Color.shields.white, "#ce9d00"),
+    3: roundedRectShield(Color.shields.white, "#ce9d00"),
+    5: roundedRectShield(Color.shields.white, Color.shields.blue),
+    6: roundedRectShield(Color.shields.white, Color.shields.green),
+    11: roundedRectShield(Color.shields.white, Color.shields.blue),
+  };
```
ZeLonewolf, line 527: Combine with line above using `...` notation
Looking at this with fresh eyes, I think it's possible to override just the one color difference rather than repeating the whole definition.

claysmalley: `textColor` is trivial to replace with `overrideByRef`, but with how it's structured, `strokeColor` would require replacing the whole `shapeBlank`. I went ahead and combined these into one declaration, but kept the helper functions in.

ZeLonewolf (approved):
On review of the comparison report, the extra size is due to bannered Georgia routes.  After this consolidation, the bannered Georgia routes also gained the overrideByRef definitions for routes 515 and 520 (the blue and green routes).  The prior code had the overrides get set later in the file and only on the main Georgia route networks, not the variants. Once #1073 is complete, this bit of extra duplication will go away. As such, I think this is OK to proceed with!

PR #1083, opened 2024-05-31
Replace Utah shield icons

claysmalley:
In preparation for SOTMUS, I figured I'd have another go at the Utah state route shield.

This improves on the icon from #259 by widening the white fill to ease text crowding, and reintroduces the two curved strokes on top that are present on the real-life shield.

image
image


1ec5 (approved):
I also appreciate that the ridges on either side are more apparent at 2×, if not also at 1×.

PR #1086, opened 2024-06-01
Assign shield icons based on ref length

### LINKED ISSUE #86: Wide shields showing up for 1 & 2 digit shields at DPR 1

claysmalley:
Fixes #86

This PR changes the behavior of shield icon selection. Instead of measuring the aspect ratio of the displayed text, it simply chooses by matching string length to the length specified at the end of the icon's filename (e.g. `shield_us_i_2`, `shield_us_i_3`). 

Certain narrow characters (currently `1`, `I`, `i`, `l` and ` `) are considered to be two-thirds of a width of all other characters. This ensures that `111`, for example, is assigned a 2-digit variant icon if available.

### Background

When I was working on #1083, it looked great on my computer screen (Firefox/Linux). Once it deployed, I checked it out on my phone (Firefox/Android), and to my disappointment, every Utah shield displayed the 3-digit variant icon. Alas, even #965 didn't fix the inconsistent text metrics across different platforms. Americana needs to choose the icon based on platform-independent factors.

[image](https://zelonewolf.github.io/openstreetmap-americana/#map=15.06/40.5228/-111.93838)


shieldlib/src/shield.ts
```diff
@@ -14,6 +14,8 @@ import {
 import { TextPlacement } from "./shield_text";
 import { StyleImage } from "maplibre-gl";
 
+const narrowCharacters = /[1Iil ]/g;
```
1ec5, line 17: Also J, since we’re using the slender version without a top bar.

Some punctuation characters are also pretty narrow.

shieldlib/src/shield.ts
```diff
@@ -51,20 +53,23 @@ function getRasterShieldBlank(
...
 
   if (Array.isArray(shieldDef.spriteBlank)) {
-    for (var i = 0; i < shieldDef.spriteBlank.length; i++) {
-      shieldArtwork = r.spriteRepo.getSprite(shieldDef.spriteBlank[i]);
+    // Certain narrow characters count as two-thirds of a character
+    let narrowCharacterCount = (routeDef.ref.match(narrowCharacters) ?? [])
+      .length;
+    let refLength = Math.ceil(routeDef.ref.length - narrowCharacterCount / 3);
+
+    // Choose icon based on optimal character length at end of filename
```
1ec5, line 61: The contributing guide should emphasize this new constraint on the file name.

1ec5 (approved):
Sad but true. I’m approving this as a workaround at least. But long-term, I don’t think this will save us from having to choose a fixed font size for each design. Some two-digit shields currently size some single digits differently than other single digits based on their optical width. Choosing a font size will free us from having to fine-tune paddings, which has got to be the most frustrating aspect of designing new shields for this codebase.

We could use some rendering tests that exercise this change, particularly the exceptions.

PR #1088, opened 2024-06-02
Add Madagascar national route shields

claysmalley:
According to [French Wikipedia](https://fr.wikipedia.org/wiki/Liste_des_routes_nationales_%C3%A0_Madagascar), national routes of Madagascar have white-on-red rectangular shields (a common scheme in areas of French influence).


ZeLonewolf (approved):
image
LGTM

PR #1089, opened 2024-06-03
Add Niger national route shields

claysmalley:
According to [French Wikipedia](https://fr.wikipedia.org/wiki/Liste_des_routes_nationales_du_Niger), national routes of Niger have white-on-red rectangular shields (a common scheme in areas of French influence).


ZeLonewolf (approved):
LGTM!
image

PR #1092, opened 2024-06-04
update package-lock.json

claysmalley:


jleedev:
Obviously this sort of nonsense should be automated, since the build uses `npm ci` it requires that the package-lock periodically be updated to pick up fixes, but can you clarify what's going on here?

ZeLonewolf:
As a general solution, perhaps we could add to the lint job to also check to see if the package-lock.json is stale?

wolfy1339:
I suggest taking a look at
https://github.com/apps/renovate

ZeLonewolf (approved):

jleedev:
In fact, `npm update` seems to be wildly unstable, or has been changed in different versions of `npm`, and I really don't care about this, except:

- Having a checked-in package lock is understood to be a good thing, so that `npm ci` will produce (hopefully) consistent results, and
- Something (either an annoying github bot, or the chaotic forces of developers running `npm install`) should periodically update the transitive and semver-compatible dependencies.

1ec5:
> an annoying github bot

Also known as @dependabot…

PR #1093, opened 2024-06-05
[WIP] Combine similar boundary style layers

ZeLonewolf:
WIP, description TBD

### DIFF
--- a/src/layer/boundary.js
+++ b/src/layer/boundary.js
@@ -3,6 +3,65 @@
 import * as Color from "../constants/color.js";
 import * as Label from "../constants/label.js";
 
+export const boundaryCasing = {
+  id: "boundary_casing",
+  type: "line",
+  paint: {
+    "line-color": [
+      "match",
+      ["get", "admin_level"],
+      2, `hsl(${Color.hueBorderCasing}, 35%, 86%)`,
+      [3, 4], `hsl(${Color.hueBorderCasing}, 30%, 90%)`,
+      5, Color.borderCasing,
+      6, Color.borderCasing,
+      Color.borderCasing
+    ],
+    "line-width": [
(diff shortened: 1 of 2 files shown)


1ec5:
Great idea!

ZeLonewolf:
The CI checks are not reporting the style size change - all deltas are zeroes.  I'm going to fix that as a preqrequisite.

src/layer/boundary.js
```diff
@@ -3,304 +3,289 @@
...
+    "line-width": [
+      "step",
+      ["zoom"],
+      ["match", ["get", "admin_level"], 2, 1, 3, 0.5, 4, 0.5, /* else */ 1],
+      4,
+      ["match", ["get", "admin_level"], 2, 1.5, 3, 0.5, 4, 0.5, /* else */ 1],
+      5,
+      ["match", ["get", "admin_level"], 2, 1.5, 3, 0.8, 4, 0.8, /* else */ 1],
+      7,
+      ["match", ["get", "admin_level"], 2, 2.115, 3, 1.1, 4, 1.1, /* else */ 1],
```
1ec5: At some zoom levels like z11, the line is thinner and fainter than it used to be.

Before | After
----|----
[Thicker](https://americanamap.org/#map=11/40.6005/-93.9365) | [Thinner](https://preview.americanamap.org/pr/1093/#map=11/40.6005/-93.9365)

ZeLonewolf:
Reviewed zoom by zoom and added additional test cases for areas noted by @1ec5.  Comparison looks clean to me in the PR preview.  Some of the exponential interpolation zoom transitions had to be converted to step transitions with linear interpolations.  So some slight differences in fractional zooms, but I don't think that will be perceptible.

Issue #1104, opened 2024-06-16
Faroe Islands shields


claysmalley:
The road network of the Faroe Islands is signed almost identically to that of Iceland, with simple black-on-white rectangle shields encompassing all numbered routes in one network. Currently, there are no `route=road` relations on the Faroe Islands. Before Americana can support shields, each route would need to be added to a relation tagged `route=road`, `network=FO`, `ref=*`.

CC @riiga in case you would like to help out.

[![image](https://github.com/ZeLonewolf/openstreetmap-americana/assets/1732117/535c1b49-a35b-4142-9f41-abd09c7b2e70)](https://www.mapillary.com/app/?lat=62.299265899999995&lng=-6.5215072999722&z=16.84425901167187&pKey=969193426982635&focus=photo&x=0.23952211645403504&y=0.44701341100630215&zoom=2.605005440696409)
[![image](https://github.com/ZeLonewolf/openstreetmap-americana/assets/1732117/bacab7e5-0910-4645-baa6-79252b25b496)](https://www.mapillary.com/app/?lat=62.00235608999998&lng=-6.800582090000034&z=17&pKey=162871565667342&focus=photo&x=0.6182885143573887&y=0.4507163267996522&zoom=1.4885745375408053)
[![image](https://github.com/ZeLonewolf/openstreetmap-americana/assets/1732117/90c6c6d2-268d-42a9-9cbb-6d8f3084d720)](https://www.mapillary.com/app/?lat=61.56361458684495&lng=-6.846056464614321&z=17.061822477621835&pKey=261825765650500&focus=photo&x=0.2800414802269362&y=0.4054936129356374&zoom=2.652221238768588)
[![image](https://github.com/ZeLonewolf/openstreetmap-americana/assets/1732117/efbb3f01-7347-4f8f-9492-1056e4cd7dda)](https://www.mapillary.com/app/?lat=62.20382339997201&lng=-6.759221799972238&z=18.341480147879285&pKey=604121017240689&focus=photo&x=0.6922736923780017&y=0.5315017235677357&zoom=2.605005440696409)

riiga:
https://www.openstreetmap.org/changeset/152762094 🚀

claysmalley:
Fixed by #1107.

PR #1107, opened 2024-06-16
Add support for Faroe Islands shields

riiga:
As per #1104, same style as Iceland.


1ec5 (commented):
Thanks! If you run `npm run status_map`, the map in the readme will highlight the Faroe Islands as being supported.

claysmalley (approved):

Issue #1101, opened 2024-06-11
National Historic Trail auto tour routes


quincylvania:
There are 21 [National Historic Trails](https://en.wikipedia.org/wiki/National_Trails_System#National_Historic_Trails) as of 2024. Unlike other members of the National Trails System, the historic trails tend to be signed as auto tour routes on modern roads rather than hiking trails. These routes are _very_ Americana and the shields would make a great addition to the style.

- [x] ~~[Ala Kahakai](https://en.wikipedia.org/wiki/Ala_Kahakai_National_Historic_Trail)~~ (offroad trail)
- [ ] [Butterfield Overland](https://en.wikipedia.org/wiki/Butterfield_Overland_National_Historic_Trail)
- [ ] [California](https://en.wikipedia.org/wiki/California_Trail)
  - [ ] Mapped
  - [ ] Shield
- [x] ~~[Captain John Smith Chesapeake](https://en.wikipedia.org/wiki/Captain_John_Smith_Chesapeake_National_Historic_Trail)~~ (water trail)
- [x] ~~[Chilkoot](https://en.wikipedia.org/wiki/Chilkoot_Trail)~~ (offroad trail)
- [ ] [El Camino Real de los Tejas](https://en.wikipedia.org/wiki/El_Camino_Real_de_los_Tejas_National_Historic_Trail)
- [ ] [El Camino Real de Tierra Adentro](https://en.wikipedia.org/wiki/El_Camino_Real_de_Tierra_Adentro_National_Historic_Trail)
- [x] ~~[Iditarod](https://en.wikipedia.org/wiki/Iditarod_National_Historic_Trail)~~ (offroad trail)
- [ ] [Juan Bautista de Anza](https://en.wikipedia.org/wiki/Juan_Bautista_de_Anza_National_Historic_Trail)
  - [ ] Mapped
  - [ ] Shield
- [ ] [Lewis and Clark](https://en.wikipedia.org/wiki/Lewis_and_Clark_National_Historic_Trail)
  - [ ] Mapped
  - [ ] Shield
- [ ] [Mormon Pioneer](https://en.wikipedia.org/wiki/Mormon_Pioneer_National_Historic_Trail)
  - [ ] Mapped
  - [ ] Shield
- [ ] [Nez Perce (Nee-Me-Poo)](https://en.wikipedia.org/wiki/Nez_Perce_National_Historic_Trail)
  - [ ] Mapped
  - [ ] Shield
[...]

PR #1116, opened 2024-06-30
Add Selma to Montgomery National Historic Trail shield

quincylvania:
This one is a little different: since the historic trail consists pretty much of extant roads, there's no "auto tour route" separate from the trail.

https://www.openstreetmap.org/relation/13302056

https://commons.wikimedia.org/wiki/File:45th_Anniversary_of_the_Civil_Rights_March_from_Selma,_Alabama_to_Montgomery,_Alabama_LCCN2010637779.tif

Re: #1101

Screenshot 2024-06-30 at 11 38 17 AM


icons/shield_us_nht_semo.svg
1ec5, line 1: Sky blue surrounded by white doesn’t have very much contrast against an already light background. Should we surround all that with another black outline, or would that be too heavy?

quincylvania: I think this does look better, though is the shield getting too large now? We should probably have consistency across all the National Historic Trail shields (and look forward to National Scenic Trail shields, which are similar in design).

Screenshot 2024-07-01 at 10 28 11 AM

quincylvania: Here is 22x21. The above is 24x23.

Screenshot 2024-07-01 at 11 07 40 AM

1ec5: This gives the shield more definition, which is great.

As an aside, maybe we should consider putting an extra outline around anything we’re currently outlining in white, including the Interstate shield. Apple Maps takes this approach; sometimes it makes the shields look more cartoonish, but the added contrast really comes in handy over light backgrounds.

ZeLonewolf (approved):
Looks good to me!
image

PR #1117, opened 2024-06-30
Add Overmountain Victory National Historic Trail Commemorative Motor Route shield

quincylvania:
Re: #1101

https://www.openstreetmap.org/relation/10967459

https://commons.wikimedia.org/wiki/Category:Signs_of_the_Overmountain_Victory_National_Historic_Trail

Screenshot 2024-06-30 at 2 48 19 PM


quincylvania:
Added a black outline per #1116. The legibility at this size is a little harder. I kept the "1780" since it's actualy readable on retina displays I think it's a sort of a fun part of the design to be able to keep.

Screenshot 2024-07-06 at 9 22 01 PM

1ec5:
The 1780 is quite intricate, but these NHT shields are already heavily graphical and distinctive based on their color and shape, so I’m less concerned about each individual element being legible on its own.

1ec5 (approved):

PR #1115, opened 2024-06-30
Add Oregon National Historic Trail Auto Tour Route shield

quincylvania:
The mapping is complete but the data is propagating to the tiles. Should be ready to review by morning.

Re: #1101 

Screenshot 2024-06-30 at 12 41 12 AM

PR #1120, opened 2024-07-07
Replace National Historic Trail shields with triangular color swatches

1ec5:
For compliance reasons, the shields introduced in #1115, #1116, and #1117 have been replaced with simple color swatches. Each color swatch is in the shape of a Reuleaux triangle, a shape that is common to many recreational trails, including U.S. Bicycle Routes. The colors continue to correspond to the primary color associated with each trail.


quincylvania:
I'm afraid using plain shields for NHTs would be too confusing. A few of the NHTs are overlapping, so users would need to refer to a legend and be able to distinguish between similar colors to tell which is which. I suggest either making our own designs or removing these entirely until we figure out if we can use the official emblems.

1ec5:
These shields do appear in the legend, and I’ve seen maps that take the same approach to simplifying the shields. But if the concern is about potential color conflicts or colorblindness issues, I suppose we could stick a monogram inside each shield.

icons/shield_us_nht_semo.svg
1ec5, line 1: Also, this fixes a path issue in the original image that had a bit of blue showing through on the other side of the white ring.

ZeLonewolf (approved):
Temporary pause until we figure out what's permitted.

icons/shield_us_nht_semo.svg
quincylvania: The dark blue intentionally overlaps the white part slightly, as per the original shield.

quincylvania:
Yeah, I guess a solid color is better than nothing for now :)

icons/shield_us_nht_semo.svg
1ec5: I was referring to this itty-bitty portion, which admittedly is too small to show up at 1×:

A light blue, triangular gap between the white and black rings.

PR #1124, opened 2024-07-10
Add fire station icon

### LINKED ISSUE #1036: How to render fire stations
Fire stations are one of the last 'community anchor' POIs (#435) for the style to render. There is no one clear iconography associated with them IMO, so I think we could go a number of directions. Therefore, I thought I'd create an issue to discuss possibilities before starting a PR. A few of the options as I see them (further ideas welcome!):

- A classic one used in other OSM maps is a flame icon. This is the approach of e.g. Carto, Maki, and OSMAnd. I find this somehow both intuitive (obviously it's fire) and counter-intuitive (wait, don't firefighters put out fires?!?). It also could be confused with legitimate fire icons we might want in the future, like for firepits.
Screen Shot 2024-02-21 at 9 53 45 PM   

- Another option would be a fire truck symbol. This more clearly represents firefighters to me, and would be pretty recognizable iconography. It might be a challenge to get a good looking 20x20px symbol though. One model might be MUTCD W11-8:

- One other option that could work is the "Firefighter's cross", something like below. This would have a nice synergy with our badge symbol for police stations. I don't know whether it's widely associated with firefighters outside the US, or recognizable enough on its own. Maybe with the flame in the center?
  

- Are there other symbols commonly associated with firefighters, but not easily confused with the literal object? Some things that came to mind are hoses, axes, and ladders, but I don't think those would be very clear/distinctive. Maybe those distinctively shaped hats could work?

Anyways, curious what others think.

wmisener:
Resolves #1036, and makes progress on #435. Renders fire stations as a shield (the same as that used for police stations) emblazoned with a flame: poi_fire_station

The icon appears at z15, the same as most of the other "community anchor" institutions, and is colored blue to match them.

Sample ([localhost](http://localhost:1776/#map=16.06/33.963912/-118.356962)):
Screen Shot 2024-07-09 at 9 58 07 PM


ZeLonewolf (approved):
This icon looks good to me and seems to have the strongest community support.

PR #1078, opened 2024-05-28
Add icons for capitals of admin_level 5 and 6

claysmalley:
Fixes #384

This PR refactors the capital symbology to apply to all values of `place=*` (not just `city`), and assigns places with `capital=5` and `capital=6` a new icon, `place_dot_in_circle`.

[image](https://preview.ourmap.us/pr/1078/#map=8.07/32.752/-97.064)

[image](https://preview.ourmap.us/pr/1078/#map=7.04/33.438/-118.158)

[image](https://preview.ourmap.us/pr/1078/#map=6.25/28.501/76.786)

PR #1127, opened 2024-07-16
Documenting use of capital, admin_level, and place

wmisener:
While adding taginfo documentation for #1119, noticed there weren't yet any taginfo entries for other `place` values, down to village (#279), as well as the various `boundary=administrative` + `admin_level=*` and the new use of `capital=5` and `6` (#1078). Adding them here so Americana's usage shows up in taginfo.


ZeLonewolf (approved):
LGTM!

PR #1131, opened 2024-07-19
Update GitHub Pages URLs

1ec5:
Replaced references to zelonewolf.github.io with americanamap.org. Although zelonewolf.github.io currently redirects to americanamap.org, Mapbox GL JS and MapLibre GL JS are apparently unable to follow the redirect due to [a CORS error](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSMissingAllowOrigin). This doesn’t affect the main deployed application, but it does affect anyone who uses the published style.json. Every fork of this repository will need to make a similar change.

PR #1134, opened 2024-07-19
Update GitHub Pages URLs for glyphs

1ec5:
As of today, the [osm-americana/fontstack66](https://github.com/osm-americana/fontstack66/) repository publishes to font.americanamap.org instead of zelonewolf.github.io. This breaks everything except the main repository’s main deployment at americanamap.org, because nothing else has CORS permissions to the old location anymore to even see the redirect. This change updates glyphs to point to the new location.

### DIFF
--- a/scripts/extract_layer.js
+++ b/scripts/extract_layer.js
@@ -24,7 +24,7 @@ const locales = opts.locales[0].split(",");
 const style = Style.build(
   config.OPENMAPTILES_URL,
   "https://americanamap.org/sprites/sprite",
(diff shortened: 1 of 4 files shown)


1ec5:
As in https://github.com/ZeLonewolf/openstreetmap-americana/pull/1131#issuecomment-2238053871, forks of this repository will need to pull in this fix as well. Otherwise, the style will fail to load.

/cc @infinatious @arahmandc @dschep

ZeLonewolf (approved):

PR #1157, opened 2024-09-18
Add a hillshade layer

### LINKED ISSUE #780: Add hill-shading to expose terrain
Much of America (and the world's) most rugged terrain also has the fewest map features. Adding hillshade (aka "hill shade") lets the terrain speak for itself and give context to the peaks, valleys, and ridges that define our landscape.

In addition to continent-scale mountain-range topography, hillshade also helps to highlight why cities might have weird blank gaps in them.

For example, Los Angeles without hillshade just has some large gaps. Why are there the blank areas without even parks? 
https://americanamap.org/#map=12.01/34.15464/-118.32441
image

When terrain and topography is highlighted by hillshade (even without contour lines which add visual noise) it becomes much more obvious to the map viewer why roads and developed areas bend around a spot:
https://www.openstreetmap.org/#map=13/34.1471/-118.3514&layers=Y
image

There may be a number of techniques that could allow adding hillshades, from translucent raster PNGs served by a separate service, to hillshades dynamically generated in Maplibre from a digital elevation map (DEM).

Ideas to consider:
- Where does the hillshade fit in the feature stack? 
  _I'm thinking that an ideal hillshade would be a grayscale translucent layer that would be placed above the landcover, landuse, water, and park layers (shading them), but be below roads, buildings, and POIs (keeping them un-shaded)._
- Multi-directional hillshade can look better than than single light-source shading.

Other discussions:
- OSM-US Slack [#american-map-style thread on hillshade desire](https://osmus.slack.com/archives/C01V02K52UX/p1630413226067000), discusses layering and multi-directional hillshade.
- OSM-US Slack [#american-map-style thread on hillshade offer from clockwork micro](https://osmus.slack.com/archives/C01V02K52UX/p1645206597476349), also includes discussion of techniques.
- OSM-US Slack [#maplibre thread with examples of hillshade rendering from DEM](https://osmus.slack.com/archives/C01G3D28DAB/p1675476578940879?thread_ts=1675464263.271319&cid=C01G3D28DAB)
- #341 - a proof of concept using 3D terrain with Americana.

Resources:
- [Maplibre hillshade layer documentation](https://maplibre.org/maplibre-gl-js-docs/style-spec/layers/#hillshade)
- [Clockwork Micro terrain styler documentation](https://docs.clockworkmicro.com/v1.html#section/Authentication)
- [AWS terrain tiles](https://registry.opendata.aws/terrain-tiles/), and an [example of using them](https://stackblitz.com/edit/web-platform-d7d27q?file=index.html).

jleedev:
What's going on here:

- Add a hillshade layer using elevation from https://registry.opendata.aws/terrain-tiles/. It goes up to z15 but limit to z13 for aesthetics and data quality, and perhaps performance.
- Upgrade maplibre-gl to get anti-anti-fingerprinting and performance improvements.
- Add a control to toggle the hillshade. Maplibre conveniently has a terrain icon.

TODO here:

- [x] Remove inline colors
- [x] Off by default

Future work:

- Use the prebaked raster hillshade if we can make it blend.
- Ponder the balance between physical and cultural map features — to me, solid green parks on top of hillshade starts to blur the line.
- Ponder the balance between global landcover and hand-mapped natural features.
- Develop the terrain toggle button some more — de-emphasize landuse fills to mainly outlines; toggle hillshade _and_ global landcover (and contour lines); desaturate road classes; emphasize cycle and hiking routes; introduce peaks and ridges and valleys and islands; etc.
- Store the selected map style in the URL hash.

Fixes #780


src/layer/index.js
```diff
@@ -83,6 +84,8 @@ export function build(locales) {
 
     lyrFerry.ferry,
 
+    lyrHillshade.hillshading,
```
1ec5, line 87: The Terrarium tiles are a bit sensitive to artificial structures like dams that look a bit strange over water, like the Lock and Dam Number 12 on the Mississippi River at Bellevue, Iowa:

[A big embankment formed by the hillshading layer in the middle of the river.](https://preview.ourmap.us/pr/1157/#map=13/42.26072/-90.4113)

If we move this layer lower in the stack, such as right below `water_line`, then the waterbody layer would tend to obscure this and other anomalies over water:

Most of the embankment gone, save for a small secondary protrusion on the landward side.

However, the tradeoff is that the “hill” shading would no longer double as a bathymetry layer at sea:

Before | After
----|----
[Seamounts and ridges on the sea floor around the Bahamas.](https://preview.ourmap.us/pr/1157/#map=5/25.5/-78.06) | Solid water fill.

While bathymetry is pretty cool, I think it’s far enough out of scope of a (land) transportation map as to be a distraction. It also looks a bit discolored without hypsometric tinting, whereas on land the absence of that tinting is quite reasonable on a non-topographic map. (General-purpose overview maps would prioritize showing the continental shelf using hypsometric tinting before bothering with elevation shading.)

src/layer/hillshade.js
```diff
@@ -0,0 +1,19 @@
...
+      ["linear"],
+      ["zoom"],
+      3,
+      0.2,
+      12,
+      0.5,
+    ],
+    "hillshade-shadow-color": "rgba(102,85,51,1)",
+    "hillshade-highlight-color": "rgba(255,255,204,1)",
+    "hillshade-accent-color": "rgba(0,0,0,1)",
```
1ec5, line 17: With these colors, some road lines don’t have a lot of contrast against the background, while the knockouts around labels stand out enough to look like halos again. Maybe we could try moving all three colors closer to the preexisting background color. This could also help the layer recede into the background while still conveying information about rugged terrain to those who look for it.

1ec5 (commented):
This is a beautiful starting point. I shared some screenshots with the AARoads Wiki folks and the reception has been quite positive.

src/layer/hillshade.js
jleedev: Tamed it a _lot_. Since the hillshade is now restricted to the luminance space, I increase the exaggeration at low zooms since it no longer fights with other colors. It fades away at high zooms to avoid ruining the foreground by being a murky blob, and to avoid getting confused by things like stadiums.

ZeLonewolf: In the flat lands, the old version made it actually look like a crumpled paper map :-D

src/layer/index.js
1ec5: Incidentally, #995 gives tunnels a different treatment that wouldn’t really exhibit the cute underlapping effect anyways.

The flash of bathymetry feels like an unsubtle Easter egg, of which we have too few in this project. 🙃 In seriousness, maybe we could load the DEM source and hillshade layer only after the main sources and layers load, to avoid jank and unnecessary bandwidth and power usage. To avoid distracting the user on land, we could fade in the hillshade layer with a `hillshade-exaggeration-transition` property, akin to deferred 3D terrain in F4Map, though I don’t know how well that would perform.

src/js/icon_control.js
```diff
@@ -0,0 +1,43 @@
+export class HillshadeControl {
+  constructor({ layerId }) {
+    this._layerId = layerId;
+  }
+
+  _onClick = () => {
+    if (this._map.getLayoutProperty(this._layerId, "visibility") == "none") {
```
1ec5, line 7: Astonishingly, my iPhone 8 is able to keep up with the map with the hillshading enabled, but it turns the battery meter into a slow trombone. Should we hide the layer by default on mobile devices, or somehow detect less powerful devices?

jleedev: I upgraded maplibre-gl, the changelog mentions some performance improvements, want to try again?

1ec5: The first render is reasonably quick, but there’s a lot more stuttering while panning than on main. The battery goes down by a few percent after about 20&nbsp;seconds of panning around the Appalachians at zoom level 8. That’s still much better than the majority of ad-laden news websites these days, but the stuttering is a bummer.

quincylvania:
The OSM US Tileservice now has pre-rendered hillside and contour lines available (thanks @jake-low!) Using these would significantly improve client-side performance. If the hillshade parameter aren't to your liking then we can consider rendering another set.

```
    "hillshade": {
      "type": "raster",
      "url": "https://dwuxtsziek7cf.cloudfront.net/hillshade.json"
    },
    "contours": {
      "type": "vector",
      "url": "https://dwuxtsziek7cf.cloudfront.net/contours-feet.json"
    }
```

1ec5:
There’s some discussion about both approaches in #780. A visual comparison would be helpful. Dynamic hillshading has the benefit of easier customization (especially if anyone forks the style for a slightly different appearance) and dynamic lighting when rotating the map, though I don’t know how frequently people rotate this map. Who knows, maybe we could use the prerendered hillshading on mobile devices and the dynamic hillshading otherwise.

jleedev:
The prerendered tiles are a lot harder to wrangle the desired look out of. Ideally we want to blend it with the beige background color, solid fills like aerodrome, park, and the fat purple boundary casing. (Or at least, that's what happened immediately when I slotted the dynamic hillshade on top of those layers, and I like it.)

As a workaround, I tried to make the background color very dark and saturated and put the hillshade jpg with 96% alpha -- this doesn't look great, and the aggressive background shows through when the hillshade tiles aren't loaded.

Another way to make it work is to change all those background layers to be alpha blended on top of the hillshade. Might work but the colors would all end up a bit murky; they're currently fully opaque and paint on top of each other.

Just spitballing, what if the hillshade tiles were instead solid black with alpha channel, so that they can blend with a background color more easily. Or perhaps some color instead of black, so that the raster-* layer properties can be used. (Either webp or png?)

(It would be very nice if maplibre supported a color matrix or blend modes to be able to do this, but that's far away.)

1ec5:
If performance is the only concern, I’d favor merging this feature as is, but disabled by default (again, maybe only for certain user agents). Then we could figure out how to make it more performant or provide alternatives before enabling it by default.

zekefarwell:
Oops I didn't mean to click the ready for review button.  Maybe that's ok though?  Anyway feel free to change if back if not...

zekefarwell:
Overall, I'm really liking the look of Americana with hill shading.  It does diverge from the paper highway map aesthetic, but I've always felt that some level of hill shading would help fill in the empty spaces.  It's just a matter of balancing the shading so it's not too overpowering, but still visible enough.  This style isn't an outdoors focused topo map after all. 

> A visual comparison would be helpful. Dynamic hillshading retains the benefits of easier customization (especially if anyone forks the style for a slightly different appearance) and dynamic lighting when rotating the map, but the effect is undermined when tilting the map.

Here's a comparison of the same spot with this PR and in OpenTrailMap which uses the raster hillshade tiles @quincylvania mentioned. Location:  [PR preview,](https://preview.ourmap.us/pr/1157/#map=12/41.85765/-122.51172)  [OpenTrailMap](https://opentrailmap.us/#map=12/41.85765/-122.51172).

| This PR | Open Trail Map |
|----------|-------|
| americana-hillshade-pr-1157 ⬆️ north up | opentrailmap-hillshade ⬆️ north up |
| americana-hillshade-pr-1157-180-deg ⬇️ north down | opentrailmap-hillshade-180-deg ⬇️ north down |

Subjectively, the OpenTrailMap hillshading is a bit more pleasing to my eye.  I'm not entirely sure why this is.  Maybe uni-directional vs multi-directional lighting?  On the other hand, the OpenTrailMap shading does look a bit strange when the map is rotated 180 degrees.  It is quite a cool how the dynamic lighting implemented in this PR causes the hillshading to change as the map is rotated.

> That said, I don’t know how frequently people rotate or tilt this map, since the paper map aesthetic doesn’t necessarily promote that kind of usage.

Unless a map has 3D terrain and buildings[^1] I find map rotation and tilting to be more of a fun gimmick than a useful feature.  I mostly trigger it by accident when using a two finger pinch to zoom on a touch screen but gesture gets interpreted as a slight rotation or tilt instead or as well.  I'd be fine with rotate and tilt being disabled. 


[^1]: Yes Americana technically does have 3D buildings set to a very short height.  I'm referring to full height 3D buildings.

1ec5:
> It does diverge from the paper highway map aesthetic, but I've always felt that some level of hill shading would help fill in the empty spaces.

Actually, there’s quite a bit of precedent for road maps out west to include subtle hillshading. Aside from filling in blank space, it communicates to the motorist whether they’ve got a climb or potentially some challenging winter road conditions ahead of them. What’s new is the modicum of hillshading in flatter regions, where paper maps tend not to bother. But for us, it’s all one map.

> Subjectively, the OpenTrailMap hillshading is a bit more pleasing to my eye. I'm not entirely sure why this is. Maybe uni-directional vs multi-directional lighting?

I assume it has more to do with us fidgeting with the exaggeration, color, and maximum zoom level (which affects the resolution).

jleedev:
As an example, here are the artifacts seen at higher zoom levels (with colors turned up for emphasis):

| 15 | 12 |
|-|-|
| Screenshot 2024-09-23 12 37 10 | Screenshot 2024-09-23 12 38 18

Limiting the resolution smooths out these weird street-shaped ridges. Could maybe go to 13 but by that point it's starting to fade out, so it would just be unnecessary tiles.

zekefarwell:
There are definitely some bad artifacts in this DEM.  In the US it's seems mostly pretty okay, but in other parts of the world it can be a bit janky.  Area of the Alps for example:

Screenshot_23-9-2024_13946_preview ourmap us


location: https://preview.ourmap.us/pr/1157/#map=12.14/46.78211/8.62852

jake-low:
@jleedev: seems like you have a good grasp already of the limitations of prerendered hillshade tiles, but I just wanted to reply with some details about what I did for OpenTrailMap in case it's useful to you.

As you mentioned, there are basically two ways MapLibre can be used to blend a prerendered hillshade tileset with another style layer like a fill layer: either you put the hillshade on top, or you put the fill layer on top. Both have drawbacks. If the hillshade is on top, it washes out the colors (reduces the saturation) of the fill layer. If the fill layer is on top, it reduces the contrast of the hillshading.

You can try to compensate against either effect, by increasing the base saturation of the fill layer style, or the base contrast of the hillshade rasters, respectively. But if the fill layer doesn't cover the whole planet (i.e. there are areas of the map that aren't covered by a polygon in the layer source), then you'll see boundaries in the rendered map where either the saturation of the fill layer or the contrast of the hillshade changes undesirably, depending on which choice you make.

Another factor that might influence this choice is how MapLibre handles opacity blending. When you set a fill layer to 50% opacity, it blends _each feature_ in the layer at that opacity, so areas where features overlap will have a higher total opacity. This may or may not be desired.

What I've personally found to work pretty well, and what I did for the hillshading I worked on for OpenTrailMap, is to set the hillshade layer above the `background` layer and above fill layers for things like landcover and landuse, and then to increase the saturation of colors used in those layers to compensate. I usually set the hillshade layer at between 25% and 50% opacity. For OpenTrailMap I chose 50%. For a map where the terrain wasn't a primary focus I'd probably choose something closer to 25%. I also usually make it so that the hillshade layer reduces opacity further as you zoom in beyond the hillshade tileset's maximum zoom, in order to hide pixelation that appears.

I find this the nicest approach from a cartography perspective because you still have direct control over the colors of all the layers placed underneath the hillshade (and don't have to worry about how they'll blend with each other, which can lead to a "muddy" effect as you mentioned). And the hillshade raster that's overlaid on top has a predictable effect on the colors below: it desaturates them by a factor that's roughly proportional to the opacity of the hillshade. That predictability makes the effect easier to correct for when picking colors.

> Just spitballing, what if the hillshade tiles were instead solid black with alpha channel, so that they can blend with a background color more easily. Or perhaps some color instead of black, so that the raster-* layer properties can be used. (Either webp or png?)

That's an interesting idea. PNG would probably not be a good choice, since it would dramatically increase the size of the tileset (hillshading has lots of high frequency detail that compresses really well as JPEG). WebP might work, though I haven't tested whether MapLibre supports alpha channels in WebP images. You'd still have to deal with the same tradeoff described above, but only in areas with lots of relief - flat areas would be mostly transparent. I don't have an intuitive idea of whether this would look pleasant or not, but it could be interesting to explore that approach.

claysmalley:
Are there any outstanding issues preventing us from merging this as-is? I really want to add `natural=ridge` labels into the style, but they'd look weird alone without hillshading or 3D terrain.

adamfranco:
> Are there any outstanding issues preventing us from merging this as-is? I really want to add `natural=ridge` labels into the style, but they'd look weird alone without hillshading or 3D terrain.

I think it currently looks great and the geography of many areas makes so much more sense with the terrain as context. In evaluating the preview there is quite a delay (14.5 seconds) in loading the terrain tiles. If it wasn't for this delay, I'd suggest making the terrain on by default.

adamfranco (approved):

jleedev:
> quite a delay (14.5 seconds) in loading the terrain tiles

That's probably more of a client issue than a server issue. The hillshade tiles set strong cache headers and load quite quickly; it may be that something in maplibre feels like waiting for the vectors to be computed before adding the raster.

adamfranco:
> > quite a delay (14.5 seconds) in loading the terrain tiles
> 
> That's probably more of a client issue than a server issue. The hillshade tiles set strong cache headers and load quite quickly; it may be that something in maplibre feels like waiting for the vectors to be computed before adding the raster.

Investigating more, I think you are right, @jleedev 👍. The browser is queuing up some requests and they sit blocked and/or waiting for TLS setup before a very sort transfer time. Definitely a client-side issue.
Screenshot 2025-01-15 at 11 06 35 AM

ZeLonewolf:
There is a small noticeable lag on my phone (<1s) when the hillshade loads, but I would consider this a very minor artifact.

PR #740, opened 2023-01-29
Add Lake Pontchartrain Causeway shield

1ec5:
Added a shield for the Lake Pontchartrain Causeway.

[Mandeville](https://zelonewolf.github.io/openstreetmap-americana/#map=10/30.3813/-90.0704) [Crossover](https://zelonewolf.github.io/openstreetmap-americana/#map=18/30.142523/-90.132655/45/40)

Fixes #734.

PR #1162, opened 2024-09-25
Add shield for Utah Federal Aid routes

claysmalley:
Federal Aid routes of Utah do not have guide signs, but are sometimes signed with mileposts that include the route number in a pentagonal "county" shield.

The definition for `network=US:UT:Wayne` is removed here because it was a mistagging by yours truly, corrected [here.](https://www.openstreetmap.org/changeset/156846582)

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -2873,7 +2873,7 @@ export function loadShields() {
       bottom: 5,
     },
   };
-  ["Wayne", "Washington"].forEach(
+  ["FA", "Washington"].forEach(
     (county) =>
       (shields[`US:UT:${county}`] = pentagonUpShield(
         3,


1ec5:
Interesting, got an example of one of these mileposts?

claysmalley:
Sadly, the best I have is [this blurry photo](https://www.bing.com/maps/?toWww=1&redig=0DA531F35E4F4F21AB5197F78570D957&cp=38.241697%7E-111.377999&lvl=19.7&pi=-6.1&style=x&dir=327.8) and a faint memory of driving through the area in 2021. The milepost can be seen below the speed limit sign, on the same assembly.

Arguably, these signs may be small enough to be considered `unsigned_ref=*`. The only county route shield I'm aware of in Utah that has full-sized reassurance signs is Washington County Road 91, a former alignment of US 91.

1ec5 (approved):
Well, your call. It sounds like it’s intentional rather than a one-off sign goof, and we do have some precedent for milepost-only shields like #740.

PR #1163, opened 2024-10-18
Add Mormon Pioneer National Historic Trail Auto Tour Route shield

quincylvania:
This is just an orange shield since we don't have permission to use NHT shields yet, but the route is mapped.


1ec5 (approved):

quincylvania:
A note of caution, some of these NHT are collinear with long stretches of interstates and US routes, so the shields may show up at relatively low zooms. This is z6 in Utah with this PR. The orange blob looks kind of weird considering there is no graphic.

Screenshot 2024-10-21 at 10 44 50 AM

quincylvania:
#1165 is now fixed, but without #1184 there is still no labels for the individual route shields.

We really need to figure out the protected insignia thing.

ZeLonewolf:
With the insignia issue solved, do we want to move forward with this or re-do it with fancier artwork?

quincylvania:
> do we want to move forward with this or re-do it with fancier artwork?

I can add icons to the shields at some point. No point in merging empty shields now.

quincylvania:
Lookin' better. I don't know why there is an extra shield space though.

Screenshot 2025-02-17 at 9 02 29 PM

quincylvania:
The blank space will be fixed by #1213.

PR #1166, opened 2024-11-14
Update Node.js in CI

jleedev:
Switch from 18.x (maintenance LTS until next spring) to 20.x (maintenance LTS for a year past that).

Rather than repeating the version in every workflow, reference the tool version by filename.

Rather than specifying an exact major version, give actions/setup-node a range of major versions.

The node-canvas dependency has prebuilt binaries for up to node 20, so limit to that in engines. They have a version 3 release candidate that switches to N-API, which is a more stable interface that will ultimately avoid a lot of this trouble.

Because of the .npmrc, this will cause npm install to loudly fail on an unsupported version, even if one is willing to e.g. build node-canvas from source. Do `--no-engine-strict` or `npm_config_engine_strict=false` in the environment to ignore this.


ZeLonewolf (approved):
Looks good to me!

PR #1153, opened 2024-09-11
Adding POI icon for restaurants and fast food

whubsch:
Finally got around to adding an icon for `restaurant`, `fast_food`, and `food_court`. Folks on the Slack channel seemed to agree that combining these, rather than using different icons for `restaurant` and `fast_food`, was preferable. I went back and forth between parallel and diagonal utensils, but I determined that the parallel ones look wonky when there are several restaurants stacked, so to speak. Since they are on the diagonal, there's not much pixel alignment to be done.

Relatedly, I think it would make sense to pursue adding the cuisine tag to OMT (discussion started in openmaptiles/openmaptiles#1527) so that we can differentiate between the various POIs and get a little more visual variety.

This PR makes progress on #692. Open to any feedback or suggestions!

Screenshot 2024-09-11 at 6 27 51 AM
Screenshot 2024-09-11 at 6 24 28 AM

PR #1170, opened 2024-11-23
Distinct food court icon

wmisener:
PR #1153, which added restaurants, also rendered food courts (`amenity=food_court`) with the same restaurant knife-and-fork icon. But food courts aren't exactly restaurants: they contain restaurants. In a sense, to dust off my SAT analogies, a restaurant is to a food court like a bus stop is to a bus station. Thus, I propose a food court icon that is the restaurant icon inscribed in a circle, following the logic Americana has already taken in #845 for bus and train stations.
poi_restaurant_circle

Old:
Screenshot 2024-11-22 at 9 50 22 PM
New:
Screenshot 2024-11-22 at 9 50 32 PM

One slightly unfortunate aspect is that I had the fork and knife icon is pretty much 20x20, so I had to shrink them a bit to fit within a 20px diameter circle. So the utensils appear a bit smaller in the food court icon than in nearby restaurant icons, but hopefully the circle's added prominence makes up for it. I guess one possible solution would be to shrink the restaurant icon so that it's not 20x20 (some icons aren't, like bus stops), but I don't know that it's necessary just for this.

The icon appears at z15, one before restaurants. It can show up without a label at that zoom, like supermarkets and gas stations. (Side note: which icons do and don't appear without labels at certain zooms is still quite ad hoc, the style would probably benefit from some stated logic for it)
Screenshot 2024-11-22 at 9 50 11 PM

More samples:
Screenshot 2024-11-22 at 9 51 31 PM
Screenshot 2024-11-22 at 9 51 48 PM
Screenshot 2024-11-22 at 9 52 42 PM

Funnily enough, the food court icon would now be the first POI that shows up in the style for many large malls. I don't think that's a flaw per se, but rather a sign that things like malls and department stores should probably be fairly high priorities for POIs the project should tackle.


claysmalley (approved):
Sure, why not?

ZeLonewolf:
@wmisener can you run code_format and merge?

PR #1174, opened 2024-12-01
Add support for place=province equivalent to place=state

zekefarwell:
This adds support for support for `place=province` nodes, giving them the same label text treatment as already supported `place=state` nodes.  The status quo for Canadian provinces is that their place nodes have been tagged `place=state` but should they be changed to `place=province` we'd still want labels on the map.   It looks like `place=province` is currently used in a few other countries including Turkey, Myanmar, and Japan.  We should make sure these are direct country subdivisions before merging this.


ZeLonewolf:
Confirmed that this works on the current planet (before restoration of place=state nodes).

I think this can be done a bit more compactly with a single "in" expression, something like:

```
filter: ["in", ["get", "class"], "state", "province"]
```

src/layer/place.js
```diff
@@ -394,7 +394,11 @@ export const state = {
     ],
     "text-halo-blur": labelHaloBlur,
   },
-  filter: ["==", ["get", "class"], "state"],
+  filter: [
+    "any",
+    ["==", ["get", "class"], "state"],
+    ["==", ["get", "class"], "province"],
+  ],
```
1ec5, line 401: For brevity, use a `match` or `in` expression:

```json
["match", ["get", "class"], ["state", "province"], true, false]
```
```json
["in", ["get", "class"], ["literal", ["state", "province"]]]
```

zekefarwell: Thank you!  I was originally trying to get an `in` expression to work, but I couldn't quite figure out what the syntax was supposed to be.  Seems I was missing the need to wrap the array in a `literal` expression.

zekefarwell:
Here's a overpass query showing all `place=province` nodes worldwide: https://overpass-turbo.eu/s/1V25

It looks to me like the `place=province` nodes are used consistently across Turkey, Thailand, South Korea, and Japan to represent direct national subdivisions.  Beyond that, the tag is used sporadically in Italy, Saudi Arabia, Burundi, and Peru.  These uses are less clear to me but there are only a few of them.  [place=province is clearly documented](https://wiki.openstreetmap.org/wiki/Tag:place%3Dprovince) as being for "a subdivision of a [country](https://wiki.openstreetmap.org/wiki/Tag:place%3Dcountry) similar to a [state](https://wiki.openstreetmap.org/wiki/Tag:place%3Dstate)".  So I'm fairly confident this is a good change and if any current uses are for lower level divisions they should be changed anyway.

1ec5 (approved):

PR #1178, opened 2024-12-10
Adding highway shields in Peru

whubsch:
This PR adds the Peruvian national, departmental, and rural route shields to the map after some [tagging changes](https://community.openstreetmap.org/t/propuesta-de-cambio-en-las-relaciones-viales-en-peru/122393). Only the national routes required a new shield.

[Localhost link](http://localhost:1776/#map=13.06/-13.53598/-71.99538)
Screenshot 2024-12-09 at 10 59 51 PM

Pan-American, national, departmental, and rural highway shields in Peru
Screenshot 2024-11-30 at 6 50 00 PM


src/js/shield_defs.js
```diff
@@ -3164,6 +3164,53 @@ export function loadShields() {
...
+    "TA",
+    "TU",
+    "UC",
+  ].forEach(
+    (department) =>
+      ([
+        shields[`PE:departmental:${department}`],
+        shields[`PE:rural:${department}`],
+      ] = [
+        escutcheonDownShield(12, Color.shields.white, Color.shields.black),
```
claysmalley, line 3209: These shields are visually closer to the state highway shields of Mexico (see line 584). Let's factor that out into a reusable shield definition and use it here.

Side note: this is making it really clear how much we need a draw function for badge shields, and maybe support for crossbars and accent colors in some of the functions...

whubsch:
Ok, turning back to this after the holidays. I've moved the definition for the crossbar'd escutcheon shield so that it can be reused, and renamed those `svg` files to have a more generic name (`shield_escutcheon_crossbar_X`).

Screenshot 2025-01-07 at 4 14 30 PM

src/js/shield_defs.js
whubsch: Thanks, this should be addressed with the latest commit.

claysmalley (approved):
Thank you!

PR #1179, opened 2024-12-15
Remove padding from the default shield for unknown networks

Blijbol:
This avoids wasting space (on each side, 3 pixels were reserved for the shield, but never actually used) and avoids unnecessarily small font sizes (e.g. 4-character refs are now much more readable).


claysmalley (approved):
This looks much better. Thank you!

[image](https://preview.ourmap.us/pr/1179/#map=7.3/52.665/13.924)

Issue #1165, opened 2024-10-21
overrideByName causes legend to omit shields


1ec5:
If a network uses `overrideByName`, its entry in the legend shows a blank space instead of any matching shields. For example, the legend for [this view](https://americanamap.org/#map=10/41.2631/-73.1864) should show a shield beside the entry for Merritt Parkway:

<img src="https://github.com/user-attachments/assets/b01cc3c1-1e66-4e6b-ad4f-6b8d6e878714" width="460" alt="Legend: Merritt Parkway without a shield beside it.">

This network’s definition relies on `overrideByName`:

https://github.com/osm-americana/openstreetmap-americana/blob/b50ef87edd165240c119e49cb939d8d532ffe800/src/js/shield_defs.js#L1106-L1113

We need to add something similar to the special case for `overrideByRef`, which is working correctly:

https://github.com/osm-americana/openstreetmap-americana/blob/b50ef87edd165240c119e49cb939d8d532ffe800/src/js/legend_control.js#L474-L478 https://github.com/osm-americana/openstreetmap-americana/blob/b50ef87edd165240c119e49cb939d8d532ffe800/src/js/legend_control.js#L503

PR #1180, opened 2024-12-16
overrideByName causes legend to omit shields #1165

joshuamanivinod:
Hi @1ec5 

I have implemented a fix for this issue. The updated logic ensures that shields appear correctly in the legend even when a network uses `overrideByName`. Specifically, the `getSortedImages` function now accounts for `overrideByName` alongside existing overrides, ensuring no blank spaces are shown.

Let me know if this solution meets your expectations or if further adjustments are needed.

Screenshot 2024-12-16 at 3 30 17 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -50,7 +50,7 @@
         "typescript": "^4.9.5"
       },
       "engines": {
-        "node": ">=18",
+        "node": ">=18 <=20",
         "npm": ">=8.3.0"
       }
     },
@@ -6300,7 +6300,7 @@
     },
     "shieldlib": {
       "name": "@americana/maplibre-shield-generator",
-      "version": "0.0.5",
+      "version": "0.0.7",
       "license": "CC0-1.0",
       "dependencies": {
         "@types/node": "^20.8.4",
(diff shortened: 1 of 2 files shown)


ZeLonewolf:
Thanks @joshuamanivinod for the contribution. Do we know why the shield is shifted to the right compared to the others?

joshuamanivinod:
It looks like two images are being displayed side by side, which is causing the image to appear slightly shifted. I'm trying to determine the reason for this issue.

Screenshot 2024-12-17 at 6 35 46 PM

joshuamanivinod:
Handle overrideByName causing legend to omit shields
- Updated logic in getSortedImages to conditionally include images.noRef 
  based on the presence of overridesByName.
- Ensures that the legend properly displays shields when overridesByName 
  is present and noRef is not required.
  
  
Screenshot 2024-12-31 at 2 15 11 PM

ZeLonewolf (dismissed):
Looks  good, thanks!  Please remove the extra comment and this is good to go.

ZeLonewolf (approved):
Looks good to me!

PR #1183, opened 2024-12-31
Widen highways at very high zoom

### LINKED ISSUE #372: Increase default highway width to better match physical width at high zoom

claysmalley:
Fixes #372

This PR changes the width of highways at zoom level 16 and above, so that they widen in direct proportion with the zoom level. At this level, even the narrowest `highway=trunk` city streets fit in their physical footprint without overlapping sidewalks.

[before](https://americanamap.org/#map=17.61/29.955931/-90.073232)/[after](https://preview.ourmap.us/pr/1183/#map=17.61/29.955931/-90.073232):

image

image


ZeLonewolf:
Thanks, this looks awesome!  There's some code in here that looks like it's modifying the ferry styles, is that intended?

claysmalley:
> There's some code in here that looks like it's modifying the ferry styles, is that intended?

It's actually there to ensure ferries maintain the same styling while other things change around it. `zoom` expressions are what they are :shrug:

wmisener:
Great work! This is a big improvement in road rendering at high z-level, and goes a long way toward addressing the longstanding issue #372 (in fact, I think I'd say it fully addresses it). In light of the discussion there, it'd be interesting to calculate out what the effective width of each way actually is. It also doesn't preclude an approach like #486 later on with mapped highway areas.

I tested around and agree with your assessment that no roads seem to be overlapping roadside features, with a big boost in attractiveness in my opinion. Here's old and new in DC:
Screenshot 2025-01-01 at 1 36 55 PM
Screenshot 2025-01-01 at 1 36 49 PM

https://github.com/osm-americana/openstreetmap-americana/issues/372#issuecomment-1150419745 raised the issue that care might be needed for places outside the mid-latitudes. I'm not sure how the current code approaches that, but I don't see any glaring issues in Hilo, HI or Fairbanks, AK:
Screenshot 2025-01-01 at 1 55 01 PM
Screenshot 2025-01-01 at 1 54 33 PM

claysmalley:
> #372 raised the issue that care might be needed for places outside the mid-latitudes.

Out of curiosity, I looked at Jakarta, a city near the equator with tons of narrow streets. Seems that we're running up against the limits in a place like this, though I believe most of the overlapping buildings here are just imperfectly aligned. On top of that, most of these streets are tagged `highway=living_street` with access restrictions, and would probably be good candidates for a [new tag for narrow streets](https://community.openstreetmap.org/t/should-narrow-roads-that-are-too-small-for-cars-still-be-tagged-as-minor-roads/117397) that we could potentially use in the future.

([location](https://preview.ourmap.us/pr/1183/#map=18.36/-6.145613/106.808996))

image

wmisener:
> Out of curiosity, I looked at Jakarta, a city near the equator with tons of narrow streets. Seems that we're running up against the limits in a place like this, though I believe most of the overlapping buildings here are just imperfectly aligned. On top of that, most of these streets are tagged highway=living_street with access restrictions, and would probably be good candidates for a [new tag for narrow streets](https://community.openstreetmap.org/t/should-narrow-roads-that-are-too-small-for-cars-still-be-tagged-as-minor-roads/117397) that we could potentially use in the future.

Other OSM renderers also struggle with areas like this:
Screenshot 2025-01-01 at 10 07 05 PM
So I don't think it should be a blocker.

wmisener:
One note, did you consider including aeroways in the scaling relation? They appear to stay constant width as the roads expand (old vs new [at this location](https://preview.ourmap.us/pr/1183/#map=17/36.081154/-115.169992)):
Screenshot 2025-01-01 at 10 14 20 PM
Screenshot 2025-01-01 at 10 14 25 PM

ZeLonewolf:
Does anyone object to moving forward with this change?  I do think it's an improvement and we can tackle the aeroways as a separate PR.

wmisener:
I'm OK with it! I only brought up the aeroways in case it was simpler to make the changes all at once, I don't think it's a blocking issue.

claysmalley:
I totally overlooked the comment about aeroways. My bad.

I agree though—we can deal with aeroways in another PR, just to make things easier for me :wink:

ZeLonewolf (approved):
Looks great, thanks!

PR #1186, opened 2025-01-03
Add several indigenous American languages to language picker

claysmalley:
This PR adds to the language picker 207 languages indigenous to the so-called United States. Since Unicode CLDR is missing language names for most of them, I've included the English name of each as a fallback.

Note that many of these languages currently have no geographic labels on OSM or Wikidata, and the map will just fall back to English. By surfacing these languages in the picker, I hope to encourage mappers to fill in the gaps.


src/js/language_label.js
```diff
@@ -40,77 +40,309 @@ let languageNames = new Intl.DisplayNames(Label.getLocales(), {
...
+  { id: "th" },
+  { id: "tr" },
+  { id: "uk" },
+  { id: "zh" },
+
+  // Indigenous languages of the United States
+
+  // These languages are grouped by linguistic family, in loose geographical
+  // order from northwest to southeast. The English name of each language is
+  // included as a fallback, in case CLDR is missing data on language names.
```
1ec5, line 117: I wonder if sorting the list alphabetically by language code would be easier to maintain than this geographical sort order. We could easily end up with duplicates.

src/js/language_label.js
```diff
@@ -40,77 +40,309 @@ let languageNames = new Intl.DisplayNames(Label.getLocales(), {
...
+
+  // Eskaleut
+  { id: "ale", name: "Aleut" },
```
1ec5, line 120: This is fine as a start, but maybe we could try to source the names dynamically from Wikidata, like we do for route networks in the legend. That would allow us to support more interface languages than just English, which seems appropriate for a feature focused on multilingualism.

claysmalley: I'm in favor of sourcing language labels from Wikidata, though I'm not experienced with async Wikidata queries. I'll make a separate issue for it.

ZeLonewolf:
In testing, I was not able to get "Aleut" to work in the language picker.

1ec5:
> In testing, I was not able to get "Aleut" to work in the language picker.

Works for me in Firefox 146 for macOS. This PR is relying on the browser to supply a language name for `ale`. Which browser and browser version are you using?

Aleut

claysmalley:
Will need a rewrite as a result of #1368.

1ec5:
The main thing left would be to extend the list of suggested items with languages that we want to support beyond browser support (along with their hard-coded names in English, I suppose).

[This Bugzilla meta bug](https://bugzilla.mozilla.org/show_bug.cgi?id=356038) tracks all the work for Firefox to support BCP 47, including more ISO 639-3 codes. But surprisingly it doesn’t even support some valid ISO 639-2 codes like `iu`.

PR #1192, opened 2025-01-20
Remove shield for Purchase Parkway (Kentucky)

claysmalley:
The [Purchase Parkway](https://en.wikipedia.org/wiki/Purchase_Parkway) is no more, after being fully upgraded to Interstate 69. No need to keep this designation around in the shield definitions.


ZeLonewolf (approved):
LGTM

PR #1193, opened 2025-02-01
Add shield for Hunebed Highway (Drenthe, Netherlands)

Blijbol:
Introduced in 2018, the Hunebed Highway is a touristic route in the Dutch province of Drenthe. It is a part of the N34 expressway. The Hunebed Highway N34 shield is signed separately in addition to the regular N34 shields.

Photos of road signs: https://www.wegenforum.nl/viewtopic.php?p=1281713#p1281713 (open the "spoiler" for many more photos)

`shield_nl_dr_hunebed` is derived from `shield_kr_expressway_2` with the border colour changed from white to black (I didn't use `shield_kr_expressway_3` because in this case, aspect ratio is more important for shield recognisability than readability of the N34 inscription).


src/js/shield_defs.js
```diff
@@ -4011,6 +4011,16 @@ export function loadShields() {
...
   );
+  shields["NL:DR:Hunebed_Highway"] = {
+    spriteBlank: ["shield_nl_dr_hunebed"],
+    textColor: Color.shields.white,
+    padding: {
+      left: 4,
+      right: 4,
+      top: 6,
+      bottom: 4,
+    },
```
1ec5, line 4022: Do we need to decrease the padding or make the overall shield larger in both dimensions? The original image was designed for only two characters, so stuffing three characters in there could make it difficult to discern the number.

Blijbol: > Or does the concurrency with the usual N34 shield makes discerning the number much less important?

Basically this, but I've now widened the shield anyway.

Blijbol:
I now derived `shield_nl_dr_hunebed` from `shield_kr_expressway_3` instead of `shield_kr_expressway_2`. This improves the readability of the "N34" text on the shield.

ZeLonewolf (approved):
Thanks, this looks great!

PR #1194, opened 2025-02-11
Re-add National Historic Trail shield graphics

quincylvania:
This PR essentially reverses #1120. These types of "official emblems" are all over Wikipedia and Wikimedia Commons [per legal argument](https://graphics8.nytimes.com/packages/pdf/us/20100803-wiki-LetterToLarson.pdf). Americana is clearly not attempting to impersonate a federal badge or identification card. I'm no lawyer, but we seem well within our rights to use low-res (or even high res) versions of these graphics for our community map project.


ZeLonewolf:
As Americana is an OSM US community project, if OSM US concurs that we're in the clear then I agree with restoring the imagery.

quincylvania:
I can say that as Tech Lead of OSM US, it is my assessment that the images are being used properly and I take responsibility for this decision.

ZeLonewolf (approved):
Seems like we have consensus with no objections

Issue #464, opened 2022-06-28
First- and second-level routes of Iran


claysmalley:
Iran has two levels of road numbering underneath freeways (#463). First-level roads have a white-on-green rectangular shield, and second-level roads have a black-on-white rectangular shield.

These two networks are conflated by the tag `network=ir:national`. For support to be added for these shields, the tagging scheme would need to be changed to distinguish first- and second-level routes.

![Iran_First_Level_Road_44](https://user-images.githubusercontent.com/1732117/176074596-2c758fee-ae5d-480a-8d35-49c2f98a6391.png) ![Iran_Second_Level_Road_122](https://user-images.githubusercontent.com/1732117/176074571-f0a54df1-acad-4827-8df4-65126f3dca66.png)

1ec5:
[Asked for help in the OSM Asia Telegram chat](https://t.me/OpenStreetMapAsia/12978).

claysmalley:
That would make sense, though we don't (and won't) support switching shield appearance based on the number of digits. We expect that `network=*` is enough to determine what the shield should look like.

There are limited, explicitly documented exceptions to this, where we override the shield for a `network=*` based on its `ref=*` value. But we do this on an individual basis, not for a whole range of numbers.

seav:
But I imagine changing the tagging scheme, which seems to work well for Iranian mappers given how it's structured and how it seems there won't be any overlap between 1st- and 2nd-level `ref=*` values, just to suit your renderer as a classic case of tagging for the renderer. It's not Iranian mappers' fault that the shield colors are different.

1ec5:
> But I imagine changing the tagging scheme, which seems to work well for Iranian mappers given how it's structured and how it seems there won't be any overlap between 1st- and 2nd-level `ref=*` values, just to suit your renderer as a classic case of tagging for the renderer.

A charge of “tagging for the renderer” shouldn’t be leveled against a suggested change solely because a tagging issue has been exposed by a renderer. Essentially, it’s our understanding that `network` has been misunderstood as a differently formatted version of `operator`, whereas in fact the national road authority operates two different highway systems. Indeed, they communicate this distinction to the public using distinctly colored shields. The color is _color-coding_, not a mere splash of decorative color.

> It's not Iranian mappers' fault that the shield colors are different.

[...]

PR #1202, opened 2025-03-16
Fixing capitalization of Iran highway network tagging

whubsch:
It looks like highway relations in Iran were updated to use the capitalized `IR` last August, so this PR updates the shield definition to align with that. I've also added shields for the second tier of road relations in Iran, which are tagged as `IR:national` and are green.

Example of roads tagged with `IR:national`
Screenshot 2025-03-16 at 7 38 31 AM

[Localhost link](http://localhost:1776/#map=9.03/34.7282/51.0156)
Screenshot 2025-03-16 at 7 42 06 AM

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -3455,8 +3455,12 @@ export function loadShields() {
   };
 
   // Iran
-  shields["ir:freeway"] = roundedRectShield(
+  shields["IR:freeway"] = roundedRectShield(
     Color.shields.blue,
+    Color.shields.white
+  );
+  shields["IR:national"] = roundedRectShield(
+    Color.shields.green,
     Color.shields.white
   );
 
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -3459,10 +3459,6 @@ export function loadShields() {
     Color.shields.blue,
     Color.shields.white
   );
-  shields["IR:national"] = roundedRectShield(
-    Color.shields.green,
-    Color.shields.white
-  );
 
   // Iraq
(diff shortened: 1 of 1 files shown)


src/js/shield_defs.js
```diff
@@ -3455,10 +3455,14 @@ export function loadShields() {
   };
 
   // Iran
-  shields["ir:freeway"] = roundedRectShield(
+  shields["IR:freeway"] = roundedRectShield(
     Color.shields.blue,
     Color.shields.white
   );
+  shields["IR:national"] = roundedRectShield(
```
1ec5, line 3462: #464 mentions that `network=IR:national` conflates two different networks that have different shields. Is that still the case?

whubsch: I had not seen that discussion honestly, but now that I'm looking, I can only see only three `IR:national` relations that clearly fall into that mistaken category. So it may not be a tagging schema issue so much as a couple mistakes? I'm willing to take a stab at retagging those three to something else (`IR:secondary` or something) too.

I can also just cut `IR:national` from this PR.

1ec5: `IR:national` doesn’t seem like a very intuitive value for either the first-level roads or second-level roads. In situations like this, we’ve typically reached out to the local communities to confirm their intention. Usually it turns out that they were just copying what they thought the scheme was supposed to be from other countries.

whubsch: Ok, I've removed that from this PR and that can be handled separately.

1ec5 (approved):

PR #1203, opened 2025-04-26
Add initialism for Bear Mountain Parkway

ZLima12:
The name in OSM was changed to remove the "State" from "Bear Mountain State Parkway", since it is usually omitted in speech and is not signed on the ground at all. However, since it is possible that someone might change it back, this commit does not remove the mapping for the old name.

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -2157,6 +2157,7 @@ export function loadShields() {
     colorLighten: Color.shields.white,
     colorDarken: Color.shields.green,
     refsByName: {
+      "Bear Mountain Parkway": "BMP",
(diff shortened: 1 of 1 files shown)


1ec5 (approved):
This is OK, but it really highlights the fragility of `refsByName` versus something more data-driven. [This relation](https://www.openstreetmap.org/relation/401758) is currently tagged `name:abbreviation=BMP`. Ideally, it would be `short_name=*`, the more prevalent key, and OpenMapTiles would expose this information even in the presence of `name=*`: https://github.com/openmaptiles/openmaptiles/pull/1576#discussion_r1327893543.

PR #1216, opened 2025-06-21
Adding Charging Stations icons to map following fuel template

GA-Kevin-Codes:
Historically, road atlases did not have charging stations as they did not exist until relatively recently. In line with the Americana goal to "Challenge the status quo by showcasing innovation and invention" the PR aims to add modern road infrastructure around EV charging and uses the icon from the OSM wiki and Wikimedia Commons. Happy to address any questions, comments, or concerns, just let me know!


ZeLonewolf:
Hi @GA-Kevin-Codes, thanks for the PR!

There are some build errors that will need to be resolved, which you can examine in the checks tab.
Also, you'll need to run `npm run code_format` to format the code, which will make the lint test pass.
Lastly, please provide screen shots or a link to the preview so we can easily test. Once the build passes, the checks tab will have a link to a "PR preview", which is a live map that will show the new rendering.

ZeLonewolf:
Screenshot:

image

1ec5:
The icon looks quite small next to the gas station icon.

ZeLonewolf:
I was thinking the same thing. Would it make sense to just take the traditional gas pump icon and put the lightning bolt on it?

GA-Kevin-Codes:
Updated the icon to look like the fuel icon with a bolt instead of gallons/price screen:

image

ZeLonewolf:
Great job with the new design, I think that makes sense! I like taking out the window so you have more room for the lightning bolt.

It might be my eyes, but it looks like the icon isn't aligned to the pixel grid, and therefore the edges aren't sharp. That's why it looks a little fuzzy. If you compare the left edge of the gas pump versus the left edge of the charger pump on the zoom in below, you can see the difference. That makes a pretty big difference in readability.

Take a look [here](https://github.com/osm-americana/openstreetmap-americana/blob/main/CONTRIBUTING.md#highway-shield-contributors-guide) on how to turn on the pixel grid in Inkscape, and see if the alignment issue can be fixed.

image

whubsch:
Thanks for adding these, when I worked on #999 I think OpenMapTiles didn't support charging stations, otherwise I would have knocked this out then. A couple points to consider:

1. It looks like the charging icon doesn't have the white halo that all the other POI icons have, and the lightning bolt should similarly be white rather than transparent to match the gas station "display". I think the lightning bolt is also potentially too big, but that is more a matter of taste, and pixel alignment might dictate that to some extent.
2. It looks like the charging stations inherited the zoom level that gas stations appear at, which is exceptionally low (or high?, always get confused with this but zoomed out), higher than even the average `shop` POI. I feel like the visual prominence of a gas station vs charging station in real life warrants a likewise less-prominent treatment for the charging station on maps. No one is really navigating based off of a charging station, which can be as simple as a glorified parking space.

Welcome discussion or thoughts on those!

ZeLonewolf:
> 2\. No one is really navigating based off of a charging station

I mean... there are quite a few electric cars on the roads these days that are _desperately_ navigating to charging stations :grinning:

1ec5:
Yes, destination charging causes drivers of electric vehicles to go even farther out of their way than drivers of gas-powered vehicles. Charging stations are one of the motorist amenity symbols that are signposted at freeway exits in several countries, including the U.S.

GA-Kevin-Codes:
image

ZeLonewolf:
Looks like we still need to tweak a couple things.  This is the sprite sheet on a black background -- the 1px outline needs to get added as noted above, and the lightning bolt should be white rather than a cut-out. I will try to see if I can adjust it when I have a few minutes (unless someone else gets to it first)

image

GA-Kevin-Codes:
Thank you, I evidently need some help after something like 10 iterations! lol

ZeLonewolf:
Ahhhhh

image

Issue #1125, opened 2024-07-11
Courthouses


wmisener:
Courthouses are the last remaining "community anchor institution" listed in #435 that has no PR. Obvious symbolism would include a gavel or the scales of justice, as seen in Temaki and Carto, respectively:
<img width="170" alt="Screen Shot 2024-07-10 at 8 17 49 PM" src="https://github.com/osm-americana/openstreetmap-americana/assets/58491489/16cedf9d-2f23-4725-901c-45e7fc72c5b2"> <img width="170" alt="Screen Shot 2024-07-10 at 8 20 08 PM" src="https://github.com/osm-americana/openstreetmap-americana/assets/58491489/4246e152-4785-412a-8c0c-28f286164f84">

I think I would be more partial to a gavel, although care will be needed to not look like some miscellaneous hammer.

Paras-5:
Hi, I'd like to work on this issue. Can you assign it to me?

wmisener:
Hi @Paras-5, sure! But feel free to open a PR that would address any issue without it needing to be assigned to you first.

PR #1219, opened 2025-06-29
Adding Courthouse

GA-Kevin-Codes:


GA-Kevin-Codes:
image
image

wmisener:
Thanks for putting this together! As the last main "community anchor institution" not yet rendered in Americana, a courthouse rendering is definitely welcome. Conceptually, as I wrote in #1125, the scales of justice motif is quite similar to Carto's, and I think it's recognizable both to a broad American audience and to the OSM community. Like I mentioned in the issue, I would've probably tried a gavel myself, but I think this works OK.

I will say, I think the icon feels a little more spindly and less strong than some of the other similar icons, especially the balances themselves. In part this might be due to the icon not being pixel-aligned, i.e., the corners and straight edges aren't snapped to the exact .000 pixel values (usually you can turn on such a grid and snap to the intersections in the software you're editing these in), similar to what @ZeLonewolf mentioned in https://github.com/osm-americana/openstreetmap-americana/pull/1216#issuecomment-2997863426. This leads to the image looking a little fuzzy when it gets rendered. Can you try snapping the edges to the pixel grid? I think it could also do with being a bit wider, it appears quite narrow to my eye right now.

Screenshot 2025-06-29 at 8 59 06 PM

If it still looks too thin, maybe try thickening the lines holding the balances? The icon can be up to 20x20 (including the halo, so really 18x18), and it's only 17x19 now, so feel free to make it wider or a little taller.

scripts/taginfo_template.json
```diff
@@ -650,6 +650,14 @@
       "doc_url": "https://openmaptiles.org/schema/#poi",
       "icon_url": "https://raw.githubusercontent.com/osm-americana/openstreetmap-americana/main/icons/poi_town_hall.svg"
     },
+    {
+      "key": "amenity",
+      "value": "courthouse",
+      "object_types": ["node", "area"],
+      "description": "A building which is home to a court of law, which administers justice according to the rule of law.",
```
wmisener, line 657: This description is of the icon itself, not the POI. So maybe something like "Courthouses are marked by an icon representing the scales of justice" or something?

ZeLonewolf:
Here is the icon laid out on the pixel grid. The thicker lines are for 1x icons and the thinner lines are for 2x icons. Ideally, we aim to align to the thicker lines where possible, and the thinner lines only if that can't be done.  This results in crisp, clean icons on all devices. Horizontal and vertical lines in particular should make every attempt to snap to the coarse grid.

Below is the current icon on the pixel grid.

image

 In Inkscape, you can turn this on by pressing Shift+Ctrl+D, which gives this dialog.  I'm not sure what the equivalent is on Mac.

image

ZeLonewolf:
Here is one example of what pixel-alignment could look like.  Notice how I snapped a number of corners and edges to the coarse parts of the grid. Obviously you'l have to play with it a bit to get a look that looks best at regular zoom.

image

File:
poi_courthouse_aligned

1ec5:
@GA-Kevin-Codes what do you think of the pixel-aligned icon suggested in https://github.com/osm-americana/openstreetmap-americana/pull/1219#issuecomment-3017490596? Or does the gavel mentioned in https://github.com/osm-americana/openstreetmap-americana/pull/1219#issuecomment-3017426329 sound appealing? Most of the maps in my collection have courthouses as generic squares or “CH” monograms, since courthouses aren’t terribly common on those maps. But either the scales or a gavel seem reasonable to me.

GA-Kevin-Codes:
@1ec5 I lean toward the scales, but I'm not married to it if others feel it should be different. The pixel aligned one looks good to my eye. Apologies, Americana icon requirements confuse me sometimes!

1ec5:
OK, then the next steps are:

* [ ] Drop the file from https://github.com/osm-americana/openstreetmap-americana/pull/1219#issuecomment-3017490596 into your branch
* [ ] Resolve the merge conflict
* [ ] Update the taginfo description per https://github.com/osm-americana/openstreetmap-americana/pull/1219#discussion_r2174057813

Let us know if you need help with any of that. Looking forward to closing the case on this feature.

PR #1222, opened 2025-07-04
Convert all scripts to typescript

ZeLonewolf:
This PR:
* Converts the remainder of the various build and helper scripts in the scripts/ folder to typescript
* Fixes an error in the stats script that was preventing it from running

I had to remove the locale option from the stats script, because this script needs to operate on a generated style.json rather than invoking the style builder code. This is needed for PR and main branch separation in the build CI. This was also the reason that the stats comparison has been broken for awhile.

Prover that the new comparison script is working (tested with a dummy change of duplicating buildings):

image

This unblocks conversion of downstream code to typescript.


src/js/shield_defs.js
```diff
@@ -24,6 +24,7 @@ import {
   bransonRouteShield,
 } from "@americana/maplibre-shield-generator";
 
+//TODO: ensure this returns ShieldSpecification
```
ZeLonewolf, line 27: This is just a note to a future developer if this file is converted to TypeScript

PR #1223, opened 2025-07-10
Update Node.js (18 → 20)

jleedev:
Node.js 18 is end of life as of March; 20 is ‘maintenance LTS’ and 22 is ‘active LTS’.

npm:canvas@3 compiles for me under Node.js 20.19.3 and 22.17.0, but it fails on 24.4.0.

(Followup from #1218; our CI uses the same version of Node.js to test both the branch and main.)


jleedev:
Ohh, in fact it just uses node-version-file from pr-branch and then tries running it in main and then NPM throws a fit because we have asked it to.

I'm sure there's a way to make CI happy but I don't feel like reading https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md

ZeLonewolf:
I've run into this before. TLDR it's a security feature to prevent a user from bypassing security controls by overtaking CI with a PR that changes the GH actions. Or something like that. 

If we're confident that this will run normally after updating the CI in main, we should bypass the merge protections.

ZeLonewolf:
In theory, the merge protections are overridable now.

ZeLonewolf:
Okay so this ended up being more complicated than I expected, and the issue wasn't what I thought it was.

I had to change some things to make the higher version of Node work, namely:
1. Make the CI runners install a bunch of stuff on the OS to make Canvas work. I'm not sure if that also means that developers need to install that stuff on their box.  Probably?  If so that likely needs to get added to the developer instructions somewhere.
2. Upgrade Canvas itself to be compatible with the newer Node version
3. There was some global.navigator references usages in label.js that broke. In newer Node.js versions, the global.navigator object is read-only so you have to use Object.defineProperty(...) instead
4. I checked in the corresponding package-lock.json
5. The PR preview explicitly uses the PR branch when generating samples and statistics.  This ensures that PR changes always use the same (newest) scripts for generating these. If you change someome about samples and stats, you want the PR change to do the sampling. If a PR has a npm version bump, that means we have to switch to the main branch npm when building or running the code, but the PR branch npm when building/running the PR code AND all of the samples/stats code

I'm not 100% sure that all of these were required, but these were the issues I worked through to get this to a green board.

1ec5 (approved):
The label.js changes look good.

jleedev:
Nice! Thanks

PR #1228, opened 2025-07-12
Add Catskill Mountains and Durham Valley Scenic Byway shields

quincylvania:
https://www.openstreetmap.org/relation/19246349

Screenshot 2025-07-12 at 9 43 42 AM
Screenshot 2025-07-12 at 9 43 58 AM

---

This one really stretches the definition of a "route", but the shield signs are real. The OSM relation could probably be adjusted to match the actual signage instead of the designated legal definition.

https://www.openstreetmap.org/relation/19246440

Screenshot 2025-07-17 at 7 57 22 PM
Screenshot 2025-07-17 at 6 25 23 PM


1ec5 (approved):
I was thinking for a moment that we could knock out the brown sign back and just use a thick brown outline for simplicity. But if picture is going to be monochromatic, then that doesn’t look as distinct, so the sign back is going to be necessary.

quincylvania:
@1ec5 Yep, I think when it's just an icon on a field then it's good to include the field. Some other scenic routes have more of a frame within a frame, and for those I'm planning to leave out the brown.

Screenshot 2025-07-18 at 9 41 45 AM
Screenshot 2025-07-18 at 9 42 00 AM
Screenshot 2025-07-18 at 9 42 08 AM

Issue #1237, opened 2025-07-28
NYC tunnel shields


quincylvania:
I recently noticed these signs in NYC. The two tunnels seem to have consistent sign colors. It would be cool to show these as shields. Americana supports a similar system with the two major bridges in Halifax.

<img width="339" height="303" alt="Image" src="https://github.com/user-attachments/assets/63979787-980e-43f4-972d-1a95e43de1b4" />
<img width="623" height="468" alt="Image" src="https://github.com/user-attachments/assets/6ddedf66-720d-4504-bffb-9f01f54a22ca" />
<img width="434" height="400" alt="Image" src="https://github.com/user-attachments/assets/5ad0452c-78fe-4997-9c84-90a504be9369" />

1ec5:
The [Battery Tunnel](https://commons.wikimedia.org/wiki/File:TBTA_BROOKLYN_BATTERY_TUNNEL.svg) and [Midtown Tunnel](https://commons.wikimedia.org/wiki/File:TBTA_QUEENS-MIDTOWN_TUNNEL.svg) also have distinctive shields.

quincylvania:
Those are cool! Though the Battery Tunnel was renamed, so I'm not sure if those signs are still maintained.

quincylvania:
Looks like at least one sign exists with the new name and the old design. This post has some snipes at the implementation, apparently mixing fonts. https://forgotten-ny.com/2019/11/hugh-carey-tunnel-sign-financial-district/

1ec5:
[They try so hard](https://commons.wikimedia.org/wiki/File:Battery_Tunnel_Is_Now_Hugh_L._Carey_Tunnel.jpg) to make it work. 😄

quincylvania:
This was completed by #1239 and #1240.

PR #1239, opened 2025-07-29
Add shields for Holland Tunnel and Lincoln Tunnel

quincylvania:
Re: #1237

I mapped these tunnels as road route relations with `network=US:PANYNJ` since they're both operated by the Port Authority of New York and New Jersey.

The shield I made only 19x19 instead of the usual 20x20 for squares, since the larger size visually dominated the other symbols in such a dense urban area.

Screenshot 2025-07-29 at 11 04 07 AM
Screenshot 2025-07-29 at 11 03 52 AM

### DIFF
--- a/src/layer/transportation_label.js
+++ b/src/layer/transportation_label.js
@@ -42,12 +42,6 @@ export const label = {
       "#333",
     ],
     "text-halo-color": [
-      ...classSelector,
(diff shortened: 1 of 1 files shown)


ZeLonewolf (approved):
LGTM!

PR #1240, opened 2025-07-29
Add shields for Queens–Midtown and Hugh L. Carey Tunnels

quincylvania:
Re: #1237

These were a bit tricky since the design doesn't really work without the words but not all the words can fit on the shield. Initialisms seem to work okay.

I went with "HLC" instead of "BB" to match the current name and signage.

Screenshot 2025-07-29 at 4 50 35 PM
Screenshot 2025-07-29 at 4 50 49 PM


claysmalley:
I think these would be recognizable without text. At the very least, we could get rid of "TUNNEL" and leave the initialisms.

If we do keep the initialisms, let's match the font on other shields. `refsByName` can get us there, and we can reuse one sprite with `colorDarken`.

quincylvania:
> I think these would be recognizable without text. At the very least, we could get rid of "TUNNEL" and leave the initialisms.

I agree they're distinct without text. I guess it's stylistic preference as to how abstracted they should be. Lately I've been leaning toward less abstraction in shields as long as the elements are generally legible (at least at 2x).

> If we do keep the initialisms, let's match the font on other shields. refsByName can get us there, and we can reuse one sprite with colorDarken.

This is cool in theory but the initials are much smaller than the regular shield numbers. I manually drew the letters to align to the pixel grid so they look good at 1x. I doubt the font would get better results. (The "TUNNEL" doesn't really work at 1x but is legible at 2x due to quarter-pixel alignment.)


> What if we abbreviate “TUNNEL” as “TUN”, as seen on some of the guide signs in the area? Then we might be able to shrink the black half slightly to make more room for the lettering in the top half.

I'm open to this for readability on the bottom half but I don't think we need to squeeze in any more elements in the top half.

Comparison:

Screenshot 2025-07-29 at 9 17 31 PM

Sources:

hugh-carey-207950561
2457262033_35dcdd147f_b-779631879

Issue #1244, opened 2025-08-02
CJK glyphs should respect language selection in Chrome


1ec5:
In https://github.com/osm-americana/openstreetmap-americana/issues/613#issuecomment-1378543457, I verified that CJK characters that have regional variants will appear in the right variant depending on the HTTP `Accept-Language` header. However, I only verified this in Firefox. @mapmeld [discovered](https://osmus.slack.com/archives/C01BU04GW7L/p1753484663456549) that it isn’t working in Chrome or Firefox: only the Simplified Chinese variant ever appears, regardless of the browser’s language preference or the language selected inside this application.

1ec5:
Closing in favor of the idea in #1445, which would be based on the actual language of the label being displayed, regardless of how it got there.

PR #1245, opened 2025-08-02
Set page language based on language selection

1ec5:
This ensures the page is explicitly set to the selected language. A full fix for #1244 still requires a change to TinySDF.

/cc @mapmeld


1ec5:
No longer blocked, but we’ll need to upgrade to [node-canvas v3.2.0](https://github.com/Automattic/node-canvas/releases/tag/v3.2.0) and [tiny-sdf v2.0.7](https://github.com/mapbox/tiny-sdf/releases/tag/v2.0.7).

1ec5:
#1288 upgraded the dependencies, but I’m still not seeing any change in Chrome. Maybe we’re setting the page language too late?

PR #1246, opened 2025-08-02
Add 3x sprite sheets

ZeLonewolf:
Add sprite sheets for 3x pixel ratios, which is needed for newer iPhones.

- I believe this is not yet supported by maplibre, so this is really just getting ahead of that presumed eventual support
- Converts the sprite generation script to typescript

### DIFF
--- a/scripts/sprites.ts
+++ b/scripts/sprites.ts
@@ -30,7 +30,10 @@ const sprites: SvgId[] = await Promise.all(
 
 console.log(`Building ${sprites.length} sprites from ${opts.icons}`);
 
-const generated: SpriteSheetResult[] = await Sprites.generate(sprites, [1, 2, 3]);
+const generated: SpriteSheetResult[] = await Sprites.generate(
+  sprites,
+  [1, 2, 3]
+);
 
 for (const result of generated) {
   const scaleText: string =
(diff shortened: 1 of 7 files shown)


scripts/sprites.ts
```diff
@@ -0,0 +1,48 @@
...
+for (const result of generated) {
+  const scaleText: string =
+    result.pixelRatio === 1 ? "" : `@${result.pixelRatio}x`;
+  const outputPng: string = `${opts.output}/sprite${scaleText}.png`;
+  const outputJson: string = `${opts.output}/sprite${scaleText}.json`;
+
+  await fs.writeFile(outputPng, result.buffer);
+  await fs.writeFile(outputJson, JSON.stringify(result.layout, null, 2));
+  const kb: string = (result.buffer.length / 1024).toFixed(1);
+  console.log(`Wrote ${kb}KiB to ${outputPng}`);
```
ZeLonewolf, line 47: The main change is here.  I now pass in an array [1,2,3] and generate sprites in a loop.  Easy extension to 4x if we ever need it.

1ec5:
> I believe this is not yet supported by maplibre, so this is really just getting ahead of that presumed eventual support

Previously tracked in mapbox/mapbox-gl-native#1008. Not sure if this has ever been requested of MapLibre GL JS. Most of the codebase theoretically supports arbitrary pixel densities, but the code that parses URL formats [assumes](https://github.com/maplibre/maplibre-gl-js/blob/cc4dfc25d57d6cec656673dbb3c9d276d5144d26/src/style/load_sprite.ts#L36) all HiDPI sprites have “2x” in the URL. QGIS also [hard-codes](https://github.com/qgis/QGIS/blob/893fd595417af6120c162b41a8a04b6968ddc7d4/src/core/vectortile/qgsvectortileutils.cpp#L368) 2× as the maximum resolution.

PR #1249, opened 2025-08-07
Minnesota shield changes

sschr15:
Adds Business MN equivalents and support for a truck banner, and moves Todd County to the always-use-white-signs section to match the majority of its signs (Inter-County sections are the only part that differ, but that may be a challenging feat to accurately reflect the real world)


sschr15:
[Business sign preview](https://preview.americanamap.org/pr/1249/#map=12.13/46.31623/-94.21684), [truck banner preview](https://preview.americanamap.org/pr/1249/#map=13.13/45.87841/-95.34577), [Todd County preview](https://preview.americanamap.org/pr/1249/#map=14.28/45.82528/-95.10389)

icons/shield_us_mn_business_2.svg
1ec5, line 1: Normally we’d outline the white in a contrasting color to ensure legibility against the land layer. But if we use a green outline, it’ll be slightly redundant with the white outline around most of the shield, so it could probably go either way.

sschr15: I was thinking about that a bit, but with the lack of outline on the regular Minnesota shields and my inability to find a version of the sign on a white background (though I haven't looked all that closely, mostly street view and MN MUTCD), I went the simple route at least initially

1ec5:
> Inter-County sections are the only part that differ, but that may be a challenging feat to accurately reflect the real world
> Technically the county does use the pentagon, but only when paired with an Inter-County sign

#1252

PR #1261, opened 2025-09-03
Attraction Land Uses

GA-Kevin-Codes:
This PR adds Theme Parks (class "theme_park"), Zoos (class "zoo"), and Water Parks (class "water_park") to Americana using the template of parks and color of attraction POIs.

### DIFF
--- a/test/sample_locations.json
+++ b/test/sample_locations.json
@@ -103,5 +103,13 @@
       "width": 400,
       "height": 400
     }
+  },
+  {
+    "location": "14/34.62355/-98.38551",
+    "name": "oklahoma_landuses",
+    "viewport": {
+      "width": 400,
+      "height": 800
+    }
   }
 ]


GA-Kevin-Codes:
Updated to brown to match attraction color
image

GA-Kevin-Codes:
Adding zoo photo:
image

wmisener:
Thanks for putting this together! I think this PR looks quite good, sorry for the delay in reviewing it.

First note (and only action item that's holding up approval): please update the [taginfo template](https://github.com/osm-americana/openstreetmap-americana/blob/main/scripts/taginfo_template.json) to cover these new tags we'll now support.

One bug is that it seems like parks within the attractions do not render correctly: the park outlines show up but not the park fill: https://preview.americanamap.org/pr/1261/#map=16.58/28.476771/-81.468597
Screenshot 2025-09-18 at 7 28 02 PM

Possibly relatedly, parks do not show up on top of piers:
https://preview.americanamap.org/pr/1261/#map=15.04/34.01007/-118.49533
Screenshot 2025-09-18 at 7 37 41 PM

I think these are both related to the similar issue we're seeing in the Cemetery PR: https://github.com/osm-americana/openstreetmap-americana/pull/329#issuecomment-3193667390, since theme parks are also in the `landuse` layer. We decided there to leave it for tail work, so I don't think it needs to be a blocker here either. Just posting these for future reference (I doubt there are any real-world examples of cemeteries within theme parks to test out, but I could be wrong!).

Theme parks do render correctly within parks though:
Screenshot 2025-09-18 at 7 30 37 PM

Another sample, really improves the look of the Orlando resort area!
Screenshot 2025-09-18 at 7 30 15 PM

wmisener:
One comment I'm again just leaving for reference: using a brown fill for attraction areas makes a lot of sense semantically for the style given our color palette and established POI coloring. But I do want to note that it does block using brown fill for native lands, as some map samples do in https://github.com/osm-americana/openstreetmap-americana/issues/105. But it seems like that discussion was perhaps leaning toward a halo treatment anyways, and there remains lots of room in our color space in any case.

1ec5:
In print maps, brown is just about the only color that ever represents tribal government boundaries, whereas tourist attractions can be other colors, such as pink. A faint pinkish fill and bolder pink labels would be consistent with that we did for ski lifts.

wmisener:
Ha, in Kevin's defense, pink was the first color suggested (via an LLM) on Slack: https://osmus.slack.com/archives/C01V02K52UX/p1756927759763679. My (and I think others') impression from the test renderings there was that it was too similar to the transit color, and that brown was already our designated attraction color in the [POI contributing guide](https://github.com/osm-americana/openstreetmap-americana/blob/main/CONTRIBUTING.md#color-scheme-1), currently in use for museums, so it made sense.

I had honestly totally forgotten that ski lifts were a different color than transport features: they're pretty similar.
Screenshot 2025-09-19 at 8 05 12 AM. To me it makes more semantic sense for theme parks and zoos to be the same color as museums than as ski lifts, but it's worth thinking about in terms of the broader color scheme of the style and what other land fills we might be interested in rendering.

In #105 there were definitely a few samples that weren't quite brown. A scattering of purples and blues, and some more yellow colors that I think could still be sufficiently different from this shade of brown.

GA-Kevin-Codes:
Hey @wmisener and @1ec5! Thanks for looking at this, and I am happy to hear this is looking good -- I agree, but I am biased! I will figure out how to update that taginfo template today so no worries there, just the theme park, water park, and zoo so 3 additions. So far as color, when I initially sent GPT to work on this I asked for purple, my favorite color, and it chose the original color. When feedback was given about brown that made sense to me so it was updated. I don't have strong feelings here, just whatever everyone would like. Though ski lifts are a type of transport if you ask me, akin to gondolas (rendered as transit in Americana). Perhaps a more clay color for aboriginal lands when the time comes? Just spitballing. Thanks again!

1ec5:
> Perhaps a more clay color for aboriginal lands when the time comes?

That’s more or less where #1289 ended up after a few iterations. It’s lighter than what @GA-Kevin-Codes came up with for tourist attractions. #1289 claims that it blocks this PR, but to know for sure, we’d need some examples of tourist attractions being labeled and filled in proximity to tribal reservations.

> brown was already our designated attraction color in the [POI contributing guide](https://github.com/osm-americana/openstreetmap-americana/blob/main/CONTRIBUTING.md#color-scheme-1), currently in use for museums

For this color scheme, we used the MUTCD sign category colors as a starting point: brown is the color for “Recreational and Cultural Interest Area” signs. The choice of MUTCD colors was pretty arbitrary, just based on our prior experience making shields. Ideally we’d base the colors on conventions in print maps, to the extent that there is a consensus among publishers. We’ve departed from the MUTCD colors where necessary, most notably on some transportation features. I suspect we could do it again by distinguishing recreational sites from cultural institutions.

Pink could work, if only because I’d expect some aerialways to be inside these areas already. Another option would be the burnt orange (“UTexas Orange”) we’re already using to represent consumerism. Either way, it should be either desaturated or lightened so that it doesn’t bleed the eyes when applied over large areas and overpower smaller features, such as at Disney World.

I don’t think it would be a big problem if cultural institutions continue to use a brown that’s relatively close to the tan we chose for tribal reservations. The tan–brown–burnt orange side of the color wheel can be about culture of all kinds.

ZeLonewolf:
Now that #105 is resolved, it would be good to look at some screen shots that have aboriginal land boundaries, large parks, and theme parks in proximity.

Edit: it would be a good idea to locate some test areas and add them to the test areas JSON file, and then they'll appear in the checks tab.

PR #1262, opened 2025-09-04
Add a progress animation to the search box

jleedev:
Add an animated background to the search input, reminiscent of an "indeterminate" progress bar, whenever we are waiting for the search results.

This can make the state more clear when the results are stale because the query has changed and the new results have yet to arrive. It also makes it somewhat more obvious that a two-character query is not going to do anything.

It's still possible to race the transition by typing three letters and then pressing backspace: It looks like the state has quiesced, but then some results appear. But this PR doesn't introduce any _new_ race conditions. :)

If the results are cached (because you've typed the exact same search string focused in the exact same viewport) then there's a flicker as the response is loaded instantly, so we use animate() to delay the background-image from changing.

### DIFF
--- a/src/search.js
+++ b/src/search.js
@@ -170,7 +170,10 @@ async function doSearch(searchQuery) {
   lastSearchRequest = controller;
 
   try {
-    searchInput.classList.add("pending");
+    if (!searchInput.classList.contains("pending")) {
+      searchInput.classList.add("pending");
+      searchInput.animate({ backgroundImage: ["none", "none"] }, 300);
+    }
     const response = await fetch(searchQuery, { signal: controller.signal });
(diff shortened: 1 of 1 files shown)


ZeLonewolf:
Looks great on mobile!

jleedev:
Fun facts:

- Some combination of relative color, system color, and transition doesn't work in Firefox, so just pick a color instead.
- transition-behavior: allow-discrete didn't make it to Firefox ESR 128
- Transitioning the background-image, a discrete property, causes a fading effect in Chrome for some reason.

ZeLonewolf (approved):
Tested also on Chrome and Firefox on Ubuntu 24.04.3 LTS

PR #1264, opened 2025-09-05
[a11y] Search box screenreader support

jleedev:
Use specific tags and roles: The search input is a [combobox], the results are a list.

When selecting the search results, use the ariaActiveDescendantElement property to cause the screen reader to highlight/announce the item. This is a very new property, but it should gracefully degrade as we are only assigning to it.

image

[combobox]: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/combobox_role


jleedev:
For what it's worth: MediaWiki uses the `aria-active-descendant` content attribute for this, which has broad support and merely requires adding an `id` to each item.

ZeLonewolf:
Great initiative! Can you recommend a method to test this?

jleedev:
Well, ChromeOS has a handy Ctrl+Alt+Z shortcut to make it start yelling everything :)

Chrome and Firefox have an accessibility inspector in the devtools, but it's pretty low level and it's not obvious. They can also find issues, though -- fixed some.

Some issues:
- On MediaWiki, one can swipe left and right to advance through the list items, while on this code it selects pieces of the results. So that could be better.
- Also on MediaWiki, hovering on the results makes them active, which is probably a good idea. Improving the focus state of the dropdown would be another enhancement in its own right.

In fact, adding the correct `listbox` and `option` role to the list and items respectively seems to have fixed that first issue -- we now swipe through the whole item.

Anyway, with ChromeVox (on ChromeOS) and TalkBack (on Android) it seems to work as expected (given my very limited experience with using the screen reader).

For any other platform, I suppose things to check are:
- When using the up and down arrow keys, the highlighted item should also display the focus ring and be spoken.
- Some sort of "next" and "previous" actions should equivalently point at each item in the list in turn. This in fact seems to be pretty hard, I note that MediaWiki doesn't make that work when using the touchscreen on ChromeOS, it just loses focus and the list disappears.

image

jleedev:
In short:

- Tool: Chrome devtools (Lighthouse), Firefox devtools, online validators.
- Read: This article <https://web.dev/articles/how-to-review>, MDN <https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/combobox_role>, the WAI-ARIA spec <https://w3c.github.io/aria/#combobox>. (Oh goody, there's a saga specifically about whether the combobox works <https://github.com/w3c/aria/blob/main/documentation/archive/1.2/Resolving%20ARIA%201.1%20Combobox%20Issues.md>).
- Test: With some screenreader and some browser

Known issues here:
- Pointing the mouse at the list items ought to highlight them as well.
- Use the content attribute with an id instead of the DOM property for broader support.
- Pressing Tab focuses the compass button, which is entirely fine except the popup should then hide itself.

ZeLonewolf (approved):
Works as expected. Tested with Chrome/Ubuntu screen reader (Super+Alt+S mode)

PR #1269, opened 2025-09-17
Add shield for Juan Bautista de Anza National Historic Trail Auto Tour Route

quincylvania:
Re: #1101

I spent some time trying to abstract this but there are too many elements for something to look readable. I ended up just leaving it pretty much as-is in order to match the signage.

Screenshot 2025-09-16 at 8 33 02 PM

IMG_2164-3486334161


quincylvania:
One potential issue here is that the NHT shield is noticeably larger than the California state route shield despite their similar shapes. The big difference is that the NHT shield has a black border around the white border (aligned with the other shields in the `US:NHT` network). Wondering if we need some standardization here, especially if we support dark mode eventually.

1ec5:
I think the black border is important for definition, since the background inside the white border is often a light color, as in this case. If the black border is too prominent, we could change it to a shade of the background, similar to navy I chose to match the sky blue background in https://github.com/osm-americana/openstreetmap-americana/pull/576#issuecomment-1328169412.

quincylvania:
I'm fine with the black border. I guess I'm wondering if every white-bordered shield in Americana should also have a dark outline (though maybe not a full pixel width). I think Apple Maps does this. This would be a larger discussion though.

1ec5:
Indeed, that is what Apple Maps does, but it’s a bit heavy on shields with dark backgrounds, like the Interstate and California state route shields. Others like Bing have previously gone with a background-matching outline on all their shields. Maybe we could explore a more neutral gray. 🤔

1ec5 (approved):

PR #1271, opened 2025-09-23
Make low-zoom highways use OpenMapTiles defaults

ZeLonewolf:
This PR does the following:
- Adds z4 trunk roads in OpenMapTiles to render, with the same style as motorway
- Remove the road layer opacity block, which was not used/necessary
- Simplify the fill color logic for trunk roads
- Remove toll selectors for zooms <9 where toll attributes aren't in the underlying tiles

Before:
image

After:
image


adamfranco:
Not sure if it's appropriate for this issue or a new one, but now that low-zoom trunks in the Trans-Canada highway are being displayed at z4, the brightness of the red for trunk looks more impactful than the motorway color:
Screenshot 2025-09-23 at 2 16 50 PM

at z5 the extra width of motorways makes them closer in terms of visual impact
Screenshot 2025-09-23 at 2 19 46 PM

and by z6 motorways look much more impactful.
Screenshot 2025-09-23 at 2 20 49 PM

The reason I mention this is that maybe the opacity lines that were removed maybe need to stay, but have the opacity gradient changed to not completely eliminate roads.

adamfranco:
This is looking quite good, with one little potential issue.

Right at z4, the trunks are super faint: https://preview.americanamap.org/pr/1271/#map=4/43.59/-92.5\
Screenshot 2025-09-28 at 7 18 16 PM

By z4.5 they are reasonably visible: https://preview.americanamap.org/pr/1271/#map=4.49/45.65/-91.07
Screenshot 2025-09-28 at 7 21 53 PM

Is that the behavior you are intending?

ZeLonewolf:
Ugh, looks like I regressed something

PR #1274, opened 2025-09-26
Add icon for Unitarian Universalist places of worship

### LINKED ISSUE #1273: Unitarian Universalist is not rendered as a place of worship
UU churches are pretty common in New England, and I was surprised to find them missing.    (OSM correctly tags them as a religion, not a Christian denomination.)

It might be nice to have a catchall for place_of_worship not otherwise specified, as a church icon (w/o a cross on top of course).  But so far, adding UU would resolve all problems I have noticed (which is 3 out of 3 UU churches not rendered).

AntiCompositeNumber:
Fixes #1273

image
image


ZeLonewolf:
Thanks for working on this! Ideally, the icon should be aligned to the 1x/2x pixel grid where possible. The base square at least should align to 1x rather than on the half pixel like it is now.  (this graphic is from the checks tab for icon changes). It's possible that the existing icons may be misaligned also -- the grid alignment graphic was a relatively recent addition.

image

quincylvania:
Thanks for working on this! I'm a little worried this looks too similar to the cross at display size? Maybe the chalice part could be cut into the square somehow and the flame could be more prominent. The actual shape of the chalice isn't uniform for UU churches.

AntiCompositeNumber:
> Thanks for working on this! I'm a little worried this looks too similar to the cross at display size? Maybe the chalice part could be cut into the square somehow and the flame could be more prominent. The actual shape of the chalice isn't uniform for UU churches.

Yeah, it's definitely not ideal at this scale. I did try moving the bowl all the way down to the building box, but it looks weird to me.
image
I think it's also less readily identifiable as a flaming chalice.[^1]

I moved the bowl down a pixel and increased the width of the flame some. This version is more distinguishable from the cross, at least.
image
image


[^1]: While UU chalices can take pretty much any shape, the traditional symbol is a bowl with a stem. More modern logos for UU congregations and related organizations often omit the stem (and sometimes even the bowl entirely), but the traditional symbol is still in common use, especially on signs & flags marking congregations. There is, as one would expect, no consistency with how these various symbols are used, and there is no particular symbolic meaning attached to the stem.

quincylvania (approved):
Revised icon looks good to me!

wmisener:
Looks great, thanks for working on this!

Screenshot 2025-09-27 at 5 25 10 PM

PR #1276, opened 2025-09-28
Preview of place classification based on urban areas

1ec5:
This is a simulation of [my proposal for overhauling place classification](https://community.openstreetmap.org/t/toward-a-national-system-for-functionally-classifying-populated-places/113674) in the United States based on urban areas.

OSM Americana currently relies on a Natural Earth–based ranking that [OpenMapTiles provides](https://openmaptiles.org/schema/#place) as a workaround for unreliable place classification in OSM. At lower zoom levels, the result is extremely sparse: just nine cities in the contiguous U.S. at zoom level 4. By contrast, print maps at an equivalent scale tend to have hundreds of labels of various sizes. We can’t easily show more than these nine based on OSM’s current classifications, because there’s been such a tendency to promote small towns and large suburbs to `place=city`.

The style intentionally shows as many labels as will fit at each zoom level, something which not every style would want to do. Of course this style can always filter more aggressively if this proposal ever gains traction, but I wanted to show that the proposed reclassifications would give style designers more confidence to surface more labels than before and perhaps explore different ways to visually distinguish them. The simulation color-codes the place labels: olive means the place still shows up despite getting demoted from `city` to `town` or from `town` to `village`; blue means the place got promoted in the other direction.

Since we haven’t retagged anything in OSM yet, I’ve pulled in a copy of [this massive GeoJSON file](https://github.com/1ec5/openstreetmap-urban-areas/blob/b8e995029e97c50c234b5e980eb8c2c2f07343be/places-ua-ranked-wgs84.geojson) of all the places in the U.S. annotated with enough metadata to determine each place’s relationship to an urban area. (The file is outdated and has some accidental omissions, since fixed in the underlying data.)

Before | After
----|----
[Continuous U.S. showing nine cities.](https://americanamap.org/#map=4/40.5/-94&language=en) | [Continuous U.S. showing scores more cities, plus Marysville, Washington, and Spring Hill, Florida, which had previously been towns. Also showing lots of towns throughout, most of which had gotten demoted to towns.](https://preview.americanamap.org/pr/1276/#map=4/40.5/-94&language=en)
[Northern California](https://americanamap.org/#map=7/38.554/-120.282&language=en) | [Northern California](https://preview.americanamap.org/pr/1276/#map=7/38.554/-120.282&language=en)
[...]

PR #1277, opened 2025-09-28
Fix object_compare handling of added layer groups

ZeLonewolf:
As we discovered in #1276, `object_compare.ts` fails when a new layer group is added to the style. This fixes that script and adds test cases to verify that `object_compare` is working properly.

Note: the `object_compare` script, is used by GitHub actions to produce a report in the checks tab showing stats on layer sizes and what was added/removed from the style.

### DIFF
--- a/scripts/object_compare.ts
+++ b/scripts/object_compare.ts
@@ -52,7 +52,10 @@ export function calculateDifference<T extends Record<string, any>>(
     if (!(key in object1!)) {
       if (typeof object2![key] === "object" && object2![key] !== null) {
         // For nested objects, recursively calculate the difference
-        (difference as any)[key] = calculateDifference(null, object2![key]) as any;
+        (difference as any)[key] = calculateDifference(
+          null,
+          object2![key]
+        ) as any;
       } else {
         (difference as any)[key] = object2![key];
       }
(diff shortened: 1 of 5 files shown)


1ec5 (commented):
I believe #1276 triggered the bug by changing three of the style layers to no longer use any source layer. (GeoJSON sources have no source layers.)

Unfortunately, rebasing #1276 atop this branch [did not](https://github.com/osm-americana/openstreetmap-americana/actions/runs/18068389020/job/51414459635?pr=1276) address the problem. The immediate error is that this function only accounts for `null`, but `val` is `undefined`:

https://github.com/osm-americana/openstreetmap-americana/blob/177e594266ff9f9bb7404467792d4239807dadfc/scripts/object_compare.ts#L136-L138

The `undefined` values are coming from here:

https://github.com/osm-americana/openstreetmap-americana/blob/5ff990c86c09e713aed8407ab2726c0642645ce3/scripts/stats_compare.ts#L102-L103

The defensive coalescing `?` is hiding the fact that there’s nothing in `layerGroup` matching `layer`. `layer` in this case is `urbanareas` – which is not a layer name – because of the following fallback to `source-layer`:

https://github.com/osm-americana/openstreetmap-americana/blob/5ff990c86c09e713aed8407ab2726c0642645ce3/scripts/stats.ts#L170-L171

PR #1284, opened 2025-10-10
Centralize all color constants

### LINKED ISSUE #1282: Refactor all colors into colors.js
In #1281, @wmisener noted:
> (Note: some of these did not make it into [color.js](https://github.com/osm-americana/openstreetmap-americana/blob/main/src/constants/color.js), it might be nice to refactor such that colors are all defined there not ad hoc)

I support such a refactor -- it will make things easier to have all colors defined in one place with sensible variable names.

ZeLonewolf:
Fixes #1282 

Moves hardcoded color values for labels and map features into src/constants/color.ts for better maintainability, and also so we can see all the colors in one place.

Also changed this constant file to typescript.


wmisener (commented):
Looks good, in that I tested and nothing seems to have changed appearance-wise on the map. Thanks for doing this.

 One note is that it looks like this PR leaves the road colors (defined in https://github.com/osm-americana/openstreetmap-americana/blob/main/src/layer/road.js) out, i.e. they're still defined in that file. Was that intentional?

src/layer/construction.js
```diff
@@ -36,9 +38,9 @@ export const road = {
       10,
       [...majorConstruction, "lightcoral", "lightslategray"],
       13,
-      [...majorConstruction, "maroon", "lightslategray"],
+      [...majorConstruction, Color.constructionMajor, "lightslategray"],
       15,
-      [...majorConstruction, "maroon", "slategray"],
+      [...majorConstruction, Color.constructionMajor, "slategray"],
```
1ec5, line 43: Did you intend to leave `lightcoral`, `lightslategray`, and `slategray` intact here?

src/layer/transportation_label.js
```diff
@@ -32,14 +32,14 @@ export const label = {
     "text-color": [
       ...classSelector,
       majorConstruction,
-      "maroon",
+      Color.constructionMajor,
       minorConstruction,
       "slategray",
```
1ec5, line 37: Did you intend to leave `slategray` intact here?

1ec5 (commented):
I could see this refactor getting us closer to proper dark mode support (provided that we even know what we want the style to look like in dark mode). But varying the colors by dark mode could require something more dynamic than global constants, so we might have to refactor in the other direction at some point.

PR #1305, opened 2025-11-24
Accept language subtags without crashing

### LINKED ISSUE #1304: Crash requesting language with extension subtag
The application crashes if I set the `language=` URL parameter to a valid language code that includes an [extension subtag or private use subtag](https://en.wikipedia.org/wiki/IETF_language_tag#Syntax_of_language_tags). For example, if I attempt to view [Buggs Island Lake](https://www.openstreetmap.org/relation/223999) in [Virginian English](https://americanamap.org/#map=11/36.5591/-78.5094&language=en-u-sd-usva), I get the following unrecoverable error:

> Uncaught RangeError: invalid language tag: "en-u"

The error occurs because we’re trying to come up with a list of acceptable locales based on the requested locale, in order from most specific to least specific:

https://github.com/osm-americana/openstreetmap-americana/blob/a4adfdf419bc86fea49d51e3d9fefc35214d1ed0/src/constants/label.js#L22-L29

Along the way, we come up with `en-u`, which is invalid because BCP&nbsp;47 requires a single-letter subtag to be followed by another subtag.

1ec5:
Ignore nonsensical partial locales derived from the user’s requested locale. Recover from errors generating localized names of locales for the language picker control.

https://preview.americanamap.org/pr/1305/#map=11/36.5591/-78.5094&language=en-t-zh,zh-u-nu-hant,en-u-sd-usnc,es-fonipa,fr-x-gallo

Fixes #1304.

### DIFF
--- a/src/constants/label.js
+++ b/src/constants/label.js
@@ -23,9 +23,17 @@ export function getLocales() {
     let components = locale.split("-");
     while (components.length > 0) {
       let parent = components.join("-");
-      if (!localeSet.has(parent)) locales.push(parent);
-      localeSet.add(parent);
+      try {
+        // Preflight the parent locale in case it’s incomplete like `en-x`.
+        new Intl.Locale(parent);
+        if (!localeSet.has(parent)) locales.push(parent);
+        localeSet.add(parent);
+      } catch {}
       components.pop();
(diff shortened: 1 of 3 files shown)


ZeLonewolf (approved):
LGTM

PR #1306, opened 2025-11-24
Switch CI image comparison to ResembleJS

ZeLonewolf:
Currently, the checks tab has a series of images that show things that have changed in a list of test areas.  This is currently done with a simple file comparison.  However, there are cases where two images can be pixel-identical but have slightly different files (for whatever reason).  This does an image-wise comparison with ResembleJS to fix this issue.

Tested by introducing and reverting a small style change to verify that diff images were generated.

### DIFF
--- a/src/constants/color.js
+++ b/src/constants/color.js
@@ -4,7 +4,7 @@ export const backgroundFillTranslucent = `hsla(30, 44%, 96%, 0.8)`;
 export const hillshadeShadow = "hsla(30, 14%, 63%, 1)";
 export const hillshadeHighlight = "hsla(30, 44%, 99%, 1)";
 
-export const waterFill = "hsl(211, 50%, 90%)";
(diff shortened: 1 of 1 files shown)


scripts/folder_diff.ts
```diff
@@ -41,30 +42,60 @@ if (!fs.existsSync(outputFolder)) {
...
+  return new Promise((resolve, reject) => {
+    const image1 = fs.readFileSync(image1Path);
+    const image2 = fs.readFileSync(image2Path);
+
+    resemble(image1)
+      .compareTo(image2)
+      .onComplete((data) => {
+        // Consider images different if there's any mismatch
+        // misMatchPercentage of 0 means images are identical
+        resolve(data.misMatchPercentage > 0);
```
1ec5: Good, we have the ability to increase this threshold in the future in case anything turns out to be slightly machine-dependent or nondeterministic.

PR #1309, opened 2025-11-27
Migrate label localization to Diplomat

### LINKED ISSUE #914: Package label localization as a plugin
The label localization functionality added in #578 #579 #592 #670 #753 etc. would be useful to a variety of maps, not just Americana. We should ship it as a separate plugin so that Mapbox/MapLibre GL JS–powered maps can benefit from the work we’ve done to prove out these features.

## Rationale

Label localization is basic functionality for any map, but GL JS makes this task so complex that most developers don’t bother or do a half-baked job of it. Americana has proven out a robust, reasonably performant implementation that other developers have shown an eagerness to reuse. There are third-party deployments of the Americana style, such as on the OSM Wiki, that lack label localization (“Americana Lite”) because this functionality can’t be used separately from the overall Americana application.

Increased adoption will also lend us credibility when [engaging with the OSM community about tagging improvements](https://community.openstreetmap.org/t/name-lists-in-osm-americana-or-how-i-learned-to-love-semicolons/7686) to facilitate this kind of functionality. As precedent, the German Mapnik transliteration code took on a life of its own, getting adopted by OpenMapTiles and influencing many project-wide decisions about translation and transliteration, because it relied on ICU functionality that was easily ported to different environments. (This influence came despite the approach’s serious flaws and cultural faux pas.)

Americana would complement the existing options for label localization. [Mapbox GL Language](https://github.com/mapbox/mapbox-gl-language/) gets the job done, but it doesn’t support dual language labeling or semicolon parsing. On the other hand, it’s written more generically, whereas Americana’s implementation assumes a typical map labeling context. [MapTiler’s fork of MapLibre GL JS](https://docs.maptiler.com/sdk-js/api/languages/) has a built-in language switcher, with a focus on location-based language switching (huh?), but its dual-language presentation is not as elegant as Americana’s.

## API design

The plugin should give styles flexibility over the font details. Similar to the [shield generator library](https://github.com/ZeLonewolf/openstreetmap-americana/tree/cf70fb81f147b7a9aa33f2217777a21378b2972d/shieldlib), it can make assumptions about the style’s preexisting `text-field` expressions, such as the presence of a `let` expression with a [`localizedName`](https://github.com/ZeLonewolf/openstreetmap-americana/blob/cf70fb81f147b7a9aa33f2217777a21378b2972d/src/constants/label.js#L90) variable. However, it should also have customization hooks for compatibility with tilesets that aren’t derived from OpenMapTiles.

## Implementation notes

The label localization code is already centralized in [label.js](https://github.com/ZeLonewolf/openstreetmap-americana/blob/cf70fb81f147b7a9aa33f2217777a21378b2972d/src/constants/label.js). There should be few dependencies on other parts of the codebase. [The unit tests](https://github.com/ZeLonewolf/openstreetmap-americana/blob/cf70fb81f147b7a9aa33f2217777a21378b2972d/test/spec/label.js) depend on the maplibre-gl-style-spec package.

If Americana later migrates to a server-side implementation of language resolution, the client-side localization code would be streamlined a bit. However, I suspect most of the functionality in this plugin would still be relevant, because it’s doing more than basic language switching.

/ref #146

1ec5:
Migrated label localization functionality to the new [Diplomat](https://github.com/osm-americana/diplomat/) plugin, which is based on the current implementation in this repository.

Fixes #914.

### DIFF
--- a/src/americana.js
+++ b/src/americana.js
@@ -2,7 +2,7 @@
 
 import config from "./config.js";
 
-import * as Label from "@americana/diplomat";
+import { getLanguageFromURL, getLocales } from "@americana/diplomat";
 
 import * as languageLabel from "./js/language_label.js";
 
@@ -79,8 +79,8 @@ function shieldDefLoad(shields) {
 
   window.addEventListener("hashchange", (event) => {
     upgradeLegacyHash();
-    let oldLanguage = Label.getLanguageFromURL(new URL(event.oldURL));
-    let newLanguage = Label.getLanguageFromURL(new URL(event.newURL));
+    let oldLanguage = getLanguageFromURL(new URL(event.oldURL));
+    let newLanguage = getLanguageFromURL(new URL(event.newURL));
     if (oldLanguage !== newLanguage) {
       console.log(`Changed to ${newLanguage}`);
       hotReloadMap();
@@ -102,7 +102,7 @@ function hotReloadMap() {
 }
 
 export function updateLanguageLabel() {
-  languageLabel.displayLocales(Label.getLocales());
+  languageLabel.displayLocales(getLocales());
   legendControl.onLanguageChange();
 }
 
--- a/src/js/language_label.js
+++ b/src/js/language_label.js
@@ -2,7 +2,7 @@
 
 import Tokenfield from "tokenfield";
 import { updateLanguageLabel } from "../americana";
-import * as Label from "@americana/diplomat";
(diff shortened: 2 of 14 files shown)


ZeLonewolf:
Tested on Ubuntu/Chrome, seems to work great!  Thanks for taking on this initiative.

ZeLonewolf (approved):

PR #1298, opened 2025-11-16
MapLibre GL JS v5.13.0

1ec5:
Upgraded to [MapLibre GL&nbsp;JS v5.13.0](https://github.com/maplibre/maplibre-gl-js/releases/tag/v5.13.0), which brings in support for the full Unicode character repertoire (maplibre/maplibre-gl-js#6640). Removed a workaround from #592 now that bidirectional isolating characters are supported.

Before | After
:--:|:--:
[那市镇](https://americanamap.org/#map=12/22.35126/105.38323&language=zh)<br>[那市镇](https://www.openstreetmap.org/node/369498042) | [ảnh](https://preview.ourmap.us/pr/1298/#map=12/22.35126/105.38323&language=zh)<br>[那𧯄市镇](https://www.openstreetmap.org/node/369498042)
[嘎味魠魚焿、十分ˋ意風味食堂](https://americanamap.org/#map=18/24.840474/121.001514/-60)<br>[嘎味魠魚焿](https://www.openstreetmap.org/node/10687926045)<br>[十分ˋ意風味食堂](https://www.openstreetmap.org/node/10687926014) | [嘎味𩵚魠魚焿、十分ㄍㄚˋ意風味食堂](https://preview.ourmap.us/pr/1298/#map=18/24.840474/121.001514/-60)<br>[嘎味𩵚魠魚焿](https://www.openstreetmap.org/node/10687926045)<br>[十分ㄍㄚˋ意風味食堂](https://www.openstreetmap.org/node/10687926014)
[Pawhuska](https://americanamap.org/#map=12/36.66284/-96.3395)<br>[Pawhuska](https://www.openstreetmap.org/node/151597427) | [Pawhuska 𐓄𐓘𐓢𐓶𐓮𐓤𐓘](https://preview.ourmap.us/pr/1298/#map=12/36.66284/-96.3395)<br>[Pawhuska 𐓄𐓘𐓢𐓶𐓮𐓤𐓘](https://www.openstreetmap.org/node/151597427)
[...]

PR #1312, opened 2025-11-29
Diplomat v0.3

1ec5:
Upgraded to Diplomat v0.3.0, which incorporates osm-americana/diplomat#9.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
         "shieldlib"
       ],
       "dependencies": {
-        "@americana/diplomat": "^0.2.0",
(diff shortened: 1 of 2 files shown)


1ec5:
This introduces an implicit dependency on MapLibre GL&nbsp;JS v5.7.0 and above due to osm-americana/diplomat#9 and maplibre/maplibre-gl-js#6209. In theory, it shouldn’t be a problem, as #1298 already introduced a hard dependency on v5.13.0 and above. But apparently some applications have been using the generated JSON stylesheet with older versions of MapLibre GL&nbsp;JS, so now they’re broken.

PR #1314, opened 2025-12-10
Upgrade macOS 13 runner to macOS 15

1ec5:
Builds have begun [failing](https://github.com/osm-americana/openstreetmap-americana/actions/runs/20110735560?pr=1312) because the macOS 13 runner has been retired: actions/runner-images#13046. This upgrades the runner to macOS 15.


ZeLonewolf (commented):
Thanks!

ZeLonewolf (approved):

PR #1316, opened 2025-12-19
Add shield for US:CO:Scenic routes

### LINKED ISSUE #1315: Colorado Scenic Byway shields
There are 26 [Scenic and Historic Byways in Colorado](https://www.codot.gov/travel/colorado-byways). I recently added all the ones that weren't in OSM and standardized them to `network=US:CO:Scenic` ([query](https://overpass-turbo.eu/s/2hse)). Adding the shield art to Americana will be very easy compared to other scenic networks since they all seem to use the same sign, showing a Rocky Mountain Columbine on a light blue field.

https://commons.wikimedia.org/wiki/File:Colorado_Scenic_Byway.png

Image

https://commons.wikimedia.org/wiki/File:Guanella_Pass_Scenic_Byway,_Colorado_(44441564484).jpg

Image

quincylvania:
Closes #1315.

Screenshot 2025-12-19 at 1 33 18 PM

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -1135,7 +1135,7 @@ export function loadShields() {
   };
   shields["US:CO:Scenic"] = {
     spriteBlank: "shield_us_co_scenic",
-    notext: true
+    notext: true,
   };
 
   // Connecticut


1ec5 (approved):
Good example showing how the outline keeps the shield from bleeding into the river.

quincylvania:
> Good example showing how the outline keeps the shield from bleeding into the river.

There's still a contrast issue with the cream background color. This has come up before, but it'd be worth considering adding some sort of uniform, thin grey outline around light shield borders.

1ec5:
Personally I’m not as concerned about light halos colliding with the background. After all, labels intentionally have cream-colored halos to mimic a knockout effect, so it feels like the same thing to me.

PR #1324, opened 2026-01-22
Add globe projection option

### LINKED ISSUE #1195: Add globe view option
The map is currently displayed in Web Mercator projection. This projection makes Alaska too big, among many other distortions. Due to this projection, social studies instructors in Nebraska are [prohibited by law](https://www.education.ne.gov/socialstudies/mercator-law-guidance/) from using OSM Americana in their classrooms, except as a demonstration of how maps can be misleading.

MapLibre supports using Vertical Perspective projection instead of Web Mercator. This is usually called “globe view” because it simulates the effect of looking straight on at a physical globe. There are still use cases for showing the whole world at once, so we still need to provide Web Mercator as an option or even leave it as the default. There should be a toggle button as well as a URL query parameter, since otherwise someone might share a URL to a view that covers a different viewport than expected. (The zoom level changes somewhat counterintuitively when panning in globe view.)

For prior art, OpenHistoricalMap/openhistoricalmap-embed#18 implements globe view in a Web application somewhat similar to Americana.

1ec5:
Added a button to toggle between Web Mercator and Vertical Perspective (“globe”) projections. Added a `#projection=mercator`/`globe` hash parameter to persist either projection.

Zoom level 2

Zoom level 3

Zoom level 4

Zoom level 5

Zoom level 6

Zoom level 7

Zoom level 8

Zoom level 9

Zoom level 10

Fixes #1195.

### DIFF
--- a/src/js/language_control.js
+++ b/src/js/language_control.js
@@ -183,6 +183,7 @@ export class LanguageControl {
     this._container.appendChild(langPicker);
     this._container.appendChild(langField);
     this._container.appendChild(langChanger);
+    this._map.once("load", (event) => this.displayLocales());
     this._map.on("americana.languagechange", (event) => this.displayLocales());
     return this._container;
   }


jleedev:
Would be nice to also add a state for the terrain toggle, although the globe is important because it materially changes what's in view.

1ec5:
I was going to continue refactoring americana.js to make `MapView` responsible for managing the hash too, but I think that’s enough churn for now.

@ZeLonewolf, can you review the language control part of this PR? Thanks!

ZeLonewolf:
Looks great! Tested and no issues with language control. Code and doc changes look fine to me.

Issue #614, opened 2022-12-10
Show tile timestamp


1ec5:
The vector tiles aren’t being generated on a short, regular interval yet, so the user doesn’t have a good sense for how old the visible data is. We should display a timestamp of when the tiles were last generated, or perhaps the timestamps of the OSM planet file and Wikidata dump that were used to generate the tiles.

Ideally, the TileJSON file v3.json would have a `metadata` field with this timestamp in it, and then we could grab it out of the vector source. However, GL JS doesn’t preserve TileJSON metadata: mapbox/mapbox-gl-js#5471. As a workaround, the attribution string could be modified to include the timestamp, which GL JS would show automatically. Alternatively, the process for generating the tiles could spit out a separate JSON file alongside v3.json that this project could fetch manually.

1ec5:
> Ideally, the TileJSON file v3.json would have a metadata field with this timestamp in it, and then we could grab it out of the vector source. 

The TileJSON now contains timestamp metadata as of onthegomap/planetiler#428. However, GL JS is still discarding this data.

PR #1327, opened 2026-01-29
Update TileJSON URL and docs links to point to tiles.openstreetmap.us

jake-low:
I noticed that the default config file for Americana points to a tile.ourmap.us TileJSON which in turn points to tiles hosted in an AWS S3 bucket and served by Cloudfront + Lambda. This tile endpoint still works, but I think it would be a good idea to switch to the new tiles.openstreetmap.us endpoint. This is the documented way to use tiles from the OSM US Tileservice and will be durable across any future changes in how the tiles are served (e.g. switching cloud providers). The new URLs also have monitoring set up so we'll know right away if they break.

I also renamed the config file to be more accurate (the new tile endpoint is served by Cloudflare, but this is an implementation detail that may change in the future), and updated links to the tile service docs and usage policy in the README.

The TileJSON returned by the new URL is _not_ identical to the one being served by tile.ourmap.us (the tiles.openstreetmap.us TileJSON is [generated using pmtiles-js](https://github.com/osmus/tileservice/blob/main/cloudflare/src/index.ts#L182-L208)), but it should be mostly equivalent.

New: `https://tiles.openstreetmap.us/vector/openmaptiles.json`
```
{
  "attribution": "<a href=\"https://www.openmaptiles.org/\" target=\"_blank\">&copy; OpenMapTiles</a> <a href=\"https:/
  /www.openstreetmap.org/copyright\" target=\"_blank\">&copy; OpenStreetMap contributors</a>",
  "bounds": [
    -180,
    -85.0511287,
    180,
    85.0511287
  ],
  "center": [
    0,
    0,
    0
  ],
  "description": "A tileset showcasing all layers in OpenMapTiles. https://openmaptiles.org",
  "maxzoom": 14,
  "minzoom": 0,
  "name": "OpenMapTiles",
  "scheme": "xyz",
  "tilejson": "3.0.0",
  "tiles": [
    "https://tiles.openstreetmap.us/vector/openmaptiles/{z}/{x}/{y}.mvt"
  ],
  "vector_layers": [...],
  "version": "3.15.0"
}
```

Old: `https://tile.ourmap.us/data/v3.json`
```
{
  "attribution": "<a href=\"https://www.openmaptiles.org/\" target=\"_blank\">&copy; OpenMapTiles</a> <a href=\"https:/
  /www.openstreetmap.org/copyright\" target=\"_blank\">&copy; OpenStreetMap contributors</a>",
  "basename": "planet.pmtiles",
  "bounds": [
    -180,
    -85.05113,
    180,
    85.05113
  ],
  "center": [
    0,
    0,
    0
  ],
  "description": "A tileset showcasing all layers in OpenMapTiles. https://openmaptiles.org",
  "format": "pbf",
  "id": "planet",
  "maxzoom": 14,
  "minzoom": 0,
  "name": "OpenMapTiles",
  "planetiler:buildtime": "2023-04-20T12:14:19.816Z",
  "planetiler:githash": "fba875f9680c58c3eb6943b2fabce4c4ce3e5bb5",
  "planetiler:osm:osmosisreplicationseq": "92967",
  "planetiler:osm:osmosisreplicationtime": "2023-08-05T12:00:00Z",
  "planetiler:osm:osmosisreplicationurl": "https://planet.osm.org/replication/hour/",
  "planetiler:version": "0.6-SNAPSHOT",
  "tilejson": "2.0.0",
  "tiles": [
    "https://dwuxtsziek7cf.cloudfront.net/planet/{z}/{x}/{y}.mvt"
  ],
  "type": "baselayer",
  "vector_layers": [...],
  "version": "3.14.0"
}
```

Noteworthy differences:
- Tiles URL (the main motivation for the PR)
- TileJSON version number (v2.0 -> v3.0)
- Added explicit `scheme: "xyz"`
- Removed planetiler generated fields (serving this info would actually be a great idea, but in this case the tile.ourmap.us fields were not up to date with the actual tile data)
- Updated OpenMapTiles version number (3.14 -> 3.15) matching what's actually in the tiles
- Removed `format: "pbf"`; I believe TileJSON does not actually specify a format field or other way to indicate whether the data is raster or vector (an unfortunate omission IMO)
- Removed `type: "baselayer"`

Also note that both TileJSONs provide `vector_layers`, I just omitted them for brevity.

I tested this by running `npm run start` and making sure the map loaded. I also grepped the repo for references to `config.aws.js` and made sure to rename them all to match the new filename.

For the sake of clarity: this PR is offered as a (hopefully helpful) bit of housekeeping, but there's no top-down directive from OpenStreetMap US to switch endpoints immediately. If you have concerns or want to do additional testing before merging this, that's fine, and let me know if you spot anything I've done wrong. 🙂

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@ node_modules
 sprites
 build
 .DS_Store
-config.js
 .parcel-cache
 dist
 download


jake-low:
I think I fixed the CI failures that were actual issues (of which there were several 🙃).

The remaining one looks like it's because `src/configs/config.osmus.js` doesn't exist on the main branch yet.

jake-low:
Yes, it should definitely be possible to include the timestamps in the tileservice's TileJSON endpoint. I just opened an issue to track that: https://github.com/osmus/tileservice/issues/148

1ec5:
> If so, we can track that as another blocker.

To be clear, the lack of a timestamp blocks #614, not this PR.

PR #1331, opened 2026-01-30
Enable Dependabot

1ec5:
Added a configuration file for Dependabot.


jleedev:
Add a cooldown? https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-

(Wait until somebody else is holding the supply chain.)

PR #1345, opened 2026-02-11
add signed shields in KR

aaei924:
Added county road shields in Korea. Because they have no legally designated road shields, we can only add shields which are found at road signs.

- KR:41:Gapyeong
IMG_9430

- KR:42:Hongcheon
image

- KR:45:Wanju
image

- KR:45:Gochang
IMG_9499

- KR:48:Namhae
IMG_9472

- KR:27:Gunwi
output_775079004

- KR:46:Sinan
IMG_9507

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -3727,7 +3727,7 @@ export function loadShields() {
     Color.shields.yellow,
     Color.shields.white,
     Color.shields.blue,
-    35,
+    30,
     0
   );
 


1ec5:
The changes so far seem fine to me, but the PR is still marked as a draft. Do you want to get these merged now, so that users can benefit from them right away, and then take care of the rest in a follow-up PR?

aaei924:
Okay. I marked this PR as draft because I was still looking for more shields to add. Now it seems that I found out enough, so no problem to proceed.

Issue #889, opened 2023-06-29
Move shield library into new repository


claysmalley:
The shield generator has taken on a life of its own. Now that it's a separate codepath and published as an NPM package, it should be given its own repository. This will make it simpler to track issues and reuse the code in other projects.

- [ ] Determine more compact format for `shields.json`
- [ ] Check in `shields.json`, remove `shield_defs.js`
- [ ] Transfer code and history to [new repository](https://github.com/ZeLonewolf/maplibre-gl-shields)
- [ ] Reconfigure Github, NPM
- [ ] Remove `shieldlib/`

ZeLonewolf:
Thanks for writing this up. I support this initiative.  I think we should also extend the work in the pending #881 to also report on the size of the `shields.json` file so we have a metric for compactness.  After all, the download time of that file is part of the overall map load time.

adamfranco:
Another benefit of this move is easier filtering/monitoring of issues -- shields significantly dominate the issue queue and not all interested in Americana are as engaged on the topic. 😉

ZeLonewolf:
To be fair, most of the shield traffic is on the design side which would still be in this repo.

quincylvania:
We have a couple potential use cases for the shield library over at OpenTrailMap. Of course we want to generate trail route shields. But another idea is to generate icons for campsites that incorporate the `ref` value… the current rendering is kind of messy. Let me know if there is any way I can help with the code separation. 

<img width="843" alt="Screenshot 2024-12-30 at 2 13 59 PM" src="https://github.com/user-attachments/assets/4cf0b788-e2b9-48d3-ada8-9acf33581380" />

ZeLonewolf:
Realistically I don't think there's really anything preventing us from moving forward with the separation.  There are some refactoring niceties describe in here and linked tickets but realistically all of this could get moved to a new repo in the org.

However, for the specific use case cited, I think you can just use the shield generator library that we [publish to npm](https://www.npmjs.com/package/@americana/maplibre-shield-generator) immediately.

quincylvania:
Oh I see, this ticket is for the Americana shield library itself. Yep, I was looking for the generator 🤦‍♂️. Thanks!

1ec5:
This monorepo is confusing Dependabot. It’s applying an application versioning strategy to the shield library, whereas the shield library needs a library versioning strategy: #1338 #1333.

1ec5:
[...]

PR #1349, opened 2026-02-19
Make Dependabot widen dependencies for shieldlib

1ec5:
Tell Dependabot to use a different dependency versioning strategy for the shield library that’s appropriate for libraries.

This is a stopgap until we spin out the shield library as its own repository: #889. It should unblock #1333 and #1338, which incorrectly narrowed the requirements for the shield library, as well as #1334 #1335 #1336 #1339 #1341, which fail to build because of a mismatch between the two package.json files.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,19 @@ updates:
   - package-ecosystem: "npm"
     directories:
       - "/"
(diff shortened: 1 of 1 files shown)


Pengor (approved):
Nice, thank you!

PR #1370, opened 2026-02-28
Redesign language control

### LINKED ISSUE #623: Improve placement of language switcher
#622 Added support for a language switcher.  This is in the lower-right, and awkwardly-sized. however, this should probably be better placed in a button, similar to the OpenMapSamples control

1ec5:
Rewrote most of the language control to encapsulate DOM manipulation within the `LanguageControl` class. Moved the token field to a modal dialog. Added help text. Moved styling from script to HTML document.

Added a Multilingual option and a Reset button to return to the browser default.

Change Language: By default, each feature on this map is labeled in your preferred language along with the local language. You can override this setting by entering one or more languages by IETF language tag or language name, in order of priority.

ip

Fixes #623.


claysmalley (approved):
Looks great! Thanks for working on this.

PR #1371, opened 2026-03-01
Suggest all known language names

1ec5:
Fixed an issue where the language picker’s dropdown menu only listed very common languages, even after #1368. The fix is to filter languages by whether the browser has a name for it in the current language, not whether the browser has a name for every language in that language.

Old English, English, American English, English (IPA Phonetics), Middle English, Jamaican Creole English, Nheengatu

### DIFF
--- a/src/js/language_control.js
+++ b/src/js/language_control.js
@@ -131,6 +131,8 @@ function getLanguageNamesByCode() {
   _languageNamesByCode = Object.entries(languageNamesByCode).map(
     ([id, name]) => ({ id, name })
   );
+  const collator = new Intl.Collator(initialLocales);
+  _languageNamesByCode.sort((a, b) => collator.compare(a.name, b.name));
   return _languageNamesByCode;
 }
 


1ec5:
Also sorted the suggestions alphabetically. Note that Chrome has support for far fewer languages and dialects than Firefox or Safari. [Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=2002423#c1) and WebKit both rely on Unicode’s CLDR project, which includes many but not nearly all the language names [in English](https://github.com/unicode-org/cldr/blob/413361b34c41aa0f081f19a9a408ea0381e92db6/common/main/en.xml#L24-L1766), while [Chromium](https://source.chromium.org/chromium/_/chromium/chromium/deps/icu/+/bbccc2f6efc1b825de5f2c903c48be685cd0cf22:filters/common.json) is based on a subset of an eight-year-old version of Unicode’s ICU project.

Firefox | Chrome | Safari
----|----|----
American English, English, English (IPA Phonetics), Jamaican Creole English, Middle English, Nheengatu, Old English | American English, English, English (FONIPA) | American English, English, English (IPA Phonetics), Jamaican Creole English, Middle English, Nheengatu, Old English

PR #1383, opened 2026-03-19
Add 'US:US:Express' shield definition

IloveRumania:
Just like Interstates, some U.S. Highways have a "local/express lane" system. I've added the appropriate banner for them.


1ec5 (approved):
Thanks!

PR #1386, opened 2026-04-10
Shields of Morocco

### LINKED ISSUE #572: National routes of Morocco
Morocco signposts its national routes with red rectangles. National routes are tagged in OSM as [`network=MA:RN`](https://taginfo.openstreetmap.org/tags/network=MA%3ARN). `ref`s begin with `RN`. Signs on the ground are inconsistent. Some contain the prefix “RN” or “N”, but most seem to omit any prefix. Ideally the `ref`s would be unprefixed, but it doesn’t seem unreasonable to include it.

[El-Argoub](https://www.flickr.com/photos/jbdodane/8589293393)

[Tan Tan](https://www.flickr.com/photos/jbdodane/8534736145)

[Tetouan](https://winairtravel.me/2017/03/14/morocco-the-blue-city-of-chaouen/)

[Ouezzane](https://www.reflectionsenroute.com/morocco-vacation-road-trip-morocco-itinerary/)

1ec5:
Added shield definitions for Morocco’s auto-, national, regional, and provincial routes. Omitted forest roads for now until we know how they’ll be tagged.

[Blue, red, and yellow, and white shields around Errahma and Oulad Azzouz.](https://preview.americanamap.org/pr/1386/#map=12/33.50778/-7.71523&language=fr)

Fixes #572.

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -3456,6 +3456,28 @@ export function loadShields() {
     Color.shields.white
   );
 
+  // Morocco
+  shields["MA:A"] = roundedRectShield(
+    Color.shields.blue,
+    Color.shields.white,
+    Color.shields.white
+  );
+  shields["MA:RN"] = roundedRectShield(
+    Color.shields.red,
+    Color.shields.black,
+    Color.shields.white
+  );
+  shields["MA:RR"] = roundedRectShield(
+    Color.shields.yellow,
+    Color.shields.black,
+    Color.shields.black
+  );
+  shields["MA:RP"] = roundedRectShield(
+    Color.shields.white,
+    Color.shields.black,
+    Color.shields.black
+  );
+
   // Niger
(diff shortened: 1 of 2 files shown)


src/js/shield_defs.js
```diff
@@ -3455,6 +3455,28 @@ export function loadShields() {
     Color.shields.red,
     Color.shields.white
   );
+  
+  // Morocco
+  shields["MA:A"] = roundedRectShield(
+    Color.shields.blue,
+    Color.shields.white,
```
1ec5, line 3462: I was unsure whether to make the border black or white. I went with white because it’s easier to discern, but we can switch it to black if that’s more accurate.

/cc @ilias52730

ilias52730: This is beautiful! Please keep it exactly as it is, white borders are used for MA:A. Absolute perfection.

ZeLonewolf (approved):

PR #1395, opened 2026-05-05
Bonneville County ID Shields

### LINKED ISSUE #1394: Bonneville County ID Shields
Digging around Idaho Falls and noticed there are a couple pentagoned routes for Bonneville County. Network is `US:ID:Bonneville`. These pentagons are slightly different than the standard ones, there is no yellow and the numbers are white.

the-spui-ninja:
Fixes https://github.com/osm-americana/openstreetmap-americana/issues/1394

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -252,7 +252,7 @@ export function loadShields() {
         spriteBlank: "shield_ca_ns_s_mdb",
       },
       "Bras d'Or Lakes Scenic Drive": {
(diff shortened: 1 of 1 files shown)


1ec5 (approved):

1ec5 (commented):
Oh, looks like you need to run `npm run code_format` so that the linter will pass.

the-spui-ninja:
That should have fixed it.

PR #1397, opened 2026-05-11
Additional Shields: addition of SPUR Banner for Idaho and normal county pentagons for Clark County ID

the-spui-ninja:
As part of my mapping in the area noticed a couple more networks that are missing. I did the blue-and-yellow pentagons similar to other states so more counties can be added if it comes up in the future.

### DIFF
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -1298,16 +1298,11 @@ export function loadShields() {
       "US:ID:Spur": ["SPUR"],
     },
   };
-  [
-    "Clark",
-  ].forEach(
-    (county) =>
-      (shields[`US:ID:${county}`] = pentagonUpShield(
-        3,
-        15,
-        Color.shields.blue,
-        Color.shields.yellow
-      ))
+  shields["US:ID:Clark"] = pentagonUpShield(
+    3,
+    15,
+    Color.shields.blue,
+    Color.shields.yellow
   );
   shields["US:ID:Bonneville"] = pentagonUpShield(
     3,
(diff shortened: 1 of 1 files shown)


src/js/shield_defs.js
```diff
@@ -1295,8 +1295,20 @@ export function loadShields() {
...
+    "Clark",
+  ].forEach(
+    (county) =>
+      (shields[`US:ID:${county}`] = pentagonUpShield(
+        3,
+        15,
+        Color.shields.blue,
+        Color.shields.yellow
+      ))
+  );
```
ZeLonewolf: Do we expect that other counties will get added to this list?  If it's just a single county, no need to have the for-each block.

the-spui-ninja: I wasn't sure, I can change it back to the single-shield format for now.

Two weeks ago I didn't think any Idaho counties had pentagons so thought I would at least make it easier in the future if more showed up

the-spui-ninja: changed

src/js/shield_defs.js
```diff
@@ -1295,8 +1295,15 @@ export function loadShields() {
     },
     bannerMap: {
       "US:ID:Business": ["BUS"],
+      "US:ID:Spur": ["SPUR"],
     },
   };
+  shields["US:ID:Clark"] = pentagonUpShield(
```
claysmalley: If this is just a single definition (for now), let's sort Clark after Bonneville.

the-spui-ninja: updated

claysmalley (approved):
Thank you!

Issue #1266, opened 2025-09-16
Oregon Scenic Byway shields


quincylvania:
A couple of these were already in OSM so I added the rest and standardized on a consistent `network=US:OR:Scenic`. The shields all seem to have [this pentagon banner design](https://commons.wikimedia.org/wiki/File:Rogue-Umpqua_Scenic_Byway_-_Rogue-Umpqua_Scenic_Byway_Road_Sign_-_NARA_-_7721384.jpg) with a unique graphic for each route. 

[Query all routes in this network](https://overpass-turbo.eu/s/2bMq)

| Done | Byway | Relation | PR |
|---|---|---|---|
| | Blue Mountain | https://www.openstreetmap.org/relation/19609813 | |
| | Cascade Lakes | https://www.openstreetmap.org/relation/15427611 | |
| | Elkhorn | https://www.openstreetmap.org/relation/19609814 | |
| | Hells Canyon | https://www.openstreetmap.org/relation/15427730 | |
| | High Desert Discovery | https://www.openstreetmap.org/relation/19607354 | |
| | Historic Columbia River Highway | https://www.openstreetmap.org/relation/19609957 | |
| | Journey Through Time | https://www.openstreetmap.org/relation/19609816 | |
| | Marys Peak to Pacific | https://www.openstreetmap.org/relation/19609815 | |
| | McKenzie Pass-Santiam Pass | https://www.openstreetmap.org/relation/15427867 | |
| | McKenzie River | https://www.openstreetmap.org/relation/19601562 | |
| | Mount Hood | https://www.openstreetmap.org/relation/15443909 | |
| | Oregon Outback | https://www.openstreetmap.org/relation/15444006 | |
| | Over The Rivers And Through The Woods | https://www.openstreetmap.org/relation/19609594 | |
| | Pacific Coast | https://www.openstreetmap.org/relation/19609972 | |
| | Rogue-Umpqua | https://www.openstreetmap.org/relation/15444134 | |
| | Trees to Sea | https://www.openstreetmap.org/relation/19599634 | |
| | Umpqua River | https://www.openstreetmap.org/relation/19609535 | |
| | Volcanic Legacy | https://www.openstreetmap.org/relation/19609531 | |
| | West Cascades | https://www.openstreetmap.org/relation/19609595 | |

quincylvania:
[...]

PR #1409, opened 2026-05-22
Add shields for Oregon Scenic Byways

quincylvania:
Re: #1266

Screenshot 2026-05-21 at 9 53 47 PM

This PR is a bit different than usual since I'm trying out raster images for the graphic components. I'm seeking feedback on this approach. Most of the graphics are highly complex and wouldn't be feasible to vectorize manually, and they all end up getting rendered to sprites anyway. 

The resolution for the bitmaps is 56x40. This is rendered down to 16x10 at 1x and 32x20 at 2x, but I left some wiggle room to support future resolutions up to 4x.

I'm not 100% sure about copyright but I would think these would fall under fair use.


1ec5:
> I'm not 100% sure about copyright but I would think these would fall under fair use.

The fair use doctrine presupposes infringement. This would be a departure from our claim of public domain status. I’d rather rely on a rationale that the designs we’re deriving are so simple that they aren’t copyrightable.

> Most of the graphics are highly complex and wouldn't be feasible to vectorize manually, and they all end up getting rendered to sprites anyway.

At this size, most of the graphics turn to mush without more contrast. It’s even worse on a 1× display. My gut instinct would be to posterize the graphics to reduce the number of colors. At that point, we’re probably back to creating new vector drawings. I don’t think it would be so bad if the shields were twice the size, but we don’t have that much room on the map for such large shields.

wmisener:
Yeah I unfortunately think these are quite illegibile as they are now, especially compared to our other shields. Just at the test I kept squinting and trying to zoom to see what was going on but it never got better haha.
Screenshot 2026-05-22 at 8 33 44 AM

PR #1412, opened 2026-05-28
fix raw language code unable to be entered

k-yle:
regarding this popup:

Image

typing or pasting a raw code and then pressing _"Done"_ doesn't work. you have to press <kbd>Enter ⏎</kbd> first, then click _"Done"_. 

This is a bit unintuitive, and made me think that raw values weren't supported. 

I've tested that validation still works: typing `123` won't be accepted, but something like `en-US` will be.


1ec5 (approved):
Good catch, thanks!

PR #1415, opened 2026-06-11
Render campground icon

### LINKED ISSUE #1126: Campgrounds and potential new POI color
Campgrounds have a long tradition of being [prominently displayed](https://github.com/ZeLonewolf/openstreetmap-americana/issues/435#issuecomment-1159582759) on American road maps, so they are very appropriate for this project. I think the icon to use is fairly obviously some sort of tent. I personally quite like the [NPS icon](https://www.nps.gov/npgallery/GetAsset/cd253093-6813-471e-9d16-042c2a373a97):
Screen Shot 2024-07-10 at 10 09 17 PM
Campgrounds (OMT `class=campsite` + `subclass=camp_site`) appear in the tiles beginning at z14, so that puts a practical limit on how early we can render them, despite American road maps sometimes displaying camping POIs even at state-wide scales.

What I'm most interested to discuss is what color to make the icon. The [contributing guidelines](https://github.com/ZeLonewolf/openstreetmap-americana/blob/main/CONTRIBUTING.md) have long left a spot open for an "outdoors" POI color, presumably a shade of green. This category could be appropriate for campsites, as well as other POIs like sports fields, picnic sites, and maybe even natural features like peaks (though that might be worth a separate discussion). If we go with green, care will be needed to choose a shade that contrasts appropriately with the park background fill many campgrounds will be on top of. Relatedly, would it be acceptable for the POI label text to be a similar shade of green as the park labels, or do they need to contrast? We'll also want to consider accessibility for color-blind users (see #1038), whether that means choosing a shade that transforms to a sufficiently different color under color-blindness or implementing some kind of color switcher.

Alternatively, campsites could plausibly be grouped with hotels in consumer-orange, or as attraction-brown.

Distinguishing campgrounds from individual campsites could be a candidate for a treatment similar to how we distinguish bus stops vs. stations: by inscribing the campground in a circle. But this is getting ahead of things, since `tourism=camp_pitch` is not present in the OMT tiles as far as I can tell.

wmisener:
Fixes #1126. Adds campgrounds (OSM `tourism=camp_site`, OMT `class=campsite` + `subclass=camp_site`). As I proposed in #1126, I've rendered campgrounds using a tent inscribed in a circle: 
poi_tent_circle

Campgrounds appear in the tiles at z14 and that's where this PR renders them. That's quite early compared to most other POIs (and compared to e.g. Carto, which renders icons at raster z16/vector z15), but campgrounds are usually in very sparse areas. American road maps often show them on state-scale maps.
37.74209/-119.58082
Screenshot 2026-06-10 at 11 10 40 PM

 I propose rendering them using our outdoors green color. Right now the only other POI icon using that is cemeteries. Campgrounds frequently appear in parks, so contrast with their background and labels is key. I think the drop shadow and the lack of bold font does the job sufficiently.
38.92803/-78.3283
Screenshot 2026-06-10 at 11 16 19 PM

The counterpart to the circle campground icon would be rendering the individual camp pitches/sites (`tourism=camp_pitch`) with just a tent icon, continuing our bus stop/station and restaurant/food court system. Unfortunately these are not present in the OMT tiles. I include the solo tent icon in the PR anyways: 
poi_tent
(EDIT: removed the pitch icon)

Campsites are also often mapped as areas, but are only present as points in the tiles, precluding area rendering for now.

### DIFF
--- a/icons/poi_tent.svg
+++ b/icons/poi_tent.svg
@@ -1,0 +1,3 @@
+<svg width="20" height="20" xmlns="http://www.w3.org/2000/svg">
+ <path d="M 15,13 10,3 5,13 H 3 v 2 h 14 v -2 z m -8,0 3,-6 3,6 z"/>
+</svg>
--- a/icons/poi_tent.svg
+++ b/icons/poi_tent.svg
@@ -1,3 +1,0 @@
-<svg width="20" height="20" xmlns="http://www.w3.org/2000/svg">
- <path d="M 15,13 10,3 5,13 H 3 v 2 h 14 v -2 z m -8,0 3,-6 3,6 z"/>
-</svg>
--- a/src/layer/poi.js
+++ b/src/layer/poi.js
@@ -309,6 +309,14 @@ var iconDefs = {
     color: Color.poi.infrastructure,
     description: "College or university",
   },
+  campsite: {
+    classes: {
+      campsite: ["camp_site"],
+    },
+    sprite: "poi_tent_circle",
+    color: Color.poi.outdoor,
+    description: "Campground",
+  },
   townhall: {
     classes: {
       town_hall: ["townhall"],
@@ -413,7 +421,10 @@ export const poi = {
         ...getSubclasses(iconDefs.pow_uu),
       ],
       Color.poi.infrastructure,
-      [...getSubclasses(iconDefs.cemetery)],
+      [
+        ...getSubclasses(iconDefs.cemetery),
+        ...getSubclasses(iconDefs.campsite),
+      ],
       Color.poi.outdoor,
       Color.poi.infrastructure,
     ],
@@ -428,7 +439,11 @@ export const poi = {
       10,
       ["station", "halt"],
       12,
-      ["bus_station", "subway"],
+      [
+        "bus_station", 
+        "subway",
+        ...getSubclasses(iconDefs.campsite),
+      ],
       14,
       [
         "bus_stop",
--- a/src/layer/poi.js
+++ b/src/layer/poi.js
@@ -439,11 +439,7 @@ export const poi = {
       10,
       ["station", "halt"],
       12,
-      [
-        "bus_station", 
-        "subway",
-        ...getSubclasses(iconDefs.campsite),
-      ],
+      ["bus_station", "subway", ...getSubclasses(iconDefs.campsite)],
       14,
       [
         "bus_stop",
--- a/icons/poi_tent_circle.svg
+++ b/icons/poi_tent_circle.svg
@@ -1,0 +1,4 @@
+<svg width="20" height="20" xmlns="http://www.w3.org/2000/svg">
+ <circle cx="10" cy="10" r="9.5" fill="#fff" stroke="#000"/>
+ <path d="M 15,13 10,3 5,13 H 3 v 2 h 14 v -2 z m -8,0 3,-6 3,6 z"/>
+</svg>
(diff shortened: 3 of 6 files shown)


icons/poi_tent.svg
1ec5, line 1: This branch isn’t using the plain tent icon. Is the idea that we’d eventually use it for [`tourism=camp_pitch`](https://wiki.openstreetmap.org/wiki/Tag:tourism%3Dcamp_pitch) (an individual campsite), whereas the circle indicates a campground (an entire site for campsites)? If so, I think we can omit the `camp_pitch` icon until we’re ready for it.

I’m reminded of how we also visually distinguish bus stops from bus stations using a circle. In both cases, I wonder if we could simplify the visual language by only showing one or the other at a given zoom level, so that it’s always the same icon without the circle. At lower zoom levels, it would slightly reduce visual complexity. But I guess we can track that idea separately as tail work.

src/layer/poi.js
```diff
@@ -413,7 +421,10 @@ export const poi = {
         ...getSubclasses(iconDefs.pow_uu),
       ],
       Color.poi.infrastructure,
-      [...getSubclasses(iconDefs.cemetery)],
+      [
+        ...getSubclasses(iconDefs.cemetery),
+        ...getSubclasses(iconDefs.campsite),
+      ],
       Color.poi.outdoor,
```
1ec5, line 428: The Outdoor category makes sense as long as we have it. My first thought was that it should be in the same category as lodging accommodations, but hotels and hostels are currently under Consumer. That would be unintuitive for campgrounds, which are more rugged and not necessarily commercial.

icons/poi_tent.svg
wmisener: I can remove the plain tent icon, you're right that it isn't being used. I was just worried I'd lose it on my computer ;). Yes, my idea would be to use it for `tourism=camp_pitch`, but those aren't in the tiles at the moment.

wmisener: As for the broader strategy, I've been a proponent of that visual language of inscribing an icon in a badge/circle for a few icons. It's currently used for bus stop/bus station, tram stop/train station, and restaurant/food court, and I was envisioning it for shop/mall as well. I see the variation on a theme as enhancing visual identity and communicating the 'center' character, rather than an unnecessary embellishment. And I think at low zooms there are cases where you'd want to see both icons, like both the name of the food court and the restaurants within it.

src/layer/poi.js
wmisener: Yeah, some other styles (like Carto) have a separate accommodation/travel color (not that that category is so well defined either, Carto uses its light blue for bus stops and parking lots too, but not for trains or rest areas). It'd be worth revisiting this choice in the future if we decide to have a more expansive color palette, but in my opinion this is the best fit for now. I'm not sure where RV parks would fit into this...but that's not the current problem.

icons/poi_tent.svg
wmisener: Removed the lone tent icon in [99291e8](https://github.com/osm-americana/openstreetmap-americana/pull/1415/commits/99291e856d4ff2a345881d51ada724b7f4f01ff1). It's just the current icon without the circle, so should be easy to re-create.

1ec5 (approved):
I guess this is reasonable. I kind of think we should remove the circle until we start showing `camp_pitch`, at which point we would need to distinguish the venue from the object. Until then, we’re implying that the user will see more detail after zooming in further, but no such detail arrives. It’s a pretty benign effect in any case.

wmisener:
Opened #1430 to track camp pitch rendering, currently blocked by their absence in the tiles

PR #1420, opened 2026-06-18
POI matching on class and subclass, not just subclass

### LINKED ISSUE #1291: University office rendering bug
University offices (OMT `class=office` `subclass=university`) get rendered in Americana with text but no icon, where they weren't intended to be rendered at all. See the screenshot, where American University and Career Services..., both tagged as `office=university` in OSM, are rendered as icon-less labels, while University of the District of Columbia has a proper mortarboard icon.

Image

I think this is because the code is doing some kind of `subclass` matching to label actual universities (OMT `class=college` `subclass=university`) without making sure the `class` also matches.

wmisener:
Fixes #1291.

Currently in poi.js, labels are rendered by matching the `subclasses` with OMT objects. However, OMT duplicates `subclass` names across different `classes`. This leads to bugs like in #1291, where labels are rendered for `class=office` + `subclass=university` because `class=college` + `subclass=university` is in our list of defined icons.

This PR fixes this bug by refactoring to match on both `class` and `subclass` using a helper function, `getClassSubclassGroups`. To do this, I reformatted all the floating subclasses in the zoom definitions and such to use the iconDefs. The logic is very similar to that already used to render the icons themselves, except the icons depend on properties defined in iconDefs rather than defined on the spot like the text labels and zoom levels. But maybe there's a way to further refactor it all into one function, I'm not sure.

The PR should produce no change in rendering, except that places tagged `office=university` should no longer render as icon-less text labels. A big refactor for a small bugfix, but this makes things much more resilient as more icon types are added, which increases the chance of accidental subclass overlap.

EDIT: there was actually a second case already in the code: `subclass=station` is in both `class=aerialway` and `class=railway`.

38.944509/-77.065173
Old
image
New
Screenshot 2026-06-17 at 11 00 47 PM
The "[American University](https://www.openstreetmap.org/node/2624290762)" and "[Student Success Center](https://www.openstreetmap.org/node/358954748)" are now gone.


src/layer/poi.js
```diff
@@ -262,11 +262,19 @@ var iconDefs = {
...
-    description: "Train or subway station",
+    description: "Train station",
+  },
+  rail_subway_station: {
+    classes: {
+      railway: ["subway"],
+    },
+    sprite: "poi_rail_circle",
+    color: Color.poi.transport,
+    description: "Subway station",
```
1ec5: This could result in the legend listing two identical icons, one as “Train station” and the other as “Subway station”. The legend supports tying an entry to multiple filters, but this file is translating them from the POI classes. Can we find a way to recombine them later on for the legend?

wmisener: Good catch, it definitely does do that. I had to separate the iconDefs for train and subway stations because they appear at different zoom levels. I'll have to look into the legend code but I bet we can find a way to coalesce them when the sprite and description are exact matches.

wmisener: Added code to deduplicate the legend entries in [358ad04](https://github.com/osm-americana/openstreetmap-americana/pull/1420/commits/358ad04cd55d9673f563bc79d79a0b4a030a1f64).
Screenshot 2026-06-22 at 11 11 18 PM
You can see "Train or subway station" renders once despite a train station and a subway station appearing here and the two appearing at different zoom levels.

1ec5 (approved):

PR #1422, opened 2026-06-19
Cyprus B/E/F road shields

### LINKED ISSUE #1421: Cyprus B/E/F road shields
Cyprus classifies non-motorway roads as B, E, and F roads. All but one of the B roads has been mapped as a route relation and tagged `network=CY:B`. [An E road](https://www.openstreetmap.org/relation/17930183) has also been tagged `network=CY:E`.

Cyprus generally uses British-style signs, so road numbers are usually just [colored text on a guide sign](https://commons.wikimedia.org/wiki/File:Agios_Tychonas_Road_Sign.jpg). However, B, E, and F road numbers sometimes appear in [a gold-trimmed box](https://www.dreamstime.com/cyprus-road-sign-distance-to-other-cities-image245155242) on distance and direction signs. Since non-motorways have blue signs, blue would be an intuitive background for these shields.

F125, B1, and A1 as gold text on a blue sign.

#558 has an implementation of Cyprus motorway shields, but the tagging is less certain.

1ec5:
Added shields for Cyprus B, E, and F roads based on the networks `CY:B`, `CY:E`, and `CY:F`, respectively. This takes care of all the non-motorway shields in Cyprus as well as Akrotiri and Dhekelia, a British Overseas Territory. The E and F roads are hardly mapped at all, but it stands to reason that they’d follow the same pattern as the B roads.

[B3](https://preview.americanamap.org/pr/1422/#map=13/34.99028/33.74445) [E327](https://preview.americanamap.org/pr/1422/#map=12/35.01121/33.98517)

Fixes #1421.

### DIFF
--- a/doc-img/shield_map_world.svg
+++ b/doc-img/shield_map_world.svg
@@ -157,6 +157,7 @@ See the end of this file for a list of available jurisdictions and their codes. 
 .bg,
 .by,
 .ch,
+.cy,
 .cz,
 .ee,
 .es,
--- a/src/js/shield_defs.js
+++ b/src/js/shield_defs.js
@@ -4101,6 +4101,16 @@ export function loadShields() {
     Color.shields.white
   );
 
+  // Cyprus
+  shields["CY:B"] =
+    shields["CY:E"] =
+    shields["CY:F"] =
+      roundedRectShield(
+        Color.shields.blue,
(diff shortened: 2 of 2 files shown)


ZeLonewolf (approved):
Looks good to me!

PR #1431, opened 2026-06-26
Military landuse rendering

### LINKED ISSUE #232: Military base styling

wmisener:
Uploading this draft PR for future reference, I was working on it a couple months ago:
Fixes #232 by adding a military landuse rendering. There's been debate in that issue and on slack (like [here](https://osmus.slack.com/archives/C01V02K52UX/p1776827070519979)) about appropriate colors. The current PR opts for a pink. Gray was also suggested and is common on print map samples, but to my eye it blended too much with buildings to be viable.
Screenshot 2026-06-25 at 9 39 44 PM

Outstanding issues are that this landuse appears lower in the stack than airport fill does, so military airports render fully on top of them. It was suggested that a second layer containing a hatching pattern could go on top, covering airfields and other landuses.

Military areas appear in OpenMapTiles as landuse areas. Unfortunately, OpenMapTiles lacks a corresponding named military POI point, meaning the names of bases are not in the tiles and cannot be rendered. This absence blocks our rendering of military bases for now, as unidentified red blobs aren't suitable for the style.


src/constants/color.js
```diff
@@ -39,6 +39,16 @@ export const airportOutline = "hsl(250, 41%, 79%)";
 export const airportRunway = "hsl(250, 41%, 79%)";
 export const airportLabel = "hsl(250, 71%, 29%)";
 
+export const militaryFill = "hsl(345, 55%, 92%)";
+export const militaryOutline = "hsl(345, 55%, 80%)";
+export const militaryLabel = "hsl(326, 40%, 35%)";
+export const militaryLabelHalo = "hsl(326, 40%, 96%)";
```
1ec5, line 45: Pink most commonly occurs on maps for tourist attractions. We’re already using pink for “transit”, mainly because of aeroways (which overlap with tourist attractions). I realize our attraction color is brown, but a layperson might see this without the legend and come to the wrong conclusion. What if we tweak this pink to something closer to a magenta hue, while desaturating it enough to avoid overpowering the map?

Issue #1425, opened 2026-06-24
Render stadiums


wmisener:
Stadiums (`leisure=stadium`) are important regional landmarks, and are usually rendered prominently on city/regional-scale paper maps. I think it would be nice to have a prominent icon and rendering for them. They appear in the tiles (`class=stadium` `subclass=stadium`) starting at z14.

On the wiki, the tag is described as for "major sports facilities with substantial tiered seating". Use in the wild seems to mostly conform to this, big pro and college sports venues and the like, although there is also noticeable usage on smaller-time facilities and even on mistagged bleachers. I partly attribute this to the poor existing mapper feedback: Carto doesn't treat this tag differently from sports fields, for example.

I think something like the Maki bowl stadium icon would be nice. This seems like it'd fit best into the 'attraction' brown category. Honestly z14 is a little zoomed-in to start the big-time arenas, but that's as early as possible.
<img width="157" height="41" alt="Image" src="https://github.com/user-attachments/assets/6c583638-fc7e-4d44-807c-e711d55560f8" />

Another possibility would be to make the icon based on the individual sports, but many arenas are multisport. I think rendering actual sport fields (`leisure=pitch`) as outdoor green with sport-specific icons would be cool too, but a separate issue.

1ec5:
Most conventional print maps I’ve seen label stadiums like any attraction: either a generic square icon with a label, or just a label over the area. City inset maps typically reinforce this symbol using an idealized representation of the stadium building footprint, sometimes emphasizing building parts. This building-centric treatment skirts the issue of distinguishing stadiums by sport, especially if the field within it already gets symbolized with a sport.

wmisener:
Yes if we don't want an icon, just labeling the area could also work. The tiles also have `class=stadium` landuse polygons: however these also include the grounds/field of the stadium, so it might be hard to emphasize the buildings themselves. Maybe if a stadium landuse layer were on top of the building layer but semi-transparent so it just darkened the building somehow? We could also link stadium labelling to that of theme parks (#1261).

[...]

PR #1435, opened 2026-07-21
Add sport fields

### LINKED ISSUE #1428: Sports fields
Sports fields (OSM `leisure=pitch`) are important neighborhood-scale POIs. They are present in OMT in two layers: as landuse areas (`class=pitch`, no subclasses) and as POIs near the center of the area `class=pitch` + `subclass=` a sport, presumably whatever is tagged in `sport=*` in OSM.

I think an approach similar to cemeteries could work (that sounds more morbid than I mean it to!). #329 rendered cemeteries as a green-tinted area labeled like a park that gains an icon at z17. One difference is that sports fields are frequently unnamed, and often occur within named parks. I'd propose a more saturated/brighter green for the fill so it sticks out against our pale green parks, with some kind of label or icon appearing within it. What Carto does is in the vein of what I mean fill-wise:
Image

It's possible we'd want to render playgrounds and/or dog parks in a similar way too. Our choices here might also have bearing on bigger areas like swimming facilities or golf courses. We'll have to be careful with the green text on green fills if we do this approach. I think the considerations for your average park basketball court are a bit different than for Madison Square Garden, so I consider stadiums separately (#1425).

I think it'd be neat to render an icon for at least some individual sports. It's always annoyed me on other map styles like Carto that you can see the fields but not the sports they're for! Tracestrack does render icons, they use people playing each sport. I was personally thinking the actual balls/equipment would be nicer-looking and fit into our existing simple style better: so like a baseball for baseball, a basketball/hoop for basketball, a football for football, etc. The action-people feel a bit overwrought compared to our style, and I'm not sure if they'd scale to 20px well. But I can see how this choice is intuitive in its own way:
Image

Sports fields are not usually that big, so their names at least can probably show up relatively late. It'd be bad to collide out park or stadium names for them! Maybe the icons can appear before any names.

wmisener:
Fixes #1428. This PR adds a saturated green fill for all sports fields (`leisure=pitch` in OSM, `class=pitch` landuse in OMT) and icons for the most popular US sports (`sport=*` in OSM, which come into OMT as subclasses of `class=pitch` POI icons). The sports I chose to give icons are those in the [top 15 in the US at this time](https://taginfo.geofabrik.de/north-america:us/tags/leisure=pitch#combinations) that "should" appear on `leisure=pitch`: tennis, baseball, basketball, pickleball, american_football, beachvolleyball, four_square, softball, equestrian, volleyball, and skateboard. This should by no means indicate that other sports could not possibly be rendered in the future, but it seemed a good place to start.

The sports field areas appear whenever they do in the tiles, which appears to be area-based but is often around z13 ([localhost](http://localhost:1776/#map=13.3/41.79552/-87.60761))
Screenshot 2026-07-20 at 10 06 36 PM
The saturated green is a new fill color, it seemed to stick out nicely against both parks and the default background. 

The icons appear mostly at z16 ([localhost](http://localhost:1776/#map=16.41/38.951244/-77.07555)):
Screenshot 2026-07-20 at 9 39 55 PM
The icons are typically a simple, recognizable piece of equipment from the game. For ball sports (baseball/softball, basketball, football, soccer/futsal, volleyball/beach volleyball), this is often the ball if it's recognizable enough. Tennis and pickleball get rackets and paddles instead. Skate parks get a skateboard, equestrian rings get a horse, and four square courts get...a four square court. The icons are 14x14, a little smaller than typical icons, to denote that these are typically subordinate features. Names appear next to the icons if the pitches are named.

Samples: 
Baseball, basketball, tennis, and soccer, with legend entries ([localhost](http://localhost:1776/#map=16.44/38.986811/-77.06198)):
Screenshot 2026-07-20 at 10 39 47 PM

Pickleball and tennis (some of the icons get collided out but appear as you zoom) ([localhost](http://localhost:1776/#map=16.8/32.753889/-117.23539)):
Screenshot 2026-07-20 at 9 50 02 PM

Baseball, football, soccer, tennis ([localhost](http://localhost:1776/#map=16.11/31.334701/-94.685144)):
Screenshot 2026-07-20 at 9 53 12 PM

Volleyball, skateboard, tennis, basketball ([localhost](http://localhost:1776/#map=16.28/33.986702/-118.472865)):
Screenshot 2026-07-20 at 9 55 39 PM

Four square, basketball, baseball, football ([localhost](http://localhost:1776/#map=18.93/37.7536793/-122.4379017)): 
Screenshot 2026-07-20 at 9 57 53 PM
Four square icons appear only at z18 rather than z16, otherwise the icons are often bigger than the individual courts. Weirdly, the four square courts are pretty square in OSM: https://www.openstreetmap.org/way/1098860548/. Must be something with OMT's polygon smoothing.

Horse, baseball, tennis ([localhost](http://localhost:1776/#map=17.04/43.47366/-110.767449)):
Screenshot 2026-07-20 at 10 03 26 PM

Single fields that are mistagged as multiple `leisure=pitch` appear many times, I think this is a feature not a bug ([localhost](http://localhost:1776/#map=16.77/42.339096/-83.048242)):
Screenshot 2026-07-20 at 10 11 19 PM


src/layer/poi.js
```diff
@@ -325,6 +325,86 @@ var iconDefs = {
...
+    classes: {
+      pitch: ["soccer", "futsal"],
+    },
+    sprite: "poi_soccer",
+    color: Color.poi.outdoor,
+    description: "Soccer field",
+  },
+  tennis: {
+    classes: {
+      pitch: ["tennis"],
```
1ec5, line 394: Should we reduce the symbol spacing between these mini POI icons? Some of the icons go missing due to closely spaced courts. We can’t completely solve this without allowing overlap, but maybe slightly closer spacing would make sense for a typical court size.

[A series of tennis courts at Bicentennial Commons. Eleven of the tennis icons are missing.](https://preview.americanamap.org/pr/1435/#map=17.01/39.10267/-84.49646)

src/layer/poi.js
```diff
@@ -325,6 +325,86 @@ var iconDefs = {
     color: Color.poi.outdoor,
     description: "Campground",
   },
+  baseball: {
+    classes: {
+      pitch: ["baseball", "softball"],
+    },
+    sprite: "poi_baseball",
+    color: Color.poi.outdoor,
+    description: "Baseball or softball field",
```
1ec5, line 334: I keep zooming into stadiums expecting to see a label, then realizing that we’re only symbolizing the nameless field within the stadium. Once we add symbols for stadiums, they’ll need to take precedence over field symbols.

[A baseball field inside a baseball stadium labeled only as a nameless field.](https://preview.americanamap.org/pr/1435/#map=17/39.096896/-84.506118)

1ec5 (commented):
Thanks for entertaining the idea of shrinking the icons a bit. It looks pretty good. Do we need to document a new kind of “mini POI” to extend to other features, or is this basically it in terms of micromapped symbols for now?

whubsch:
Do we want to get into the mess of sorting out multi-sport fields that are tagged with a semicolon separated list? For example, at [Catholic University's sports complex](https://preview.americanamap.org/pr/1435/#map=16.12/38.944699/-77.000598), the baseball/softball fields have an icon but the others don't because they are tagged as `soccer;lacrosse` and `american_football;field_hockey;soccer`. I think the only reasonable way to do that would be just to take the first value and assume that's the primary use, and this may not be a super common case.
Screenshot 2026-07-21 at 8 35 22 PM

1ec5:
We do technically have the ability to split on semicolons now, though ideally OpenMapTiles would provide something more structured for us. What would a visual representation of multiple sports look like? We could implement multiple icons in a row as part of the label, similar to route concurrencies, but it would behave as part of the label rather than a separately aligned icon group.

wmisener:
> Thanks for entertaining the idea of shrinking the icons a bit. It looks pretty good. Do we need to document a new kind of “mini POI” to extend to other features, or is this basically it in terms of micromapped symbols for now?

This is a good idea I think, especially as we push toward increasing detail and thus more minor features. I'm just now noticing the [contributing guide](https://github.com/osm-americana/openstreetmap-americana/blob/main/CONTRIBUTING.md#points-of-interest) doesn't even say what the normal POI icon size is (20x20, same as the shields). As far as I recall the only other 'small' POI icon the style currently uses is for bus stops, which are also 14x14.

src/layer/poi.js
wmisener: I'm just now realizing that when I shrunk the icons, I didn't actually shrink the canvas size, so while transparent it's still a 20x20 icon. Maybe shrinking the icon canvases will help alleviate some of the collisional issues, I'll try to test that shortly (in the next couple days).

src/layer/poi.js
wmisener: For sure! Tagging #1425 for reference. I expect stadiums will show up at much lower z in any case, given their local landmark status. Although to make things even more exciting for us, some teams have become so innovative in finding ways to sell sponsorships that the field has a different name than the stadium!
Screenshot 2026-07-21 at 10 48 23 PM
(Uniqlo Field at Dodger Stadium, of course)

wmisener:
> Do we want to get into the mess of sorting out multi-sport fields that are tagged with a semicolon separated list?

Haha, good catch, I was avoiding trying to wade into it, but it's worth thinking about.

> I think the only reasonable way to do that would be just to take the first value and assume that's the primary use, and this may not be a super common case.

> We could implement multiple icons in a row as part of the label, similar to route concurrencies, but it would behave as part of the label rather than a separately aligned icon group.

I tend to think picking the first value would be a reasonable thing to do, at least as a first pass. Showing all listed values would be fun but a little unwieldy I imagine. It might be interesting to design custom icons for particularly common combinations too (like `american_football;soccer`). But IMO addressing this in detail would be fine to be tail-work.

> We do technically have the ability to split on semicolons now

Oh I didn't know that! But now I see https://github.com/maplibre/maplibre-style-spec/pull/1518/. Is it just as simple as replacing ["get", "subclass"] with ["split", ["get", "subclass"], ";"] to match on? I'm not sure then which one it will end up matched on...maybe someone understands better than me.

1ec5:
> Is it just as simple as replacing ["get", "subclass"] with ["split", ["get", "subclass"], ";"] to match on? I'm not sure then which one it will end up matched on...maybe someone understands better than me.

This would result in an array that you’d index into using an `at` expression. It’s easy enough to read just the first value in the list and discard the rest, but reading each of the values would require generating the expression in a loop, as we do in highway_shield.js.

src/layer/poi.js
wmisener: Shrinking the icons to actually 14x14 helps a lot! I'll change all of them to be this way
Yours:
Screenshot 2026-07-23 at 9 09 01 PM
The San Diego one I had in the preview:
Screenshot 2026-07-23 at 9 04 52 PM

wmisener: Resolved by [aab5713](https://github.com/osm-americana/openstreetmap-americana/pull/1435/commits/aab5713297f8b8e72de22dbc6a9c88f1335edf49)

wmisener:
The splitting on semicolons works great! Implemented in [53f4b85](https://github.com/osm-americana/openstreetmap-americana/pull/1435/commits/53f4b85ea56db24536c8bfe38f4e4c4d22c0ef50). I just take the first value, so here it renders as soccer for `soccer;lacrosse` and American football for `american_football;field_hockey;soccer`.

Screenshot 2026-07-23 at 10 16 38 PM

I'm implementing it here for all POIs, not just sports fields. I don't know if any others come in lists through OMT (actually, it seems like shops tagged with semicolons don't show up in the tiles at all...), but this seems like a universally good way to treat them as a first pass. I think this is a nice improvement to the PR, so glad it got brought up!

I wouldn't be opposed to attempting to show all the tagged sports, but it seems like a heavy lift technically and design-wise that I don't personally want to take on right now. I don't think it's a blocker for this PR though.

src/layer/poi.js
```diff
@@ -341,7 +421,10 @@ function iconImageDefs() {
     let classes = iconDefs[key].classes;
     for (var poiClass in classes) {
       if (!mapping[poiClass]) {
-        mapping[poiClass] = ["match", ["get", "subclass"]];
+        mapping[poiClass] = [
+          "match",
+          ["at", 0, ["split", ["get", "subclass"], ";"]], // Split if it does
```
1ec5, line 426: I’m unsure if OSM has settled on always putting the “main” sport first in the list, but it’s easy to see that being the most common case, and it’s the best we can do for now.

1ec5 (approved):

src/layer/poi.js
wmisener: Yeah I think strict OSM convention would be that semicolon delimited lists are non-ordered, but I think this is a reasonable approach, it renders more information than any current osm.org style. We could certainly revisit it if this were to ever lead to edit wars 😬

quincylvania:
@wmisener Nice work! Where did you grab the icons from? I'm looking to sync these to Pinhead :)

wmisener:
Oh I drew some of them myself and adapted the rest from other sources! I often started from the Pinhead coverage table in the first place, so that'll be a good feedback loop. A few representative examples, just as an idea: the soccer ball is almost exactly the Pinhead one, just scaled to 14x14px. The football I think I took from Pinhead/Maki but with adaptations, I tweaked the dimensions and the seams. The skateboard I just drew myself in Inkscape, couldn't find any public domain icons that matched what I was going for in this icon set. Happy for you to put them in Pinhead, if you need anything from me lmk (on OSMUS Slack or here).

quincylvania:
@wmisener Nice!! I love the skateboard icon, excited to add that to Pinhead. The only other icon I had questions about was the horse.

wmisener:
For the horse I think I started from https://pinhead.ink/#person_riding_horse, but I wanted icons without any people, so I took out the person and made it look like it was trotting.

I think the only other fully new ones that didn't have a close Pinhead equivalent were the baseball and the basketball, where I wanted just the balls not anything else.

PR #1437, opened 2026-07-27
Add park tags to taginfo file

wmisener:
When making #1435, I noticed that the tags that the style interprets as parks didn't make it into the taginfo file. This PR fixes that, adding `leisure=nature_reserve`, `boundary=national_park`, `boundary=protected_area`, and `leisure=park` to the list of tags the style uses. The first three were rendered in #49, and the last in #306.

### DIFF
--- a/scripts/taginfo_template.json
+++ b/scripts/taginfo_template.json
@@ -978,6 +978,34 @@
       "icon_url": "https://raw.githubusercontent.com/osm-americana/openstreetmap-americana/main/icons/poi_gravestone.svg"
     },
     {
+      "key": "boundary",
+      "value": "protected_area",
+      "object_types": ["node", "area"],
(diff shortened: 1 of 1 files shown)


ZeLonewolf:
These should probably just be "area" and not "node" - to the extent that we draw icons, they're computed from the underlying areas.

wmisener:
For `leisure=park` a label is actually rendered even if it's just a node, but that doesn't appear true for e.g. `leisure=nature_reserve`. Compare https://www.openstreetmap.org/node/356581070 (https://americanamap.org/#map=17.31/38.850864/-77.307083) and https://www.openstreetmap.org/node/358960091 (https://americanamap.org/#map=15.98/38.943352/-77.104141). I'll change the first three, but I think `node`, `area` is correct for `leisure=park`.

ZeLonewolf (approved):

PR #1448, opened 2026-07-29
Render mountain peaks

### LINKED ISSUE #1362: Label mountain peaks
Transportation maps often label mountain peaks, which are major landmarks. Typically peak labels are italicized like other natural features are marked by a small triangle (or no icon at all). Even without an icon, they’re identifiable as topographical features because of an elevation indicator.

DeLorme Southern & Central California Atlas & Gazetteer

Best Western Nevada  New Mexico Department of Transportation Rand McNally Hallwag Nevada Department of Transportation 2004

OpenMapTiles exposes peaks in the [`mountain_peak`](https://openmaptiles.org/schema/#mountain_peak) layer. OpenMapTiles provides the elevation in both meters and feet. By convention, maps of the U.S. list elevations in feet, even when published overseas. Maps of other places that originate in the U.S. may list elevations in either feet or meters depending on the audience. As a starting point, we can label everything in feet. Later, we can add a unit option that affects both the peak labels and the scale bar (#898). As of maplibre/maplibre-style-spec#1335, we can format the elevation without having to hard-code a locale. osm-americana/diplomat#32 would automatically adjust the number formatting locale according to the preferred locale.

wmisener:
Fixes #1362. In that issue, all the posted samples from US road maps used an unfilled black triangle. They also mostly used italics (like we do for natural features like lakes) and had the elevation displayed underneath the label. So I've followed suit in this PR:
Screenshot 2026-07-28 at 8 59 01 PM
Screenshot 2026-07-28 at 8 58 54 PM

Peaks appear in OMT in their own layer starting at z7, and come with a `rank` attribute that somehow decides their local (by tile) importance. The ranking method is a black box as far as I can tell, and my subjective assessment is that it does an OK job at picking the most important peaks, although with the odd, indecipherable choice thrown in here and there. So I've decided to filter aggressively on it, only using `rank=1` until late zoom levels. But as the tiles get smaller, more peaks reach `rank=1`, so there ends up being a gradual increase in peaks appearing as you zoom in. All peaks are exposed at z>14.
z9:
Screenshot 2026-07-28 at 9 45 04 PM
z11:
Screenshot 2026-07-28 at 9 43 36 PM
z14:
Screenshot 2026-07-28 at 9 19 29 PM

I decided to start rendering at z8, a "regional" level. I considered starting at z7, which in my opinion showed too many low importance peaks at a statewide scale in flatter areas (Kansas is not littered with mountains). Names appear whenever the peak renders, I kinda hate e.g. Carto's method showing a bunch of unlabeled triangles if many peaks are mapped.

Peak elevations render in local units (denoted by ft or m following the elevation). This is accomplished using OMT's `customary_ft` attribute, which is set to 1 in the US and False everywhere else as far as I can tell, and OMT containing elevations in both meters and feet everywhere. In the future, perhaps we could allow users to toggle units.
Screenshot 2026-07-28 at 9 13 03 PM

Peaks with an elevation but no name are labeled with just the elevation, and vice versa:
Screenshot 2026-07-28 at 9 15 06 PM

Volcanoes (OSM `natural=volcano`) and normal mountains (OSM `natural=peak`) render identically in this PR. They are separate classes in the tiles and so could be rendered distinctly in the future
Screenshot 2026-07-28 at 9 17 59 PM

I opted to gloss the names in the local language(s), reasoning that these are place names of a sort, and that there will often be a mix of translated and untranslated peaks in close proximity:
Screenshot 2026-07-28 at 9 25 32 PM

Happy to hear any feedback!


ZeLonewolf:
I love how this came out.  I wasn't sure what an American cartographic depiction of a mountain peak was before this PR but seeing it, this definitely speaks to paper map.  :+1: 

image

src/layer/peak.js
```diff
@@ -0,0 +1,78 @@
...
+  ["<=", ["get", "rank"], 1], //Only rank 1 (dynamic) to z13
+  13,
+  ["<=", ["get", "rank"], 2], //Start showing rank 1+2 at z13
+  14,
+  [">=", ["get", "rank"], 1], //Show all past z14
+];
+
+const eleUnits = [
+  "case",
+  ["==", ["get", "customary_ft"], 1], //customary_ft variable is essentially whether the object is in the US or not
```
1ec5: > In the future, perhaps we could allow users to toggle units.

This would be in scope for osm-americana/diplomat#32.

src/layer/peak.js
```diff
@@ -0,0 +1,78 @@
...
+  ["concat", ["get", "ele_ft"], " ft"], //If customary_ft, return ele in ft
+  ["concat", ["get", "ele"], " m"], //Otherwise return it in m
```
1ec5: Until osm-americana/diplomat#32 is implemented, we should manually surround the `get` expression in a `number-format` expression. This will add digit grouping and decimal separators based on the user’s preferred locale. More importantly, if the language uses non-European numerals, it’ll use those numerals instead. You can set the `unit` property to append the unit symbol as needed.

src/layer/peak.js
```diff
@@ -0,0 +1,78 @@
...
+    peakStepFilter,
+  ],
+  layout: {
+    "text-font": ["Americana-Italic"],
+    "text-size": 9,
+    "icon-image": "peak",
+    "icon-size": 1.0,
+    "text-field": peakTextExpression,
+    "text-anchor": "bottom",
+    "text-variable-anchor": ["right", "left"],
```
1ec5: In a left-to-right writing system, right-aligned text can be more difficult to read when it’s spread across multiple lines. Let’s switch this to prefer `left` before `right`. We might also consider `top` so the label appears below the icon.

src/layer/peak.js
```diff
@@ -0,0 +1,78 @@
...
+  "case",
+  ["==", ["get", "customary_ft"], 1], //customary_ft variable is essentially whether the object is in the US or not
+  ["concat", ["get", "ele_ft"], " ft"], //If customary_ft, return ele in ft
+  ["concat", ["get", "ele"], " m"], //Otherwise return it in m
+];
+
+var peakTextExpression = [
+  "case",
+  ["all", ["has", "name"], ["has", "ele"]], //If name and ele present, use name with ele on next line
+  ["concat", localizedNameWithLocalGloss, "\n", eleUnits],
```
1ec5: > I opted to gloss the names in the local language(s), reasoning that these are place names of a sort, and that there will often be a mix of translated and untranslated peaks in close proximity:

I’d even go a bit further: many atlases tend to label any natural feature in (a transliteration of) the local language, so a dual language label is quite reasonable for peaks. The main downside I see is that Wikidata tends to abbreviate “Mt.” in English whereas OSM spells it out, causing the name to be duplicated. We have the same issue with “St.” in place names in any case: https://github.com/osm-americana/openstreetmap-americana/pull/592#issuecomment-1336810468.

wmisener: Yes, I've noticed quite a few places where this rendering surfaces spelling/orthography differences between OSM and Wikidata. Maybe they're more common since peak names are fairly informal compared to, say, towns. I guess every renderer is a validator right

src/layer/peak.js
wmisener: Done in [a9c35d7](https://github.com/osm-americana/openstreetmap-americana/pull/1448/commits/a9c35d70028363a97e0f1f27d7589437a607bf97). This adds thousands separators, which I don't honestly love (I thought it looked cleaner before) but am fine with in the name of better tooling. I couldn't honestly find a language to switch into that didn't use Arabic numerals.
Screenshot 2026-07-30 at 9 02 37 PM

src/layer/peak.js
wmisener: Switched to prefer left in [c8861c4](https://github.com/osm-americana/openstreetmap-americana/pull/1448/commits/c8861c4ba71cdec47439d3ed2624390065289a46). I didn't see much advantage in including top after some testing

src/layer/peak.js
```diff
@@ -0,0 +1,78 @@
...
+  ["<=", ["get", "rank"], 2], //Start showing rank 1+2 at z13
+  14,
+  [">=", ["get", "rank"], 1], //Show all past z14
+];
+
+const eleUnits = [
+  "case",
+  ["==", ["get", "customary_ft"], 1], //customary_ft variable is essentially whether the object is in the US or not
+  ["number-format", ["get", "ele_ft"], { unit: "foot" }], //If customary_ft, return ele in ft
+  ["number-format", ["get", "ele"], { unit: "meter" }], //Otherwise return it in m
```
1ec5: For consistency, format the number according to the selected locale, which isn’t necessarily the browser locale. Inside the options object, set `locale` to `getLocales()[0]` (exported from Diplomat).

src/layer/peak.js
1ec5: > This adds thousands separators, which I don't honestly love (I thought it looked cleaner before) but am fine with in the name of better tooling.

It makes more of a difference at five digits than at four:

[Mount Williamson at 14,380 feet, Mount Whitney at 14,500 feet, and an unnamed peak at 5,509 feet.](https://preview.americanamap.org/pr/1448/#map=10/36.6373/-118.2195)

[`number-format`](https://maplibre.org/maplibre-style-spec/expressions/#number-format) is modeled after `Intl.NumberFormat` in JavaScript. However, that constructor’s [`useGrouping`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#usegrouping) option isn’t part of the style specification yet. I think we could reasonably ask MapLibre to add more options to the `number-format` syntax, but it needs to be judicious. Not every platform has the same underlying options built into the standard library for MapLibre to use.

1ec5 (commented):
Once you fix the locale setting, you’ll be able to set the language to `uk` to see different unit symbols, or `ar` or `hi` to see different numerals.

src/layer/peak.js
wmisener: Hm, this seems to be giving me issues. Setting this then changing the language in my test environment fails to change the unit symbol. Changing `locale` manually in the code does change it.

I was hoping it was just a test environment issue, but it fails to build a PR preview here at all.

`locale: getLocales()[0]`, with language set to Ukrainian
Screenshot 2026-07-31 at 8 02 06 AM

`locale: 'uk'`
Screenshot 2026-07-31 at 8 02 27 AM

1ec5: Ah, it’s because the headless build/test environment doesn’t have `window`: osm-americana/diplomat#66.

```
file:///home/runner/work/openstreetmap-americana/openstreetmap-americana/pr-branch/node_modules/@americana/diplomat/index.js:95
  let parameter = getLanguageFromURL(window.location)?.split(",");
                                     ^

ReferenceError: window is not defined
    at getLocales (file:///home/runner/work/openstreetmap-americana/openstreetmap-americana/pr-branch/node_modules/@americana/diplomat/index.js:95:38)
    at file:///home/runner/work/openstreetmap-americana/openstreetmap-americana/pr-branch/src/layer/peak.js:24:29
    at ModuleJob.run (node:internal/modules/esm/module_job:439:25)
    at async node:internal/modules/esm/loader:643:26
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)
```

1ec5: [Diplomat v0.6.0](https://github.com/osm-americana/diplomat/releases/tag/v0.6.0) includes the fix for this error.

wmisener: This allows the checks to pass, but it still doesn't seem to have the expected behavior in the PR preview: switching languages does not switch the unit labels.

1ec5: Looks like it works when refreshing, but not dynamically. 🤔

wmisener: Oh weird, yeah I reproduce that as well

src/layer/peak.js
```diff
@@ -0,0 +1,83 @@
...
+  [">=", ["get", "rank"], 1], //Show all past z14
+];
+
+const eleUnits = [
+  "case",
+  ["==", ["get", "customary_ft"], 1], //customary_ft variable is essentially whether the object is in the US or not
+  [
+    "number-format",
+    ["get", "ele_ft"],
+    { unit: "foot", locale: getLocales()[0] },
```
1ec5: This only localizes the references to name fields in the `text-field` property’s `format` expression:

https://github.com/osm-americana/openstreetmap-americana/blob/1345510328835b86f675ad06e9d538488485cbad/src/js/map_view.ts#L11

Since we’re setting the `locale` manually here, we need to explicitly change the locale by replacing that part of the expression at the same time. To make this easier, move `eleUnits` into a `let` expression and refer to it by a `var` expression everywhere you’re currently referring to an `eleUnits` JavaScript variable. Something like this:

```js
[
  "let",
  "eleUnits",
  getEleUnits(),
  [
    "case",
    ["all", …],
    ["concat", localizedNameWithLocalGloss, "\n", ["var", "eleUnits"]],
    …
  ],
]
```

Then we can inject an updated number format expression when setting a new language:

```js
import updateVariable from "@americana/diplomat";

let peakTextExpression = this.getLayerProperty("peak", "text-field");
peakTextExpression = updateVariable(peakTextExpression, "eleUnits", getEleUnits());
this.setLayerProperty("peak", "text-field", peakTextExpression);
```

This hard-codes a particular layer, which is fine for Americana but not generic enough if we want to upstream it to Diplomat. Once maplibre/maplibre-gl-js#7933 is implemented, we’ll be able to move `number-format` expression’s format options object to [global state](https://maplibre.org/maplibre-style-spec/root/#state), making this step unnecessary. Instead, we’ll be able to create that option dynamically and have it rely on the same global state as everything else:

```js
[
  "number-format",
  ["get", "ele_ft"],
  [
    "semiliteral",
    {
      unit: "foot",
      locale: ["at", 0, ["global-state", "locales"]],
    },
  ],
]
```

src/layer/peak.js
```diff
@@ -0,0 +1,83 @@
...
+  },
+  minzoom: 8,
+  source: "openmaptiles",
+  "source-layer": "mountain_peak",
+};
+
+export const legendEntries = [
+  {
+    description: "Peak",
+    layers: [peak.id],
```
1ec5, line 84: The legend omits the peak symbol for some reason:

[Cerro Colorado Mountains (4,843 ft): Peak](https://preview.americanamap.org/pr/1448/#map=17.42/31.703058/-111.255055&language=en-US)

wmisener: Hm yeah I see that too. I can't for the life of me figure out why it's not catching the icon here, the legendEntries code seems exactly the same as e.g. the `place` or `poi` layers and nothing sticks out in `legend_control.js` that would be failing here.

src/layer/peak.js
wmisener: > Then we can inject an updated number format expression when setting a new language:
> 
> import updateVariable from "@americana/diplomat";
> 
> let peakTextExpression = getLayerProperty("peak", "text-field");
> peakTextExpression = updateVariable(peakTextExpression, "eleUnits", getEleUnits());
> setLayerProperty("peak", "text-field", peakTextExpression);

Sorry, where is this block is supposed to go?

I've reformatted the eleUnits constant into a getEleUnits function successfully

src/layer/peak.js
wmisener: Ah, `icon-image` has to be an expression for the legend code to pick it up, so changing `"peak"` to `["literal", "peak"]` resolves the symbol not showing up

Screenshot 2026-08-05 at 10 25 21 AM

1ec5: Ah, this is the culprit:

https://github.com/osm-americana/openstreetmap-americana/blob/1345510328835b86f675ad06e9d538488485cbad/src/js/legend_control.js#L292

When it’s an expression, then the evaluated expression looks like:

```js
{ name: "peak", available: false }
```

But when it’s just a constant value, it becomes just the string `"peak"`. This code would handle both cases:

```js
let imageValue = symbol.layer.layout["icon-image"];
let imageName = imageValue?.name ?? imageValue;
```

src/layer/peak.js
1ec5: The import goes at the top of map_view.js. The rest goes before or after this function call:

https://github.com/osm-americana/openstreetmap-americana/blob/1345510328835b86f675ad06e9d538488485cbad/src/js/map_view.ts#L11

src/layer/peak.js
```diff
@@ -15,26 +15,32 @@ const peakStepFilter = [
...
-  ], //If customary_ft, return ele in ft
-  ["number-format", ["get", "ele"], { unit: "meter", locale: getLocales()[0] }], //Otherwise return it in m
-];
+function getEleUnits() {
+  return ["case",
+    ["==", ["get", "customary_ft"], 1], //customary_ft variable is essentially whether the object is in the US or not
+    [
+      "number-format",
+      ["get", "ele_ft"],
+      { unit: "foot", locale: getLocales()[0] },
```
1ec5: We could inline more of `getEleUnits()`, so that the only things being created by JavaScript functions are the two `{unit: …, locale: …}` objects.

src/layer/peak.js
wmisener: Yes this works! Done in [0d7716a](https://github.com/osm-americana/openstreetmap-americana/pull/1448/commits/0d7716a67ce233267900ec6ebe16add2c31cf8c6)

src/layer/peak.js
wmisener: This did not seem to work, I get the same behavior as before. I thought the problem might be that getLocales() was called as you were setting a new locale, but my attempt to pass the new locale through didn't fix the problem either. Uploaded this failed attempt as [ccf1bde](https://github.com/osm-americana/openstreetmap-americana/pull/1448/commits/ccf1bde2ac4b421267f452548d8f6fb3bd2e3beb) anyways.

src/layer/peak.js
wmisener: Do you mean like moving the "number_format" up into peakTextExpression? The ["get", ...] statements are also different between the two cases, because ele and ele_ft are separate OMT variables.

src/layer/peak.js
1ec5: I pushed a fix for the bug. There had been a few typos in my earlier code suggestion, and a function needed to be imported. I’m surprised TypeScript didn’t catch the error at build time.

src/layer/peak.js
1ec5: I was thinking we’d isolate the format options objects as two different variables. But on second thought, that won’t work, because `updateVariable()` expects the replacement value to be an expression, not an object.

1ec5 (approved):
Thanks for hanging in there with the localization hurdles!

PR #1463, opened 2026-08-10
Add a rank filter for parks at low z

### LINKED ISSUE #1290: Too many parks show up at low zoom
As described [on Slack](https://osmus.slack.com/archives/C01V02K52UX/p1756839706360909), the map is very cluttered at low zooms with insignificant park units. These both overwhelm the map and make it hard to find the actually important parks. A clear example is the Western US:

Image

Park labels currently appear at z5. These parks are tagged in OSM in a large variety of ways, making it hard to use real tagging to differentiate what gets rendered when. One possible solution is to set all park labels with minzoom=7, so nothing comes in until a couple of zoom levels later than they do currently. Another possibility that I think has a lot of potential is aggressively filtering based on OMT `rank`. `rank=1` typically includes National Parks and a few other units, while a lot of the forest units and wilderness areas have high `rank` values. The exact values might take some fine tuning, but rendering, say, `rank<2` at z5, `rank<3` at z6, etc. could go a long way toward making the map less cluttered.

wmisener:
Fixes #1290. Implements a `rank` filter on parks similar to what was implemented in [peak.js](https://github.com/osm-americana/openstreetmap-americana/blob/main/src/layer/peak.js). In practice, this removes (EDIT: the labels of) a bunch of relatively insignificant wilderness areas, state/regional parks, and protected forests from very low zooms, reducing clutter on the map while preserving a fairly park-forward rendering.

I picked `rank<=2` for z5 through 9 subjectively after testing various options. It seemed to e.g. surface most big national parks at z5, although the exact way OpenMapTiles chooses `rank` is still a black box to me. A `rank=1` filter eliminates some national forests and such, although not all, plus some national parks from z5, so I went with the more inclusive one. After z10 everything appears like it did before, but most big park areas of any type render before that once they take up enough of a tile.

Also removes the `aboriginal_lands` filter added in #1019, I think this has been vestigial since https://github.com/openmaptiles/openmaptiles/pull/1604 when OpenMapTiles separated out indigenous areas from the parks layer.

Samples:
Western US z5
Current:
Screenshot 2026-08-09 at 10 35 57 PM

New:
Screenshot 2026-08-09 at 10 37 13 PM

Upper Midwest z5.5, referenced in https://github.com/aaroads-wiki/openstreetmap-americana/issues/7
Current:
Screenshot 2026-08-09 at 10 43 31 PM

New:
Screenshot 2026-08-09 at 10 43 42 PM

Colorado z7:
Current:
Screenshot 2026-08-09 at 10 52 20 PM

New:
Screenshot 2026-08-09 at 10 52 27 PM

France z5
Current:
Screenshot 2026-08-09 at 11 08 30 PM

New:
Screenshot 2026-08-09 at 11 08 45 PM

### DIFF
--- a/src/layer/park.js
+++ b/src/layer/park.js
@@ -3,11 +3,16 @@
 import { localizedName } from "@americana/diplomat";
 import * as Color from "../constants/color.js";
 
-const parkLayerFilter = ["!=", ["get", "class"], "aboriginal_lands"];
+const parkStepFilter = [
+  "step",
+  ["zoom"],
+  ["<=", ["get", "rank"], 2], //Only rank 2 (dynamic) to z10
+  10,
+  [">=", ["get", "rank"], 1], //Show all past z10
+];
 
 export const fill = {
   id: "protected-area_fill",
-  filter: parkLayerFilter,
   type: "fill",
   paint: {
     "fill-color": Color.parkFill,
@@ -18,7 +23,6 @@ export const fill = {
 
 export const outline = {
   id: "protected-area_outline",
-  filter: parkLayerFilter,
   type: "line",
   paint: {
     "line-color": Color.parkOutline,
@@ -31,7 +35,7 @@ export const outline = {
 export const label = {
   id: "protected-area_label",
   type: "symbol",
-  filter: ["all", ["has", "rank"], parkLayerFilter],
+  filter: parkStepFilter,
   paint: {
     "text-color": Color.parkLabel,
     "text-halo-blur": 1,


src/layer/park.js
```diff
@@ -3,11 +3,16 @@
 import { localizedName } from "@americana/diplomat";
 import * as Color from "../constants/color.js";
 
-const parkLayerFilter = ["!=", ["get", "class"], "aboriginal_lands"];
+const parkStepFilter = [
+  "step",
+  ["zoom"],
+  ["<=", ["get", "rank"], 2], //Only rank 2 (dynamic) to z10
+  10,
+  [">=", ["get", "rank"], 1], //Show all past z10
```
1ec5, line 11: Would more steps make the transition at z10 less dramatic?

1ec5 (commented):
Looks reasonable, especially since it only affects the labels, not the fills or outlines.

src/layer/park.js
wmisener: Do you have a test locale in mind that looks especially dramatic? I didn't find it to be very abrupt when I was poking around. The funny thing about filtering by OMT `rank` is that the ranks adjust at each z level, so the value of, say, a `rank=5` object at z5 gets lower and lower as it takes up more of a tile at increasing z. By z10, most parks seem to be `rank=1` or `2`. So even when filtering by constant `rank` you get the effect of gradually increasing the number of labels anyways.

1ec5: I see, so if there is a spot where it transitions abruptly, it would probably need to be somewhere with a high density of parks to label once we get to z10. Seems OK to me then.

1ec5 (approved):

wmisener:
I by no means think this is a perfect solution to how to render parks at low z. It'd be really good to do it such that only National Parks get rendered at certain levels, etc, but that's a challenge to generalize worldwide. I think this is definitely an improvement on the status quo though!

